PR #100, opened 2026-05-19
feat(tui): preview idea text in grid with `i`

ivankuznetsov:
## Summary

Pressing `i` on a focused task in the TUI grid now shows that task's original idea text in the bottom strip, and any key dismisses it. Operators no longer have to leave the TUI to remember what a task in `3-plan/` or `7-finalize/` was originally about.

The gesture works on any focused row regardless of stage because the handler reads `<row.folder>/idea.md` directly, the file that `hive new` writes once in `1-inbox/` and that travels with the folder through every later stage.

## Design notes

The change is intentionally additive: a new `:idea_preview` mode hangs on the existing MVU pipeline (model fields, message, key binding, side-effect handler, update branch, view). No existing message, view, or update branch changes shape; each one only grows a new case.

A few decisions worth the reviewer's attention:

- **Read-only by construction.** `KeyMap#idea_preview_message` returns `Messages::BACK` for *every* key without inspecting it, so there is no surface for accidental edits and no whitelist that a future Bubble Tea release could outgrow.
- **Reuses the bottom-strip pathway.** `compose_idea_preview_view` composes through the same `compose_two_pane_view(footer: ...)` plumbing that `Views::FilterPrompt` and `Views::NewIdeaPrompt` use, so the preview shares layout with the existing prompts.
- **Failures flash, not crash.** Missing folder, missing `idea.md`, missing `original_text` key, malformed YAML, or unreadable file all flash a slug-scoped message and stay in `:grid`. `Psych::Exception` is rescued at the handler so the operator sees the slug, not the generic "internal error" surface.
- **Bounded render.** `original_text` is truncated at `Model::NEW_IDEA_BUFFER_MAX_CHARS` (4096) before storing, and the body view caps at `MAX_VISIBLE_ROWS = 6` wrapped lines, matching the composer's window. Scrolling longer ideas is explicitly out of scope.
- **Snapshot stability.** The preview captures `idea_preview_text` and `idea_preview_slug` at open time, so the 1Hz snapshot poll moving the cursor underneath does not change the preview content until dismissed.

## What's not in scope

- No CLI surface (`hive show ...` or similar) — TUI-only by design.
- No display of stage documents (`brainstorm.md`, `plan.md`, ...). Only the source idea's `original_text`.
- No edit affordances or stage re-trigger from inside the preview.
- No scrolling or pagination for ideas longer than the visible window.
- Image-attachment placeholders (`[imageN]`) render verbatim; the preview does not resolve them.

## Tests

Unit coverage was added per implementation unit:

- `test/unit/tui/model_test.rb` — new fields default to `nil` and survive `with(...)`.
- `test/unit/tui/messages_test.rb` — `OpenIdeaPreview` carries the row.
- `test/unit/tui/key_map_test.rb` — `i` in `:grid` opens preview; `i` in `:filter` and `:new_idea` still types as text; every key in `:idea_preview` returns `BACK`.
- `test/unit/tui/bubble_model_test.rb` — happy path reads `original_text`, plus flash branches for missing folder, missing `idea.md`, missing key, malformed YAML, oversized text truncation, and a roundtrip open-then-dismiss test that also pins the no-subprocess contract.
- `test/unit/tui/update_test.rb` — `BACK` from `:idea_preview` clears the preview fields, returns to `:grid`, and preserves cursor and scope.
- `test/unit/tui/views/idea_preview_test.rb` — header, body, hint, width truncation, row cap, and nil-text defensive render.
- `test/unit/tui/help_test.rb` and `test/unit/tui/views/help_overlay_test.rb` — `?` documents the new gesture.

Full suite: `2371 runs, 8169 assertions, 0 failures, 0 errors, 1 skips`. Manual interactive dogfood of `bin/hive tui` has not been run yet.

---


ivankuznetsov:

## High
- [ ] Diff includes broad non-TUI behavior outside the authoritative idea-preview plan: `main..HEAD` changes status, daemon, bot, finalize, schema, docs, and README files even though the plan limits this slice to `lib/hive/tui/**` and its tests, so this branch would ship unrelated CLI/daemon/bot behavior with the TUI preview.

## Medium
- [ ] Enter on `7-finalize` rows that are still `ready_to_finalize` opens the summary instead of dispatching finalize: `TaskAction` now classifies `COMPLETE is_draft=true` rows as `ready_to_finalize`, but `KeyMap.finalize_complete_row?` still checks only `stage == "7-finalize"` and `marker == "complete"`, so contextual Enter returns `OpenSummary` and bypasses the ready action.
- [ ] Refresh diagnosis is exposed for markerless `PLAN_MISSING_OUTPUT` rows but cannot succeed: `TaskAction` emits an `error` diagnostic/retry for marker `:none` plan output gaps, while `hive status --diagnose --write` goes through `DiagnosisAgent`, which rejects every marker outside review/error/execute-stale recovery markers.

## Nit

PR #125, opened 2026-05-22
feat(drop): hard-delete tasks via Shift+X and `hive drop`

ivankuznetsov:
## Summary

Abandoning an active task previously meant killing the agent, removing the task folder, pruning the worktree, deleting the branch, and closing the draft PR by hand — five steps across three surfaces. `hive drop <slug>` and the rebound `Shift+X` in `hive tui` collapse that into one idempotent action: a row vanishes within ~2s, the agent is signalled cleanly, and the worktree, branch, locks, logs, and draft PR are gone. Re-running after an interrupt converges; there is no dropped bucket, no undo, and no prompt beyond the Shift gesture itself.

### What changed

- `Hive::Commands::Drop` (new) performs the cleanup sequence and emits a `hive-drop` JSON envelope. Both the CLI verb and the TUI handler delegate to it, so behavior stays in lockstep.
- `hive tui` grid-mode `X` now dispatches `hive drop <slug> --project <name> --from <stage> --json` against the focused row. Lowercase `x` remains a no-op. Empty grid and archived (`9-done`) rows flash a refusal instead of acting.
- `Hive::Daemon::Dispatcher#dispatch_or_block` skips rows whose folder has vanished between snapshot capture and dispatch, closing the window where a mid-tick drop could spawn a fresh agent against a freshly deleted slug.
- Process-kill logic lifted from `Commands::Bot` into a shared `Hive::ProcessKill` module so Drop, Bot, and any future caller share one PID-reuse guard (`process_start_time` match) and one TERM/KILL escalation.
- `Hive::GitOps#delete_branch!` and `Hive::Worktree#remove!(force:)` + `prune_worktrees!` give Drop the pieces it needs without ad-hoc shellouts.
- `schemas/hive-drop.v1.json`, `wiki/commands/drop.md`, and a `CHANGELOG.md` entry document the new verb and the breaking TUI rebind.

### Design decisions worth a second look

| Decision | Rationale |
|---|---|
| **Cleanup order is fixed: kill → close PR → worktree → branch → folders → logs → audit commit.** | Every step except the audit commit is best-effort and treats "already gone" as success. A mid-flight interrupt + retry always converges. The audit commit is the only step that must surface failure, because it's the durable `git log hive/state` record of what was removed. |
| **PID-reuse guard refuses to signal when `process_start_time` doesn't match.** | Same guard `Hive::Lock#stale_lock?` and `Daemon::Commands.pid_owned_by_us?` already use. Envelope surfaces `agent_kill_skipped_reason: "pid_reuse_guard"` so operators can see the choice was deliberate. |
| **Daemon guards on `File.directory?(row.folder)` per dispatch, not a denylist.** | The natural "folder is gone → next snapshot omits it" effect handles the steady state. One stat per dispatch closes the in-flight-tick race without introducing a denylist that would need its own lifecycle. |
| **No confirmation modal beyond Shift.** | Shift+letter is not a casual keystroke; the friction is the modifier. The Help overlay row explicitly documents "no undo, no confirmation", so the surprise surface is discovery, not first-time use. |
| **Breaking TUI rebind: `X` no longer drops missing-project registry entries.** | That binding was a `mktemp -d` escape hatch (see the comment that was removed from `BubbleModel`). The replacement is `hive forget NAME` from the shell, which is already documented and covers every CLI workflow. CHANGELOG calls this out under `Unreleased`. |

### Refusals

| Case | Exit | `error_kind` |
|---|---:|---|
| Unknown slug or path | 64 | `invalid_task_path` |
| Ambiguous slug across projects (no `--project`) | 64 | `ambiguous_slug` |
| `--from` doesn't match the resolved stage | 4 | `wrong_stage` |
| Task already at `9-done` | 64 | `already_archived` |
| Hive-state commit lock contention | 75 | `concurrent_run` |
| Internal failure | 70 | `internal` |

Archived tasks are out of scope on purpose: `9-done` is the audit record, not active work. The TUI flashes the same refusal instead of acting on the row.

## Test plan

- `test/unit/commands/drop_test.rb` covers the happy path across every active stage (parameterised over `ACTIVE_STAGE_DIRS`), archived guard, unknown/ambiguous slug, PID-reuse safety, missing `gh`, `gh` returning non-zero, missing branch, worktree force-fallback, malformed `worktree.yml`, partial-cleanup interrupt + idempotent re-run, and a hand-built `claude_pid == pid` lock to pin the group-kill merge rule.
- `test/integration/drop_command_test.rb` exercises the CLI verb end-to-end against a fixture project, including a `4-execute` task with worktree+branch+log dir, a slug occupying two active stages simultaneously, and the `commit_action` value the envelope returns under `hive/state` lock contention.
- `test/unit/process_kill_test.rb` (new) pins the shared kill module directly: invalid-PID, `EPERM`, `ESRCH`, the empty-`live` short-circuit, and the TERM→KILL escalation arms.
- `test/unit/daemon/dispatcher_test.rb` adds the folder-missing skip case and asserts no spawn / no policy decision when the row's folder has vanished.
- `test/e2e/tui_drop_task_test.rb` drives `Shift+X` against the live TUI by spawning the real `hive drop` subprocess (no in-process stub) and asserts the row disappears within 2s for both a `2-brainstorm` task (no worktree/branch) and a `4-execute` task (full cleanup, including markers-lock removal).
- `test/unit/tui/key_map_test.rb`, `test/unit/tui/bubble_model_test.rb`, and `test/unit/tui/messages_test.rb` cover the per-row dispatch, the empty-grid and archived-row flashes, the left-pane focus guard, the `flash_set_at` update on the "dropping…" flash, and the non-zero `SubprocessExited` arm.
- `test/unit/schema_files_test.rb` cross-checks `schemas/hive-drop.v1.json` against `Hive::Schemas::DropErrorKind` so the closed enum and the documented exit codes can't drift.

## Review summary

Two passes of multi-agent code review (`claude-ce`, `codex-ce`, `pr-review-toolkit`) plus two escalation rounds. Findings landed in three buckets:

- **Auto-fixed in pass 1 and pass 2** (`d4a96200`, `04329cb6`). PID-reuse claude-pid candidate clobbering its own group-kill flag; `ACTIVE_STAGE_DIRS` hardcoding `"9-done"` instead of deriving from `Stages::DIRS.last`; `guard_archived!` returning true on an empty `folders` array; `delete_branch!` regex missing the "checked out at" wording; `integer_like?` accepting whitespace and PID 0; malformed `worktree.yml` aborting the otherwise idempotent cleanup; dispatcher collapsing `folder = nil` rows into the same `folder_missing` signal as genuine drop races; `gh pr close` swallowing non-zero status; agent-writable `worktree.yml` pointer validated through `Hive::Worktree.validate_pointer_path` before any destructive op; `schemas/hive-drop.v1.json` adding exit 75 + `lock_path` / `holder` for `ConcurrentRunError`; `Hive::Worktree.canonical_root` exposed and reused so Drop and the worktree subsystem can't drift; plus a long tail of doc/comment/test enrichment items. See the commit bodies for the full list.
- **Resolved no-fix.** Plan-pinned ordering and envelope shape were the most common reason: U1 step (vii) explicitly accepts that cleanup runs before the audit commit (and surfaces the commit as the one non-best-effort step); U1's envelope shape pins single booleans (`pr_closed`, `worktree_removed`, `agent_killed`) so partial-result fields would be a schema break. Other no-fixes reflect deliberate convention alignment (e.g. `"error"` as the catch-all `error_kind` matches `RunErrorKind` / `StatusErrorKind`) or scope (e.g. consolidating `ProcessKill` into Bot / ChildSupervisor / DiagnosisAgent is follow-up work outside Drop's diff).
- **Escalated, deferred.** Two related items reached escalation rounds 1 and 2 without an answer and are tracked in `docs/solutions/drop-pid-reuse-followup.md` for a follow-up PR:
  1. The `claude_pid` cleanup path has no PID-reuse guard because the lock never records `claude_process_start_time`. Drop currently signals the recorded `claude_pid` without a start-time match. Fixing it requires a lock-schema change (record `claude_process_start_time` when `claude_pid` is captured) or a reader-side live probe; both are out of scope for this PR.
  2. `Hive::ProcessKill.pid_owned_by_recorded_start?` fails OPEN when the live `process_start_time` can't be probed (containerised hosts with `/proc` hidden). The right policy is a cross-module security tradeoff that affects every future caller of the extracted module.

## Linked task

- Slug: `i-need-an-easy-way-260522-568b`
- Task folder (post-finalize): `.hive-state/stages/8-finalize/i-need-an-easy-way-260522-568b/`
- Brainstorm + plan are committed to `~/Dev/hive-private/docs/{brainstorms,plans}/i-need-an-easy-way-260522-568b.md` per the project convention; reviews live alongside the task folder under `reviews/`.

---


ivankuznetsov:

# Review — pass 1 (claude-ce-code-review)

Scope: `git diff origin/main..HEAD` (4 commits, ~1550+ LOC across 30 files)
Plan: `Shift+X` + `hive drop` hard-delete task command (R1–R13, U1–U7).

## High

- [ ] `Hive::Commands::Drop#kill_recorded_agents` sends SIGTERM/SIGKILL to the recorded `claude_pid` **without any PID-reuse guard** (`lib/hive/commands/drop.rb:241-251` hardcodes `process_start_time: nil`). `Hive::ProcessKill.pid_owned_by_recorded_start?` returns `true` whenever `recorded_start_time.to_s.empty?`, so a stale `.lock` whose `claude_pid` has been reused by an unrelated process will cause drop to `kill(-pgid, SIGTERM/SIGKILL)` against that unrelated process group. The `pid` field has a real guard (R-2), but `claude_pid` (written by `lib/hive/agent.rb:123` via `update_task_lock`) does not carry a sibling `claude_pid_start_time`, so the current code has no defense for it. This is the exact `kill someone else's process` risk R-2 was meant to prevent — and the worse one, because `claude_pid` is the one we group-kill.

## Medium

- [ ] `kill_recorded_agents` collapses the process-group kill when `lock["claude_pid"] == lock["pid"]` (`lib/hive/commands/drop.rb:245-258`). Both branches index `by_pid` by the same integer key, and the second assignment unconditionally overwrites the first with `group: false` (the marker branch uses `||=` to avoid this, but the lock branches don't). Result: only a single-process SIGTERM gets sent — orphans any children Claude spawned. Either guard with `by_pid[..] ||= ...` or upgrade the existing entry's `group:` instead of overwriting.
- [ ] Plan R12 / CLI long_desc / wiki promise *cleanup is idempotent — re-running drop after an interrupt completes the remaining steps and exits 0*, but the implementation cannot keep that promise once cleanup has progressed past `remove_task_folders` (`lib/hive/commands/drop.rb:199`). If the process dies after `rm_rf` of stage folders but before `record_drop_commit!`, a second `hive drop <slug>` raises `InvalidTaskPath` (exit 64) because `resolve_context` finds no folder. The promise should either be tightened ("commit step needs operator follow-up via `git -C .hive-state add -A && commit`") or the resolver should treat "no folder present + recent drop in hive/state" as a no-op success.
- [ ] Several plan-required test scenarios are missing from the new test files. The plan's U1 "Test scenarios" lists `gh missing on host → pr_closed: false, stderr warning, exit 0`, `branch missing → branch_deleted: false, exit 0`, and `worktree force-fallback`. None are exercised in `test/unit/commands/drop_test.rb` or `test/integration/drop_command_test.rb`, so the documented best-effort behaviors are unverified — the `--force` retry path in `remove_one_worktree` in particular has no coverage.
- [ ] `Hive::Commands::Drop#worktree_paths` adds the derived worktree path via `Hive::Worktree.new(context.project_root, context.slug).path` (`lib/hive/commands/drop.rb:357`), which calls `Hive::Config.load(@project_root)` and can raise `Hive::ConfigError` if the project's `.hive-state/config.yml` is malformed. That exception escapes `remove_worktrees` and aborts the whole drop with a non-typed crash, even though the rest of the cleanup is best-effort. Either rescue `ConfigError` here (and emit a warn) or precompute the worktree_root once and reuse it.
- [ ] `AlreadyArchived = Class.new(Hive::InvalidTaskPath)` (`lib/hive/commands/drop.rb:25`) silently treats archived tasks as "invalid path" for any caller that catches `Hive::InvalidTaskPath` without knowing the subclass. The `case error when AlreadyArchived` branch in `error_kind_for` only works because the subclass is matched first; any future rescue/case ordering shuffle would silently re-classify archived tasks. Consider making it a sibling of `InvalidTaskPath` rather than a subclass.

## Nit

- [ ] `BubbleModel#drop_focused_task` (`lib/hive/tui/bubble_model.rb` new code) calls `dispatch_command(message)`, which already builds a `flash: "running 'hive drop <slug>'…"`, then immediately overwrites it with `"dropping #{row.slug}..."`. The first flash is wasted work, and the trailing ASCII `...` is inconsistent with `dispatch_command`'s Unicode `…` used everywhere else in the file.
- [ ] `BubbleModel#drop_focused_task` has a defensive `return [flashed("select a task first..."), nil] if row.nil?` even though `KeyMap` already flashes the exact same message before emitting `DropFocusedTask` (`lib/hive/tui/key_map.rb:113`). The branch is unreachable in production code — either delete it or wire `KeyMap` to delegate to BubbleModel for that flash so the strings can't drift.
- [ ] `lib/hive/tui/help.rb` describes `X` with an ASCII hyphen (`... close draft PR - no undo ...`) while every other multi-clause entry in the file uses an em-dash. Minor visual inconsistency in the help overlay.
- [ ] CLI long_desc lists only exit codes `0 / 4 / 64 / 70` (`lib/hive/cli.rb:344-348`), but `schemas/hive-drop.v1.json`'s ErrorPayload allows `[1, 4, 64, 70, 78]` and `error_kind_for` maps `Hive::ConfigError → "config"` (exit 78) and unmatched errors → `"error"` (exit 1). Either document the full set or trim the schema enum to match what the command actually can emit.
- [ ] `Hive::Commands::Drop#remove_task_folders` (`lib/hive/commands/drop.rb:365-368`) walks every `ACTIVE_STAGE_DIRS` entry instead of just the stages in `context.from_stages`. Harmless (`rm_rf` is a no-op on missing dirs), but it makes the audit trail harder to reason about and duplicates work `resolve_*_context` already did.
- [ ] `raise_wrong_stage_if_slug_exists_elsewhere!` (`lib/hive/commands/drop.rb:146-164`) returns early when the slug exists in two projects, so `hive drop <slug> --from <wrong-stage>` against a cross-project collision surfaces as `invalid_task_path` instead of `ambiguous_slug` or `wrong_stage`. Currently no test covers this combination; either document the precedence or surface the ambiguity explicitly.
- [ ] `wiki/commands/drop.md` says the cleanup order is "active stages, then per-slug logs, then audit commit", but the JSON envelope's `from_stages` is **always** sourced from `resolve_*_context`'s scan (path target hits `ACTIVE_STAGE_DIRS`, slug target re-filters to active stages). For a slug that lives in `9-done` only, `guard_archived!` fires before any cleanup — good — but for the path-target case targeting `9-done`, `collect_stage_folders` already excludes it via `ACTIVE_STAGE_DIRS`, then `folders` is empty, then `guard_archived!`'s `all?` predicate fires on an empty array (returns `true` for vacuous truth), raising `already_archived`. That's the right outcome but for a non-obvious reason; a comment would help future readers.

ivankuznetsov:

## High
- [ ] `lib/hive/commands/drop.rb:52` records the audit commit only after deleting the task folders: if `hive drop` is interrupted after `remove_task_folders` at lines 199/364-367 but before log cleanup or `record_drop_commit!`, the retry cannot resolve the slug anymore and cannot complete R12's required cleanup/audit remainder.
- [ ] `lib/hive/commands/drop.rb:245` kills `.lock` `claude_pid` process groups without a start-time guard: `process_start_time` is set to nil, and `Hive::ProcessKill.pid_owned_by_recorded_start?` treats blank recorded times as trusted, so a stale/reused `claude_pid` can receive TERM/KILL for its whole process group.

## Medium
- [ ] `lib/hive/commands/drop.rb:100` restricts `folders` to `--from` before cleanup discovery: TUI always passes `--from`, so a multi-stage slug only has locks, `pr.md`, and `worktree.yml` read from the focused stage while `remove_task_folders` deletes every active-stage folder, leaving sibling-stage agents/PRs/worktrees unhandled and under-reporting `from_stages`.
- [ ] `lib/hive/commands/drop.rb:349` trusts `worktree.yml` paths without `Hive::Worktree.validate_pointer_path`: a corrupted or agent-edited pointer to another registered worktree for the same project can make drop remove that unrelated worktree instead of only the per-task worktree.

## Nit

ivankuznetsov:

## High
- [ ] [silent-failure] `remove_one_worktree` reports `worktree_removed: false` on a successful removal — `lib/hive/commands/drop.rb:331-347`: the guard `!File.directory?(path) && e.message.start_with?("git worktree remove failed")` is effectively `!File.directory?(path)` (only `WorktreeError`'s `raise` site uses that prefix), so an already-pruned worktree returns false when it is in fact gone; operator sees the envelope claim removal failed when it did not.
- [ ] [silent-failure] `process_candidates` silently downgrades a claude group-kill to a single-pid kill — `lib/hive/commands/drop.rb:252-258`: the `lock["pid"]` branch uses bare `=` (not `||=`), so when `lock["pid"] == lock["claude_pid"]` the `group: true` entry is overwritten with `group: false`; children of the agent survive `kill(TERM, pid)` while envelope reports `agent_killed: true`.
- [ ] [silent-failure] `close_draft_prs` collapses "no PR" and "tried-and-failed" into `pr_closed: false` — `lib/hive/commands/drop.rb:297-316`: a non-zero `gh pr close` or `Hive::GhError` only warns to stderr (lost under `--json` capture); envelope cannot distinguish "no PR existed" from "draft PR is still open"; add a `pr_close_skipped_reason` field paralleling `agent_kill_skipped_reason`.
- [ ] [silent-failure] `delete_branch!` conflates "branch never existed" and "branch deletion recovered from error" into `branch_deleted: false` — `lib/hive/commands/drop.rb:360-362` + `lib/hive/git_ops.rb:186-194`: the regex-swallowed not-found path returns the same boolean as a benign no-op; add a tri-state return + `branch_delete_skipped_reason` so operators can distinguish; also pin `LC_ALL=C` so non-English git locales don't escalate to `GitError`.
- [ ] [code] CLI `drop` long_desc and wiki exit-code table omit code 78 (CONFIG) — `lib/hive/cli.rb:344-348` + `wiki/commands/drop.md:24-30`: `Drop#error_kind_for` maps `Hive::ConfigError → DropErrorKind::CONFIG` (exit 78) and the published schema enum includes 78, but the docs advertise only 0/4/64/70; agents reading the doc will mis-handle a config-error envelope.
- [ ] [tests] Idempotence (R12 / AC6) — no test exists for "re-run drop after partial cleanup converges and exits 0"; the `delete_branch!` / `remove_task_folders` / `remove_one_worktree` already-gone branches in `lib/hive/commands/drop.rb:191-200,331-347` are entirely uncovered.
- [ ] [tests] Worktree force-fallback path is uncovered — `lib/hive/commands/drop.rb:331-347` has a three-state `remove! → remove_force! → swallow` flow but no test arranges a dirty worktree to exercise the force retry; the plan explicitly lists this scenario.
- [ ] [tests] `gh` missing / `Hive::GhError` rescue path is uncovered — the only PR test stubs `Hive::Gh.capture3` to return success; the failure branch at `lib/hive/commands/drop.rb:312-314` would crash silently on any host without `gh` on PATH if the rescue were broken.
- [ ] [tests] Branch-already-absent case is not explicitly tested — `lib/hive/git_ops.rb:186-194`'s "branch not found" regex recovery has no dedicated test; only happy-path branch deletion is covered, leaving locale/regex regressions invisible.
- [ ] [comments] CLI long_desc / class doc / CHANGELOG / wiki describe drop as closing a "draft PR" — `lib/hive/commands/drop.rb:17-19`, `lib/hive/cli.rb:331`, `CHANGELOG.md:11`, `wiki/commands/drop.md:10`: the implementation calls `gh pr close <url>` against any recorded `pr_url`, draft or not; an operator dropping a 6-review task with an open non-draft PR will silently lose work.
- [ ] [comments] CHANGELOG breaking change is not under the `### Breaking changes` section — `CHANGELOG.md:9-12`: the X-rebind is announced inline under `### Changed — task drop` but the file already maintains a `### Breaking changes` section at L60; release-notes tooling keyed on section headers will miss it.

## Medium
- [ ] [silent-failure] `worktree_paths` swallows `Hive::WorktreeError` from a malformed pointer file — `lib/hive/commands/drop.rb:349-358`: a corrupted `worktree.yml` silently downgrades to the canonical-path fallback, so a non-standard layout pointed to by the file is missed and the operator gets a "success" envelope.
- [ ] [silent-failure] `lock_payload` and `marker_for` rescue corruption and yield empty state without an envelope signal — `lib/hive/commands/drop.rb:273-291`: a torn-read `.lock` mid-write looks identical to "no agent recorded"; envelope reports `agent_kill_skipped_reason: "no_pid"` when the truthful answer is "could not parse state files"; distinguish `lock_unreadable` / `marker_unreadable`.
- [ ] [silent-failure] `process_candidates` marker-pid branch nullifies the PID-reuse guard when `lock["pid"] != marker_pid` — `lib/hive/commands/drop.rb:264-268`: empty `process_start_time` in `pid_owned_by_recorded_start?` returns true (guard disabled); under PID reuse the wrong process is killed and the envelope claims success. Either capture live start-time via `Hive::Lock.process_start_time(marker_pid)` or emit `skipped_reason: "marker_pid_no_start_time_guard"`.
- [ ] [silent-failure] EPERM during `safe_kill` is not propagated to `terminate_process`'s skipped_reason — `lib/hive/process_kill.rb:54-55,82-86`: when `Process.kill` raises EPERM, `safe_kill` swallows it; the outer rescue at line 54 only catches EPERM from `Integer(pid)`, so a permission-denied signal returns `killed: false, skipped_reason: nil` instead of `"permission_denied"`; envelope loses the actionable reason.
- [ ] [silent-failure] Dispatcher `folder_missing` skip has no rate-limit, escalation, or metric — `lib/hive/daemon/dispatcher.rb:354-360`: a stuck snapshot that keeps yielding a vanished slug logs the same skip on every tick with no fatal escalation; mirror the `@legacy_layout_logged` first-sight-only pattern already in the dispatcher or add an N-tick fatal.
- [ ] [tests] No direct unit tests exist for `Hive::ProcessKill` — `grep terminate_process test/` returns zero matches; TERM→KILL escalation, group kill, PID-reuse guard branches, and EPERM/ESRCH rescues are only indirectly exercised via one drop happy-path test.
- [ ] [tests] `test_drop_pid_reuse_guard_refuses_to_signal_recorded_pid` writes `Process.pid` (the test runner) into the lock — `test/unit/commands/drop_test.rb:160-177`: if a regression weakened the guard, `terminate_process` would SIGTERM minitest itself; the build would crash without the assertion firing. Spawn a real `sleep` child with a wrong-start-time lock and assert it survives.
- [ ] [tests] E2E test stubs `Hive::Tui::Subprocess.dispatch_background` — `test/e2e/tui_drop_task_test.rb:80-88`: the keystroke→Thor→bin/hive pipeline is never exercised end-to-end; the `--from row.stage` literal vs Thor's `APPROVE_TO_ENUM` value match is unverified; per CLAUDE.md "use real APIs" the dispatch should fork a real subprocess.
- [ ] [tests] Schema success-payload test only validates one shape — `test/unit/schema_files_test.rb:1040-1060`: only the `agent_killed: false, killed_pids: [], pr_closed: false` shape is round-tripped; killed-agent, closed-PR, and multi-stage payloads are not validated against the JSON schema; parametrize over at least three shapes.
- [ ] [tests] `test_drop_removes_execute_task_worktree_branch_logs_and_agent` may be flaky on slow CI — `test/unit/commands/drop_test.rb:43-79`: relies on `sleep` honoring SIGTERM within the 2s `TERM_GRACE_SECONDS` window. Assert weaker (`Process.kill(0, pid)` raises ESRCH) or inject a shorter grace via a ProcessKill seam.
- [ ] [tests] Archived-guard test does not assert "no side effects" — `test/integration/drop_command_test.rb:64-76` checks the folder survives but not that `hive/state` HEAD is unchanged; add `assert_equal head_before, head_after` to lock in the contract.
- [ ] [comments] `lib/hive/process_kill.rb` has no module-level documentation — every other shared helper module in the codebase carries a rationale paragraph; future maintainers won't know why the guard fails open on empty start time, why TERM→KILL uses two grace windows, or why `pid_alive?` returns true on EPERM.
- [ ] [comments] Daemon dispatcher's new folder-existence guard has no rationale comment — `lib/hive/daemon/dispatcher.rb:354-360`: a maintainer reading this in isolation may think the re-stat is paranoid and remove it, re-opening the snapshot-vs-drop TOCTOU window the guard closes.
- [ ] [comments] CLI `drop` long_desc topic shift is jarring — `lib/hive/cli.rb:336-342`: "already at 9-done refused with exit 64" lands inside the "drop is irreversible" paragraph rather than the exit-codes block; cosmetic but harms agent parseability.

## Nit
- [ ] [code,silent-failure] `delete_branch!` regex contains an unreachable alternative — `lib/hive/git_ops.rb:191`: `/branch .* not found|not a valid branch name|branch .+ not found/i` — the third branch is fully subsumed by the first; collapse to two alternatives.
- [ ] [comments] `schemas/hive-drop.v1.json:68-71` — `commit_action` is always the literal string `"dropped"`; pin via `{ "const": "dropped" }` or update the description to say so.
- [ ] [comments] `schemas/hive-drop.v1.json:64-67` — `agent_kill_skipped_reason` description names only three values but `lib/hive/process_kill.rb` emits five (`pid_reuse_guard`, `not_alive`, `no_pid`, `permission_denied`, `invalid_pid`); either enumerate or list all five in the description.
- [ ] [comments] `wiki/commands/drop.md:46` should note that `stage_name: "dropped"` in the audit commit is a synthetic label — no on-disk `dropped` stage exists, which will confuse a future maintainer doing `git log --grep="dropped/"` archaeology on `.hive-state`.
- [ ] [comments] `lib/hive/tui/help.rb:47` description for `X` is ~190 chars vs the table's typical <80; word-wrap may break the overlay layout. Shorten and fix the "draft PR" inaccuracy at the same time.
- [ ] [comments] `lib/hive/tui/messages.rb:344-347` — `DropFocusedTask` comment lost the four-gate documentation (right-pane focus, non-nil row, non-9-done stage, modifier) that the old `DropScopedProjectIfMissing` carried; add a one-liner pointing at KeyMap's gating.
- [ ] [tests] Test `test_drop_closes_pr_url_from_pr_md_best_effort` uses a hand-rolled singleton-method stub — `test/unit/commands/drop_test.rb:202-208`: a raise between `define_singleton_method` and the `ensure` could leak the stub into the next test; minor brittleness.
- [ ] [tests] TUI drop tests assert exact flash text (`"dropping drop-me..."`) — `test/unit/tui/bubble_model_test.rb`: exact-string matches force test churn on every copy edit; prefer `assert_match(/dropping drop-me/, ...)`.
- [ ] [tests] No test for "task folder present but state file misnamed" path — `lib/hive/commands/drop.rb:282-291`'s `state_file_for_folder` rescue is not exercised; cheap unit test would lock in "drop never crashes on broken task folders".
- [ ] [silent-failure] Stderr-only warnings from `close_draft_prs` are invisible to `--json` consumers — `lib/hive/commands/drop.rb:310,313`; covered by the per-step skipped-reason recommendations above.

ivankuznetsov:

## High
- [ ] `lib/hive/commands/drop.rb:268-275` — `claude_pid` PID-reuse guard is structurally bypassed. `process_candidates` registers the `claude_pid` candidate with `process_start_time: nil`, and `Hive::ProcessKill.pid_owned_by_recorded_start?` (lib/hive/process_kill.rb:29) treats a blank recorded start time as "owned" (`return true`). The result: when `claude_pid != lock["pid"]` the kill proceeds without any reuse check, so a recycled `claude_pid` will receive `kill(-pgid, TERM/KILL)` against an unrelated process group. The orchestrator `pid` branch is correctly guarded because `process_start_time` is recorded in `.lock`; the `claude_pid` branch has no equivalent writer. Plan R-2 mitigation explicitly requires the guard for any signalled PID. Escalations Q1 (pass-1) still has an empty `### A1.` — the gap remains unresolved in this diff.
- [ ] `lib/hive/process_kill.rb:28-35` — `pid_owned_by_recorded_start?` fails OPEN when the live start time cannot be probed: `return true if live.to_s.empty?` lets the kill proceed even though the plan's R-2 says to "refuse to signal when the recorded process_start_time differs from the live one". On a container/`/proc`-restricted host the guard silently degrades from "checked" to "skipped" with no `agent_kill_skipped_reason` differentiator (callers see the same envelope as a successful guard). Escalations Q2 (pass-1) has an empty `### A2.` — still unanswered, still in the diff.

## Medium
- [ ] `lib/hive/commands/drop.rb:382-396` — malformed `worktree.yml` can crash drop. `Hive::Worktree.read_pointer` calls `YAML.safe_load(File.read(pointer))` with no rescue for parse errors; the surrounding `rescue Hive::WorktreeError` block at drop.rb:394 only catches the "must be a hash" branch. A truncated or hand-edited `worktree.yml` propagates `Psych::SyntaxError` up and out of the otherwise best-effort cleanup, leaving folders/branches/logs in place against the plan's "cleanup is idempotent / converges" contract (R12). Either rescue `Psych::Exception, SystemCallError, IOError` here, or harden `Worktree.read_pointer` to return nil on parse failure (mirroring `Drop#lock_payload`'s rescue list).
- [ ] `lib/hive/daemon/dispatcher.rb:355` — `row.folder.to_s` swallows nil-folder rows into the same `:skipped reason: "folder_missing"` event used for genuinely-vanished folders. A `Row` with `folder: nil` (status-bug or malformed snapshot) is now indistinguishable from a post-drop race in the logs. Cheap fix: bail with a distinct `reason: "folder_missing_nil"` (or a separate guard) so an operator scanning daemon logs after a stuck dispatch can tell "drop just ran" apart from "the snapshot is malformed". Plan U5 only asked for the race guard; the silent nil-collapse is collateral.

## Nit
- [ ] `test/e2e/tui_drop_task_test.rb:176-177` — `original_home = ENV["HIVE_HOME"]` followed by `ENV["HIVE_HOME"] = ENV["HIVE_HOME"]` is a no-op write whose intent (preserve the value across the test) is unclear from reading. Either delete both lines (the rescue restores it anyway only when non-nil) or write an explicit comment explaining why the dance is needed. Mirror in pattern: the surrounding `BUNDLE_GEMFILE` save/restore *does* mutate the variable.
- [ ] `lib/hive/commands/drop.rb:160-166` — `collect_stage_folders` runs `File.directory?(folder)` then `File.realpath(folder)` with no rescue in between. A concurrent unlink (another drop, an operator `rm -rf`) lands as `Errno::ENOENT` out of realpath rather than the same "treat as gone" semantics the rest of the file follows. `rescue Errno::ENOENT` and `next` would collapse the race into the existing idempotent path.
- [ ] `lib/hive/commands/drop.rb:354-360` — `remove_worktrees` constructs `Hive::Worktree.new(context.project_root, context.slug)` without passing `worktree_root:`, while the sibling `worktree_paths` does pass it. The bare constructor re-resolves `worktree_root` via `Hive::Config.load(project_root)` inside `worktree_root`, which is harmless but adds a redundant disk read per drop and is asymmetric with `worktree_paths`. Plumb `canonical_worktree_root(context)` through both constructors or memoize on the context.
- [ ] `lib/hive/commands/drop.rb:364-366` — `remove_one_worktree` calls `wt.list_worktree_paths` (one `git worktree list --porcelain` subprocess) inside the loop body before each remove. Cheap, but if a slug ever has >1 worktree path candidate the cost grows linearly. Hoist the list once per drop or pass it in alongside `wt`.
- [ ] `wiki/commands/drop.md:38` — "Tasks in the last stage (`9-done`) are archive records, not active work" duplicates information already in the refusals table two lines above. Consider folding the prose into the table caption to keep the page tight; otherwise this is a Nit-of-a-Nit.
- [ ] `lib/hive/commands/drop.rb:419-434` — `record_drop_commit!` returns `result.to_s` which is `"committed"` for the normal path and `"nothing_to_commit"` only when `allow_empty: false` short-circuits — but the call sets `allow_empty: true`, so `nothing_to_commit` is unreachable here. The envelope's `commit_action` field therefore only ever carries `"committed"`, making the field shape (string) heavier than it needs to be. Either drop the to_s coercion (return the symbol) or document the closed set in `schemas/hive-drop.v1.json` so consumers know it is single-valued in practice.

ivankuznetsov:

## High
- [ ] `lib/hive/process_kill.rb:37` accepts PID `0` from drop cleanup: `Drop#integer_like?` allows `0`, so a malformed `.lock` or marker can make `Process.kill("TERM", 0)` or a group kill target the current process group instead of the task agent.
- [ ] `lib/hive/commands/drop.rb:268` kills `claude_pid` without a start-time guard: the agent only records `claude_pid`, and Drop passes `process_start_time: nil`, so stale locks can signal a reused PID/process group despite the plan's PID-reuse safety requirement.

## Medium
- [ ] `lib/hive/commands/drop.rb:382` lets malformed `worktree.yml` abort drop: `Worktree.read_pointer` YAML parse errors are not rescued, so corrupted pointer metadata can stop otherwise idempotent cleanup before folder/log removal.
- [ ] `lib/hive/daemon/dispatcher.rb:355` treats missing `row.folder` as the drop-race case: `row.folder.to_s` turns nil into `""`, emitting `folder_missing` and skipping dispatch for malformed snapshots instead of surfacing the bad row shape.

## Nit

ivankuznetsov:

## High

- [ ] [pr-test-analyzer] `lib/hive/process_kill.rb` has zero direct unit tests: 106 lines of security-sensitive PID kill/reuse logic only get incidental coverage via Drop; EPERM/ESRCH/invalid_pid arms (lines 52-79) untested — regressions in guard arms would let drop signal foreign PIDs silently.
- [ ] [pr-test-analyzer] No test exercises the marker-only PID path in `drop.rb:264-288` (`<!-- AGENT_WORKING pid=N -->` with no `.lock` file): a regression that broke `marker_for` would leave an agent alive while drop reports `ok=true`.
- [ ] [pr-test-analyzer] "Branch checked out elsewhere / used by worktree" benign-skip path at `lib/hive/git_ops.rb:198-201` is uncovered: only the "branch not found" arm gets hit via the idempotence test; a regex regression against modern git's "used by worktree at" would convert benign skip into a fatal raise mid-drop.
- [ ] [pr-test-analyzer] `gh pr close` returning non-zero status branch at `lib/hive/commands/drop.rb:340-344` is uncovered: only the `GhError` raise path (binary missing) and the success path are tested; a regression swallowing the warn would silently leave PRs unclosed without surface.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:330-349` `close_draft_prs` returns one `closed_any` boolean: a 1-of-N URL failure looks identical to full success in the JSON envelope; agent consumers can't detect partial close after auth expiry / 5xx / rate-limit on any single URL.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:345-347` `rescue Hive::GhError` masks both binary-missing and real network/auth/timeout failures — `Hive::Gh.capture3` raises GhError for SystemCallError AND wall-clock timeout; a 60s hang gets the same "skipped" treatment as missing CLI with no envelope signal.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:306-313` `lock_payload` rescues `SystemCallError` (superclass of every `Errno::*`): `EACCES`/`EISDIR`/`ENAMETOOLONG`/corrupt YAML all silently degrade to "no PID recorded" while `agent_kill_skipped_reason` reports `"no_pid"` — the agent keeps running while drop reports success.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:315-320` `marker_for` rescues `Hive::InvalidTaskPath, SystemCallError, IOError`: a 4-execute task that lost its task.md or a renamed-stage drift returns `name: :none`, the marker-derived PID candidate is skipped, and the agent may not be killed — no log, no envelope flag.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:382-396` `worktree_paths` silently treats out-of-root pointer escapes as "no worktree": a pointer escape is a security-relevant signal reduced to `worktree_removed: false` with zero log of the rejected path.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:373-377` `remove_one_worktree` returns false silently if `force_error.message` doesn't start with the exact prefix: a future git rewording or a different WorktreeError class would leave the worktree behind with the envelope showing `worktree_removed: false` indistinguishable from "already gone".
- [ ] [silent-failure-hunter] `lib/hive/git_ops.rb:198-201` `delete_branch!` regex-matches English stderr to decide "benign-skip vs real failure" — locale-dependent (`LC_MESSAGES`), version-dependent, and silently drops protected-branch / hook-refusal / half-rebase cases as `branch_deleted: false` with no warning.
- [ ] [comment-analyzer] `lib/hive/commands/drop.rb:23-31` load-bearing claim "the schema test will catch the drift" has no actual test reference; if `schema_files_test.rb` is renamed or the assertion changes, this comment misleads — replace with file + assertion, or drop the second sentence.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:35-38` `TaskContext` cannot defend its core invariant (`from_stages == folders.map { |f| f[:stage] }`): both fields are independent `Struct.new(..., keyword_init: true)` slots, defaulting to nil, accepting any types — `TaskContext.new(slug: "x")` constructs silently with five nils.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:235,256-261` `kill_recorded_agents` emits inconsistent shape: `agent_pid: candidates.first[:pid]` is set even when nothing was killed; readers can't tell whether the field means "PID we killed" or "PID we considered" — model as `KillResult = Data.define(:status, :pid, :killed_pids, :skipped_reason)` with `status ∈ {:killed, :skipped, :no_pid}`.
- [ ] [type-design-analyzer] `lib/hive/process_kill.rb:51` `Result.skipped_reason` is nil on a non-killed result when TERM+KILL grace expires without ESRCH/EPERM: `killed=false` + `skipped_reason=nil` cannot be distinguished from "haven't run yet" — schema advertises a closed enum but a value path produces nil meaning "we tried and failed".

## Medium

- [ ] [code-reviewer] `lib/hive/commands/drop.rb:330-349` `close_draft_prs` (and `cli.rb:328` long_desc) advertise "close draft PR" only but `gh pr close` is unconditional on the `pr_url` in pr.md frontmatter — a non-draft PR from past `5-open-pr` will be silently closed; either filter on `gh pr view --json isDraft` or rename to "close associated PR".
- [ ] [code-reviewer] `lib/hive/commands/drop.rb:266-288` `process_candidates` merges stale agent metadata from leftover older-stage folders: a `.lock` in `2-brainstorm` with a recycled `claude_pid` (no `claude_process_start_time` recorded) bypasses the pid-reuse guard via the unconditional empty-start-time path → group-kill via `kill -<pgid>` to an unrelated process group.
- [ ] [pr-test-analyzer] Cross-project wrong-stage path at `drop.rb:179` (slug exists in 2 projects, `--from` mismatches both) is unexercised: contract falls through to InvalidTaskPath instead of WrongStage; no test pins this.
- [ ] [pr-test-analyzer] Worktree pointer validation reject branch at `drop.rb:393-395` (Worktree.validate_pointer_path raises WorktreeError) is uncovered: plan calls this out as a security guard ("never trust the agent-writable pointer file") — needs explicit fixture.
- [ ] [pr-test-analyzer] Same-slug spanning active + 9-done simultaneously is untested at `drop.rb:146-149`: the asymmetric silent drop of the archive entry could be inverted by regression so drop accidentally removes the 9-done copy.
- [ ] [pr-test-analyzer] TUI E2E doesn't test row advancing between snapshot and dispatch (WRONG_STAGE arm): operator sees a generic `SubprocessExited` flash with no row-level feedback when the row moved.
- [ ] [pr-test-analyzer] Integration test (`drop_command_test.rb`) lacks a 4-execute happy-path case with worktree+branch: all 3 ints use 2-brainstorm or 9-done; full chain (worktree-remove subprocess, branch-delete subprocess, hive/state commit) never validated end-to-end through `bin/hive`.
- [ ] [pr-test-analyzer] `record_drop_commit!` failure (commit-lock contention, `Hive::Lock` timeout) is untested at `drop.rb:419-434`: documented exit code 75 has no test asserting the contract; regression removing/rewrapping the lock would break the documented exit code.
- [ ] [silent-failure-hunter] `lib/hive/process_kill.rb:13-20` `pid_alive?` returns `false` on `ArgumentError, TypeError`: malformed PID silently treated as "not alive" → skip reason `"not_alive"` rather than `"invalid_pid"`; blends two distinct failure modes — let those propagate so `terminate_process` reports `invalid_pid`.
- [ ] [silent-failure-hunter] `lib/hive/process_kill.rb:99-104` `reap_if_child_exited` rescues only `Errno::ECHILD`: an EINTR mid-syscall propagates out and aborts the entire terminate_process; drop bubbles up `InternalError` instead of a clean kill result.
- [ ] [silent-failure-hunter] `lib/hive/process_kill.rb:82-86` `safe_kill` silently swallows `Errno::EPERM`: kernel refusal (setuid binary / UID change after PID reuse) → SIGKILL → still alive → result reports `killed: false, skipped_reason: nil` with no `"permission_denied_signal"` distinction.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:255-261` `skipped_reason` only surfaces the FIRST reason when multiple candidate PIDs exist (claude_pid + pid differ): users debugging "why didn't drop kill all three?" get one clue out of three — schema should expose `agent_kill_skipped_reasons: []`.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:351-362` `remove_worktrees` collapses N-worktree result into one boolean: 1-of-N force-remove failure looks like full success in `worktree_removed: true`.
- [ ] [silent-failure-hunter] `lib/hive/daemon/dispatcher.rb:355-360` `:skipped` event is now overloaded for 4+ reasons (folder_missing / baseline_recorded / plan_approval_invalid / plain skip): operators must grep specifically; also verify status snapshot stops emitting the row post-folder-removal to avoid daemon.log spam each tick.
- [ ] [silent-failure-hunter] `lib/hive/git_ops.rb:206-209` `prune_worktrees!` raises GitError on any non-zero exit: invoked AFTER per-worktree remove — a prune failure aborts the whole drop, no envelope emit, user sees `Hive::Error` exit code, half-cleaned state with no signal that prune is the failing step.
- [ ] [silent-failure-hunter] `lib/hive/git_ops.rb:186-204` `delete_branch!` discards the absorbed stderr when the regex matches: no breadcrumb for "why is `branch_deleted: false`?" postmortems — add a `warn "[hive] drop: branch -D #{name} treated as no-op: #{combined.strip}"`.
- [ ] [comment-analyzer] `lib/hive/process_kill.rb` has zero comments despite subtle work (PID-reuse guards, group-vs-process kill, TERM→KILL escalation, WNOHANG reaping): `pid_owned_by_recorded_start?` in particular deserves a one-line WHY about Linux PID reuse.
- [ ] [comment-analyzer] `lib/hive/daemon/dispatcher.rb:352-358` new folder-missing guard has no comment: rest of file is heavily commented; one line explaining "task folder may have vanished between status snapshot and dispatch (concurrent `hive drop` or `hive forget`)" would match tone.
- [ ] [comment-analyzer] `lib/hive/cli.rb:331-352` `long_desc`'s "Cleanup is idempotent — re-running drop after an interrupt completes the remaining steps" is load-bearing for operators but only sometimes tested; worth a dedicated test that asserts re-running drop on a partially-cleaned task returns ok=true.
- [ ] [comment-analyzer] `lib/hive/commands/drop.rb:290-294` `register_candidate` doc comment over-documents merge mechanics; only the parenthetical "claude_pid==pid lock would clobber group-kill" is non-obvious — trim to that one WHY.
- [ ] [comment-analyzer] `lib/hive/commands/drop.rb:387-392` "Mirror Stages::Execute/Review/Diagnosis" reference will rot if any caller is renamed; consider moving the rationale onto `Worktree.validate_pointer_path` and shortening this site to a cross-ref.
- [ ] [type-design-analyzer] `lib/hive/process_kill.rb` is presented as "extracted from Bot" but `bot/child_supervisor.rb:266-270`, `diagnosis_agent.rb:528`, `tui/subprocess.rb:640` keep their own duplicates — the new module has exactly one caller (Drop); either consolidate the four sites or rename to its actual role.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:402-407` reimplements the canonical `worktree_root` resolution duplicating `lib/hive/worktree.rb:19-25`: will drift the day someone adds a fallback — expose `Worktree.canonical_root(project_root)` and call from both sites.
- [ ] [type-design-analyzer] `lib/hive/git_ops.rb:186-204` `delete_branch!` benign-failure detection is a free-form regex over combined stdout+stderr: locale- and version-dependent; lock `LC_ALL=C` for the capture path OR probe via `git rev-parse --verify` + `git worktree list --porcelain` first.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:64` `do_call` is `public` (the `private` boundary appears at line 80 below it): leaks the envelope/no-envelope split into the public API — rename `_do_call` with `protected`, or fold into a private method invoked via `send`.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:330-349` `close_draft_prs` swallows `Hive::GhError` per-URL but cannot surface partial close: `pr_closed` is `true` if any URL closed and `false` otherwise — for a destructive idempotent op consumed by agents the partial-success signal matters.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:89-91` `path_target?` discriminator duplicates path-or-slug logic that `Hive::TaskResolver` also implements: align via `TaskResolver.path_like?` (or expose one) to prevent future drift.
- [ ] [code-reviewer] `lib/hive/commands/drop.rb` cleanup chain is not atomic against `ConcurrentRunError` from `with_commit_lock` at the end (line 415, `record_drop_commit!`): on lock contention the destructive cleanup is done but the audit commit is skipped → exit 75; consider wrapping the lock around the whole sequence.

## Nit

- [ ] [code-reviewer] `lib/hive/commands/drop.rb:52-76` `do_call` should sit under the `private` declaration on line 79 — consistent with sibling command classes that keep entry-points minimal.
- [ ] [code-reviewer] `lib/hive/git_ops.rb:184-202` `delete_branch!` regex anchors on lowercase `cannot` with `/i` but a future git rewording (`is being used by`, `is checked out by worktree`) would not match — broaden to `(checked out|used by|in use)`.
- [ ] [code-reviewer] `lib/hive/commands/drop.rb:330-349` per-URL `rescue Hive::GhError` continues but `warn` line is misleading when followed by a successful URL — cosmetic.
- [ ] [pr-test-analyzer] `test/e2e/tui_drop_task_test.rb:184-188` violates project CLAUDE.md E2E rule against `sleep` ("NEVER use sleep or hard-coded timeouts") with `sleep 0.05` poll under a 5s deadline — refactor to an explicit-condition wait.
- [ ] [pr-test-analyzer] `test/unit/tui/bubble_model_test.rb:5223-5246` over-stubs `dispatch_background`; doesn't verify `flash_set_at` was updated (the visible-timeout for the "dropping…" flash) — regression leaving flash_set_at nil would freeze the flash forever.
- [ ] [pr-test-analyzer] `test/e2e/tui_drop_task_test.rb:107-124` plants a `.markers-lock` but only asserts the folder is gone, not the markers-lock cleanup specifically.
- [ ] [pr-test-analyzer] `test/unit/commands/drop_test.rb:200-217` PID-reuse guard test uses `Process.pid` and is indirect: a future change to `process_start_time` returning nil/empty silently degrades the guard to no-op (`pid_owned_by_recorded_start?` short-circuits on empty `live`); test wouldn't catch that regression.
- [ ] [pr-test-analyzer] `test/unit/tui/key_map_test.rb:848-886` doesn't pin the `pane_focus: :left` + `key: "X"` + nil row combo — falls through silently to the focus guard, undocumented.
- [ ] [pr-test-analyzer] `test/unit/daemon/dispatcher_test.rb:171-187` folder-missing test could additionally assert `sup.spawned.size == 0` AND that Policy.decide didn't run, for better discriminative power.
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:307` `lock_payload` returns `{}` for both "no lock file" and "unreadable lock file" — collapses two distinct conditions, precludes a future "warn if lock existed but was unreadable" diagnostic.
- [ ] [silent-failure-hunter] `lib/hive/process_kill.rb:22-26` `process_start_time` rescues silently; combined with empty-string check, a malformed PID effectively bypasses the pid-reuse guard — worth a defence-in-depth comment.
- [ ] [silent-failure-hunter] `lib/hive/git_ops.rb:326-328` (pre-existing) `fetch_default_branch` `rescue StandardError` returns false with no log; a stderr warn would help debug "why didn't my worktree pick up origin/main?".
- [ ] [silent-failure-hunter] `lib/hive/git_ops.rb:509-511` `run_git_with_timeout` tail rescue reports `Process.spawn` ENOENT (no git binary) as "rebase failed with stderr <msg>" — distinguish "couldn't even start git".
- [ ] [silent-failure-hunter] `lib/hive/commands/drop.rb:339` `gh pr close ... --comment "task dropped"` is generic; include the slug for reviewers to link back to the task.
- [ ] [silent-failure-hunter] `lib/hive/tui/bubble_model.rb:614-620` `drop_focused_task` flashes "dropping…" but verify `dispatch_background` surfaces a non-zero exit back to the model so the TUI doesn't claim success on failure (check Hive::Tui::Subprocess.dispatch_background).
- [ ] [comment-analyzer] `lib/hive/commands/drop.rb:86-88` Slugs comment duplicates `Stages::SLUG_RE`; shorten to "Slugs match Stages::SLUG_RE — any /, ~, or . means a path. Windows \\ unsupported".
- [ ] [comment-analyzer] `lib/hive/commands/drop.rb:40-43` YARD-style param doc violates project "default to writing no comments"; only `@from` has hidden semantics — trim to a one-liner.
- [ ] [comment-analyzer] `lib/hive/worktree.rb:62-70` no comment on `remove_force!` explaining it exists for drop's "worktree-locked / index-missing / dirty-tree" fallback — readers must dig into `drop.rb#remove_one_worktree`.
- [ ] [comment-analyzer] `lib/hive/git_ops.rb:162-165` (pre-existing) `hive_commit` comment says "scoped add: only stage files under stages/<stage_name>/<slug>/" but with `pathspecs:` the method now stages arbitrary pathspecs — update.
- [ ] [type-design-analyzer] `lib/hive/tui/messages.rb:349` `DropFocusedTask` doesn't validate `row` like sibling `NewIdeaImageAttached` (lines 304-312); either trust the KeyMap guard and remove the defensive nil-check in `bubble_model.rb:609`, or add a constructor raise.
- [ ] [type-design-analyzer] `lib/hive/process_kill.rb:9` `Result` is a `Struct` with `keyword_init: true` but the TUI uses `Data.define` for immutable records — `Data.define(:pid, :killed, :skipped_reason)` gives free immutability + equality.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:290-304` `register_candidate`'s `group: true` precedence is documented but not type-enforced; a `Candidate = Data.define(:pid, :process_start_time, :group)` with `merge_with(other)` makes the rule executable.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:26` `ARCHIVE_STAGE = Hive::Stages::DIRS.last` couples behaviour to array order; adding a post-archive sentinel silently rebinds it — prefer a named `Stages::ARCHIVE` constant.
- [ ] [type-design-analyzer] `schemas/hive-drop.v1.json:38-50` `from_stages` enum is a parallel source of truth for `ACTIVE_STAGE_DIRS`; the schema test catches drift only when run — consider generating the enum from `Hive::Stages::DIRS` at build time.
- [ ] [type-design-analyzer] `lib/hive/worktree.rb:72-74` `remove_force!` is sugar duplicating `remove!(force: true)` — pick one; if `remove_force!` is the intended surface, call git `--force` directly.
- [ ] [type-design-analyzer] `lib/hive/commands/drop.rb:431` + `lib/hive/git_ops.rb:165-184` `record_drop_commit!` always passes `allow_empty: true` — a one-caller flag widening `hive_commit`'s signature; fold into the caller unless another site materialises.

PR #126, opened 2026-05-22
feat(events): per-task event log foundation

ivankuznetsov:
# feat(events): per-task event log foundation

## Summary

Stage runners and agent spawns now emit structured lifecycle events to a per-task `events.jsonl` log, with a derived `status.md` snapshot rewritten after each event. Operators get a machine-readable, append-only record of every stage enter/exit, every agent start/end (with profile + log label + exit status), brainstorm and plan round transitions, and error landings — answering "which sub-agent is review running right now?" without parsing free-form log files.

The TUI drill-down (plan U5) is deferred: the plan reserved keystroke `s`, but `s` is already bound to `Messages::OpenInAgent` (manual steering) in `lib/hive/tui/key_map.rb`. Rather than silently rebind a documented user-facing keystroke, execute stopped at U5 and shipped U1–U4 so the backing data is captured today and a future PR can pick a free key (`D`, `T`, or grid `Enter`) and add the view.

What landed:

| Unit | Scope | Status |
|------|-------|--------|
| U1 | `Hive::Events` emitter — append-only `events.jsonl` + atomic `status.md` rewrite | landed |
| U2 | `Stages::Base.with_stage_events` — `stage_enter` / `stage_exit` / `error`, plus brainstorm + plan `round_waiting` / `round_complete` and per-phase review brackets | landed |
| U3 | `Hive::Agent#run!` brackets every spawn with `agent_start` / `agent_end` (profile, full cwd, timeout, exit status) | landed |
| U4 | `hive status` JSON contract pinned by snapshot test — no schema change | landed |
| U5 | TUI drill-down view | **deferred** (keybinding conflict needs a plan amendment) |

Design notes:

- **Single-syscall appends.** Each `emit` performs one `syswrite` of `JSON + "\n"`, bypassing stdio buffering so the on-disk record corresponds to one `write(2)`. Messages are truncated to ~1 KiB before encoding to stay well under the Linux atomic-append threshold for O_APPEND on regular files. The module docstring states this contract so future contributors don't silently break it with multi-write transactions.
- **`status.md` is derived state.** It is rewritten via tmpfile + `File.rename` (mirroring `Hive::Markers#write_atomic`) and tails the last 20 events for display while walking up to 200 lines back for "Current agent" so a long-running spawn does not silently render as `—`.
- **Disk-full / read-only mounts must not hijack the orchestrator.** The emitter rescues `SystemCallError`, warns to stderr, and returns `nil` instead of raising — matching the existing `warn_isolation_reduced` posture in `lib/hive/stages/base.rb`.
- **No JSON schema change.** `hive status` text/JSON output is byte-for-byte identical; `test/integration/status_test.rb` pins this against a fixture.

Until U5 ships, operators tail the artifacts directly:

```
tail -f .hive-state/stages/<N>-<name>/<slug>/events.jsonl
cat   .hive-state/stages/<N>-<name>/<slug>/status.md
```

## Test plan

- `bundle exec rake test test/unit/events_test.rb` — emit contract, snapshot rendering, ENOSPC handling, EACCES (skipped under root), concurrent appends across 5 threads, 1000-emit perf regression guard.
- `bundle exec rake test test/unit/agent_test.rb` — `agent_start` / `agent_end` bracketing across normal exit, timeout, and exception paths.
- `bundle exec rake test test/integration/run_brainstorm_test.rb` — end-to-end `stage_enter` → `agent_start`/`_end` → `round_waiting`/`_complete` → `stage_exit` sequence.
- `bundle exec rake test test/integration/run_plan_test.rb` — `round_waiting` / `round_complete` symmetry on the plan stage.
- `bundle exec rake test test/integration/run_review_test.rb` — per-reviewer `agent_start`/`_end` pairs carrying the reviewer's configured name, per-phase brackets for CI/triage/fix/browser, and CI-stale → `error` event with reason.
- `bundle exec rake test test/integration/status_test.rb` — `hive status --json` byte-for-byte snapshot.
- Manual: ran `hive brainstorm` / `hive plan` / `hive review` on a fixture project and confirmed the event sequence reads naturally; verified the prior-stage `events.jsonl` survives a stage transition.

## Review summary

Two passes of `pr-review-toolkit` + `claude-ce-code-review` + `codex-ce-code-review` (round 2 escalated to the user on whether to ship U5 in this PR — answer: defer with an amended plan).

Resolved during review:

- **Atomicity contract honored end-to-end.** Round 1 found two separate `file.write` calls under O_APPEND violating the docstring; round 2 switched to `syswrite` (single syscall) and added a ~1 KiB message truncation so records stay well under the kernel's atomic-append threshold.
- **"Current agent" stays honest on long event logs.** Walk now reads up to 200 lines back, with the byte window scaled to the requested line limit instead of a flat 16 KiB cap.
- **Stage brackets balance on raise paths.** `with_stage_events` now emits a closing `stage_exit` after the rescue's `error` event, and `emit_rescue_close` is wrapped in its own `rescue` so a secondary failure inside event emission can never mask the original control-flow exception.
- **Per-phase review instrumentation landed.** Each `mark_working` site in `Hive::Stages::Review` (CI / per-reviewer / triage / fix / browser) is now bracketed with `agent_start`/`agent_end` carrying `phase=<name> pass=<N>` (or the reviewer's configured name for the per-reviewer loop). Green-path integration assertion added alongside the existing CI-stale error coverage.
- **Round events route through a stage-level allow-list, not a hard-coded string match.** `ROUND_EVENT_STAGES` lives next to the helper and is honored by `emit_marker_event`.
- **Test isolation tightened.** Removed the global `File.open` and `Time.now` monkey-patches in favor of scoped stubs / dependency-injected clocks; added a `skip_if Process.uid.zero?` guard to the EACCES test so it does not silently no-op under root.
- **Stale-branch silent-revert blocker fixed.** The branch was forked before PRs #130 and #132 landed on `origin/main`; rebased onto current main so this PR no longer reverts the telegram-bot background-by-default change or the daemon-first TUI README quickstart.
- **Cosmetics & docstrings.** Em-dash placeholders aligned with the plan, redundant `require` removed, `cwd` restored to the full path per the plan's U3 contract, narrow rescues in `close_phase_event!` and `write_atomic` so programmer bugs still surface, provenance markers stripped from `review.rb` source.

Escalated to the user (round 2): "Should plan U5 land in this PR?" — answer: defer. The agent burned three execute attempts on the keybinding conflict; shipping U1–U4 now unblocks future TUI work without further delay. Plan amended to mark U5 deferred.

Documentation: `wiki/modules/events.md` describes the emitter, the JSONL schema, and the U1–U4 wiring; `wiki/stages/review.md` notes the per-phase brackets; `wiki/index.md` and `wiki/log.md` updated accordingly.

## Linked task

Task: `i-want-to-improve-statis-260513-0610`
Stage: `8-finalize`
Idea: "I want to improve hive status and the TUI so I can see *which sub-agent within a stage* is currently running, not just the stage name."

Plan and brainstorm artifacts live in `~/Dev/hive-private/docs/plans/` and `~/Dev/hive-private/docs/brainstorms/` per the project's convention (hive's own `docs/plans/` and `docs/brainstorms/` are gitignored).

---


ivankuznetsov:

## High
- [ ] U5 (TUI drill-down view) is not implemented: Plan lists U5 as in-scope with concrete files (`lib/hive/tui/views/task_detail.rb`, `lib/hive/tui/model.rb`/`bubble_model.rb`/`key_map.rb`/`help.rb` edits, `D` keybinding, dedicated smoke + unit tests). No TUI files are in the diff (`git diff --name-only` shows only events/agent/base + tests). Without U5 the Requirements Trace items "hive status launches an interactive polling TUI (overview + drill-down)" and "Drill-down shows current agent + sub-pass + last action ts + recent events" are not satisfied, and the brainstorm Acceptance criterion ("drilling into one slug reveals every agent_start/end since stage entry") has no consumer. Plan does not defer U5.
- [ ] U2's per-phase review instrumentation is missing: `lib/hive/stages/review.rb` was not modified, but plan U2 requires emitting `agent_start`/`agent_end` at each `mark_working` site (CI / reviewers-per-reviewer / triage / fix / browser) with `agent: "<phase> pass=<N>"`. The lower-level `Hive::Agent#run!` bracket from U3 only fires for agent spawns, so the CI phase (which runs a shell command, not an agent) produces zero events in the drill-down history. The plan's integration scenario "review run with three reviewers configured produces one `agent_start`/`agent_end` per reviewer, plus the phase-level pairs for CI, triage, and (if reached) fix and browser" is not met — `test/integration/run_review_test.rb` only checks the CI-stale error path, not per-reviewer or per-phase pairs.
- [ ] `Hive::Events.render_status!` reads the entire `events.jsonl` file on every emit: `read_recent_events` calls `File.readlines(path, chomp: true).last(limit)`, materializing the whole file in memory before dropping all but 20 lines. Combined with plan-mandated "retain forever, no rotation" (Risk 5 explicitly acknowledged unbounded growth) and the 1000-emit-under-5s benchmark in Risk 7, this is an O(n²) cost over a task's lifetime — a 4-execute task that runs many passes will get progressively slower at every emit. Use the same tail strategy specified for U5 ("last ~16KB, split on newlines") here.

## Medium
- [ ] `current_agent` in `lib/hive/events.rb:86` is fed only the last 20 events (`STATUS_TAIL_LINES`), so a long-running agent whose `agent_start` is older than 20 records ago will silently render as `Current agent: -` even while still running. Plan U5 specs a 200-line walk for exactly this reason; the same depth should apply when building `status.md` snapshots so the snapshot is honest. Compute `current_agent` over a wider tail than the 20-line display tail.
- [ ] `with_stage_events` in `lib/hive/stages/base.rb:136-153` never emits `stage_exit` on the rescue paths — only `error`. The plan says `stage_exit` lands "on terminal-marker landing", and `:error`/`:review_error` are terminal markers (`report_json` raises after landing them). Drill-down consumers reading the events tail see an `error` event but no closing `stage_exit`, leaving the stage life-cycle unclosed. Either emit `stage_exit` after the rescue's `error` event, or document that `error` is the closing event for failure paths.
- [ ] `event_slug` / `event_stage` in `lib/hive/agent.rb:459-469` add `respond_to?` defensive fallbacks for `@task.slug` / `@task.stage_index` / `@task.stage_name`, but `@task` is always a `Hive::Task` (set in `initialize`). This is dead defensive code per CLAUDE.md ("Don't add error handling … for scenarios that can't happen"). Drop the fallbacks and call directly.
- [ ] `emit_marker_event` filters round_waiting/round_complete by hard-coded `%w[brainstorm plan]` stage-name allow-list (`lib/hive/stages/base.rb:161`). The check is duplicated from the stage runners and will silently drift if a new stage starts using `:waiting`/`:complete`. Move the round mapping into each affected stage runner (as U2 originally specified) or key the allow-list off a constant on the stage itself.
- [ ] Tests monkey-patch global singletons: `test/unit/events_test.rb` aliases `File.singleton_class#open` and `test/integration/status_test.rb#with_fixed_time` aliases `Time.singleton_class#now`. Both are process-global and will race or leak if a parallel test (or the surrounding suite) touches `File.open` or `Time.now` during the patched window. Use Minitest mocks scoped to the test, or inject a clock/IO seam into the emitter.
- [ ] `stage_exit` message drops marker attrs (`lib/hive/stages/base.rb:133`): only `marker.name` is serialized. When stage_exit lands on `:review_waiting reason=fix_guardrail pass=2` or `:review_error phase=fix reason=fix_tampered`, the drill-down loses the reason/phase/pass that the operator needs. Include marker.attrs in the message (or at minimum reason/phase/pass when present).

## Nit
- [ ] `lib/hive/events.rb:64,70,78-79` uses ASCII `-` for the "no value" placeholder; plan U5 specs em-dash (`—`). Cosmetic; pick one and stay consistent across the snapshot.
- [ ] `lib/hive/commands/run.rb:9` adds `require "hive/stages/base"` but `stages/base` is already loaded transitively via `hive/stages` (line 8). Redundant explicit require.
- [ ] `agent_start_message` in `lib/hive/agent.rb:471` interpolates the absolute `@cwd` path into every emit. With retain-forever logs this is verbose; a basename or worktree-relative path keeps records under the plan's "~140 chars" target.
- [ ] `write_atomic` in `lib/hive/events.rb:132-134`'s `ensure` calls `File.exist?(tmp)` then `File.delete(tmp)`, which is racy against the rename above; if rename succeeded the existence check is redundant, and if it failed `File.delete` is what should run. Simpler: `File.delete(tmp) rescue nil`.
- [ ] `lib/hive/events.rb:127-129`'s `rescue StandardError` around `fsync` swallows every error class. If fsync is best-effort, narrow the rescue to `Errno::*` / `IOError`.

ivankuznetsov:

## High
- [ ] Missing TUI drill-down implementation: `origin/main..HEAD` changes no `lib/hive/tui/*`, `lib/hive/tui/views/task_detail.rb`, keymap/help wiring, or TUI smoke/unit tests, so the plan-required `D` task detail view cannot exist.

## Medium
- [ ] Review phase transitions are not emitted to `events.jsonl`: `lib/hive/stages/review.rb:149`, `:331`, `:352`, `:422`, and `:519` only write `REVIEW_WORKING`, while events are emitted only by `Commands::Run`/`Hive::Agent`, so CI/browser/skipped phases and phase start/end history are absent from the planned drill-down data.
- [ ] `events.jsonl` appends are not atomic per record: `lib/hive/events.rb:45-47` performs separate writes for JSON and newline under `O_APPEND`, allowing concurrent emitters to interleave bytes and create unparsable JSONL despite the single-write concurrency contract.

## Nit

ivankuznetsov:

## High
- [ ] [scope-guardian] U5 (TUI drill-down view) is plan-required but unimplemented: the plan's Requirements Trace maps the "Drill-down shows current agent + sub-pass + last action ts + recent events" goal to U5, yet the diff makes zero changes to `lib/hive/tui/` — no `lib/hive/tui/views/task_detail.rb`, no new `task_detail` mode in `model.rb`/`bubble_model.rb`, no `D` keystroke in `key_map.rb`, no `help.rb` entry, and no `test/unit/tui/views/task_detail_test.rb` / smoke-test wiring. U5 is not deferred anywhere in the plan; the headline brainstorm acceptance test ("3 slugs in 5-review … overview shows each slug's current review agent and pass on its row. Drilling into one slug reveals every `agent_start`/`agent_end` since stage entry") therefore cannot be exercised.
- [ ] [correctness] Per-phase review-stage instrumentation specified by U2 is missing: U2 said "Review's per-phase emission lives at the `mark_working` call sites already in `lib/hive/stages/review.rb` … we add an `agent_start` event at the same call site with `agent: \"<phase> pass=<N>\"` and an `agent_end` at the matching phase-exit", but `lib/hive/stages/review.rb` is untouched in this diff. The five `mark_working(task, phase: …, pass: …)` sites at review.rb:148-149, 330-331, 351-352, 421-422, 518-519 never emit phase-level events, so the drill-down (when built) cannot show "pass 2 — reviewers — claude-ce-code-review started 14:02:11" as the plan's Requirements Trace promises; only spawn-level events from `Hive::Agent#run!` (U3) survive, and those carry `claude <log_label>`, not the phase / pass / reviewer-name labels.

## Medium
- [ ] [reliability] `lib/hive/events.rb:45-48` opens with `O_APPEND` but performs two separate `file.write` calls (`JSON.generate(record)` then `"\n"`). POSIX atomicity covers a single `write()` syscall — a concurrent emitter from another process (the daemon racing a manual `hive run`, which the plan's Risk #2 explicitly contemplates) can append between them, producing a corrupt line like `{"…":…}{"…":…}\n\n` that fails JSON parse. The module docstring on lines 22-27 self-describes as "Each emit performs one O_APPEND write", so the code violates its own documented contract. Use a single `file.write(JSON.generate(record) + "\n")`. The `test_concurrent_emits_keep_every_line_parseable` smoke test cannot exercise this — Ruby's GVL serializes writes from in-process threads.
- [ ] [observability] `lib/hive/events.rb:57,86-101` computes "Current agent" from only `STATUS_TAIL_LINES = 20` events because `read_recent_events` is called with that limit and `current_agent` operates on the result. For any task where the running agent's `agent_start` is more than 20 events back (review with multiple reviewers easily clears that), the snapshot will silently render `Current agent: -` even while the agent is still running. Decouple the "Current agent" walk (the plan's U5 suggested 200 lines) from the 20-line "Recent events" tail.
- [ ] [contract] `stage_exit` is never emitted when the runner raises: `with_stage_events` in `lib/hive/stages/base.rb:136-153` only emits an `error` event before re-raising. Consumers that bracket the stage on `stage_enter`/`stage_exit` pairs (the natural read of the contract, and what `test_ci_failures_yield_review_ci_stale_after_cap` asserts on the non-raising stale path) will see unbalanced events on every exceptional exit. Either emit a final `stage_exit` after the rescue's `error`, or document that `error` closes the bracket and adjust readers accordingly.
- [ ] [test-fragility] `test/unit/events_test.rb:71-83` monkey-patches `File.open` on the entire `File` class for the duration of the test body, raising `Errno::ENOSPC` for every File.open in Ruby/Minitest/test_helper that happens between the alias and the ensure. The blast radius is far broader than the unit under test; any unrelated call to `File.open` during that window will explode. Stub only the production path under test (e.g. wrap the append in a helper that the test can replace) instead of globally replacing the File class method.
- [ ] [test-fragility] `test/integration/status_test.rb:117-126` monkey-patches `Time.now` on the global `Time` class for the duration of the block. `Time` is process-wide global state; if the suite ever parallelizes or any unrelated production code reads `Time.now` during the window, it sees the frozen value. Prefer freezing time via a SUT-scoped accessor or via a per-test helper that wraps a tighter surface.
- [ ] [coverage] Plan U2's acceptance test "review run with three reviewers configured produces one `agent_start`/`agent_end` per reviewer, plus the phase-level pairs for CI, triage, and (if reached) fix and browser. The per-reviewer events carry the reviewer's name in the `agent` field" is not exercised — the only events.jsonl assertion in `test/integration/run_review_test.rb` is on the CI-stale path. Add a green-path assertion that per-reviewer pairs land and that the agent label carries the reviewer's configured name.
- [ ] [coverage] Plan U2 calls out plan-stage `round_waiting`/`round_complete` symmetry with brainstorm, but no integration test exists for the plan stage. Only `test_brainstorm_writes_round1_and_waiting` covers brainstorm, and `test_brainstorm_complete_marker` does not check events.jsonl. The plan-stage branch in `emit_marker_event` (`%w[brainstorm plan].include?(task.stage_name)` at `lib/hive/stages/base.rb:161`) has no behavioral test.

## Nit
- [ ] [maintainability] `lib/hive/agent.rb:459-469` — `@task.respond_to?(:slug)` and `@task.respond_to?(:stage_index)` defenses add dead branches: `Hive::Task` (constructed by every caller, including all tests) always satisfies them. Inline the attribute reads.
- [ ] [maintainability] `lib/hive/commands/run.rb:9` adds an explicit `require "hive/stages/base"` even though `require "hive/stages"` already loads it transitively via each runner. Redundant.
- [ ] [docs] `lib/hive/events.rb:22-27` docstring states "Each emit performs one O_APPEND write of a short JSON line", but the implementation performs two writes (see the reliability finding above). Either fix the code (preferred) or update the comment.
- [ ] [maintainability] `lib/hive/events.rb:62-83` `render_status_body` hard-codes inconsistent column padding (`"Stage:        "` has 8 spaces, `"Updated:      "` has 6, `"Last event:   "` has 3). Either align all labels or stop padding.
- [ ] [maintainability] `lib/hive/agent.rb:455-457` `event_agent_label` builds `"#{@profile.name} #{@log_label}"` with no separator collapse — when `@log_label` is nil or empty the label ends with a stray trailing space (and the test fixture exercises this via the `log_label: nil` default in some specs). `[@profile.name, @log_label].compact.join(' ')` or similar avoids the trailing whitespace.

ivankuznetsov:

## High
- [ ] U5 (TUI drill-down view) is STILL not implemented after the round-1 fix pass: the plan's Requirements Trace lists "`hive status` launches an interactive polling TUI (overview + drill-down)" and "Drill-down shows current agent + sub-pass + last action ts + recent events" as scope items met by U5; the diff against `origin/main` adds no files under `lib/hive/tui/views/` or edits to `lib/hive/tui/model.rb` / `bubble_model.rb` / `key_map.rb` / `help.rb`, so the brainstorm's load-bearing acceptance criterion ("drilling into one slug reveals every `agent_start`/`agent_end` since stage entry") has zero consumer in the hive UX. The fix-pass commit message (`cc7d588`) explicitly punts: "U5 TUI drill-down view: out of scope for a fix pass" — that is a runtime triage decision, not a plan-level deferral; the plan does not defer U5 to a separate task and U1–U4 produce events.jsonl + status.md only for U5 to render.
- [ ] `with_stage_events` rescue paths emit a generic exception message and drop the structured marker context the runner's own rescue just landed (`lib/hive/stages/base.rb:136-142,149-165`): plan U2 specifies "Every existing `rescue` that lands a REVIEW_ERROR / EXECUTE_STALE / ERROR marker emits an `error` event with `message` = the same `reason=` attr that lands on the marker", but on Review's top-level rescue (`lib/hive/stages/review.rb:559-569` sets `:review_error phase=ci reason=runner_exception exception_class=RuntimeError` then re-raises), `emit_rescue_close` writes only `"RuntimeError: synthetic helper failure"` — losing `phase=ci reason=runner_exception` from both the `error` event and the closing `stage_exit`. A drill-down reader scanning events.jsonl for the failure reason must fall back to reading `task.md` to recover the structured `phase`/`reason` the marker has on disk. Re-read the marker inside `emit_rescue_close` and surface its attrs in the message (the no-raise path already does this via `emit_marker_event` + `marker_event_message`).

## Medium
- [ ] `CURRENT_AGENT_WALK_LINES = 200` is unreachable given `STATUS_TAIL_BYTES = 16 * 1024` (`lib/hive/events.rb:18-28,72-78,127-153`): each emitted record is ~180-220 bytes (UTC iso8601 ts + slug + stage + agent label + message), so a 16 KiB tail holds ~80 records, not 200. `read_recent_events(path, 200)` reads the trailing 16 KiB, splits, then `.last(200)` returns at most ~80. Round-1 fix #4 widened the walk specifically so "a long-running agent whose agent_start scrolled past the recent-events tail still renders honestly" — but the byte cap re-imposes the same problem at ~80 events instead of 20. Either raise `STATUS_TAIL_BYTES` to fit 200 records (~48 KiB), or document the effective ceiling and reduce `CURRENT_AGENT_WALK_LINES` to match.
- [ ] `marker_event_message` (`lib/hive/stages/base.rb:201-207`) emits a single attr while `stage_exit_message` (`lib/hive/stages/base.rb:213-219`) emits three, producing inconsistent detail in the same emit pair: for `:review_error phase=fix reason=fix_tampered pass=2` the `error` event message reads `"review_error fix_tampered"` (phase + pass dropped) while the immediately-following `stage_exit` reads `"status=review_error phase=fix reason=fix_tampered pass=2"`. An operator scanning event messages for `phase=` cannot rely on which event holds it. Pick one approach (probably the three-attr form) and apply it uniformly across both helpers.

## Nit
- [ ] `"events.jsonl"` is hard-coded twice in `lib/hive/events.rb` (lines 56 and 73) and `"status.md"` twice (lines 77 and a-future-reader-may-look-for-it). Extract `EVENTS_FILENAME` / `STATUS_FILENAME` constants alongside `STATUS_TAIL_LINES` so a future filename change touches one site.
- [ ] `Hive::Stages::Review.close_phase_event!` rescues bare `StandardError` (`lib/hive/stages/review.rb:660-664`) — the same anti-pattern round-1 fix #19 narrowed in `Hive::Events.write_atomic`'s fsync block. A programming bug (NoMethodError on `@open_phase_event[:task]` if state corrupts, KeyError, etc.) is silently swallowed instead of surfaced. Narrow to `SystemCallError` / `IOError` consistent with the rest of the events plumbing.
- [ ] `close_phase_event!(_task = nil)` (`lib/hive/stages/review.rb:647`) takes an unused parameter; the function reads `@open_phase_event[:task]` from instance state. Drop the parameter and update the two callers (lines 575, 630) — the underscore-prefixed name signals "ignored" but a future maintainer still has to follow the trail to confirm it.
- [ ] `"Recent events (last #{STATUS_TAIL_LINES})"` (`lib/hive/events.rb:90`) hard-codes "last 20" in the header even when the file holds only 1-2 events. Either render the actual rendered count (`events.size`) or phrase as `"Recent events (up to #{STATUS_TAIL_LINES})"` so the header doesn't overpromise on a fresh task folder.

ivankuznetsov:

## High
- [ ] TUI drill-down mode is missing: Plan U5 requires `D` to open a task detail view over `events.jsonl`, but the diff adds no TUI view/model/keymap wiring and `lib/hive/tui/key_map.rb:81` still has no `:task_detail` mode.

## Medium
- [ ] Error events do not preserve the required failure detail: `lib/hive/stages/review.rb:167` writes `REVIEW_CI_STALE` with only attempts and `lib/hive/stages/base.rb:136` emits exception text on raised `REVIEW_ERROR` paths, so drill-down cannot show the CI failure or marker reason the plan calls for.

## Nit

ivankuznetsov:

## High
- [ ] [code-reviewer] U5 TUI drill-down view not implemented: Plan lists U5 as in-scope ("In scope: New file `lib/hive/tui/views/task_detail.rb`, new TUI keystroke `D`, mode transitions in `bubble_model.rb`/`model.rb`, keymap binding, help overlay, tests"). Diff contains zero TUI changes (no `task_detail.rb`, no `D` binding, no `:task_detail` mode); plan does NOT defer this to another task and the brainstorm acceptance test explicitly requires "drilling into one slug reveals every agent_start/agent_end since stage entry" — primary user-facing requirement unmet despite emitter being plumbed.
- [ ] [pr-test-analyzer] Plan-required performance regression benchmark missing: Plan risk #7 mandates a benchmark in `test/unit/events_test.rb` asserting "1000 emits complete in under 5 seconds as a regression guard". `events_test.rb` contains no such benchmark — without it, a future change that adds a per-emit fsync/lock can silently regress the hot path and the plan's documented performance contract is unverified.

## Medium
- [ ] [silent-failure-hunter] `Hive::Events.emit` rescues `SystemCallError` too broadly (`lib/hive/events.rb:67-69`): The rescue catches every Errno subclass (including Errno::ENOMEM, Errno::EBADF from a programming bug that leaks a closed fd) and converts them to a stderr warning + nil return. Narrow to the disk-state errors documented in the plan (Errno::ENOSPC, EROFS, EACCES, EDQUOT) so a genuine programming error surfaces instead of being absorbed.
- [ ] [silent-failure-hunter] `Hive::Stages::Review.close_phase_event!` swallows all `StandardError` (`lib/hive/stages/review.rb:660-664`): The comment justifies "don't mask control flow" but `rescue StandardError; nil` will also swallow NoMethodError / NameError introduced by future refactors — narrow to the IO/SystemCallError families that the underlying `Hive::Events.emit` may actually raise.
- [ ] [code-reviewer] Plan-trace divergence — rescue-path `error` event does not carry the marker's `reason` attr (`lib/hive/stages/base.rb:149-165`): Plan U2 specifies "Every existing rescue that lands a REVIEW_ERROR / EXECUTE_STALE / ERROR marker emits an error event with message = the same reason= attr that lands on the marker." `emit_rescue_close` emits `"#{e.class}: #{e.message}"`, not the marker's `reason=`. Happy-path `emit_marker_event` does it correctly via `marker_event_message`; rescue-path readers see the exception summary but lose the marker reason classification.
- [ ] [pr-test-analyzer] Concurrent-emit test uses Ruby threads, not processes (`test/unit/events_test.rb:115-131`): Plan risk #2 calls out "concurrent appends from sibling agents on the same slug" — the real risk surface is daemon + manual `hive run` racing as separate OS processes. Five Ruby threads serialise under the GIL for the syscall portion and don't validate the POSIX append-atomicity claim the emitter docstring relies on. Fork 5 children writing 10 lines each to exercise the documented contract.
- [ ] [code-reviewer] Module-instance state in `Hive::Stages::Review` is racy under any future multi-task-per-process model (`lib/hive/stages/review.rb:91-98`, `:622-640`): `@open_phase_event` and `@current_phase` live on the Review module singleton. Reset at `run!` entry keeps single-task callers safe today, but the invariant is not enforced (any thread-pool daemon would clobber both). At minimum, document the invariant at the module level; better, hoist them into a per-task object passed through the loop.
- [ ] [comment-analyzer] Docstring claims "single O_APPEND write" but emit also re-renders status.md (`lib/hive/events.rb:35-39`): "Each emit performs one O_APPEND write of a short JSON line" — true for `events.jsonl`, but the same `emit` call follows up with an `read 16 KiB → write tempfile → fsync → rename` sequence on `status.md` that future readers will miss. Either qualify ("for the append; status.md is derived") or move the perf-critical claim closer to the events.jsonl write.
- [ ] [code-reviewer] `write_atomic` skips parent-directory fsync (`lib/hive/events.rb:155-184`): The tempfile rename is atomic with respect to a concurrent reader, but on power loss the rename may not be persisted because the dir entry is not fsync'd. Acceptable for derived `status.md` per plan, but the docstring claim of "atomic rename" misleads a reader who expects crash-safety; either fsync the directory or document the limitation.

## Nit
- [ ] [comment-analyzer] `Hive::Events::EM_DASH` is `module_function`-public but used only by tests + internal renderer (`lib/hive/events.rb:30`): Constants exported only for tests are usually a sign of test coupling — assert against the literal `"—"` instead, and keep EM_DASH private.
- [ ] [pr-test-analyzer] No direct unit test exercises `read_recent_events` against a malformed/truncated line: The "skip unparseable lines" branch (`lib/hive/events.rb:144-148`) is only hit incidentally by concurrent emits, never deterministically — a one-line test that writes `{"valid":...}\n{garbage` and asserts the parser returns just the valid record would lock the documented "torn-write recovery" guarantee.
- [ ] [pr-test-analyzer] No green-path browser-phase agent_start/agent_end test: `run_review_test.rb` covers CI, reviewers, triage, and fix-phase brackets; the U2 plan-trace ("Phase 5: browser-test") has no positive-path assertion that browser also emits its phase pair.
- [ ] [code-reviewer] `emit_rescue_close` re-resolves `stage` defensively (`lib/hive/stages/base.rb:149-150`): `stage ||= stage_label(task)` — both call sites already pass `stage`. Either drop the `||=` or drop the kwarg; the dual-shape signature is mildly confusing for one private helper.
- [ ] [code-reviewer] `agent_start_message` truncates `cwd` to basename without surfacing the truncation (`lib/hive/agent.rb:479-483`): `File.basename("/home/asterio/Dev/foo.worktrees/slug")` and `File.basename("/tmp/foo/slug")` produce identical `slug` records — readers correlating events from two co-named tasks lose disambiguation. The plan trade-off is documented in the comment but a `cwd_basename=` attr (or a one-shot full path in stage_enter) preserves it without inflating every record.
- [ ] [code-reviewer] `event_stage` falls back to filesystem layout (`lib/hive/agent.rb:469-477`): When neither `stage_index` nor `stage_name` is available, the stage label becomes `File.basename(File.dirname(task.folder))` — couples the event payload to disk layout. Today this only happens for `Hive::Reviewers::SyntheticTask`, whose `stage_name` is always set, so the branch is dead in practice; either delete it or assert non-empty stage_name at construction.

PR #127, opened 2026-05-22
feat(tui): full-screen info panel and footer hint for [i]

ivankuznetsov:
## Summary

Pressing `i` on a focused task previously opened a 6-row bottom strip of `original_text` that any keystroke dismissed, and the binding was not advertised in the footer. This PR promotes `i` to a discoverable, full-screen, read-only info panel with stage-aware content and an explicit close set.

- **Footer advertises the binding.** `[i] info` is inserted between `[?] help` and `[q] quit` in `BubbleModel#footer_hint`, so the key is discoverable without opening the `?` overlay.
- **Full-screen, read-only panel.** `Hive::Tui::Model::InfoPanelState` carries the panel payload; `Views::IdeaPreview` renders an identity block (slug, stage, `created_at`, absolute stage folder, latest log path, original idea) plus a stage-specific extras block:

  | Stage | Extra content |
  |---|---|
  | `1-inbox` | (no extras) |
  | `2-brainstorm` | `brainstorm.md`, capped |
  | `3-plan` | `plan.md`, capped |
  | `4-execute` | tail of latest `.hive-state/logs/<slug>/execute-*.log` (4 KiB) |

- **Open-time gather, render-time no I/O.** `BubbleModel#open_idea_preview` reads `idea.md` frontmatter once, locates the latest log via `Hive::Task#log_dir`, pulls the stage extra, and packs everything into `InfoPanelState`. The view is a pure function of the captured state.
- **Sticky close-key gating.** `KeyMap#idea_preview_message` returns `BACK` only for `q`, `Esc`, or `i`; every other key is `NOOP`. Accidental typing no longer dismisses the panel.
- **Composition swap.** `compose_idea_preview_view` joins the panel body with `default_footer` directly instead of using `compose_two_pane_view(footer:)`, so the panel owns the body area while the footer hint line stays visible below it.
- **Help reflects the new behaviour.** `Help::BINDINGS` replaces the `key: "any"` row with explicit `q` / `Esc` / `i` entries and rewrites the grid-mode `i` description to enumerate the panel's fields.

The handler is pure read: `File.read` for `idea.md`, `File.open(..., "rb")` for extras, and `File.mtime` for the latest-log scan. No writes, no subprocesses, no markers, no network. Missing or unreadable extras fall through to `nil` and still render a useful identity block; pathological `idea.md` shapes (directory, non-UTF-8 bytes, oversize path) flash `"could not read idea"` instead of crashing the render loop. C0/C1 control bytes in extras are stripped to keep the panel frame intact.

## Test plan

- [x] `bundle exec rake test TEST='test/unit/tui/**/*_test.rb'` — 850 runs / 2860 assertions, green.
  - `bubble_model_test`: per-stage open paths (`1-inbox`, `2-brainstorm`, `3-plan`, `4-execute`), missing/unreadable extras, missing `idea.md`, empty `original_text`, unknown-stage safe fallback, read-only mtime invariant, `default_footer` truncation at cols=80 and cols=76.
  - `views/idea_preview_test`: per-stage layouts, overflow truncation reserves room for the stage-extra block, divider/ellipsis behaviour.
  - `key_map_test`: `q` / `Esc` / `i` → `BACK`; every other key (`x`, `Enter`, arrows…) → `NOOP`.
  - `help_test`: `BINDINGS` contains the three explicit `:idea_preview` close entries and the new grid-mode `i` description; no `key: "any"` row.
  - `model_test` / `update_test`: `InfoPanelState` member order pinned, `apply_back` clears `info_panel_state`, `Model.initial.info_panel_state` is `nil`.
- [x] Full suite `bundle exec rake test` — green after rerunning one unrelated flaky `DiagnosisAgentTest` timeout.
- [x] Manual smoke `bin/hive tui` on a project with cards in each of the four stages — footer shows `[i] info`, panel opens/closes only on `q` / `Esc` / `i`, accidental keys are inert, cursor returns to the originating card.

## Review summary

Two `/ce-code-review` passes plus the `pr-review-toolkit` reviewer ran across the branch. All High and Medium findings were auto-fixed in-tree:

- Restored the daemon-first README quickstart and appended (rather than replaced) the existing `wiki/log.md` entry — an unintended rebase regression.
- Hardened `BubbleModel#open_idea_preview` rescues to cover `Errno::EISDIR` / `Errno::ENOTDIR` / `Errno::ENAMETOOLONG` / `Encoding::InvalidByteSequenceError` so pathological `idea.md` shapes never crash the render loop.
- `latest_log_for` now narrows its outer rescue to `Errno::ENOENT` / `Errno::EACCES` (NFS `EIO` / `EMFILE` no longer silently masquerade as "no log"); the `4-execute` stage extra filters to `execute-*.log` so a stale `plan-*.log` can't masquerade as a fresh execute tail.
- `tail_capped` handles `Errno::ESPIPE` for non-seekable files and `read_capped` / `tail_capped` guard `File.file?` against directory/FIFO shapes.
- `scrub_utf8` strips C0/C1 control bytes (newlines/tabs preserved) so agent-authored markdown can't corrupt the TUI frame.
- `idea_created_at` falls back through `Time#utc.iso8601` and a tightened `frontmatter_scalar` regex (whitespace-anchored `#`-strip), and the fallback string is `.strip`ped so block scalars don't leak a trailing newline into the field grid.
- Deleted the dead `idea_preview_text` / `idea_preview_slug` mirror fields and `Views::IdeaPreview.legacy_state`; `InfoPanelState` is now the single source of truth.
- `field_line` reads its label-column width from a constant computed from the actual labels (no more silent 12→1 collapse for new 12+ char labels).
- `with_ellipsis` runs before styling so the truncation indicator stays inside the dim divider style on a real TTY.
- Test-side: stubbed unreadable-extra coverage at the `File.open` boundary (no longer depends on `chmod 0` semantics that fail under root in CI); added unknown-stage extra and cols=80 footer boundary tests; pinned `InfoPanelState.members` order; tightened the `frontmatter_scalar` comment-strip unit test for quoted-`#` and non-UTC `created_at` cases.
- Comment hygiene: stripped CLAUDE.md-disallowed multi-line docstrings, positional cross-references ("~1700 lines below"), and the history-narrating comment in `bubble_model_test`.

Items deferred with rationale (see review notes): display-width chunking for CJK / emoji, scrolling inside the panel, divider-row ellipsis cosmetics, and review-local `<task>/logs` discovery — all out of plan scope.

## Linked task

- Task slug: `add-i-key-with-legend-260522-ca28`
- Plan + reviews: `/home/asterio/Dev/hive/.hive-state/stages/8-finalize/add-i-key-with-legend-260522-ca28/`

---

### DIFF
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -2886,7 +2886,12 @@ module Hive
 
       def compose_idea_preview_view
         usable = [ @hive_model.cols.to_i - 1, 1 ].max
-        compose_two_pane_view(footer: prompt_footer(Views::IdeaPreview.render(@hive_model, width: usable), usable))
+        panel_height = [ @hive_model.rows.to_i - 1, 1 ].max
+        Lipgloss.join_vertical(
+          Lipgloss::TOP,
+          Views::IdeaPreview.render(@hive_model, width: usable, height: panel_height),
+          default_footer(usable)
+        )
       end
 
       # New-idea mode: same composition; footer = the inline prompt with
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -3017,7 +3017,7 @@ module Hive
       end
 
       def footer_hint
-        "[Tab] switch  [Enter] action  [n] new  [/] filter  [?] help  [q] quit"
+        "[Tab] switch  [Enter] action  [n] new  [/] filter  [?] help  [i] info  [q] quit"
       end
 
       # Compute pane widths and join horizontally. Left pane is clamped
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -72,10 +72,7 @@ module Hive
       # lives on `Hive::Markers::KILL_CLASS_EXIT_CODES` so KeyMap's
       # Enter-routing predicate and this auto-healer never drift.
       KILL_CLASS_EXIT_CODES = Hive::Markers::KILL_CLASS_EXIT_CODES
-      # Read window for the 4-execute info-panel log tail (see `tail_capped`
-      # ~1700 lines below). Sized for ~32 lines of recent agent output —
-      # enough operator context for the no-scroll panel without letting
-      # large logs dominate render-time memory.
+      # Sized for ~32 lines of recent agent output — bounds render-time memory on large logs.
       INFO_PANEL_EXECUTE_TAIL_BYTES = 4 * 1024
       # Lowercase snapshot-row marker keys → uppercase CLI marker names
       # accepted by `hive markers clear --name`. Mirrors the upcase
@@ -1547,7 +1544,7 @@ module Hive
           @hive_model.with(mode: :idea_preview, info_panel_state: state),
           nil
         ]
-      rescue Errno::ENOENT, Errno::EACCES, Psych::Exception
+      rescue Errno::ENOENT, Errno::EACCES, Errno::EISDIR, Errno::ENOTDIR, Errno::ENAMETOOLONG, Encoding::InvalidByteSequenceError, Psych::Exception
         [ flashed("could not read idea for #{row.slug}"), nil ]
       end
 
@@ -1564,15 +1561,7 @@ module Hive
         parsed.is_a?(Hash) ? parsed : {}
       end
 
-      # `created_at` is rendered as-is when the frontmatter carries a
-      # plain scalar (so a non-UTC `+03:00` timezone survives verbatim
-      # for the operator). The fallbacks cover edge cases the scalar
-      # parser can't represent — block scalars (`|` / `>`) or YAML-typed
-      # values that already parsed to a Time/Date — by formatting from
-      # `data`, which has been through `YAML.safe_load`. Time values are
-      # normalized to UTC iso8601 since a typed Time has lost the
-      # original literal; non-UTC offsets must come through the
-      # verbatim path above to be preserved.
+      # Verbatim plain scalar preserves non-UTC offsets; YAML-typed fallback normalizes Time to UTC iso8601.
       def idea_created_at(contents, data)
         raw = frontmatter_scalar(contents, "created_at")
         return raw unless raw.to_s.empty?
@@ -1582,57 +1571,38 @@ module Hive
         return value.utc.iso8601 if value.is_a?(Time)
         return value.iso8601 if value.respond_to?(:iso8601)
 
-        value.to_s
+        value.to_s.strip
       end
 
-      # Tightly-scoped scalar extractor used only by `idea_created_at`
-      # so a non-UTC timestamp (e.g. `+03:00`) survives the
-      # YAML.safe_load round-trip verbatim. Intentionally narrow: it
-      # only handles plain and matched-quote scalars on a single line.
-      # Block scalars (`|` / `>`), multi-line strings, escaped quote
-      # sequences, and YAML tags fall through to `nil`, which sends
-      # `idea_created_at` to its typed-value fallback path. Do not
-      # extend this for general use — for anything beyond `created_at`
-      # verbatim preservation, read from the already-parsed
-      # `idea_frontmatter` hash.
+      # Narrow plain/matched-quote scalar extractor — block scalars and multi-line strings return nil.
       def frontmatter_scalar(contents, key)
         match = contents.match(/\A---[ \t]*\r?\n(.*?)\r?\n---[ \t]*(?:\r?\n|\z)/m)
         return nil unless match
 
         match[1].lines.each do |line|
-          next unless (field = line.match(/\A#{Regexp.escape(key)}:[ \t]*(.*?)[ \t]*(?:#.*)?\r?\n?\z/))
+          next unless (field = line.match(/\A#{Regexp.escape(key)}:[ \t]*(.*?)\r?\n?\z/))
 
-          value = field[1].strip
-          return nil if value.empty? || value == "|" || value == ">"
+          raw = field[1].strip
+          return nil if raw.empty? || raw == "|" || raw == ">"
 
-          quote = value[0]
-          return value[1...-1] if value.length >= 2 && (quote == "\"" || quote == 39.chr) && value[-1] == quote
+          if raw.length >= 2 && [ "\"", "'" ].include?(raw[0])
+            quote = raw[0]
+            closing = raw.index(quote, 1)
+            return closing ? raw[1...closing] : nil
+          end
 
-          return value
+          stripped = raw.sub(/[ \t]+#.*\z/, "").strip
+          return stripped.empty? ? nil : stripped
         end
         nil
       end
 
-      # Returns the most recently modified log file under
-      # `.hive-state/logs/<slug>/` regardless of stage. Used for the
-      # generic `latest log` line in the info panel header — the
-      # 4-execute extra panel uses `latest_execute_log_for` instead so
-      # a stray `plan-*.log` mtime cannot be tailed as "execute log".
-      # The `*.log` glob matches `Hive::Agent#log_path`'s naming
-      # convention (`<log_label>-<ts>.log`); a non-`.log` artefact
-      # dropped into the logs dir is intentionally invisible here so
-      # cron-style sidecar files (`.tmp`, `.rotated`) do not surface as
-      # the operator-visible "latest log" pointer.
+      # Generic latest log used in the header; 4-execute extra uses `latest_execute_log_for` to avoid stray plan-*.log mtimes.
       def latest_log_for(row)
         latest_log_matching(row) { |_path| true }
       end
 
-      # Same shape as `latest_log_for` but restricted to files whose
-      # basename starts with `execute-` (matches the `log_label:
-      # "execute-impl"` convention in `Hive::Stages::Execute`). Keeps
-      # the 4-execute info-panel tail from picking up a freshly-written
-      # `plan-*.log` whose mtime happens to outrank the prior execute
-      # log.
+      # Restricted to `execute-*` basenames so a fresh plan-*.log mtime cannot masquerade as the execute log.
       def latest_execute_log_for(row)
         latest_log_matching(row) { |path| File.basename(path).start_with?("execute-") }
       end
@@ -1655,7 +1625,7 @@ module Hive
         end
 
         with_mtimes.max_by(&:last).first
-      rescue Errno::ENOENT, Errno::EACCES, SystemCallError
+      rescue Errno::ENOENT, Errno::EACCES
         nil
       end
 
@@ -1688,29 +1658,28 @@ module Hive
       def tail_capped(path)
         return nil if path.to_s.empty? || !File.file?(path)
 
-        # The seek is byte-aligned, not line-aligned: the tail's first
-        # rendered line may begin mid-character on a log line. This is
-        # acceptable for the read-only panel — it is documented here so
-        # a future contributor does not "fix" it by adding line-aware
-        # framing that re-introduces a full-file scan on large logs.
+        # Byte-aligned seek — the first rendered line may begin mid-character; do not add line-aware framing.
         text = File.open(path, "rb") do |file|
-          file.seek(-INFO_PANEL_EXECUTE_TAIL_BYTES, IO::SEEK_END)
-          file.read
-        rescue Errno::EINVAL
-          file.rewind
-          file.read
+          begin
+            file.seek(-INFO_PANEL_EXECUTE_TAIL_BYTES, IO::SEEK_END)
+            file.read
+          rescue Errno::EINVAL
+            begin
+              file.rewind
+              file.read
+            rescue Errno::ESPIPE
+              nil
+            end
+          end
         end
+        return nil if text.nil?
+
         scrub_utf8(text)
-      rescue Errno::ENOENT, Errno::EACCES, Errno::EISDIR, Errno::ENXIO
+      rescue Errno::ENOENT, Errno::EACCES, Errno::EISDIR, Errno::ENXIO, Errno::ESPIPE
         nil
       end
 
-      # `force_encoding` mutates in place, so dup first — callers may
-      # pass shared strings now or in the future. Also strips C0 control
-      # bytes (preserving TAB / LF / CR) and DEL so an agent-authored
-      # markdown body or log tail carrying ANSI escape sequences cannot
-      # corrupt the info-panel frame. Lone C1 bytes (0x80-0x9F) are
-      # invalid UTF-8 leads and are already removed by `String#scrub`.
+      # Dup before force_encoding (callers may share); also strips C0/DEL so ANSI escapes cannot corrupt the panel frame.
       def scrub_utf8(text)
         scrubbed = text.to_s.dup.force_encoding(Encoding::UTF_8).scrub
         scrubbed.gsub(/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/, "")
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -72,6 +72,7 @@ module Hive
       # lives on `Hive::Markers::KILL_CLASS_EXIT_CODES` so KeyMap's
       # Enter-routing predicate and this auto-healer never drift.
       KILL_CLASS_EXIT_CODES = Hive::Markers::KILL_CLASS_EXIT_CODES
+      INFO_PANEL_EXECUTE_TAIL_BYTES = 4 * 1024
       # Lowercase snapshot-row marker keys → uppercase CLI marker names
       # accepted by `hive markers clear --name`. Mirrors the upcase
       # convention in `Hive::Markers::KNOWN_NAMES`. A new recoverable
@@ -1520,18 +1521,30 @@ module Hive
         idea_path = File.join(row.folder, "idea.md")
         return [ flashed("no idea.md for #{row.slug}"), nil ] unless File.exist?(idea_path)
 
-        data = idea_frontmatter(File.read(idea_path))
+        contents = File.read(idea_path)
+        data = idea_frontmatter(contents)
         original_text = data["original_text"].to_s
         if original_text.empty?
           return [ flashed("idea has no original_text for #{row.slug}"), nil ]
         end
 
         capped_text = original_text[0, Hive::Tui::Model::NEW_IDEA_BUFFER_MAX_CHARS]
+        latest_log_path = latest_log_for(row)
+        state = Hive::Tui::Model::InfoPanelState.new(
+          slug: row.slug,
+          stage: row.stage,
+          created_at: idea_created_at(contents, data),
+          original_text: capped_text,
+          folder_path: File.expand_path(row.folder),
+          latest_log_path: latest_log_path,
+          stage_extra: stage_extra_for(row, latest_log_path)
+        )
         [
           @hive_model.with(
             mode: :idea_preview,
             idea_preview_text: capped_text,
-            idea_preview_slug: row.slug
+            idea_preview_slug: row.slug,
+            info_panel_state: state
           ),
           nil
         ]
@@ -1550,6 +1563,105 @@ module Hive
           aliases: false
         ) || {}
         parsed.is_a?(Hash) ? parsed : {}
+      end
+
+      def idea_created_at(contents, data)
+        raw = frontmatter_scalar(contents, "created_at")
+        return raw unless raw.to_s.empty?
+
+        value = data["created_at"]
+        return nil if value.nil?
+        return value.utc.iso8601 if value.is_a?(Time)
+        return value.iso8601 if value.respond_to?(:iso8601)
+
+        value.to_s
+      end
+
+      def frontmatter_scalar(contents, key)
+        match = contents.match(/\A---[ \t]*\r?\n(.*?)\r?\n---[ \t]*(?:\r?\n|\z)/m)
+        return nil unless match
+
+        match[1].lines.each do |line|
+          next unless (field = line.match(/\A#{Regexp.escape(key)}:[ \t]*(.*?)[ \t]*(?:#.*)?\r?\n?\z/))
+
+          value = field[1].strip
+          return nil if value.empty? || value == "|" || value == ">"
+
+          quote = value[0]
+          return value[1...-1] if value.length >= 2 && (quote == "\"" || quote == 39.chr) && value[-1] == quote
+
+          return value
+        end
+        nil
+      end
+
+      def latest_log_for(row)
+        log_dir = task_log_dir_for(row)
+        return nil if log_dir.nil? || !File.directory?(log_dir)
+
+        candidates = Dir.glob(File.join(log_dir, "*.log"))
+        return nil if candidates.empty?
+
+        with_mtimes = candidates.filter_map do |path|
+          [ File.expand_path(path), File.mtime(path) ]
+        rescue Errno::ENOENT
+          nil
+        end
+        if with_mtimes.empty?
+          candidate = candidates.find { |path| File.exist?(path) }
+          return candidate ? File.expand_path(candidate) : nil
+        end
+
+        with_mtimes.max_by(&:last).first
+      rescue Errno::EACCES
+        nil
+      end
+
+      def task_log_dir_for(row)
+        Hive::Task.new(row.folder).log_dir
+      rescue Hive::InvalidTaskPath
+        nil
+      end
+
+      def stage_extra_for(row, latest_log_path)
+        case row.stage.to_s
+        when "2-brainstorm" then read_capped(File.join(row.folder, "brainstorm.md"))
+        when "3-plan" then read_capped(File.join(row.folder, "plan.md"))
+        when "4-execute" then tail_capped(latest_log_path)
+        else nil
+        end
+      rescue Errno::ENOENT, Errno::EACCES
+        nil
+      end
+
+      def read_capped(path)
+        return nil if path.to_s.empty? || !File.exist?(path)
+
+        text = File.open(path, "rb") { |file| file.read(Hive::Tui::Model::NEW_IDEA_BUFFER_MAX_CHARS) }
+        scrub_utf8(text)
+      rescue Errno::ENOENT, Errno::EACCES
+        nil
+      end
+
+      def tail_capped(path)
+        return nil if path.to_s.empty? || !File.exist?(path)
+
+        # 4 KiB is enough recent execute context for the no-scroll panel
+        # without letting large logs dominate render-time memory.
+        text = File.open(path, "rb") do |file|
+          file.seek(-INFO_PANEL_EXECUTE_TAIL_BYTES, IO::SEEK_END)
+          file.read
+        rescue Errno::EINVAL
+          file.rewind
+          file.read
+        end
+        scrub_utf8(text)
+      rescue Errno::ENOENT, Errno::EACCES
+        nil
+      end
+
+      def scrub_utf8(text)
+        text.to_s.force_encoding(Encoding::UTF_8).scrub
       end
 
       def resolve_agent_label(row)
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -72,6 +72,10 @@ module Hive
       # lives on `Hive::Markers::KILL_CLASS_EXIT_CODES` so KeyMap's
       # Enter-routing predicate and this auto-healer never drift.
       KILL_CLASS_EXIT_CODES = Hive::Markers::KILL_CLASS_EXIT_CODES
+      # Read window for the 4-execute info-panel log tail (see `tail_capped`
+      # ~1700 lines below). Sized for ~32 lines of recent agent output —
+      # enough operator context for the no-scroll panel without letting
+      # large logs dominate render-time memory.
       INFO_PANEL_EXECUTE_TAIL_BYTES = 4 * 1024
       # Lowercase snapshot-row marker keys → uppercase CLI marker names
       # accepted by `hive markers clear --name`. Mirrors the upcase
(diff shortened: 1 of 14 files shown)


ivankuznetsov:

# Code Review — add-i-key-with-legend-260522-ca28 (pass 1)

Diff base: `origin/main` (HEAD `3e90c8c`, 6 commits, 14 files, +715/-128).

The diff implements IU1–IU5 from the plan: footer hint, `InfoPanelState`, full-screen `IdeaPreview` renderer, the rewritten `open_idea_preview` side-effect handler, and the tightened close-key gate. IU6 is plan-deferred. All plan-listed test scenarios appear covered, including 70/76/120-col footer widths, per-stage extras, mtime-based latest-log selection, the read-only mtime/content invariant (`test_open_idea_preview_does_not_dispatch_or_mutate_files`), and explicit close-key gating.

## High
- (none)

## Medium
- [ ] Dead legacy preview path retained in `lib/hive/tui/views/idea_preview.rb:100-117` + dual writes in `lib/hive/tui/bubble_model.rb:1632-1640`: `BubbleModel#open_idea_preview` always populates `info_panel_state`, so `Views::IdeaPreview#legacy_state` and the `idea_preview_text` / `idea_preview_slug` mirrors are unreachable in production yet kept frozen into `Model` schema and the `apply_back` clear (`lib/hive/tui/update.rb:741`). The plan asked to grep and delete outright if there are no out-of-tree callers; this retention re-introduces the dual-source-of-truth the plan tried to remove and the deprecation contract is undocumented (`# deprecated legacy preview ... mirror` comment is the only signal).
- [ ] `BubbleModel#latest_log_for` (`lib/hive/tui/bubble_model.rb:128-148`) restricts the candidate set with `Dir.glob(File.join(log_dir, "*.log"))`: the plan specifies "most recent file under `.hive-state/logs/<slug>/`" with no extension filter, so rotated/compressed siblings (`*.log.1`, `*.log.gz`, `stdout`, `stderr.txt`) or other artifacts written by the daemon would silently produce `latest_log_path: nil` even when the user expects a hit; no test pins the `*.log`-only contract, so this can drift.

## Nit
- [ ] `BubbleModel#frontmatter_scalar` (`lib/hive/tui/bubble_model.rb:112-126`) strips `(?:#.*)?` as an inline YAML comment without requiring a preceding space, so an unquoted value containing `#` (e.g. a URL fragment, path, or commit-style scalar) is silently truncated; for `created_at` ISO timestamps this is benign, but the helper is generic and the fallback to `idea_frontmatter` Time→iso8601 only kicks in when the raw scalar is empty.
- [ ] `BubbleModel#idea_created_at` (`lib/hive/tui/bubble_model.rb:100-110`) preferring the raw frontmatter scalar over the YAML-parsed Time means timestamps with a non-UTC offset are displayed verbatim (no `Time#utc.iso8601` normalization); this is plan-compliant ("verbatim from idea.md frontmatter") but the asymmetry between the raw-string and Time branches is undocumented and may surprise readers when one task shows `+0500` and another shows `Z`.
- [ ] `BubbleModel#tail_capped` (`lib/hive/tui/bubble_model.rb:176-191`) handles "file smaller than 4 KiB" via an in-block `rescue Errno::EINVAL` around `file.seek(-N, SEEK_END)`; a `File.size(path) < INFO_PANEL_EXECUTE_TAIL_BYTES` pre-check would convey intent without relying on the platform-specific EINVAL contract of `seek` for small files.
- [ ] `Views::IdeaPreview#extra_lines` (`lib/hive/tui/views/idea_preview.rb:47-56`) invokes `extra_title(state)` twice (for the title line and for the divider width); harmless allocation-wise, but a single local binding would read cleaner and pre-empts divergence if the title helper grows logic.
- [ ] `Views::IdeaPreview#render` uses `DISMISS_HINT.rjust(usable)` then `truncate(..., usable)` to right-align the close hint (`lib/hive/tui/views/idea_preview.rb:29`); at `width < DISMISS_HINT.length` the `rjust` is a no-op and the close hint is left-aligned and truncated mid-token. The plan accepts truncation, but no test pins this narrow-width close-hint behavior — easy to drift.
- [ ] `Views::IdeaPreview` tests do not cover the legacy `idea_preview_text` / `idea_preview_slug` render path (`legacy_state` is only reached via `state: nil` in `test_returns_empty_string_without_panel_state`); if the dead-code retention above stays, that path should at least be pinned, otherwise it can silently break without any test failing.

ivankuznetsov:

## High
- [ ] 4-execute info panel can tail a non-execute log: `stage_extra_for` tails `latest_log_path` from `latest_log_for`, which selects the newest `*.log` across brainstorm/plan/execute logs, so a new `4-execute` card can show a `plan-*.log` under the `execute log` section instead of the latest execute log (`lib/hive/tui/bubble_model.rb:1694`, `lib/hive/tui/bubble_model.rb:1722`).

## Medium
- [ ] Latest log path ignores review-local task logs: `latest_log_for` only scans `Hive::Task#log_dir`, while existing review sub-spawns and the log-tail path also use `<task>/logs`, so common info for `6-review` rows can show an older global log or `(none)` despite fresh review logs (`lib/hive/tui/bubble_model.rb:1503`, `lib/hive/tui/bubble_model.rb:1691`, `lib/hive/reviewers/synthetic_task.rb:26`).
- [ ] Optional extra reads can crash on non-regular paths: `read_capped`/`tail_capped` treat `File.exist?` as sufficient and only rescue `ENOENT`/`EACCES`, so a directory or FIFO named `plan.md`, `brainstorm.md`, or `*.log` raises outside the read-only best-effort path (`lib/hive/tui/bubble_model.rb:1729`, `lib/hive/tui/bubble_model.rb:1738`).
- [ ] Info panel renders file/log content without terminal-control sanitization: `scrub_utf8` preserves valid ANSI/control bytes and `wrapped_body` prints them directly, so agent-authored markdown or log tails can corrupt the TUI frame (`lib/hive/tui/bubble_model.rb:1755`, `lib/hive/tui/views/idea_preview.rb:117`).

## Nit

ivankuznetsov:

## High

## Medium
- [ ] [silent-failure-hunter] Outer `rescue Errno::ENOENT, Errno::EACCES, Psych::Exception` in `BubbleModel#open_idea_preview` (lib/hive/tui/bubble_model.rb:1645) is broad enough to swallow failures from `latest_log_for`/`stage_extra_for`. Helpers have inner rescues but do not cover every operation (e.g. `Dir.glob` or `File.directory?` against a permission-locked logs root, or a non-canonical `row.folder` raising something other than `Hive::InvalidTaskPath`); an unexpected EACCES would flash "could not read idea" and abort the open instead of leaving `latest_log_path: nil` / `stage_extra: nil` per plan-trace requirement A6.
- [ ] [code-reviewer] `Views::IdeaPreview.legacy_state` (lib/hive/tui/views/idea_preview.rb:103) is dead code on this branch: every `open_idea_preview` write path now populates `info_panel_state` alongside `idea_preview_text`/`idea_preview_slug`, and `apply_back` clears all three together. Plan IU2 explicitly authorises deletion when no out-of-tree callers exist; keeping it (plus the legacy `Model` fields and the dual-source render path) doubles the surface area without testing the fallback.
- [ ] [code-reviewer] `with_ellipsis` + `truncate` are applied unconditionally to `visible[-1]` in `truncate_rows` (lib/hive/tui/views/idea_preview.rb:71), but the last visible line can be a `Styles::HINT.render(...)` output (header divider, "brainstorm.md" title, etc.). `Format.truncate` counts code units, including ANSI escape bytes on a TTY, so the truncation point can land inside an escape sequence and corrupt terminal rendering for the rest of the frame; apply ellipsification only to plain content rows.
- [ ] [code-reviewer] `idea_created_at` (lib/hive/tui/bubble_model.rb:1662) re-runs a full frontmatter regex via `frontmatter_scalar` immediately after `idea_frontmatter` already invoked `YAML.safe_load`, and the motivation (preserving verbatim ISO timestamps that YAML→Time round-tripping would reformat) is not stated in code. Future readers will likely "simplify" the regex path away and silently regress the verbatim-string contract the plan asks for.
- [ ] [type-design-analyzer] `Model::InfoPanelState` (lib/hive/tui/model.rb:90) has seven nullable fields with no constructor validation or invariants — every consumer (view + tests + side-effect handler) re-implements `present?`/nil-check defaults. Combined with the surviving `idea_preview_text`/`idea_preview_slug` mirror fields and `Views::IdeaPreview.legacy_state`, the info-panel data flow has three competing sources of truth. Consider a `from_row` factory that enforces the contract once.
- [ ] [pr-test-analyzer] Plan IU2 requires an assertion that `Model::InfoPanelState` is frozen, but `test_info_panel_state_round_trips_fields_and_is_frozen` (test/unit/tui/model_test.rb:526) only asserts `state.frozen?` — it does not verify that mutating individual array/string members raises (Data.define values are shallow-frozen; if a future contributor swaps a field for an Array it would silently become mutable). Add a hash-frozen and mutation-attempt assertion.

## Nit
- [ ] [code-simplifier] `extra_title(state)` is invoked twice inside `extra_lines` (lib/hive/tui/views/idea_preview.rb:52-53) — once for the title, once for the divider length. Cache into a local.
- [ ] [code-reviewer] `field_line` hardcodes a 12-char label column (lib/hive/tui/views/idea_preview.rb:88); fine for today's three labels but adding any label longer than 11 chars (e.g., "latest agent log") collapses the column to a single space without warning.
- [ ] [pr-test-analyzer] `Views::IdeaPreview.legacy_state` has no direct unit coverage; if it survives the legacy-mirror cleanup it should be exercised, otherwise delete with the mirrors.
- [ ] [pr-test-analyzer] `test_default_footer_truncates_from_right_at_narrow_width` (test/unit/tui/bubble_model_test.rb:126) asserts `[i] info` survives at 76 cols and the trailing token is not split, but does not pin where exactly the truncation falls — a future hint change that silently clipped `[q] quit` entirely while still ending in `…` would pass. Add an assertion on the surviving suffix.
- [ ] [comment-analyzer] Model comments (lib/hive/tui/model.rb:43-44) mark `idea_preview_text` / `idea_preview_slug` as "deprecated legacy preview text/slug mirror" without naming the replacement (`info_panel_state`) or the removal timeline. If they are kept across the next release, point readers at the new field.
- [ ] [code-reviewer] `latest_log_for` filters strictly on `*.log` (lib/hive/tui/bubble_model.rb:1694) but the plan trace says "latest log path (most recent file under .hive-state/logs/<slug>/)" without an extension constraint. Either tighten the plan wording or relax the glob; flag for the author to confirm intent.
- [ ] [code-reviewer] `tail_capped` (lib/hive/tui/bubble_model.rb:1738) reads up to `INFO_PANEL_EXECUTE_TAIL_BYTES` from the end of the log; the seek lands mid-line and produces a partial first line. Acceptable per plan, but worth a one-line code comment so it is not "fixed" by a contributor who expects line-aligned tails.
- [ ] [comment-analyzer] `INFO_PANEL_EXECUTE_TAIL_BYTES` (lib/hive/tui/bubble_model.rb:75) is defined at the top of `BubbleModel` but only referenced from `tail_capped` ~1700 lines below; either move the constant near the helper or add a short pointer comment so it does not look orphaned.

ivankuznetsov:

## High

## Medium

## Nit
- [ ] Footer clips the new info legend at the supported 70-column boundary: `compose_two_pane_view` passes `cols - 1` into `default_footer`, so an actual 70-column grid renders `[i] inf…` instead of the required `[i] info` token.

ivankuznetsov:

# Code review — pass 2

Scope: `git diff origin/main..HEAD` against `add-i-key-with-legend-260522-ca28` (7 commits, 14 files, +863/−153). Source touched: `bubble_model.rb`, `model.rb`, `views/idea_preview.rb`, `key_map.rb`, `help.rb`, `update.rb`. Tests touched: matching `_test.rb` files. Wiki: `wiki/commands/tui.md`, `wiki/log.md`.

Intent: replace the 6-row bottom-strip idea preview with a full-screen read-only info panel; advertise `[i] info` in the footer; gate close-keys to `q` / `Esc` / `i`. Plan units IU1–IU5 implemented; IU6 (optional e2e) deferred per plan. All Goals / Requirements-Trace items from `plan.md` appear addressed in the diff and tests.

Personas applied (synthesised in-process, since orchestrator skill was invoked): correctness, testing, maintainability, project-standards, kieran-rails (Ruby idiom), reliability (I/O rescue surface), security (read-only invariant + control-byte scrub).

## High

_(none)_

## Medium

- [ ] `frontmatter_scalar` returns block-scalar chomp/keep indicators verbatim instead of falling through: `lib/hive/tui/bubble_model.rb:1693` only treats bare `|` and `>` as the "fall through to nil" sentinels, so `created_at: |-` (or `>+`, `|+`, `>-`) returns the literal `"|-"` string as `created_at`, which is then rendered to the operator as the timestamp. Either expand the guard to `%w[| > |- |+ >- >+].include?(value)` or strip the chomp indicator before the empty/sentinel check; today the typed-value fallback never fires for these shapes. Low-likelihood in real `idea.md` frontmatter, but it's the same code path the new feature relies on for "non-UTC tz survives verbatim", so a fix is cheap.
- [ ] `frontmatter_scalar` mishandles quote-balanced values with embedded escapes: same site, the matched-quote branch strips the outer quotes but leaves embedded `\"`/`\'` as literal backslash-quote sequences in the returned string. The method comment claims "escaped quote sequences ... fall through to `nil`", so behavior and comment disagree. Practical impact for `created_at` is small, but either tighten the regex to refuse a value containing a backslash-escape (then the typed fallback runs YAML's unescaping) or update the comment to match what the code actually returns.
- [ ] `INFO_PANEL_EXECUTE_TAIL_BYTES` lives ~1700 lines from `tail_capped` with a forward-reference comment (`see tail_capped ~1700 lines below`): `lib/hive/tui/bubble_model.rb:75-79`. The comment acknowledges the distance, but the constant is feature-local (only `tail_capped` reads it) and the file's other tunables (e.g. `KILL_CLASS_EXIT_CODES`) are cross-cutting. Co-locating the constant with `tail_capped` (or pulling info-panel constants into a small module/section) removes the forward reference without changing behaviour.

## Nit

- [ ] `IdeaPreview#wrapped_body` calls `truncate(line, width)` on each chunk emitted by `chunk_buffer(line, capacity)` where `capacity == width`: `lib/hive/tui/views/idea_preview.rb:119-123`. Each chunk is already at most `width` characters, so the `truncate` is a no-op. Either drop the redundant call or rely on it exclusively and skip the chunk-then-truncate double pass.
- [ ] `extra_title` returns `"details"` for any stage outside `2-brainstorm` / `3-plan` / `4-execute`, but `stage_extra_for` only ever populates `state.stage_extra` for those three stages: `lib/hive/tui/views/idea_preview.rb:110-117` + `lib/hive/tui/bubble_model.rb:1791-1798`. The `"details"` branch is unreachable through the normal open-panel flow; either remove it or guard `extra_lines` to never render for unknown stages explicitly so the dead branch can't quietly come back to life on a future stage rename.
- [ ] `Help::BINDINGS` for `:idea_preview` lists only the three close keys; nothing tells a help-overlay reader that other keys are intentional no-ops: `lib/hive/tui/help.rb:63-65`. A one-line `:idea_preview` "everything else: no-op" pseudo-entry (mirroring the `key: "any"` pattern that was removed) would document A7 ("unmapped keys are a no-op inside the panel") directly in the cheatsheet instead of leaving it inferable.
- [ ] `idea_preview_message` keeps its method name even though the panel is now an "info panel" — `lib/hive/tui/key_map.rb:362`. The plan deliberately kept the `:idea_preview` mode symbol for blast-radius reasons, so the name is consistent with the symbol, but the doc comment above could note that "idea preview" == "info panel" for future readers grep-ing for the new feature name.
- [ ] No direct unit test for `scrub_utf8`: control-byte stripping (`\x00-\x08\x0B\x0C\x0E-\x1F\x7F`) and the dup-before-`force_encoding` invariant are only exercised through the read flow on ASCII inputs. A small standalone test that pipes `"hello\x1B[31mred\x1B[0m\n"` and an invalid-UTF8 byte through `BubbleModel#send(:scrub_utf8, ...)` would pin the security-flavoured promise the comment makes.
- [ ] `test_open_idea_preview_created_at_falls_back_to_typed_value` only exercises the `value.to_s` branch (YAML parses the block scalar to a String); the `Time#utc.iso8601` and `respond_to?(:iso8601)` branches in `idea_created_at` (`lib/hive/tui/bubble_model.rb:1668-1672`) are untested. Adding a typed-Time frontmatter case (e.g. `created_at: 2026-05-22 22:40:00 +03:00` without quoting so YAML returns a `Time`) would pin the UTC-normalisation comment.

ivankuznetsov:

## High

- [ ] [silent-failure-hunter] `tail_capped` inner rescue only catches `Errno::EINVAL` — `lib/hive/tui/bubble_model.rb:1788-1796`: if `file.rewind`/`file.read` raise `IOError` or `Errno::EBADF`, the error escapes the outer rescue list (no `IOError`) and crashes `open_idea_preview` since its top-level rescue also omits `IOError`.
- [ ] [silent-failure-hunter] `open_idea_preview` top-level rescue omits `IOError` — `lib/hive/tui/bubble_model.rb:1644`: new helper surface (`read_capped`, `tail_capped`, `latest_log_matching`) plus `File.read(idea_path)` can produce `IOError`; rescue list is sized for the pre-change surface and a transient I/O failure would crash the handler instead of flashing.
- [ ] [comment-analyzer] `scrub_utf8` comment is factually wrong — `lib/hive/tui/bubble_model.rb:1814-1815`: claims "Lone C1 bytes (0x80-0x9F) … are already removed by `String#scrub`", but `String#scrub` *replaces* them with U+FFFD (which the subsequent gsub does not strip), so C1 bytes survive as visible replacement glyphs.
- [ ] [comment-analyzer] `INFO_PANEL_EXECUTE_TAIL_BYTES` doc-comment references "see `tail_capped` ~1700 lines below" — `lib/hive/tui/bubble_model.rb:75-76`: line-count reference is exactly the rot-prone form CLAUDE.md warns against; drop the count and keep only the method name.
- [ ] [pr-test-analyzer] No end-to-end open-then-close test with non-default cursor (A5) — `test/unit/tui/bubble_model_test.rb:3604` round-trips on default `[0,0]`, and `update_test.rb:1161` only exercises a synthesized `:idea_preview` via `apply_back`; a regression mutating `cursor:` inside `open_idea_preview` would ship undetected.
- [ ] [type-design-analyzer] `stage` is a free-form `String` but only the 9 entries in `Hive::Stages::DIRS` are legal — `lib/hive/tui/model.rb:90`: `InfoPanelState#initialize` should validate membership (mirroring `Model::Attachment` pattern at `model.rb:79`) so a typo like `"4-exectue"` is impossible and the view's stage-switching code is grounded in a closed set.
- [ ] [type-design-analyzer] `original_text` cap enforced by caller, not the type — `lib/hive/tui/bubble_model.rb:1629` vs `model.rb:88`: only the one caller applies `NEW_IDEA_BUFFER_MAX_CHARS`; a future constructor (likely a test fixture) silently violates the documented cap.
- [ ] [code-simplifier] `latest_log_for` / `latest_execute_log_for` / `latest_log_matching` trio collapses to one method — `lib/hive/tui/bubble_model.rb:1718-1751`: two one-line shims for a method that already takes a block; with only two call sites, the named shims add surface without saving the inline predicate.
- [ ] [code-simplifier] `fit_panel_lines` collapses to one branch — `lib/hive/tui/views/idea_preview.rb:64-70`: `truncate_rows(base + extra, capacity, width)` is behaviour-identical because `truncate_rows` already keeps `rows.first(capacity)` when oversize; the `base.length >= capacity` short-circuit is redundant.

## Medium

- [ ] [code-reviewer] Non-ANSI-aware secondary truncate on styled lines after ellipsis insertion — `lib/hive/tui/views/idea_preview.rb:77`: `truncate(with_ellipsis(visible[-1].to_s), width)` counts raw bytes including SGR codes, so on real ttys it can chop into the trailing `\e[…m` reset and bleed dim style into later output; tests don't catch it because non-tty test envs strip ANSI.
- [ ] [code-reviewer] `fit_panel_lines` drops extra section without `…` indicator when `base.length == capacity` — `lib/hive/tui/views/idea_preview.rb:64-69`: `truncate_rows(base, capacity, …)` returns rows unchanged when `rows.length <= capacity`, so on narrow/short terminals the entire brainstorm/plan/execute extra block disappears with no visual hint that content was hidden.
- [ ] [code-reviewer] `scrub_utf8` ANSI escape stripping is partial — `lib/hive/tui/bubble_model.rb:1807`: gsub kills `\e` (0x1B) but leaves the following `[2m…[22m` literal characters visible in the panel; comment claims "frame cannot be corrupted" (true for style flow) but operators see literal `[2m`/`[22m` runs.
- [ ] [pr-test-analyzer] `scrub_utf8` ANSI/control-byte stripping not behaviorally tested — `lib/hive/tui/bubble_model.rb:1800-1809` documents intent against ANSI in agent-authored content, but no test feeds `\e[31m`/BEL into `brainstorm.md` and asserts the bytes are removed from `state.stage_extra`; future regex simplification could silently re-admit them.
- [ ] [pr-test-analyzer] `tail_capped` `Errno::EINVAL` fallback for short logs untested — `lib/hive/tui/bubble_model.rb:1788-1794`: existing execute-tail test uses a >4KB file; a sub-4KB log that exercises the `seek` rescue + `rewind` path has no coverage.
- [ ] [pr-test-analyzer] `latest_log_matching` race rescue path (per-file `Errno::ENOENT` during mtime sort) untested — `lib/hive/tui/bubble_model.rb:1741`: filter_map's rescue is the documented mitigation for plan R7 but no test stubs `File.mtime` to raise and assert the survivor wins.
- [ ] [pr-test-analyzer] `frontmatter_scalar` quoted-scalar branches uncovered — `lib/hive/tui/bubble_model.rb:1701`: matched single/double quotes are stripped, but tests only cover bare + block-scalar paths; a refactor breaking quote handling would silently leak literal quotes into the panel header.
- [ ] [pr-test-analyzer] `frontmatter_scalar` inline-comment branch uncovered — `lib/hive/tui/bubble_model.rb:1696`: `(?:#.*)?` trailing-comment stripping has no test; a regex change turning the lazy `.*?` greedy would silently keep `value # note` text.
- [ ] [pr-test-analyzer] `with_ellipsis` ANSI-style preservation branch untested at view level — `lib/hive/tui/views/idea_preview.rb:86-96`: existing truncation test asserts `…` appears but not that it lands inside the trailing `\e[…m` reset (the actual operator-visible path through `Styles::HINT.render(DIVIDER * …)`).
- [ ] [silent-failure-hunter] `latest_log_matching` outer rescue is redundant — `lib/hive/tui/bubble_model.rb:1750`: `Errno::ENOENT, Errno::EACCES, SystemCallError` lists two Errnos as siblings of their parent `SystemCallError`; keep only the parent (or narrow to the specific Errnos actually expected) for consistency.
- [ ] [silent-failure-hunter] `stage_extra_for` rescue redundant + omits `IOError` — `lib/hive/tui/bubble_model.rb:1767`: lists four Errnos before `SystemCallError` (their parent); also missing `IOError` which the H1/H2 escape paths can produce; pick one convention and align with `read_capped`/`tail_capped`.
- [ ] [silent-failure-hunter] `task_log_dir_for` only rescues `Hive::InvalidTaskPath` — `lib/hive/tui/bubble_model.rb:1756`: any other exception from `Hive::Task.new` (e.g. `ArgumentError`) escapes `latest_log_matching` (`SystemCallError` won't catch it) and crashes `open_idea_preview`; assumption deserves an explicit comment or wider rescue.
- [ ] [type-design-analyzer] Three-state `String/nil/""` contract leaks into the view — `lib/hive/tui/views/idea_preview.rb:150-152` (`present?` used at 42-44, 53): producer already returns `nil` for missing data, but view defends against `""` anyway; normalizing blank to `nil` in `InfoPanelState#initialize` would let the view read two-state.
- [ ] [type-design-analyzer] `slug` and `original_text` could be required non-empty — `lib/hive/tui/model.rb:88-89`: callers already guard via `flashed(...)` returns, so encoding the invariant in the constructor would catch a future bypass at the type level.
- [ ] [type-design-analyzer] `stage_extra` is polymorphic across rendered markdown vs log tail — `lib/hive/tui/model.rb:95` written by `bubble_model.rb:1763-1771`, read by `views/idea_preview.rb:60` + `extra_title:111-117`: the meaning is split across `stage` + `stage_extra` and the view reconstructs it twice; an `extra_title` method on the type would consolidate.
- [ ] [type-design-analyzer] `extra_title` belongs on `InfoPanelState`, not the view — `lib/hive/tui/views/idea_preview.rb:111-117`: pure function of `state.stage`; sibling `Data.define` types in this same file (`Attachment`, `RedStatusDetailState`) already use `do…end` blocks for behaviour.
- [ ] [comment-analyzer] `idea_created_at` comment drifts into restating each branch — `lib/hive/tui/bubble_model.rb:1668-1675`: the first two sentences are good WHY (preserve non-UTC offsets) but the rest enumerates each branch of the code that follows; trim to the WHY clause.
- [ ] [comment-analyzer] `frontmatter_scalar` comment enumerates unsupported cases — `lib/hive/tui/bubble_model.rb:1680-1686`: "Block scalars (`|` / `>`), multi-line strings, escaped quote sequences, and YAML tags fall through to `nil`" restates code; keep the "Do not extend this for general use" scoping sentence, drop the enumeration.
- [ ] [comment-analyzer] `latest_log_for` first sentence describes WHAT — `lib/hive/tui/bubble_model.rb:1694-1702`: method name + one-line body already convey it; trim to the second-half WHY about the `execute-` filter rationale.
- [ ] [comment-analyzer] "Pre-fix" reference in test rots — `test/unit/tui/bubble_model_test.rb:3589-3592`: violates CLAUDE.md "don't reference the current fix/prior implementation"; rewrite as forward-looking WHY (`File.chmod(0)` is a no-op for root which CI runs as).
- [ ] [comment-analyzer] wiki/log.md "now"/"instead of" references prior implementation — `wiki/log.md:5+body`: "the view is now a full-screen … instead of a 6-row bottom strip" will read confusingly once the bottom strip is forgotten.
- [ ] [code-simplifier] `with_ellipsis` argument: drop the `to_s` round-trip — `lib/hive/tui/views/idea_preview.rb:77`: `visible[-1]` is already a String; `with_ellipsis(visible[-1])` suffices.
- [ ] [code-simplifier] `idea_created_at` typed-value branch — `respond_to?(:iso8601)` subsumes the `is_a?(Time)` check — `lib/hive/tui/bubble_model.rb:1670-1680`: collapsing to two duck-typed branches removes the class check; defensible to keep for grep-ability though.
- [ ] [code-simplifier] `info_state` test factory `UNSET` sentinel is heavy — `test/unit/tui/views/idea_preview_test.rb:8-22`: replace with a `**overrides` merge over a defaults hash so `folder_path: nil` becomes an explicit override; removes the const + `.equal?` boilerplate.

## Nit

- [ ] [code-reviewer] `height: model.rows.to_i - 1` default mirrors caller math — `lib/hive/tui/views/idea_preview.rb:22`: making `height:` required prevents future drift if a new caller forgets to pass it.
- [ ] [code-reviewer] `DIVIDER * [title.length, width].min` is not piped through `truncate` — `lib/hive/tui/views/idea_preview.rb:59`: inconsistent with line 47's `DIVIDER * "Original idea".length` path; safe today but unifying would be future-proof.
- [ ] [code-reviewer] Header `gap < 2` fallback can exceed width — `lib/hive/tui/views/idea_preview.rb:102`: `"#{prefix}  #{stage}"` overflows on narrow widths and relies on the outer `truncate(header_line, width)` to clamp; worth a one-line note.
- [ ] [code-reviewer] Three near-identical help entries for q/Esc/i — `lib/hive/tui/help.rb:63-65`: could mirror the existing `key: "j/k"` idiom (`new_idea_project` row) and collapse to one entry, saving help-overlay vertical space.
- [ ] [code-reviewer] `UNSET = Object.new.freeze` sentinel could use a one-line WHY comment — `test/unit/tui/views/idea_preview_test.rb:8`: clarifies that `nil` is a meaningful explicit value distinct from "use the default path".
- [ ] [pr-test-analyzer] Missing logs directory not separately tested for non-execute stages — `lib/hive/tui/bubble_model.rb:1734`: only `4-execute` exercises the missing-dir path via `test_open_idea_preview_execute_without_log_has_nil_log_fields`.
- [ ] [pr-test-analyzer] `stage_extra_for` `Errno::EISDIR` rescue untested — `lib/hive/tui/bubble_model.rb:1767`: existing stub raises EACCES; fold an EISDIR variant into the same test.
- [ ] [pr-test-analyzer] Unknown-stage fallthrough returns `nil`/`"details"` untested — `lib/hive/tui/bubble_model.rb:1765` + `views/idea_preview.rb:115`: a future stage rename in `Hive::Stages::DIRS` would silently hit the catch-all without anyone noticing.
- [ ] [silent-failure-hunter] `scrub_utf8` silently collapses binary input — `lib/hive/tui/bubble_model.rb:1806`: no defect today (glob restricted to `*.log` / `*.md`) but worth a one-line note so a future glob widening doesn't surprise.
- [ ] [silent-failure-hunter] `latest_log_matching` fallback `File.file?` can itself raise — `lib/hive/tui/bubble_model.rb:1748`: relies on the outer rescue; an explicit `rescue false` inside the `find` block would make the chain local.
- [ ] [silent-failure-hunter] `with_ellipsis` comment over-promises for mid-line ANSI resets — `lib/hive/tui/views/idea_preview.rb:84-96`: behaviour is correct today but the comment assumes reset is always trailing, which only holds because `Styles::HINT.render` is the sole producer.
- [ ] [type-design-analyzer] Sibling `Data.define` types in this file use a `do…end` block — `lib/hive/tui/model.rb:75-86` (`Attachment`) and `:98-109` (`RedStatusDetailState`) carry behaviour; `InfoPanelState` is the only one without one despite needing `extra_title`/`present?` helpers.
- [ ] [type-design-analyzer] `folder_path` is typed as nullable but always present in practice — `lib/hive/tui/bubble_model.rb:1636`: guarded by the empty-folder early return at `:1617`; can be tightened to non-null alongside `slug`/`original_text`.
- [ ] [code-simplifier] `LABEL_COL = [...].max + 2` magic constant — `lib/hive/tui/views/idea_preview.rb:18`: name the gutter (`LABEL_GUTTER = 2`) to clarify the geometry.

PR #128, opened 2026-05-22
feat(tui): simplify red-status detail to Recover + Open in agent

ivankuznetsov:
## Summary

The red-status detail screen now leads with what the operator can do, not how hive talks to itself. The previous Q/A layout exposed marker names like `REVIEW_ERROR phase=fix pass=2`, `$EDITOR`, and a four-key footer (`[Enter] autofix / retry  [f] manual fix  [R] refresh diagnosis  [q] back`) whose affordances varied with `action_key`. Every red-status row now shows the same panel: a one-sentence "why", `[Enter] Recover`, and `[o] Open in agent` (which names the project's configured agent, e.g., `claude` or `codex`). After either action, the screen closes and the operator lands back on the grid.

### What changed

- `lib/hive/tui/views/red_status_detail.rb` — one layout for every row: header, project / stage / why lines, two labelled actions, footer. Wrapped in `Styles::PANE_FOCUSED_BORDER` for terminals ≥ 40 cols; narrower terminals fall back to plain text. The footer is reserved outside the body-height trim so the action-keys hint cannot be clipped. The "why" line strips ANSI/control bytes and replaces marker-shape summaries (`REVIEW_ERROR phase=fix pass=2`) with a plain-English fallback.
- `lib/hive/tui/red_status_detail_keys.rb` (new) — single source of truth for the two action keys, shared by the view footer and the KeyMap refusal-flash hint so a key rename can't drift between surfaces.
- `lib/hive/tui/key_map.rb` — `red_status_detail_message` routes `Enter → RedStatusAutofix`, `o → OpenInAgent`, `Esc`/`q → BACK`. The retired `f` and `R` plus the grid-mode `s` and verb keys all flash a hint that points at `o`.
- `lib/hive/tui/bubble_model.rb` — both `RedStatusAutofix` and `OpenInAgent` are wrapped so that, when dispatched from `:red_status_detail`, the screen closes back to `:grid` in the same tick (success, refusal, and "no recipe" all close). For Open-in-agent, the foreground takeover Cmd is still returned, so the suspend/resume cycle ends on the grid. The recover-execute refusal flash now reads "Hive has no automatic recovery for this state — try Open in agent" instead of the bare `no autofix action available`.
- Dead surface removed: `Messages::OpenManualFix`, `Messages::RefreshRedStatusDiagnosis`, the `open_manual_fix` / `refresh_red_status_diagnosis` handlers, the `marker_signature` / `acknowledged_marker_signature` / `refreshing` state, the `recover_execute` ENTER_FLASH_MESSAGES entry, and the `SubprocessExited(folder:)` kwarg.
- Agent label is resolved once at open time in `BubbleModel#open_red_status_detail` (not on every render or inside `Update`), cached on `RedStatusDetailState#agent_label`, and protected by a wide rescue (`Hive::ConfigError`, `Hive::AgentProfiles::UnknownAgent`, `Hive::AgentError`, `Hive::InvalidTaskPath`, `Psych::Exception`, `SystemCallError`) so any bad config falls back to "your project's development agent" without crashing the bubbletea loop.
- Wiki + docs/solutions/ refreshed: `wiki/commands/tui.md`, `wiki/decisions.md` (ADR-027), `wiki/modules/diagnosis_agent.md`, `docs/solutions/architecture-patterns/red-status-diagnose-then-act-2026-05-16.md`, and `wiki/log.md` all now describe the two-action contract; the stale `FOOTER_NO_ENTER` comment in `task_action.rb` was removed.

### Design notes

- `recover_execute` rows now advertise Recover too. Previously the footer dropped `[Enter]` for `EXECUTE_STALE` because there is no auto-retry recipe. The unified surface trades that branch for a flash from `red_status_autofix` when no recipe exists, so the operator's gesture stays "let hive try" and the flash points them at Open in agent.
- `o`, not `s`. Grid-mode already uses `s` to steer the focal row into the dev agent. Reusing `s` here would muddle the mode boundary; `o` reads as "open in" and the `s`-key hint surfaces the rename to anyone with grid-mode muscle memory.
- Close-on-dispatch is gated on the source mode being `:red_status_detail`. Grid-mode `s` (which dispatches the same `OpenInAgent` message) and grid-mode `RedStatusAutofix` both pass through the wrapper unchanged.

## Test plan

- `bin/rake test TEST=test/unit/tui/views/red_status_detail_test.rb` — new copy, agent-label lookup, `ConfigError` fallback, missing-diagnostic fallback, marker-summary suppression, ANSI sanitization, bordered render on wide terminals, plain-text fallback on narrow terminals, footer survival under short rows and long summaries.
- `bin/rake test TEST=test/unit/tui/key_map_test.rb` — pins `o → OpenInAgent`, `Esc`/`q → BACK`, `f`/`R → NOOP`, the `s`-key "press o for Open in agent" hint, and the verb-keys generic hint.
- `bin/rake test TEST=test/unit/tui/bubble_model_test.rb` — pins screen-close on Recover success and refusal from detail mode; close on Open-in-agent success (Cmd still returned) and refusal; grid-mode pass-through of the close wrapper.
- `bin/rake test TEST=test/unit/tui/help_test.rb` — pins the exact `:red_status_detail` BINDINGS (Enter / o / q / Esc) and the absence of `f` / `R`.
- `bin/rake test TEST=test/unit/tui/update_test.rb` and `bin/rake test TEST=test/unit/tui/messages_test.rb` — confirm the removed surface (refresh-diagnosis, OpenManualFix, SubprocessExited.folder) is gone without collateral damage.

## Review summary

Two review passes (claude + codex + pr-review-toolkit) ran against this branch. All High and Medium findings were auto-fixed and all Nit findings were either fixed or resolved with explicit triage notes — see `reviews/` in the task folder for the full audit trail. Highlights:

- Plan Units 3 (help-overlay sweep), 4 (Recover closes the screen), 5 (Open-in-agent closes the screen), and 6 (delete dead surface) were initially missed in pass 1 and landed in pass 2 (commit `1450d12`).
- Plan Risk #4 (cache agent-label at open time, wide rescue list) and Risk #5 (border threshold + footer survival) were addressed in the same pass.
- Pass 2 hardening (commit `1e5354f`) moved the agent-label lookup out of `Update.apply` into `BubbleModel`, extracted `RedStatusDetailKeys` so KeyMap no longer reaches into a view module, tightened `MARKER_SUMMARY_PATTERN` so all-caps verdicts like `ABORTED` aren't suppressed, and aligned wiki + `docs/solutions/` with the new contract.
- Pass 2 tests (commit `4cd1654`) added the plan-mandated happy-path close tests, narrow-terminal regression, footer-truncation regression, and grid-mode pass-through coverage.

## Linked task

- Slug: `red-status-debug-screen-where-260519-4e22`
- Idea: red status debug screen where i'm choosing between recovery and editor not looking very good, we need to make it easy to understand for the user
- Task folder: `.hive-state/stages/8-finalize/red-status-debug-screen-where-260519-4e22/`

---

Claude Code

### DIFF
--- a/docs/solutions/architecture-patterns/red-status-diagnose-then-act-2026-05-16.md
+++ b/docs/solutions/architecture-patterns/red-status-diagnose-then-act-2026-05-16.md
@@ -105,12 +105,13 @@ For ambiguous red rows, grid Enter should open a detail view, not immediately mu
 1. Why is this red?
 2. What can Hive do next?
 
-Then offer explicit actions:
+Then offer a deliberately narrow, unified action surface:
 
-- `Enter`: existing autofix/retry path.
-- `f`: open worktree in `$EDITOR` for manual fix.
-- `R`: refresh diagnosis via the headless command.
+- `Enter`: run the automated recovery for this row, then close the detail screen. Rows with no auto-recovery recipe surface a refusal flash that names the manual fallback ("Open in agent") and still close — the operator's binary gesture should never strand them on a stale view.
+- `o`: open the task in the project's configured development agent (the same takeover path the grid `s` key uses), then close the detail screen as the TUI suspends.
 - `q`/`Esc`: return.
+
+Two actions beat three because every additional gesture demands a justification on the screen. Refreshing diagnosis is a shell-level affordance (`hive status --diagnose <slug> --write`) rather than a TUI keybinding — the detail screen is for "act on what we already know", and the headless diagnosis command is for "ask the agent to look again." Separating them keeps the in-TUI action surface boring and predictable.
 
 Preserve direct paths for deterministic states where a detail view would add friction:
 
--- a/lib/hive/task_action.rb
+++ b/lib/hive/task_action.rb
@@ -439,14 +439,13 @@ module Hive
     # The diagnostic shape exists for ALL three recovery action keys per
     # ADR-027 and wiki/commands/status.md — recover_review, error, AND
     # recover_execute (EXECUTE_STALE). The TUI's red_status_detail view
-    # renders all three; recover_execute deliberately omits the [Enter]
-    # autofix affordance because EXECUTE_STALE has no auto-retry recipe
-    # (see red_status_detail.rb#FOOTER_NO_ENTER and
-    # suggested_next_action_payload below — manual_fix / command:null).
-    # Bot, daemon, and external `--json` consumers also rely on this
-    # field for EXECUTE_STALE rows the autofix path cannot resolve;
-    # emitting nil would defeat the diagnose-then-act feature for one
-    # of the three red states it covers.
+    # renders all three under a unified [Enter] Recover / [o] Open in
+    # agent contract; recover_execute rows surface the Risk-#3 mitigation
+    # flash when [Enter] Recover is pressed (no auto-retry recipe), then
+    # the screen closes. Bot, daemon, and external `--json` consumers
+    # also rely on this field for EXECUTE_STALE rows the autofix path
+    # cannot resolve; emitting nil would defeat the diagnose-then-act
+    # feature for one of the three red states it covers.
     def diagnostic_action?
       %w[recover_execute recover_review error].include?(key.to_s)
     end
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -368,12 +368,14 @@ module Hive
           dispatch_command(message)
         when Hive::Tui::Messages::OpenLogTail
           open_log_tail(message.row)
+        when Hive::Tui::Messages::OpenRedStatusDetail
+          open_red_status_detail(message)
         when Hive::Tui::Messages::RecoverReview
           recover_review(message.row, force: message.force)
         when Hive::Tui::Messages::RecoverError
           recover_error(message.row)
         when Hive::Tui::Messages::RedStatusAutofix
-          red_status_autofix_from_detail(message.row)
+          dispatch_red_status_autofix_then_close_detail(message.row)
         when Hive::Tui::Messages::OpenInputEditor
           open_input_editor(message.row)
         when Hive::Tui::Messages::OpenTaskFolder
@@ -381,7 +383,7 @@ module Hive
         when Hive::Tui::Messages::OpenIdeaPreview
           open_idea_preview(message.row)
         when Hive::Tui::Messages::OpenInAgent
-          open_in_agent_from_detail(message.row)
+          dispatch_open_in_agent_then_close_detail(message.row)
         when Hive::Tui::Messages::AgentSteerExited
           archive_steer(message)
         when Hive::Tui::Messages::OpenSummary
@@ -861,8 +863,12 @@ module Hive
       # Wrap red_status_autofix so the detail screen closes after the
       # keypress (Recover refusal flash still surfaces — the operator's
       # gesture was binary, leaving them on the screen contradicts the
-      # plan's "screen closes after the keypress" requirement).
-      def red_status_autofix_from_detail(row)
+      # plan's "screen closes after the keypress" requirement). The
+      # wrapper is also reached from grid mode (Enter routes
+      # `RedStatusAutofix` through `handle_side_effect`) — the close
+      # branch is mode-guarded so grid-mode invocations are a no-op
+      # pass-through.
+      def dispatch_red_status_autofix_then_close_detail(row)
         model, cmd = red_status_autofix(row)
         close_red_status_detail_on_dispatch(model, cmd)
       end
@@ -871,7 +877,9 @@ module Hive
       # (success and refusal alike). The takeover Cmd is preserved so
       # the foreground takeover still suspends the TUI; on return the
       # operator lands on the grid rather than a stale detail view.
-      def open_in_agent_from_detail(row)
+      # Reached from grid mode (`s` key) and detail mode (`o` key); the
+      # close branch is mode-guarded so grid-mode `s` is unchanged.
+      def dispatch_open_in_agent_then_close_detail(row)
         model, cmd = open_in_agent(row)
         close_red_status_detail_on_dispatch(model, cmd)
       end
@@ -881,9 +889,8 @@ module Hive
         return [ model, cmd ] unless model.mode == :red_status_detail
 
         closed = model.with(mode: :grid, red_status_detail_state: nil)
-        snap = closed.snapshot
-        if snap
-          visible = snap.scope_to_project_index(closed.scope).filter_by_slug(closed.filter)
+        visible = Hive::Tui::Update.visible_snapshot(closed)
+        if visible
           cursor = Hive::Tui::Update.reclamp_cursor(visible, closed.cursor)
           closed = closed.with(cursor: cursor)
         end
@@ -895,7 +902,7 @@ module Hive
         return [ flashed("Hive has no automatic recovery for this state — try Open in agent"), nil ] if command.nil?
 
         if command.match?(/[;&|]/)
-          return [ flashed("diagnostic retry command is not directly dispatchable; use the shell command from details"), nil ]
+          return [ flashed("diagnostic retry command is not directly dispatchable; run it from a shell or use Open in agent"), nil ]
         end
 
         argv = Shellwords.split(command)
@@ -1511,6 +1518,38 @@ module Hive
           aliases: false
         ) || {}
         parsed.is_a?(Hash) ? parsed : {}
+      end
+
+      # Resolve the configured development-agent label for the detail
+      # screen, then delegate to Update so the model transition stays in
+      # the pure MVU layer. Lives here (not in Update or the view) so
+      # the no-I/O Update contract is preserved and the view stays a
+      # pure projection. Rescue list is intentionally broad — a corrupt
+      # project config (chmod 000, hand-edited YAML aliases, dangling
+      # path) must never crash the bubbletea runner when the user opens
+      # the detail screen; fall back to the canonical "your project's
+      # development agent" label instead. Plan Risk #4.
+      def open_red_status_detail(message)
+        label = resolve_agent_label(message.row)
+        new_message = Hive::Tui::Messages::OpenRedStatusDetail.new(
+          row: message.row,
+          agent_label: label
+        )
+        new_model, _cmd = Hive::Tui::Update.apply(@hive_model, new_message)
+        [ new_model, nil ]
+      end
+
+      def resolve_agent_label(row)
+        task = Hive::Task.new(row.folder.to_s)
+        cfg = Hive::Config.load(task.project_root)
+        agent_name = cfg.dig("execute", "agent") || "claude"
+        profile = Hive::AgentProfiles.lookup(agent_name, cfg: cfg)
+        basename = File.basename(profile.bin.to_s)
+        basename.empty? ? Hive::Tui::Model::RedStatusDetailState::AGENT_FALLBACK : basename
+      rescue Hive::ConfigError, Hive::AgentProfiles::UnknownAgent,
+             Hive::AgentError, Hive::InvalidTaskPath,
+             Psych::Exception, SystemCallError, IOError
+        Hive::Tui::Model::RedStatusDetailState::AGENT_FALLBACK
       end
 
       def open_in_agent(row)
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -138,10 +138,9 @@ module Hive
         # in the spawn-thread's `ensure` so retries unblock once the
         # first pass settles. Reuses @healed_folders_mutex so all
         # background-thread bookkeeping fields (@healed_folders,
-        # @heal_threads, @review_recovery_inflight, @error_recovery_inflight,
-        # @diagnosis_inflight) share a single lock.
+        # @heal_threads, @review_recovery_inflight,
+        # @error_recovery_inflight) share a single lock.
         @error_recovery_inflight = Set.new
-        @diagnosis_inflight = Set.new
         # Once-per-session latch (reset in #stage_image on success).
         @clipboard_tool_hint_shown = false
         # Consecutive `wl-paste`/`xclip` timeouts: a single timeout
@@ -374,11 +373,7 @@ module Hive
         when Hive::Tui::Messages::RecoverError
           recover_error(message.row)
         when Hive::Tui::Messages::RedStatusAutofix
-          red_status_autofix(message.row)
-        when Hive::Tui::Messages::OpenManualFix
-          open_manual_fix(message.row)
-        when Hive::Tui::Messages::RefreshRedStatusDiagnosis
-          refresh_red_status_diagnosis(message.row)
+          red_status_autofix_from_detail(message.row)
         when Hive::Tui::Messages::OpenInputEditor
           open_input_editor(message.row)
         when Hive::Tui::Messages::OpenTaskFolder
@@ -386,7 +381,7 @@ module Hive
         when Hive::Tui::Messages::OpenIdeaPreview
           open_idea_preview(message.row)
         when Hive::Tui::Messages::OpenInAgent
-          open_in_agent(message.row)
+          open_in_agent_from_detail(message.row)
         when Hive::Tui::Messages::AgentSteerExited
           archive_steer(message)
         when Hive::Tui::Messages::OpenSummary
@@ -667,37 +662,12 @@ module Hive
       # `Update.apply_subprocess_exited` keeps its existing default
       # flash behavior.
       def diagnose_subprocess_exit(msg)
-        if msg.verb.to_s == "status"
-          # Evict EXACTLY the completing folder (passed through by
-          # refresh_red_status_diagnosis's dispatcher wrap). Clearing
-          # the whole Set on every status exit cascaded across rows:
-          # a finish on folder A wiped B's inflight slot while B was
-          # still running, letting a second R-press on B spawn a
-          # duplicate agent.
-          evict_diagnosis_attempt(msg.folder) if msg.folder
-          model = clear_detail_refreshing(@hive_model)
-          if msg.exit_code.to_i.zero?
-            return [ model.with(flash: "red-status diagnosis refreshed", flash_set_at: Time.now), nil ]
-          end
-
-          diagnostic = Hive::Tui::Subprocess.diagnose_recent_failure(msg.verb)
-          text = diagnostic || "`status --diagnose` exited #{msg.exit_code} — tail #{Hive::Tui::Subprocess.log_path}"
-          return [ model.with(flash: text, flash_set_at: Time.now), nil ]
-        end
-
         return nil if msg.exit_code.nil? || msg.exit_code.zero?
 
         diagnostic = Hive::Tui::Subprocess.diagnose_recent_failure(msg.verb)
         return nil if diagnostic.nil?
 
         [ @hive_model.with(flash: diagnostic, flash_set_at: Time.now), nil ]
-      end
-
-      def clear_detail_refreshing(model)
-        state = model.red_status_detail_state
-        return model if state.nil?
-
-        model.with(red_status_detail_state: state.with(refreshing: false))
       end
 
       # Scan a fresh snapshot for tasks whose `:error` marker came from
@@ -884,13 +854,45 @@ module Hive
         when "error"
           recover_error(row)
         else
-          [ flashed("no autofix action available for #{row.slug}"), nil ]
+          [ flashed("Hive has no automatic recovery for this state — try Open in agent"), nil ]
         end
+      end
+
+      # Wrap red_status_autofix so the detail screen closes after the
+      # keypress (Recover refusal flash still surfaces — the operator's
+      # gesture was binary, leaving them on the screen contradicts the
+      # plan's "screen closes after the keypress" requirement).
+      def red_status_autofix_from_detail(row)
+        model, cmd = red_status_autofix(row)
+        close_red_status_detail_on_dispatch(model, cmd)
+      end
+
+      # Wrap open_in_agent so the detail screen closes on dispatch
+      # (success and refusal alike). The takeover Cmd is preserved so
+      # the foreground takeover still suspends the TUI; on return the
+      # operator lands on the grid rather than a stale detail view.
+      def open_in_agent_from_detail(row)
+        model, cmd = open_in_agent(row)
+        close_red_status_detail_on_dispatch(model, cmd)
+      end
+
+      def close_red_status_detail_on_dispatch(model, cmd)
+        return [ model, cmd ] unless model
+        return [ model, cmd ] unless model.mode == :red_status_detail
+
+        closed = model.with(mode: :grid, red_status_detail_state: nil)
+        snap = closed.snapshot
+        if snap
+          visible = snap.scope_to_project_index(closed.scope).filter_by_slug(closed.filter)
+          cursor = Hive::Tui::Update.reclamp_cursor(visible, closed.cursor)
+          closed = closed.with(cursor: cursor)
+        end
+        [ closed, cmd ]
       end
 
       def dispatch_diagnostic_retry(row)
         command = diagnostic_retry_command(row)
-        return [ flashed("no autofix action available for #{row.slug}"), nil ] if command.nil?
+        return [ flashed("Hive has no automatic recovery for this state — try Open in agent"), nil ] if command.nil?
 
         if command.match?(/[;&|]/)
           return [ flashed("diagnostic retry command is not directly dispatchable; use the shell command from details"), nil ]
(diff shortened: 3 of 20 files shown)


ivankuznetsov:

## High
- [ ] Unit 3 not implemented — `lib/hive/tui/help.rb:54-58` still lists the removed bindings `[Enter] autofix / retry`, `[f] manual fix`, `[R] refresh diagnosis`: the help overlay diverges from the screen and the keymap; plan Requirements Trace requires this update and the plan does not defer it.
- [ ] Unit 4 not implemented — `bubble_model.rb#red_status_autofix` (L876) does not flip mode back to `:grid`: pressing Enter from `:red_status_detail` leaves the operator stranded on the detail screen, violating the plan requirement "screen closes after the keypress and control returns to TUI" and the explicit close-on-dispatch helper called for in Unit 4.
- [ ] Unit 5 not implemented — `bubble_model.rb#open_in_agent` (L1649) does not close the detail screen on dispatch: after the takeover returns, the user lands back on the stale detail screen instead of the grid; plan Requirements Trace lists this as an in-scope requirement with no deferral.
- [ ] Footer can be dropped on narrow terminals — `red_status_detail.rb:38-41` appends `FOOTER` to `lines` then trims via `lines.first(body_height)`, so the action-keys hint is cut off when `rows` is small (previously the footer was joined after the trim and was always visible): regresses the only on-screen reference for `[Enter] Recover / [o] Open in agent / [Esc] back`.

## Medium
- [ ] Agent profile resolved on every render — `red_status_detail.rb:55-64` calls `Hive::Config.load` + `Hive::AgentProfiles.lookup` inside `render`, which fires per poll/redraw: plan Risk #4 explicitly mandates caching the resolved label on `RedStatusDetailState` at open time so `render` stays a pure projection; current implementation pays the YAML+disk cost on every frame.
- [ ] `recover_execute` Recover refusal flash text not updated — `bubble_model.rb:887` still flashes the bare `no autofix action available for #{row.slug}`: plan Risk #3 mitigation requires the friendlier `Hive has no automatic recovery for this state — try Open in agent` because the detail screen now always advertises `[Enter] Recover` for these rows, and the current flash reads like a hive bug.
- [ ] Test stubs mutate real module singleton state — `red_status_detail_test.rb` `with_config_load_stub` / `with_agent_profile_lookup_stub` redefine singleton methods on `Hive::Config` / `Hive::AgentProfiles` via `define_singleton_method` with no `Mocha`/Minitest-mock isolation: an `ensure`-time failure or parallel test would leak the stub into other tests; safer with `Minitest::Mock`, a real fixture project root, or a small caching seam on the view.
- [ ] Unit 6 not implemented — `Messages::OpenManualFix`, `Messages::RefreshRedStatusDiagnosis`, `bubble_model.rb#open_manual_fix`, `refresh_red_status_diagnosis`, and the `RedStatusDetailState#refreshing` / `acknowledged_marker_signature` fields are now dead code: plan recommends option (a) "delete now" to keep the surface honest; carrying half-wired plumbing while the wiring was already removed in Unit 2 invites accidental re-introduction.

## Nit
- [ ] `f` and `R` keys silently NOOP in `red_status_detail_message` — `key_map.rb:309-314` flashes the hint only for `s` and `VERB_KEYS`; `f` / `R` (the previous bindings) return `Messages::NOOP` with no feedback: reuse the same hint flash for muscle-memory drift, matching the `s` branch.
- [ ] `agent_label` does not guard empty `profile.bin` — `File.basename(profile.bin.to_s)` returns `""` for nil/blank `bin`, producing the label `launch  in the task worktree` (double space): treat blank basename as the fallback `AGENT_FALLBACK`.
- [ ] `refute_includes output, "marker"` in `test_missing_diagnostic_uses_plain_english_fallback` is fragile — any future copy mentioning "marker" in the friendly fallback prose would break the test without indicating a real regression: scope the assertion to specific old strings (`marker_signature`, `Marker:`).
- [ ] Hint flash omits `q` — `red_status_detail_message` hint says "Esc to close" but `q` also closes (and the footer reads `[Esc] back`): align hint wording with the footer (`Esc / q to close`) for consistency.
- [ ] No regression test for the bordered-vs-plain render branch — `outer_width >= 40` threshold in `red_status_detail.rb:24` is uncovered: a small terminal would silently skip the border, but the assertion gap means a future refactor could flip the threshold without test failure.

ivankuznetsov:

## High
- [ ] Red-status detail still leaks marker summaries and attrs in normal rows (`lib/hive/tui/views/red_status_detail.rb:32`): the view renders `diagnostic["summary"]` verbatim, while `Hive::TaskAction#marker_summary` still produces strings like `REVIEW_ERROR phase=fix pass=2`, violating the plan's no-marker/no-debug-copy requirement.
- [ ] Recover/Open-in-agent do not close the detail screen (`lib/hive/tui/bubble_model.rb:376`, `lib/hive/tui/bubble_model.rb:388`): the handlers still return `red_status_autofix`/`open_in_agent` directly, so mode remains `:red_status_detail` and `red_status_detail_state` remains set after the keypress.
- [ ] Help overlay still advertises removed red-detail actions (`lib/hive/tui/help.rb:54`): it lists `autofix`, `f` manual fix, and `R` refresh diagnosis with `$EDITOR` wording instead of the required Recover/Open-in-agent/back bindings.

## Medium
- [ ] `RedStatusDetail` is no longer self-contained to require (`lib/hive/tui/views/red_status_detail.rb:2`): requiring the file directly loads `hive/agent_profiles` before `Hive::ConfigError` is defined and raises `NameError`.
- [ ] Invalid project config can still crash render instead of using the agent fallback (`lib/hive/tui/views/red_status_detail.rb:55`): `Hive::Config.load` can raise `Psych::SyntaxError` from malformed YAML, which is not rescued by `agent_label`.

## Nit
- [ ] Stale recover-execute comment still documents the retired `R` refresh/manual-fix surface (`lib/hive/tui/key_map.rb:236`): the comment now contradicts the new two-action red-detail keymap and will mislead the next maintainer.

ivankuznetsov:

## High
- [ ] [code-reviewer] Plan Unit 4 not implemented — `Messages::RedStatusAutofix` handler in `lib/hive/tui/bubble_model.rb:377` does not flip `:red_status_detail` back to `:grid` after dispatch, so the Requirements-Trace row "Screen closes after the keypress and control returns to TUI" is violated for Recover.
- [ ] [code-reviewer] Plan Unit 5 not implemented — `Messages::OpenInAgent` handler in `lib/hive/tui/bubble_model.rb:389` does not close the detail screen before launching the takeover; after the agent exits the operator returns to a stale detail view instead of the grid.
- [ ] [code-reviewer] Plan Unit 3 not implemented — `lib/hive/tui/help.rb:54-58` still advertises `[f] manual fix ($EDITOR)` and `[R] refresh diagnosis` for `:red_status_detail`, but the keymap no longer routes those keys, so the `?` overlay now ships outright wrong bindings.
- [ ] [code-reviewer] Plan Unit 6 not done — `OpenManualFix` / `RefreshRedStatusDiagnosis` messages (`lib/hive/tui/messages.rb:172,176`) and their handlers (`lib/hive/tui/bubble_model.rb:378,380`) plus `refreshing` state plumbing are now unreachable from any key path; plan recommends deletion to avoid resurrecting half-wired behavior.
- [ ] [code-reviewer] `agent_label` does per-frame disk I/O — `lib/hive/tui/views/red_status_detail.rb:55-64` invokes `Hive::Config.load` (walks parent dirs + reads YAML) inside `render`, which is called on every TUI redraw; the plan explicitly flagged this as risk #4 and mandated caching on `RedStatusDetailState` at open time.

## Medium
- [ ] [code-reviewer] Footer (action-key hint) now lives inside the truncated `lines` array — `lib/hive/tui/views/red_status_detail.rb:38-42` puts FOOTER into `lines`, then `lines.first(body_height)` truncates; on narrow terminals or long wrapped "Why" blocks the footer is clipped, hiding the only affordance hint the operator has.
- [ ] [code-reviewer] Recover-execute UX mitigation missing — plan risk #3 required the no-recipe flash to read "Hive has no automatic recovery for this state — try Open in agent"; `red_status_autofix` is unchanged, so operators still see the legacy "no autofix action available" flash and have no nudge toward the second action.
- [ ] [silent-failure-hunter] Over-broad rescue in `agent_label` — `lib/hive/tui/views/red_status_detail.rb:61-63` swallows `SystemCallError`, `IOError`, and `Hive::InvalidTaskPath` beyond the plan's listed exceptions; a real Errno::EACCES on `config.yml` is silently masked as "your project's development agent" with no operator signal.
- [ ] [pr-test-analyzer] No test exercises the unbordered (`outer_width < 40`) fallback — both new tests use `cols: 100`, so the dual-path render at `lib/hive/tui/views/red_status_detail.rb:24-44` and the narrow-terminal regression risk (#5 in plan) are uncovered.
- [ ] [pr-test-analyzer] No regression test pins "footer survives wrapping" — with a long summary or small `model.rows`, `lines.first(body_height)` can drop the FOOTER line; no test fails when that happens, so a copy or layout edit can silently regress the action-key hint.
- [ ] [pr-test-analyzer] Bordered render not asserted — tests check string content but never that `Styles::PANE_FOCUSED_BORDER` was applied, so the visual chrome called out as a requirement can regress without test failure.
- [ ] [pr-test-analyzer] Stubbing helpers in `red_status_detail_test.rb:30-44` rely on a dead conditional — `sentinel = Hive::Config.method(:load)` is always truthy, so the `ensure if sentinel` is misleading; if `yield` raises after stub install on a `module_function`, the restore rebinds it as a plain singleton method and subtly changes lookup semantics for sibling tests.

## Nit
- [ ] [code-reviewer] Bordered threshold off-by-2 vs plan — `outer_width >= 40` where `outer_width = cols - 2` means the real threshold is `cols >= 42`; plan said "below 40 cols". Cosmetic but documented intent differs from code.
- [ ] [code-simplifier] Affordance copy duplicated — `FOOTER` ("[Enter] Recover   [o] Open in agent   [Esc] back") in `red_status_detail.rb:13` and `hint` ("press Enter to recover, o to open in agent, Esc to close") in `key_map.rb:311` encode the same key set twice with different wording; a key rename must be done in both places or they drift.
- [ ] [comment-analyzer] Design rationale lost — the previous file documented why `recover_execute` Enter was a no-op (FOOTER_NO_ENTER block); the new unified "always show [Enter] Recover, flash on no recipe" contract has no comment explaining the decision, so a future reader will not know it is intentional.
- [ ] [code-reviewer] `Hive::Task.new(row.folder.to_s)` in `agent_label` builds a full Task object just to read `project_root`; a lighter helper (e.g., `Hive::Task.project_root_for(path)`) would avoid the `InvalidTaskPath` rescue surface and the construction cost on every render.
- [ ] [code-reviewer] `summary_text` calls `safe(...).strip` even though `safe` already routes through `Text.sanitize`; harmless but suggests the contract of `safe` is unclear and worth tightening.
- [ ] [pr-test-analyzer] Removed coverage not replaced — `test_markerless_retry_error_keeps_enter_affordance`, `test_manual_fix_error_omits_enter_affordance`, and `test_footer_omits_enter_affordance_for_recover_execute_rows` were deleted; add a positive test that pins "Enter affordance is always shown regardless of action_key" so the new unified contract cannot regress.
- [ ] [code-reviewer] `Messages::Flash` for the `s` key in `key_map.rb:312` shares its wording with the VERB_KEYS branch ("press Enter to recover, o to open in agent, Esc to close"); the plan asked specifically for an "use o for Open in agent" nudge on `s` so muscle-memory drift is visible — current text addresses it generically but loses the "you pressed the grid-mode steer key" signal.

ivankuznetsov:

## High
- [ ] Wiki not updated to match the new red-status detail contract — `wiki/commands/tui.md` (L78, L158, L160, L167) still documents `[Enter] autofix / retry`, `f` opens `$VISUAL`/`$EDITOR`/`vi`, `R` dispatches `hive status --diagnose ... --write`, and the "marker changed — refresh (R)" flash flow; `wiki/log.md` has no entry for this change. The project CLAUDE.md mandates updating `wiki/` + `wiki/log.md` when behavior changes, and the plan does not defer this. The wiki is now a misleading source of truth for the two-action surface this PR delivers.

## Medium
- [ ] `Views::RedStatusDetail.resolve_agent_label` rescue list (`lib/hive/tui/views/red_status_detail.rb:101-103`) is narrower than the `manual_fix_candidate` it functionally supersedes — missing `SystemCallError` (config-file permission/ENOENT during `Hive::Config.load` or `Hive::Task#project_root` walking) and `Psych::DisallowedClass` (config YAML carrying disallowed Ruby tags). Either escapes from `Update.apply_open_red_status_detail` (no rescue around `resolve_agent_label` at the call site `update.rb:706`) and crashes the detail-screen open path, defeating the plan Risk #4 mitigation that "the screen never breaks on an unhealthy config".
- [ ] `BubbleModel#close_red_status_detail_on_dispatch` (`lib/hive/tui/bubble_model.rb:879-891`) duplicates the `:red_status_detail` branch of `Update.apply_back` (`update.rb:719-722` — same `mode: :grid`, `red_status_detail_state: nil`, `visible_snapshot`, `reclamp_cursor` shape). Plan Unit 4 explicitly called for "extracted into a private helper so Unit 5 can reuse it"; the helper exists but `apply_back` still rolls its own copy. Two implementations can drift independently (nil-snapshot handling, future cursor-policy tweaks). Extract one shared helper or have the wrapper call into the `Update` module.
- [ ] Plan-prescribed help-overlay regression test missing — plan Unit 3 Test scenarios required `test/unit/tui/help_test.rb` assertions that the `:red_status_detail` section contains exactly four entries and exactly the two action symbols `:recover` and `:open_in_agent` (plus `:back` twice). `help_test.rb` only validates grid-mode bindings; a future revert of `help.rb` red-detail bindings would not break any test.

## Nit
- [ ] `BubbleModel#open_in_agent_from_detail` (`lib/hive/tui/bubble_model.rb:874`) is the dispatch handler for ALL `Messages::OpenInAgent` deliveries (including grid-mode `s`); the `_from_detail` suffix advertises a constraint the method does not enforce (the `mode == :red_status_detail` guard inside the wrapper is what gates the close behavior). Rename to `open_in_agent_and_maybe_close_detail` (or similar) so the call-site routing in the `case` block reads honestly.
- [ ] `MARKER_SUMMARY_PATTERN = /\A[A-Z][A-Z0-9_]+(?:\s+[a-z_]+=\S+)*\z/` (`lib/hive/tui/views/red_status_detail.rb:30`) requires at least two uppercase chars in the marker name (one for `[A-Z]`, one+ for `[A-Z0-9_]+`); single-letter markers ("`E pass=1`") would bypass the leak guard. Cheap to widen to `[A-Z][A-Z0-9_]*` so the detection matches `TaskAction#marker_summary`'s output regardless of marker-name length.
- [ ] `red_status_detail_message` flashes the verbose `red_status_detail_hint` ("press Enter to recover, o to open in agent, Esc / q to close") for `f` and `R` (`key_map.rb:320`) while `s` gets the targeted "press o for Open in agent" (L316). For consistency with the plan Risk #1 mitigation ("`f`-press fires a diagnosable signal"), give `f`/`R` the same targeted "press o for Open in agent" / "press Enter to Recover" wording so muscle-memory drift surfaces the exact replacement, not the full keymap.
- [ ] `red_status_autofix_force?` (`lib/hive/tui/bubble_model.rb:927-931`) is reachable from the detail screen only when a snapshot poll mutates the row into a `max_passes_stale_with_escalations?` shape mid-session; on the open path `red_detail_row?` short-circuits those rows to false. The current behavior is correct, but worth a one-line comment noting the only live entry point is the snapshot-poll mutation case so a future reader doesn't conclude the helper is dead.
- [ ] `[o]     Open in agent — …` (`lib/hive/tui/views/red_status_detail.rb:60`) uses five spaces to visually align with `[Enter] Recover …`; if `Enter` is ever renamed (e.g. `Return`) or the key letters change, alignment will drift silently. Prefer `Lipgloss`-style column padding or a small helper so the alignment stays a single source of truth.

ivankuznetsov:

## High
- [ ] `wiki/commands/tui.md:78` still documents the removed `f`/`R` red-detail bindings and `$EDITOR` flow: the project wiki is an authoritative surface per AGENTS/CLAUDE, and it now contradicts the two-action Recover/Open-in-agent contract.

## Medium
- [ ] `lib/hive/tui/views/red_status_detail.rb:101` does not catch all config-read failures for the fallback agent label: `Hive::Config.load` can still raise `Psych::DisallowedClass` or `Errno::*`, so malformed or unreadable project config can abort opening the detail screen instead of showing "your project's development agent".
- [ ] `test/unit/tui/help_test.rb` and `test/unit/tui/bubble_model_test.rb` are missing plan-prescribed regression coverage: Unit 3's exact red-detail help binding assertion and Unit 5's successful Open-in-agent close-path test are not pinned.

## Nit
- [ ] `lib/hive/tui/bubble_model.rb:898` still tells operators to "use the shell command from details": the redesigned detail screen no longer renders that command or stays open on refusal, so this fallback copy is stale and unactionable.

ivankuznetsov:

## High
- [ ] [reliability] resolve_agent_label rescue list misses Errno::* and Psych::DisallowedClass — `lib/hive/tui/views/red_status_detail.rb:101-103` rescues `Hive::ConfigError, Hive::AgentProfiles::UnknownAgent, Hive::AgentError, Hive::InvalidTaskPath, Psych::SyntaxError`. `Hive::Config.load` can raise `Errno::EACCES` (chmod 000 config.yml), `Errno::EISDIR` (config-as-directory), `Errno::ENOENT` (TOCTOU race), `Psych::DisallowedClass` (hand-edited `!ruby/object:` tag), `Psych::BadAlias` — none of these are subclasses of the rescued set, so a hand-edited or permission-denied config.yml crashes the TUI on Enter into the detail screen.
- [ ] [project-standards] wiki/commands/tui.md still documents the removed [f]/[R] red-status detail bindings — `wiki/commands/tui.md:78,158-160` describe `Enter autofix / retry  f manual fix  R refresh diagnosis` as the detail-mode contract; the new contract is `[Enter] Recover  [o] Open in agent`. CLAUDE.md mandates wiki updates whenever a decision changes; the PR contains no wiki edits.
- [ ] [project-standards] wiki/log.md not appended for the red-status detail contract change — `wiki/log.md` newest entry is 2026-05-22T13:30:00Z, predating these commits; CLAUDE.md mandates an append on every learning / decision.
- [ ] [project-standards] wiki/modules/diagnosis_agent.md describes a TUI entry point that no longer exists — `wiki/modules/diagnosis_agent.md:10,63` says diagnosis_agent is "Invoked by … the TUI red-status detail view's R key" and details `refresh_red_status_diagnosis` + `@diagnosis_inflight`, all removed by this PR. The wiki is now factually wrong about the integration.
- [ ] [maintainability] Stale code comment in `lib/hive/task_action.rb:444` references deleted `FOOTER_NO_ENTER` constant — comment block on `diagnostic_action?` still says "see red_status_detail.rb#FOOTER_NO_ENTER and suggested_next_action_payload below — manual_fix / command:null"; the constant and the per-action_key footer branching it pointed at are gone. Misleads the next reader about what the view does.

## Medium
- [ ] [correctness] Footer truncated below ~50 cols hides `[Esc] back` affordance — `lib/hive/tui/views/red_status_detail.rb:15-19,39`: the border branch fires at `cols >= 40`, but the joined `FOOTER` is ~52 chars. With `cols=40` inner_width=36, `truncate(FOOTER, inner_width)` drops `[Esc] back` (and possibly `Open in agent`). Keys still work, but the screen's "consistent on-screen reference for the action keys" goal regresses on narrow terminals; the footer-survives-short-rows test pins cols=80 only.
- [ ] [correctness] `MARKER_SUMMARY_PATTERN` false-positive on legitimate all-caps short verdicts — `lib/hive/tui/views/red_status_detail.rb:23`: regex `/\A[A-Z][A-Z0-9_]+(?:\s+[a-z_]+=\S+)*\z/` matches bare tokens of length ≥2 (e.g., `ABORTED`, `TIMEOUT`, `AGENT_TIMEOUT`), silently swallowing a future short verdict and replacing it with the "no diagnosis yet" fallback.
- [ ] [reliability] `Hive::Config.load` runs synchronously on the dispatch thread when opening detail — `lib/hive/tui/update.rb:700-709` calls `resolve_agent_label` inline; a slow NFS / stalled FUSE / disk-pressure read freezes the bubbletea Update loop indefinitely with no timeout. Plan Risk #4 documents caching for the render path, but the OPEN keystroke itself is still on the synchronous path.
- [ ] [reliability] Flash for `s`/`f`/`R`/verb keys in `:red_status_detail` mode is invisible until the operator Esc back to grid — `lib/hive/tui/key_map.rb:313-329` returns `Messages::Flash` for these refusals; `Views::RedStatusDetail.render` never consults `model.flash`. So pressing `s` muscle-memory-from-grid silently buffers the flash; the operator only sees "press o for Open in agent" after they back out — by then the muscle-memory loop has already corrected. This defeats the explicit-refusal contract documented in the keymap comments.
- [ ] [maintainability] `RedStatusDetailState#marker_signature` is now dead data — written but never read — `lib/hive/tui/model.rb:89-93`: removing `acknowledged_marker_signature` deleted the only consumer of `marker_signature` in `Update.apply_snapshot_arrived`. The field is still required on init, still recomputed on every snapshot poll (`update.rb:185-186, 705`), and still asserted by `test/unit/tui/update_test.rb:139` — but no production code reads it. Either drop the field or document the intended future reader.
- [ ] [maintainability] `_from_detail` wrappers are mis-named — `open_in_agent_from_detail` also runs on grid-mode `s` press — `lib/hive/tui/bubble_model.rb:383,874-891`: `Messages::OpenInAgent` comes from both grid `s` and detail `o`; the dispatcher routes both through the `_from_detail` wrapper, which inspects `model.mode` and short-circuits when not `:red_status_detail`. The name lies about the call site; either rename (`dispatch_open_in_agent_then_close_detail`) or inline the four-line wrapper.
- [ ] [maintainability] `ACTION_KEYS` exposes both rendering surface AND flash hint copy — cross-module coupling between view and key_map — `lib/hive/tui/key_map.rb:5` now `require`s `hive/tui/views/red_status_detail` and reaches into a view constant to build flash text. KeyMap is the input layer, the view is the output layer; the shared concern is the key-binding spec, not the footer rendering. Extract a tiny `RedStatusDetailKeys` module both depend on instead.
- [ ] [maintainability] `resolve_agent_label` is filesystem/config logic living inside a presentation module — `lib/hive/tui/views/red_status_detail.rb:94-103` loads `Hive::Config`, instantiates `Hive::Task`, looks up an `AgentProfile`. Views::RedStatusDetail is `module_function` glue; this belongs in `Update` (or a sibling resolver). The view module shouldn't need `require "hive/config"` + `"hive/task"` + `"hive/agent_profiles"`.
- [ ] [maintainability] `agent_label:` defaults to nil with view-side fallback — duplicate fallback contract — `lib/hive/tui/model.rb:90` defaults to nil; `lib/hive/tui/views/red_status_detail.rb:60` re-applies `state.agent_label || AGENT_FALLBACK` on every render. Two valid representations of "no resolved label" — either default the keyword to `AGENT_FALLBACK` or require callers to always pass a value.
- [ ] [maintainability] `MARKER_SUMMARY_PATTERN` sanitization belongs upstream, not in the view — `lib/hive/tui/views/red_status_detail.rb:30`: detecting that `diagnostic["summary"]` is a `TaskAction#marker_summary`-shape string is a data-quality concern, not a rendering concern. CLI/bot/JSON consumers reading the same artifact emit the raw debug copy; the TUI silently lies about what's in the file. Move the fallback to the producer (`task_action.rb` / `diagnosis_agent.rb`).
- [ ] [project-standards] `docs/solutions/architecture-patterns/red-status-diagnose-then-act-2026-05-16.md:102-113` prescribes the removed `[f]/[R]` action surface — pattern doc still lists the canonical actions as "Enter: existing autofix/retry path; f: open worktree in $EDITOR; R: refresh diagnosis via the headless command"; stale and will mislead the next implementer.
- [ ] [project-standards] `wiki/decisions.md` ADR-027 (~lines 306-319) still cites marker_signature freshness behavior that no longer fires in the TUI — `Update.apply` no longer flashes "marker changed since you opened this view — refresh (R)"; `RedStatusDetailState` lost `acknowledged_marker_signature` and `refreshing`. ADR-027's operator-visible refresh affordance presupposition is gone.
- [ ] [testing] No test asserts screen-close on happy-path Recover / Open-in-agent from detail mode — `test/unit/tui/bubble_model_test.rb:2225-2253,3280-3303` only pin refusal branches. Happy path (`recover_review` with marker clear, `OpenInAgent` with worktree present) never asserts `mode == :grid` + `red_status_detail_state.nil?` together with the returned Cmd. A regression that returns the Cmd but forgets to close (early return bypassing the wrap) would not be caught.
- [ ] [testing] No test for grid-mode pass-through of `close_red_status_detail_on_dispatch` — `lib/hive/tui/bubble_model.rb:873-882` early-returns `[model, cmd]` when `mode != :red_status_detail`. The grid-mode `s` and grid-mode `RedStatusAutofix` (if any) paths now route through this wrapper but no test asserts that cursor/scope/filter are preserved when called from grid mode.
- [ ] [testing] `MARKER_SUMMARY_PATTERN` regex has no positive-summary edge case test — `test/unit/tui/views/red_status_detail_test.rb:129-139` pins only the suppression case (`REVIEW_ERROR phase=fix pass=2`); no test verifies a legitimate sentence summary (e.g., "Build failed: out of memory") survives unmodified, nor a bare upper-case verdict like `ABORTED`.
- [ ] [testing] `test_red_status_detail_s_key_flashes_use_o_hint` assertions are weak — `test/unit/tui/key_map_test.rb:729-732` asserts `\bo\b` + case-insensitive "open in agent"; both pass against either the s-key specific text (`"press o for Open in agent"`) or the generic `red_status_detail_hint`. Pin the exact text so the s-key vs. f/R-key divergence is load-bearing.
- [ ] [testing] Stray `SubprocessExited` with `verb == "status"` no longer covered after refresh-diagnosis removal — `lib/hive/tui/bubble_model.rb:664-672`: the prior special-case branch was removed; the replacement behavior (fall through to standard `exit_code.zero?` short-circuit + generic diagnose flash on non-zero) has zero replacement coverage. Deleted tests `test_diagnose_subprocess_exit_*` left a gap.

## Nit
- [ ] [correctness] Agent label cached at open-time can go stale if config changes during the session — `lib/hive/tui/views/red_status_detail.rb:89-103`: if the operator edits `.hive-state/config.yml`'s `execute.agent` while the screen is open, the label diverges from what `[o] Open in agent` will actually launch (`open_in_agent` re-reads config at dispatch). Short-lived screen, but residual risk.
- [ ] [correctness] `open_in_agent_from_detail` returns a foreground takeover Cmd alongside `mode=:grid` model — `lib/hive/tui/bubble_model.rb:874-891`: brief flash of grid → suspended terminal → grid on return. Plan-mandated, but worth pinning in a test exercising the full open/o/exit cycle.
- [ ] [correctness] Stale wiki reference to deleted `RefreshRedStatusDiagnosis` / `@diagnosis_inflight` — `wiki/modules/diagnosis_agent.md:10,63` (additional staleness already raised at High by project-standards).
- [ ] [maintainability] `ACTION_KEYS`' Esc/q hint string is duplicated and inconsistent — `lib/hive/tui/views/red_status_detail.rb:17-21`: stated purpose is "single source so a key rename in one surface cannot drift away from the other"; footer says `[Esc] back` but hint says `Esc / q to close` — already drifted on day one.
- [ ] [maintainability] Footer-pin test asserts a literal containing three spaces — fragile — `test/unit/tui/views/red_status_detail_test.rb:174` pins `"[Enter] Recover   [o] Open in agent   [Esc] back"` exactly; any future tweak to `ACTION_KEYS`' join breaks it cosmetically.
- [ ] [testing] `test_renders_with_border_on_wide_terminals` couples to Lipgloss box-drawing glyphs — `test/unit/tui/views/red_status_detail_test.rb:157-167` pins `╭` / `╰`; a Lipgloss theme/capability swap to ASCII corners breaks the test without changing the "border is drawn" contract.
- [ ] [testing] `with_singleton_method_stub` doesn't fully restore the target — `test/unit/tui/views/red_status_detail_test.rb:207-213` re-installs the captured `Method` as a *new* singleton method rather than removing the stub. Cross-test contamination possible if another test depends on `singleton_class.method_defined?(:load)` semantics; use `remove_method` in ensure.
- [ ] [testing] `agent_label`-missing test does not pin the literal fallback string — `test/unit/tui/views/red_status_detail_test.rb:65-71` asserts the copy "your project's development agent" rather than `RedStatusDetail::AGENT_FALLBACK`; rewording the fallback breaks the test even though the contract holds.
- [ ] [reliability] `close_red_status_detail_on_dispatch` cursor lands on adjacent row when snapshot rotated row X out — `lib/hive/tui/bubble_model.rb:877-891`: if the row recovers concurrently between Open and Recover, `reclamp_cursor` may land on whatever row now occupies the old index; not a regression vs. `apply_back_to_grid`, but worth noting since the operator just dispatched against a row that's no longer there.

PR #129, opened 2026-05-23
feat(claude): global claude.mode launch setting for every stage

ivankuznetsov:
## Summary

Before this PR, only `2-brainstorm` could launch Claude inside an interactive tmux session backed by the operator's logged-in account; every other Claude-driven stage (plan, execute, open-pr, review sub-spawns, artifacts, finalize) was locked to headless `claude -p` and its separate billing/auth path. After this PR a single project-global key, `claude.mode` (`tmux` | `headless`), governs every Claude launch in the pipeline, with sequential session sharing for `6-review` reviewer passes.

Fresh projects default to `tmux`. Existing projects with no key are treated as `tmux`. Switching is a config edit plus stage restart. When `claude.mode: tmux` is set on a host without a runnable tmux, the stage fails hard with `reason: "tmux_unavailable"` — no silent fallback to headless. Daemon and other non-TTY hosts must set `claude.mode: headless` explicitly (called out in `CHANGELOG.md` and `docs/adrs/030-global-claude-launch-mode.md`).

### How stages dispatch

```mermaid
graph TB
  Cfg[".hive-state/config.yml<br/>claude.mode"] --> Base["Hive::Stages::Base.spawn_claude!"]
  Stage2["2-brainstorm"] --> Base
  Stage3["3-plan"] --> Base
  Stage4["4-execute"] --> Base
  Stage5["5-open-pr"] --> Base
  Stage6["6-review reviewers"] --> Base
  Stage7["7-artifacts"] --> Base
  Stage8["8-finalize"] --> Base
  Base -->|claude profile| Launcher["Hive::ClaudeLauncher.launch!"]
  Base -->|non-claude profile| Spawn["Base.spawn_agent (headless)"]
  Launcher -->|claude.mode = tmux| Tmux["tmux session + Stop hook"]
  Launcher -->|claude.mode = headless| Spawn
```

`spawn_claude!` is the single funnel. Stages whose resolved `AgentProfile` is not Claude (e.g. `execute.agent: codex`) keep their existing headless path regardless of `claude.mode`.

### What changed

| Area | Change |
|------|--------|
| `lib/hive/config.rb` | New `claude.mode` key in `DEFAULTS`, `CLAUDE_MODES`, `validate_claude_mode!`, and `Config.claude_mode(cfg)` single read path. `claude` added to `HASH_SHAPED_KEYS`. |
| `lib/hive/claude_launcher.rb` | New shared module. `launch!` branches on mode; `with_shared_session` opens one tmux session and yields a `SessionHandle` whose `send_and_wait!` can be reused across reviewers. Lift-and-rename of the tmux scaffolding previously inlined in `brainstorm_tmux.rb`. |
| `lib/hive/stages/base.rb` | `spawn_claude!` helper; raises if a non-Claude profile reaches it. `spawn_claude_with_tmux_marker!` propagates any `Hive::AgentError` so Plan/OpenPr/Finalize/Artifacts surface a uniform `:error` marker. |
| Stage runners | `brainstorm`, `plan`, `execute`, `open_pr`, `artifacts`, `finalize` now route Claude through `spawn_claude!`. `brainstorm_tmux.rb` shrinks to a thin shim. Execute uses `:exit_code_only` under tmux so the runner-owns-`EXECUTE_COMPLETE` invariant survives. |
| `lib/hive/stages/review.rb` | Claude reviewers under `claude.mode: tmux` run sequentially inside one shared session per pass via `with_shared_session`; non-Claude reviewers stay headless. tmux-unavailable bubbles to the `:review_error` rescue rather than masquerading as per-reviewer infra errors. |
| `lib/hive/commands/init/prompts.rb` | `prompt_claude_mode` replaces the brainstorm-only runtime question; non-TTY default is `tmux`. |
| `lib/hive/commands/doctor.rb` | `claude/tmux` row is active whenever `claude.mode: tmux`; legacy `brainstorm.runtime` triggers a deprecation advisory. |
| `templates/project_config.yml.erb` | Renders top-level `claude.mode`; old `brainstorm.runtime` line is dropped from the live template. |

### Design decisions worth scrutinizing

- **One global key, not per-stage overrides.** The brainstorm answers were explicit: switching modes is intended to be rare and project-wide. `stages.<name>.claude.mode` is intentionally not supported. Future Claude-backed stages inherit the setting by calling `spawn_claude!`.
- **6-review uses one shared tmux session per pass, not one session per reviewer.** Reviewers in the same pass send sequential prompts into the same pane. This matches the operator's "watch the same pane" intent and minimizes setup/teardown. `prepare_claude_session!` is re-invoked between reviewers as a safety reset. If session-state contamination shows up in smoke testing, the fallback is one session per reviewer (still sequential); the launcher API already accommodates it.
- **Hard fail when tmux is missing, no silent fallback.** Daemon/service hosts that cannot run tmux must set `claude.mode: headless`. `hive doctor` surfaces a `claude/tmux` row and a clear failure mode rather than letting the daemon trip on every task.
- **`brainstorm.runtime` deprecated, dual-read for one release.** When `claude.mode` is unset but `brainstorm.runtime: tmux_interactive` is present, brainstorm still honors the legacy value. `hive doctor` warns and points the operator at the new key. The `HIVE_BRAINSTORM_TMUX_*` env-var aliases are kept on the same removal timeline.
- **Stage-aware `--allowedTools`.** `wrapper_command` takes the allowed-tools and add-dirs per call site. Brainstorm/plan keep the narrow `Read,Write,Edit,LS` set via `Hive::ClaudeLauncher::PLANNER_ALLOWED_TOOLS`; execute and downstream stages use `IMPLEMENTER_ALLOWED_TOOLS` (`Read,Write,Edit,Bash,LS,Glob,Grep`). No new policies invented — both sets mirror what each stage already used headless.
- **tmux result envelope matches headless.** `ClaudeLauncher.launch!` returns the same `{status:, final_message:, final_message_source:, log_label:, ...}` shape `Hive::Agent` returns, so research-mode 4-execute and the `## Execute Output` block in `task.md` behave identically across modes.

See `docs/adrs/030-global-claude-launch-mode.md` for the recorded rationale.

### Migration

1. Run `hive doctor`. The advisory row flags any legacy `brainstorm.runtime` key.
2. Replace it with the new top-level block:
   ```yaml
   claude:
     mode: tmux  # or headless
   ```
3. Restart any in-flight stage; the next Claude launch picks up the new value.

## Test plan

- [x] `bundle exec rake test` (full suite): `2562 runs, 9010 assertions, 0 failures, 0 errors, 2 skips`.
- [x] New unit coverage: `test/unit/claude_launcher_test.rb`, `test/unit/stages/claude_mode_dispatch_test.rb`, `test/unit/stages/review/run_reviewers_test.rb`.
- [x] Doctor, init prompt, and brainstorm-runtime tests updated to cover the new key, the dual-read fallback, and the deprecation advisory.
- [x] Per-stage `--allowedTools` and tmux session-name matrices asserted across brainstorm/plan/execute/open-pr/finalize/artifacts/review-fix/triage/ci-fix/browser-test/shared-reviewer-session.
- [x] Shared-session mid-pass reviewer failure + `:all_failed` path covered; `:review_error reason="tmux_unavailable"` rescue covered directly.
- [x] `hive doctor` exercised against three config shapes: fresh (`claude.mode: tmux` default), legacy (`brainstorm.runtime` only), explicit headless.

## Review summary

Two review passes ran (one codex `ce-code-review`, one `pr-review-toolkit` multi-agent sweep per pass).

### Pass 1

- **High — tmux 4-execute could not signal completion.** Execute pinned `:state_file_marker` but the execute prompt forbids the agent from writing markers, so every tmux run timed out. Fixed by switching to `:exit_code_only` under tmux while keeping the runner-owns-`EXECUTE_COMPLETE` invariant.
- **High — Stop hook installed outside Claude's settings lookup path.** Worktree-cwd stages installed `.claude/settings.json` under the task folder; Claude never found it, so `.done`/`result.json` never landed. Fixed by installing settings in cwd as well.
- **High — tmux-unavailable was downgraded to a browser warning.** `spawn_claude!` rescued tmux preflight failures locally, so review browser-test treated them as soft warnings. Fixed by propagating `Hive::AgentError` so R7's hard-fail contract holds.
- **High — `AGENT_WORKING` marker not stamped under tmux.** Added the marker write before `send_prompt` so status/TUI/daemon recovery match headless.
- **Medium/Nit cluster (resolved).** Migration string had a literal `\n`; `tmux_unavailable_error?` was too broad; `wait_for_done_signal` ignored `result.json`; ensure-block cleanups swallowed exceptions; reviewer marker dropped `e.message`; `sweep_orphan_processes` swallowed `ENOENT`; `cleanup_scratch` swallowed `EACCES`; init/CLI docs were stale; brainstorm headless threat-model comment was mis-attributed.
- **One escalation (Q1) resolved in-pass** by choosing Option A (tmux execute → `:exit_code_only`).

### Pass 2

- **High — research-mode 4-execute regressed under tmux** because the tmux result envelope omitted `:final_message` / `:final_message_source`. `Execute.append_implementation_output` silently no-opped for the same reason. Fixed by populating the same keys in both modes (envelope parity); a stubbed-launcher integration test now exercises `Execute.run_pass` end-to-end under `claude.mode: tmux` to keep parity in CI.
- **High — non-tmux `Hive::AgentError`s leaked from the launcher** (session-start failures, ready-prompt timeouts, pane inspection errors) past Plan/OpenPr/Finalize/Artifacts which had no rescue, leaving stale `AGENT_WORKING` markers. Fixed via a uniform wrapper that writes `:error` with attributed reason and clears `AGENT_WORKING`.
- **Medium cluster.** Pane-unreadable timeouts were being overwritten by generic `timeout`; shared-session `prepare_claude_session!` window was too tight against long-running reviewers; `explicit_claude_mode?` fallback could lie when callers synthesised cfgs; the wide allowed-tools string was duplicated across 11 call sites and consolidated into `IMPLEMENTER_ALLOWED_TOOLS` / `PLANNER_ALLOWED_TOOLS`.
- **Nits resolved.** Dead `tmux_session_name` fallback removed; `ORCHESTRATOR_OWNED_PREFIXES` deduplicated; `HIVE_BRAINSTORM_TMUX_*` removal timeline pinned in CHANGELOG/ADR; daemon-must-set-`headless` callout added; collision test weakened to match its actual invariant.
- **Resolved/no-fix.** `BrainstormTmux` retirement is plan-deferred (kept as a thin shim per U2 dual-read); `runtime_for` non-Claude `:headless` mapping matches plan U7's "no warn for internally consistent configs"; one pr-review-toolkit comment claim about `wiki/log.md` deletions was factually incorrect against the diff.

Two reviewer infra failures (`claude-ce-code-review` exit_code=1 in both passes) are recorded in `errors-0{1,2}.md` and were treated as infra noise; both passes recorded `fix-success` sentinels.

## Linked task

- Task slug: `we-added-tmux-mode-for-260519-69be`
- Hive pipeline state: `.hive-state/stages/8-finalize/we-added-tmux-mode-for-260519-69be/`
- ADR: `docs/adrs/030-global-claude-launch-mode.md`
- Plan: `plan.md` in the task folder
- Brainstorm: `brainstorm.md` in the task folder

---


ivankuznetsov:

## High

## Medium
- [ ] Per-reviewer log files vanish in shared tmux session (`lib/hive/claude_launcher.rb:130-138`): `send_prompt_and_wait!` accepts `log_label:` but never writes a per-reviewer log; in headless mode `Hive::Agent` writes `logs/review-<name>-passNN.log`, so tmux-mode reviewers lose the structured per-reviewer log file used for debugging. Pane scrollback is the only artifact, and it is multiplexed across all reviewers in the pass.
- [ ] `prepare_claude_session!` ready-check is brittle across sequential reviewers (`lib/hive/claude_launcher.rb:282-318`): `claude_ready_prompt?` rejects any pane whose last 8KB tail contains the substring "Do you want to" — a phrase a reviewer's `/ce-code-review` output can legitimately produce in findings (e.g., quoting permission prompts) — causing the NEXT reviewer's prepare check to fail until timeout. Plan R1 mitigation (`prepare_claude_session!` between sends) does not address the false-positive on benign reviewer content.
- [ ] No safeguard against Claude context-window accumulation across reviewers (`lib/hive/stages/review.rb:828-847`): U6 runs three Claude reviewers sequentially in one session — each adds the full skill prompt + diff + skill output (~10–50KB) to a shared context. There is no `/clear` or context-reset between sends, so the third reviewer can land mid-auto-compact, distorting earlier reviewer state. Plan R1 acknowledges this but ships nothing to detect or bound it.
- [ ] No integration test exercises the real sequential send/wait cycle (`test/unit/stages/review/run_reviewers_test.rb:222-319`): `with_stubbed_claude_session` mocks `Hive::ClaudeLauncher.with_shared_session` whole, so `send_prompt_and_wait!`, `prepare_claude_session!` between reviewers, `wait_for_expected_output`, and `reset_signal_files` are never exercised in the multi-reviewer path. The biggest behavioural change in this PR has no test coverage of the integrated flow.
- [ ] Concurrent / orphaned tmux session error is not classified as tmux_unavailable (`lib/hive/claude_launcher.rb:26-32, 184-192`): "tmux session ... already exists" is not in `TMUX_UNAVAILABLE_PATTERNS`, so the helpful recovery hint ("attach with `tmux attach -t ...` or run `tmux kill-session -t ...`") in the AgentError message gets re-raised, lands as `runner_exception exception_class=Hive::AgentError` in the REVIEW_ERROR marker attrs, and the operator loses the recovery instructions in the marker payload.
- [ ] Explicit-key tracking has two code paths with no Config.load coverage (`lib/hive/config.rb:254-292`, `test/unit/stages/brainstorm_runtime_test.rb`): `explicit_claude_mode?` / `explicit_brainstorm_runtime?` use `EXPLICIT_*_KEY` set by `Config.load`, but unit tests pass raw cfgs that exercise only the `cfg.dig(...) != nil` fallback. A bug that broke EXPLICIT_*_KEY population in `Config.load` would not fail any test, while production routing depends on it.

## Nit
- [ ] `DEFAULTS["brainstorm"]["runtime"] = "headless"` still rendered in defaults (`lib/hive/config.rb:61-65`): keeps the deprecated key live in every merged cfg even though templates no longer write it; consider scheduling removal alongside the doctor advisory's deprecation window so the surface area shrinks.
- [ ] tmux_unavailable classification is regex-based on message text (`lib/hive/claude_launcher.rb:26-32, 222-227`): a future reword of any preflight error silently breaks the classification. A typed `Hive::TmuxUnavailableError < Hive::AgentError` (raised inside `preflight_tmux!` / `wait_until_session_exists!`) would be more robust than `TMUX_UNAVAILABLE_PATTERNS`.
- [ ] `safe { ... }` in `with_shared_session`'s ensure swallows StandardError without logging (`lib/hive/claude_launcher.rb:122-127, 523-527`): a failed `runner.kill_session` (tmux server crash, EACCES) silently leaves an orphan session that will collide with the next pass's preflight; piping the error to `orphan_sweep_log_path` would surface the cause.
- [ ] `cfg_with_claude_mode` shallow-merge produces a cfg where `explicit_claude_mode?` flips to true (`lib/hive/stages/brainstorm.rb:99-104`): no current caller re-checks `explicit_claude_mode?` on the merged hash, but future code that does (e.g., reviewer dispatch helpers) will see the synthesised mode as user-explicit. Consider also copying / clearing the `EXPLICIT_CLAUDE_MODE_KEY` so the helper preserves the original "was the user explicit?" signal.
- [ ] `"tmux session ... did not start"` may be transient rather than unavailable (`lib/hive/claude_launcher.rb:31, 257-264`): startup failure under load (OOM during fork, busy disk) is classified as a permanent tmux-unavailable error and writes `:error reason=tmux_unavailable`, when a retry might succeed. Low impact in practice.

ivankuznetsov:

## High
- [ ] tmux state-file launches can return stale terminal markers: `lib/hive/claude_launcher.rb:130` sends prompts without first replacing an existing `WAITING`/`COMPLETE` marker with `AGENT_WORKING`, while `marker_from_sentinel_tail` trusts the current marker if it appears in the pane tail; brainstorm/plan prompts contain marker examples, so a resume can finish immediately on the previous marker before Claude updates the file.
- [ ] 4-execute in tmux mode times out after successful Claude implementations: `lib/hive/stages/execute.rb:215` sends execute through `:state_file_marker`, but `templates/execute_prompt.md.erb:20` tells the implementer not to write `EXECUTE_COMPLETE`; the tmux launcher therefore ignores the Stop-hook `.done` signal unless a terminal marker appears and eventually writes a timeout error.

## Medium
- [ ] tmux launches do not write standard per-spawn logs: `Hive::Agent` creates `<task.log_dir>/<log_label>-<timestamp>.log`, but `lib/hive/claude_launcher.rb:97` starts tmux sessions without creating an equivalent log file, leaving tmux-mode stages without the plan-required identical log location or normal TUI/log-tail diagnostics.

## Nit

ivankuznetsov:

## High
- [ ] [code-reviewer] tmux-unavailable misclassification across review sub-spawns (`lib/hive/stages/base.rb:188-196` + `lib/hive/stages/review.rb:1192,1196`, `review/triage.rb:88-99`, `review/ci_fix.rb:361-371`, `review/browser_test.rb:131-140`): the `:tmux_unavailable` marker written by `spawn_claude!` is overwritten by each caller's generic `:review_error reason=fix_failed/triage_failed/browser_unexpected`, defeating plan R7 "hard-fail with a clear error" — operators see the wrong reason.
- [ ] [silent-failure-hunter] `safe { ... }` helper at `lib/hive/claude_launcher.rb:523-527` swallows EVERY teardown failure (`kill_session`, `sweep_orphan_processes`, `cleanup_scratch`, `cleanup_done`) with no logging: leaked tmux sessions and leaked Claude PIDs leave no diagnostic trail.
- [ ] [silent-failure-hunter] `sweep_orphan_processes` at `lib/hive/claude_launcher.rb:443-456` rescues `Errno::ENOENT` and returns silently — on hosts lacking `pgrep`/`pkill` every Claude child leaks per teardown with no surfaced warning to the operator.
- [ ] [silent-failure-hunter] `record_claude_pid` at `lib/hive/claude_launcher.rb:266-280` swallows `Hive::TmuxError` and gives up silently on timeout; the PID is load-bearing for orphan cleanup so a missing capture should at minimum warn.
- [ ] [silent-failure-hunter] `marker_from_sentinel_tail` at `lib/hive/claude_launcher.rb:350-359` swallows `Hive::TmuxError` from `capture_pane_tail`; a dead tmux session is masked as a generic poll timeout in `wait_for_terminal_marker`.
- [ ] [silent-failure-hunter] `wait_for_expected_output` at `lib/hive/claude_launcher.rb:367-372` swallows pane-capture failures during the ready-prompt check and returns the misleading `"expected output file missing or empty"` envelope when the real cause is a dead tmux session.
- [ ] [silent-failure-hunter] `wait_for_terminal_marker` at `lib/hive/claude_launcher.rb:320-344` unconditionally sets a `:error reason=timeout` marker on poll exhaustion, potentially clobbering a real Claude-written `:error reason="agent_failed"` and returning it indistinguishably to callers.
- [ ] [silent-failure-hunter] `log_orphan_sweep`, `log_orphan_sweep_warning`, and `rotate_orphan_sweep_log` at `lib/hive/claude_launcher.rb:462-498` end in `rescue StandardError; nil` — if disk logging fails (ENOSPC/EROFS/EACCES) the orphan-sweep audit trail vanishes silently, defeating the only diagnostic for a best-effort cleanup.
- [ ] [pr-test-analyzer] no test covers `Hive::ClaudeLauncher.launch!` happy-path tmux flow end-to-end (plan U2 AC "tmux mode opens session + polls for marker + tears down"); `claude_launcher_test.rb` only covers dispatch/predicate paths, leaving the central spawn point of every Claude stage uncovered against a regression in `with_shared_session`/`send_and_wait!`/teardown.
- [ ] [pr-test-analyzer] `test/integration/run_brainstorm_tmux_test.rb:11` conditionally skips when tmux is missing — direct violation of project rule "NEVER skip tests conditionally based on environment availability"; combined with the gap above, the tmux happy path can regress on a tmux-less CI runner with zero red tests.
- [ ] [pr-test-analyzer] no test that `ClaudeLauncher.preflight_tmux!` actually raises `Hive::AgentError` with the expected message shapes for missing-tmux / too-old-tmux / not-runnable-binary cases — a wording drift would silently bypass `tmux_unavailable_error?`'s regex match in `spawn_claude!` rescue.
- [ ] [pr-test-analyzer] `run_reviewers_test.rb::test_claude_reviewers_share_one_tmux_session_when_claude_mode_tmux` stubs `with_shared_session` so the production path through `SessionHandle#send_and_wait!` → `prepare_claude_session!` is never executed — a regression that drops the per-prompt contamination-reset between sequential reviewers would not be caught (the very U6 risk R1 the plan called out).
- [ ] [comment-analyzer] `wiki/modules/config.md:43-48` DEFAULTS quote is materially incomplete — omits `brainstorm.skill`, the full `plan.skill_by_agent` hash, and other present keys; will mislead maintainers about the defaults shape.
- [ ] [comment-analyzer] `wiki/state-model.md:184-235` per-project YAML example uses old `pr:` key and omits `artifacts`/`open_pr`/`finalize` budget+timeout entries; doc and code disagree on the documented config shape.

## Medium
- [ ] [code-reviewer] `lib/hive/stages/brainstorm_tmux.rb` is now 157 lines of pure 1-line delegation methods with no header comment explaining "compatibility shim retained for one release"; either annotate or delete since it's dead weight.
- [ ] [code-reviewer] Eight separate sites (`stages/{brainstorm,plan,execute,open_pr,artifacts,finalize}.rb`, `review.rb:1191-1201`, `review/{triage,ci_fix,browser_test}.rb`) duplicate the same `if profile.name == :claude then spawn_claude!(...) else spawn_agent end` branch with subtle per-site kwarg differences — extract `Hive::Stages::Base.spawn_for_profile(...)` to close drift risk.
- [ ] [code-reviewer] `prepare_claude_session!` at `lib/hive/claude_launcher.rb:282-307` will spam `runner.send_keys("Enter")` ~120 times across the 30s budget on a sticky/repeating trust prompt; add a max-Enter-presses counter so operators see "trust prompt did not clear after N attempts" instead of a generic timeout.
- [ ] [code-reviewer] `lib/hive/stages/review.rb:828` `shared_claude_reviewer_session?` re-evaluates `Hive::Config.claude_mode(cfg)` per spec — cache once per pass, and the helper's `specs` arg is unused by the per-spec path.
- [ ] [silent-failure-hunter] `cleanup_expected_output` at `lib/hive/claude_launcher.rb:439-441` only rescues `Errno::ENOENT`; `Errno::EACCES` or other IO errors will crash the teardown — inconsistent with the surrounding "best-effort" pattern.
- [ ] [silent-failure-hunter] `tmux_unavailable_error?` in `stages/base.rb` pattern-matches `e.message` against regexes (`TMUX_UNAVAILABLE_PATTERNS`); fragile vs. message edits. A typed `Hive::TmuxUnavailableError` class would be safer.
- [ ] [silent-failure-hunter] `fix_success_fresh?` at `lib/hive/stages/review.rb:716-722` rescues `SystemCallError, IOError` and returns `true` — i.e. on a perms/FS-unmounted read failure the method LIES that the fix-success sentinel is fresh, advancing past a possibly-incomplete pass.
- [ ] [silent-failure-hunter] `collect_legacy_checked_escalations` and `parse_escalation_questions` at `lib/hive/stages/review.rb:1067-1123` silently treat permission errors as empty content/empty list — user's manually-ticked `[x]` answers can disappear with no warning.
- [ ] [silent-failure-hunter] `git_head` / `worktree_dirty?` at `lib/hive/stages/review.rb:1353-1361` fail-closed but emit no warning on underlying `git` failures; an `approval_head_mismatch` envelope can hide a corrupted worktree / missing git binary root cause.
- [ ] [silent-failure-hunter] Top-level `rescue StandardError` in `lib/hive/stages/review.rb:561-571` records `exception_class` but drops `e.message` — diagnosability suffers for cases like `Errno::ENOSPC`. Add `message: e.message[0, 500]` to marker attrs.
- [ ] [pr-test-analyzer] `Hive::Config.claude_mode(cfg)` has its own runtime raise (`lib/hive/config.rb:267`) reachable by callers that bypass `Config.load`; no unit test pins this behavior.
- [ ] [pr-test-analyzer] `claude_mode_dispatch_test.rb` never asserts that a non-claude stage agent + `claude.mode=tmux` skips the global tmux preflight at the spawn boundary (only via doctor surface); a future regression wiring tmux preflight into the shared base path would pass.
- [ ] [pr-test-analyzer] no test that a mid-run resume keeps using the claude launcher (dispatch should be sticky across resume) — a refactor re-resolving the launcher inside the loop could silently fall back to headless `spawn_agent`.
- [ ] [pr-test-analyzer] `run_reviewers_test.rb` only tests "tmux session not opened when wall-clock already exceeded"; the U6-required "open one session, mid-loop budget elapses → tear down before next reviewer" scenario is not covered.
- [ ] [pr-test-analyzer] no test for the re-raise branch of `spawn_claude!`'s rescue (non-tmux-unavailable `Hive::AgentError` passes through); broadening the rescue would silently mask real agent failures as `:tmux_unavailable`.
- [ ] [pr-test-analyzer] `doctor_test.rb` lacks a `refute_match(%r{claude/tmux}, ...)` for `claude.mode=headless` configs; dropping the headless guard at `lib/hive/commands/doctor.rb:83` would silently surface a stale tmux row.
- [ ] [comment-analyzer] `wiki/commands/doctor.md:33` lumps the `ANTHROPIC_API_KEY/CLAUDE_API_KEY` leak warning together with the legacy `brainstorm.runtime` advisory as if both are conditional on tmux mode, but the legacy-runtime advisory fires unconditionally whenever the key exists.
- [ ] [comment-analyzer] `HIVE_CLAUDE_TMUX_* → HIVE_BRAINSTORM_TMUX_*` env-var fallback at `lib/hive/claude_launcher.rb:423-427` is undocumented in CHANGELOG / README / wiki — silent comment-rot waiting to happen; document the back-compat or remove it.
- [ ] [comment-analyzer] `wiki/stages/artifacts.md:20` "If the current marker is already `COMPLETE`, return without a new commit action" is ambiguous — clarify to "short-circuits — no agent spawn, no new marker; returns `{ commit: nil, status: :complete }`".
- [ ] [comment-analyzer] `docs/adrs/030-global-claude-launch-mode.md:23` does not mention that tmux-unavailable surfaces as a `:tmux_unavailable` marker rather than a process exit — important for daemon operators reading the ADR for the hard-fail contract.

## Nit
- [ ] [code-reviewer] `SessionHandle` Struct at `lib/hive/claude_launcher.rb:34-47` exposes `task` and `runner` publicly; external callers could bypass orchestration. Hide internals behind a narrow API.
- [ ] [code-reviewer] `lib/hive/commands/init/prompts.rb:29-30` duplicates the mode list between `CLAUDE_MODES = Hive::Config::CLAUDE_MODES` and `CLAUDE_MODE_CHOICES = ["tmux", "headless"]`; alias to one source so UI cannot drift from validator.
- [ ] [code-reviewer] `READY_WAIT_TIMEOUT_SEC = 5` in `lib/hive/claude_launcher.rb:14-23` may be aggressive for cold-start (first run after auth refresh); consider raising default to 10s.
- [ ] [code-reviewer] `tmux_session_name` at `lib/hive/claude_launcher.rb:160-162` truncates to 250 bytes via `[0, 250]` without validating that the result remains tmux-legal (no `:` or `.` introduced); add a sanity check.
- [ ] [code-reviewer] `Hive::Config.claude_mode(cfg)` raises at runtime even though `validate_claude_mode!` validates at load time — two-tier validation is fine but worth documenting the invariant.
- [ ] [pr-test-analyzer] `claude_launcher_test.rb#test_headless_mode_delegates_to_base_spawn_agent` uses a leaky `Method#call` capture-wrap stub pattern; reapply the `instance_method`/`define_method` pattern from `run_brainstorm_tmux_test.rb:196-198`.
- [ ] [pr-test-analyzer] no test pins the `tmux_session_name` `[0, 250]` truncation cap (one-line addition).
- [ ] [pr-test-analyzer] `claude_mode_dispatch_test.rb` does not cover open-pr's non-claude branch (only claude profile is exercised) — both arms exist at `lib/hive/stages/open_pr.rb:36`.
- [ ] [comment-analyzer] `CHANGELOG.md:32` "Legacy `brainstorm.runtime` remains readable for one release" is slightly misleading — it's only honored as a fallback when `claude.mode` is absent; tighten phrasing.
- [ ] [comment-analyzer] `wiki/commands/init.md:58` "Codex/Pi stages remain on their normal headless profile path" implies stage-level switching when the actual dispatch is per-profile-resolution at each spawn site.
- [ ] [comment-analyzer] `lib/hive/stages/brainstorm_tmux.rb` shim lacks a top-of-file comment explaining "thin compatibility shim retained for one release; callers should target `Hive::ClaudeLauncher`".
- [ ] [comment-analyzer] The phrase "runtime is legacy read-back-compat" is duplicated across `wiki/modules/config.md:43`, `wiki/state-model.md:214`, and `wiki/stages/brainstorm.md:10` — pick one source of truth (likely ADR-030) and link the others.

ivankuznetsov:

## High
- [ ] 4-execute tmux mode can never complete cleanly: `spawn_implementation` waits for a state-file terminal marker, but the execute prompt tells the agent not to write `EXECUTE_COMPLETE`; `wait_for_terminal_marker` discards `.done` unless a terminal marker already exists, so clean exits time out.

## Medium
- [ ] tmux `:output_file_exists` ignores Claude stop-hook failure status: `wait_for_expected_output` returns ok whenever `.done` and a non-empty file exist without reading `result.json`, unlike the headless exit-code contract, so failed partial reviewer/triage/browser outputs can be accepted.
- [ ] Mixed review passes can open or fail tmux after the wall-clock budget is already exhausted: after non-Claude reviewers run, there is no budget check before `with_shared_session`, so an expired pass can report `tmux_unavailable` instead of `review_stale reason=wall_clock`.

## Nit

ivankuznetsov:

## High
- [ ] [code-reviewer] Research-mode 4-execute is broken under `claude.mode: tmux`: `Hive::ClaudeLauncher.wait_for_status` returns `{ status: marker.name, log_label: ... }` with no `:final_message` / `:final_message_source` (lib/hive/claude_launcher.rb:185-197), so `Execute.run_pass` falls into the `missing_research_output` branch whenever the agent did not commit (lib/hive/stages/execute.rb:166-169) — research tasks that worked headless now write `execute_waiting reason="missing_research_output"` instead of `execute_complete`, breaking R6 ("tmux and headless produce identical artifacts / completion signalling").
- [ ] [code-reviewer] `Execute.append_implementation_output` silently no-ops in tmux mode: it reads `result[:final_message]` (lib/hive/stages/execute.rb:282-298) but the tmux result envelope never carries that key (lib/hive/claude_launcher.rb:185-197). Operators upgrading to tmux lose the `## Execute Output` block in task.md and the long-tail context it gave reviewers — silent regression vs. headless, also a violation of R6.
- [ ] [silent-failure-hunter] `Stages::Base.spawn_claude_with_tmux_marker!` only catches the tmux-unavailable subset (lib/hive/stages/base.rb:202-213); any other `Hive::AgentError` from `ClaudeLauncher.launch!` — `tmux session … did not start`, `tmux session … already exists`, `claude interactive prompt did not become ready`, `could not inspect claude tmux session` — propagates uncaught through `Plan.run!`/`OpenPr.run!`/`Finalize.run!`/`Artifacts.run!`, which have no `rescue Hive::AgentError`. Pre-existing headless flows always returned `{status: :error, ...}` from `spawn_agent` so a stale `AGENT_WORKING` marker never lingered; the new tmux path leaves the marker on disk and surfaces as a generic crash via cli.rb. New failure mode introduced by this PR — needs either a stage marker or a uniform rescue in the wrapper.

## Medium
- [ ] [silent-failure-hunter] `wait_for_terminal_marker` loses the tmux-pane-unreadable cause on deadline: `marker_from_sentinel_tail` writes `:error reason="tmux_pane_unreadable"` then returns nil (lib/hive/claude_launcher.rb:429-439); the outer loop keeps polling and on `Time.now >= deadline` overwrites with `:error reason="timeout"` (lib/hive/claude_launcher.rb:409-412), masking the real root cause from the operator. Either break the loop on the tmux raise or skip the timeout marker when a prior `tmux_pane_unreadable` marker is already on disk.
- [ ] [code-reviewer] `prepare_claude_session!` is re-invoked before every shared-session reviewer send (lib/hive/claude_launcher.rb:156, lib/hive/stages/review.rb:865-883), but `claude_ready_prompt?` requires `❯` to be visible and `CLAUDE_READY_WAIT_TIMEOUT_SEC` defaults to 30s. If a long-running reviewer is still streaming when the next reviewer's send arrives (legitimate for 600s-timeout reviewers), `prepare_claude_session!` raises and the next reviewer is recorded as `:error` in `errors-NN.md`. The retry only re-runs `prepare`, not the previous claude work — needs either a longer per-send ceiling, or a "claude busy" detector that's distinct from "claude crashed".
- [ ] [type-design-analyzer] The tmux result envelope is structurally incompatible with the headless one — headless returns `{status:, final_message:, final_message_source:, ...}` (Hive::Agent), tmux returns `{status:, log_label:}` (lib/hive/claude_launcher.rb:185-197). Every caller has to either tolerate missing keys or branch on mode. Encode the contract as a `Hive::Agent::Result` data type with explicit nil-able fields, so the gap that produces High #1/#2 surfaces at the type boundary rather than as a silent caller-by-caller bug.
- [ ] [pr-test-analyzer] No test exercises the tmux result envelope. `test/unit/stages/claude_mode_dispatch_test.rb` only asserts dispatch parameters (session_name, allowed_tools, launcher kind); none of the Execute/research-mode codepaths are exercised under `claude.mode: tmux`. Add at least one integration that stubs `wait_for_terminal_marker` and walks `Execute.run_pass` end-to-end so High #1/#2 are caught in CI.
- [ ] [code-reviewer] `Hive::Config.explicit_claude_mode?` fallback always returns true after `Config.load`: the merged cfg carries `claude.mode` from DEFAULTS, so `cfg.dig("claude", "mode") != nil` (lib/hive/config.rb:282-286) cannot distinguish "user wrote it" from "default kicked in". The `EXPLICIT_CLAUDE_MODE_KEY` flag covers the load path, but in-process callers that synthesise a cfg directly (tests, daemon helpers) silently pretend the user was explicit. Tighten the fallback or document the contract.
- [ ] [code-simplifier] The allowed-tools string `"Read,Write,Edit,Bash,LS,Glob,Grep"` is duplicated in 8 sites (execute.rb:223, finalize.rb:107, open_pr.rb:86, artifacts.rb:41, review.rb:871/1248, review/triage.rb:95, review/ci_fix.rb:366, review/browser_test.rb:136); the narrow `"Read,Write,Edit,LS"` in 3 (brainstorm.rb:95, plan.rb:48, brainstorm_tmux.rb:33/68). Extract `Hive::ClaudeLauncher::IMPLEMENTER_ALLOWED_TOOLS` / `PLANNER_ALLOWED_TOOLS` so a policy change can't drift between stages — R4 in the plan specifically calls out the implementer/planner split.

## Nit
- [ ] [code-simplifier] `Hive::Stages::BrainstormTmux` is now a 158-line delegation shim (lib/hive/stages/brainstorm_tmux.rb); if nothing outside the test suite depends on it, fold the few unique behaviors into `Brainstorm` and retire the file.
- [ ] [code-simplifier] `Hive::ClaudeLauncher.safe` is dead inside the launcher — every production call site uses `safe_with_log` instead (lib/hive/claude_launcher.rb:679-683). Only the BrainstormTmux shim re-exports it; remove both once the shim goes.
- [ ] [pr-test-analyzer] `test_tmux_session_names_collide_on_shared_prefix_within_same_stage` pins a collision as a behavioural contract (test/unit/claude_launcher_test.rb:239-251). A future refactor that hashes the slug tail to avoid the collision (a reasonable mitigation) would have to delete this test even though the behaviour change is strictly safer. Weaken to "if the tail differs the names may collide on a 250-byte boundary" rather than asserting equality.
- [ ] [code-reviewer] `ORCHESTRATOR_OWNED_PREFIXES` is defined in both `Hive::Stages::Review` (lib/hive/stages/review.rb:61) and `Hive::Stages::Review::Triage` (lib/hive/stages/review/triage.rb:44). They must stay in lockstep; the current PR keeps them in sync, but the duplication is brittle. Reference the single source from Triage.
- [ ] [code-reviewer] `HIVE_BRAINSTORM_TMUX_*` env-var fallback in `tmux_env` (lib/hive/claude_launcher.rb:554-558) carries no removal timeline. CHANGELOG / ADR-030 should pin the version after which the legacy prefix is dropped, otherwise it ages indefinitely.
- [ ] [comment-analyzer] The plan's R10 says "auto-fallback tmux → headless is explicitly out of scope" — but neither CHANGELOG.md nor docs/adrs/030-global-claude-launch-mode.md state that operators running on daemon hosts MUST set `claude.mode: headless` (R2 in Risks). Without that callout, daemon operators upgrading from `brainstorm.runtime: headless` will hit the hard-fail on every task. A doctor-level warning when stdin is non-TTY would also help.
- [ ] [code-reviewer] `Hive::Stages::Base.spawn_claude!`'s session_name fallback (`Hive::ClaudeLauncher.tmux_session_name(task.stage_name, task)`) at base.rb:182 is dead in practice — every production caller passes an explicit `session_name:`. Either remove the fallback or document why `task.stage_name` is the right shape (e.g. for a future caller that wants the default).
- [ ] [code-reviewer] `tmux_unavailable_error?` regex `/tmux not runnable/` is a sub-pattern of `/tmux binary not runnable/` only by happenstance (different infix words); a future error string like "tmux runnable check failed" could match `/tmux .* runnable/` if a contributor "tidies" the patterns. Add a comment pinning the exact-message contract or anchor the patterns more tightly.

PR #131, opened 2026-05-23
feat(tui): redesign red-status recovery detail screen

ivankuznetsov:
# feat(tui): redesign red-status recovery detail screen

## Summary

The `:red_status_detail` mode rendered as an unstructured wall of text: a single header line, four plain `Project:` / `Stage:` / `Marker:` / `Status:` rows, the Q/A diagnosis, an artifact list (or `Artifacts: none`), and a footer of keybindings — none of it visually separated, with no log preview. Identifying the failure and picking an action took longer than it should have.

The screen now renders as a paneled layout: one color-coded header row, a bordered reason panel carrying the existing Q/A diagnosis, a bordered scrollable preview of the last 50 lines from the task's latest log, an optional compact artifacts list, and a separated action bar pinned to the bottom.

Recovery logic, the action set, and the existing `Enter` / `f` / `R` / `q` keybindings are unchanged. The one widening is six scroll-only keys for the new log panel (↑/↓ for 1 line, PgUp/PgDn for 10, Home/End to saturate). They don't change any existing affordance.

### What changed

- **Header bar.** One line of `RED · <project>/<stage> · <slug> · <worktree>` in bold red. Truncation drops the worktree first, then the slug, but never the `RED ·` prefix so the urgency cue always survives narrow terminals.
- **Reason panel.** The existing Q/A diagnosis is wrapped in a rounded border (`PANEL_BORDER`) with the relocated `Marker:` / `Status:` facts pinned above as a faint meta row.
- **Log preview panel.** Loads the last 50 lines via `Hive::Tui::LogTail::Tail` at mode entry, then closes the FD immediately. Snapshot, not live tail. Omitted entirely when no log exists (no empty bordered box). Title row shows `Log · last N of M lines (64 KiB tail · snapshot · R refreshes)` so the operator knows the contract.
- **Artifacts.** Compact list capped at 5 entries with `… (N more)` overflow. Section omitted entirely when the row has zero artifacts; the old `Artifacts: none` placeholder is gone.
- **Action bar.** Chips with `[Enter]` rendered in reverse-video and the rest in faint. Footer wording preserved verbatim so existing assertions and operator muscle memory still work. Wraps to two rows below ~60 cols.
- **Responsive composer.** Action bar reserves rows at the bottom; header reserves 1 at the top; the rest is split between reason and log via a shared `RedStatusDetailLayout` helper so the view and the scroll-clamp arithmetic stay in sync. Log panel is dropped first when the budget is tight; reason panel and action bar stay visible down to very short terminals.

### Design decisions

- **Snapshot at entry, not live tail.** The failing agent has already stopped writing — a `Tick`-driven refresh would just churn an FD. Pressing `R` re-enters the detail state and re-reads the log, which is the only path where new lines could appear. The panel title surfaces this contract.
- **Reuse `LogTail::Tail` instead of writing a fresh "read last N lines" helper.** The `:log_tail` view and the recovery panel render byte-identical lines, and the shared 64 KiB backbuffer ceiling that protects `:log_tail` from huge files protects this view too.
- **`f` keeps `OpenManualFix` ($EDITOR).** The brainstorm hinted at routing `f` to an agent for worktree-bound stages, but the out-of-scope clause froze the action set. This PR is presentation-only — action rewiring belongs in a separate brainstorm.
- **Capacity-aware scroll clamp.** `apply_red_status_detail_scroll` reuses `RedStatusDetailLayout` to compute the rendered log capacity from `model.cols`/`model.rows` (the same math the composer runs) before clamping the offset, so the operator can't scroll past visible data even on resize.

## Test plan

Full suite passes locally: `2571 runs, 9127 assertions, 0 failures, 0 errors, 2 skips`.

New unit coverage:

- **View** (`test/unit/tui/views/red_status_detail_test.rb`): header truncation order at widths 100/40/20, reason panel border, log slice + offset clamping, responsive reflow at 100×30 / 80×24 / 80×12 / 60×24 / 40×40 with the "last rendered line belongs to the action bar" invariant asserted at every size, chip composition with and without the `[Enter]` affordance, artifacts omission at 0/3/5/7 paths, and action bar single-row vs. wrap thresholds.
- **BubbleModel** (`test/unit/tui/bubble_model_test.rb`): `OpenRedStatusDetail` loads ≤50 lines from a 100-line fixture; missing log dir yields empty state without raising; an unreadable log (`chmod 000`, skipped under root) collapses to an empty panel; integration test pins composition order (header → reason → log → artifacts → action) at both 100×30 and 80×24.
- **Update** (`test/unit/tui/update_test.rb`): `RedStatusDetailScroll` decrement, up-saturation, down-clamp at offset 0, empty-log no-op, capacity recomputation on resize.
- **KeyMap** (`test/unit/tui/key_map_test.rb`): ↑/↓/PgUp/PgDn/Home/End in `:red_status_detail` mode emit `RedStatusDetailScroll` with the exact named-constant amounts.
- **Styles** (`test/unit/tui/styles_test.rb`): predicate asserts on `RECOVERY_HEADER_STYLE`, `ACTION_CHIP_PRIMARY`, `ACTION_CHIP_MUTED`, and `PANEL_BORDER`.

Manual verification: launched the TUI against a project with a red-status task and eyeballed the layout at 100×30, 80×24, 60×24, and 40×40 to confirm the action bar stays pinned and the log panel drops cleanly when the budget collapses.

## Review summary

Two automated review passes ran (Codex CE + pr-review-toolkit, plus claude-CE on pass 02 after pass-01 reviewer infra failure). Every finding was either auto-fixed in pass 02 or marked resolved with rationale:

- **Pass 01 fixes:** post-render slicing of the reason panel (now trims inside the panel before render), action bar wrapping at 80 cols (separator + threshold tightened so 80c stays single-row), PageUp/PageDown escape sequences (`\e[5~`/`\e[6~`) added to `InputDecoder`, scroll keys added to the in-app help overlay.
- **Pass 02 fixes:** breadcrumb on `InvalidTaskPath` now includes `row.folder`, snapshot-not-live design surfaced in the panel title, expanded filesystem rescue list (`EISDIR`, `ENAMETOOLONG`, `ELOOP`, `ENOTDIR`, `EBADF`, `EIO`) shared between `LogTail#poll!` and `red_status_detail_log_snapshot` via a new `LogTail::FILESYSTEM_RESCUE` constant, `Tail#close!` hardened so an FD that raced shut never escapes the ensure block, layout constants and `auto_action_available?` consolidated into the `RedStatusDetailLayout` module so view + update share one source of truth, refresh-diagnosis path now re-snapshots the log, action chip width math switched to raw text (no ANSI inflation), 64 KiB backbuffer extracted as a shared constant feeding both the snapshot read and the panel label, test coverage tightened (exact-boundary widths 40/20, 5-paths boundary, scroll-down-from-0 clamp, multi-size action-bar invariant, unreadable-log fixture).
- **Resolved without code change:** rebase-artifact findings from `codex-ce-code-review-02` and `pr-review-toolkit-02` (bot defaults, README quickstart, wiki/log entries) — `git diff <base>..HEAD` confirms the branch doesn't touch any of those paths; they disappear on rebase to current main. A handful of style nits resolved per CLAUDE.md (snapshot label placement subjective; root-only test skip is the standard chmod-can't-deny-root pattern; four debug breadcrumbs under helper-extraction threshold).

No high-severity findings remain open. Escalations files for both passes are empty (no user-blocking questions).

## Linked task

Hive task: `we-need-to-improve-this-260522-db23` (stage 8-finalize)

Origin idea: `we need to improve this screen (agent recovery) -- right now it's just a bunch of text, not formatted` (2026-05-22, screenshot in `assets/bug-1.png`).

---


ivankuznetsov:

## High

## Medium
- [ ] Short layouts can cut the rendered reason panel instead of trimming inside it (lib/hive/tui/views/red_status_detail.rb:94): `reason_height_budget` forces an 8-row panel even when less space remains, so `render` slices the already-rendered block and 80x12 output loses the bottom border plus the action Q/A.

## Nit
- [ ] The action bar wraps at 80 columns despite the plan reserving wrapping for narrow `<60c` layouts (lib/hive/tui/views/red_status_detail.rb:120): the four chip labels plus dot separators require 85 characters, so an 80-column terminal spends two footer rows and shrinks the log preview outside the intended narrow case.

ivankuznetsov:

## High

- [ ] [code-reviewer] `wrap_action_chips` measures styled (ANSI-inflated) strings — `red_status_detail.rb:122-138` calls `Styles::ACTION_CHIP_*.render(text)` then compares `candidate.length <= max_width`, so on a real tty escape bytes inflate the length: chips wrap prematurely while `Update.red_status_detail_action_rows:780` measures unstyled labels, desyncing reserved action-bar rows from rendered rows and corrupting scroll capacity.
- [ ] [code-simplifier] Cross-module duplication of red-status detail layout math and constants — `update.rb:759-831` mirrors `red_status_detail.rb:84-97` (height budget) plus `auto_action_available?` / `diagnostic_retry_command` / chip-label list / artifact-row count; constants `MIN_REASON_ROWS`/`MIN_LOG_PANEL_ROWS`/`MAX_LOG_PANEL_ROWS` exist verbatim at `red_status_detail.rb:20-22` and as `RED_STATUS_*` at `update.rb:27-29`; any drift desyncs Home/End/PgUp/PgDn clamping from the visible window (flagged again by code-reviewer Med, comment-analyzer Med).
- [ ] [silent-failure-hunter] `red_status_detail_log_snapshot` rescue list is narrower than `Tail#open!`'s actual error surface — `bubble_model.rb:1542` misses `Errno::EISDIR` (log dir is a directory), `Errno::ENAMETOOLONG`, `Errno::ELOOP`, `Errno::ENOTDIR`, so a malformed log path on Enter crashes the TUI rather than collapsing to an empty panel.
- [ ] [silent-failure-hunter] `Tail#close!` in the `ensure` block can re-raise after a partial `open!` failure — `log_tail.rb:265-272` only rescues `IOError`, so a partially-opened tail (e.g. `File.stat` raced to `ENOENT` after `File.open` succeeded) lets `Errno::EBADF`/`Errno::EIO` bubble through `bubble_model.rb:1546`'s ensure *after* the rescue already ran, crashing the very recovery surface the rescue was meant to keep alive.
- [ ] [silent-failure-hunter] Debug breadcrumb omits everything diagnostic — `bubble_model.rb:1543` logs only `e.class.name.split('::').last`, dropping the row slug, the resolved log path, and `e.message`; six months later the operator can't tell whether the empty panel came from no log, a permission issue, or which of `task.log_dir` vs the `logs/` fallback failed.
- [ ] [pr-test-analyzer] U6 "last rendered line is part of the action bar" invariant unasserted at 80×24, 60×24, 40×40 — `red_status_detail_test.rb:200,208` only pin it at 100×30 and 80×12, leaving the explicit safety net for lipgloss height drift uncovered at three of the five sizes the plan enumerates.
- [ ] [pr-test-analyzer] U4 permission-denied / unreadable-log branch untested — `red_status_detail_log_snapshot` rescues `EACCES`/`EBADF`/`EIO`/`NoLogFiles`/`InvalidTaskPath` at `bubble_model.rb:1538` but `bubble_model_test.rb` covers only the happy 100-line case (line 5091) and the "no log file" case (line 5121); a `chmod 000` regression on the detail open path would not be caught.
- [ ] [comment-analyzer] `FOOTER`/`FOOTER_NO_ENTER` constants in `red_status_detail.rb:11-19` are no longer read by `render` — only `footer_for` references them and the renderer now composes chips via `action_bar_lines`; the "PR #84 review row 25" rationale is stale and the constants are effectively dead in production code (the plan deliberately kept them for test substring assertions — say so or remove them).

## Medium

- [ ] [code-reviewer] `open_red_status_detail` does synchronous file I/O on the Bubbletea update thread — `bubble_model.rb:1522-1547` opens a `LogTail::Tail`, seeks, reads up to 64 KiB, and closes inline on Enter; on slow disks/sshfs this is perceptible keystroke latency and `Hive::Task.new(nil)`'s `TypeError` is not in the rescue list.
- [ ] [silent-failure-hunter] `log_panel` silently degrades to a labelled empty panel — `red_status_detail.rb:200` slices `lines[start_index...end_index]` which returns `[]` when `end_index == 0`, leaving a "Log · last 0 of N lines" header above an empty bordered box with no fallback hint; the `|| []` defends against an inverted range that `clamp_log_offset` already prevents (dead defensive code).
- [ ] [silent-failure-hunter] `apply_red_status_detail_scroll` has 5 distinct no-op return paths conflated — `update.rb:736-757` returns the unchanged model for `mode != :red_status_detail`, `state.nil?`, empty lines, `capacity <= 0`, and clamped-at-boundary; the first four signal state inconsistency but the operator gets no flash and no `Debug.log` breadcrumb when PgUp/PgDn silently does nothing.
- [ ] [silent-failure-hunter] `red_status_detail_log_capacity` returns 0 from inconsistent-state branches without distinguishing them — `update.rb:759-778` collapses "no state", "no lines", and "budget too small" into the same `0`; the first is a coding error that should `Debug.log` rather than silently render no scroll capacity.
- [ ] [pr-test-analyzer] U2 width-40 and width-20 boundary cases tested at off-by-one offsets — `red_status_detail_test.rb:70,80` use widths 41 and 21 instead of the plan's 40 and 20, and `header_line:152-168` uses `<` (not `<=`) so the exact boundary is exactly what's at risk and is not pinned.
- [ ] [pr-test-analyzer] U10 integration test pins composition at only one size — `bubble_model_test.rb:368` asserts header/panel/log/artifact/action ordering at 100×30 only; a second pass at 80×24 would mirror U6's invariant and catch composition-order regressions across the responsive boundary.
- [ ] [pr-test-analyzer] U9 scroll `:down` from offset 0 (clamp-at-zero) not pinned — `update_test.rb` covers up-saturation, decrement-from-5, and empty-log no-op, but not the `[next_offset, 0].max` clamp at the bottom edge of the window.
- [ ] [pr-test-analyzer] U7 exactly-5-paths boundary unpinned — tests cover 0/3/7 paths but not 5; a regression to `>= 5` at `red_status_detail.rb:287` would print an unnecessary "… (0 more)" line and slip past the existing test cases.
- [ ] [comment-analyzer] `RED_STATUS_*` and `MIN_*`/`MAX_*` constants land without a WHY — `update.rb:27-30` and `red_status_detail.rb:20-22` declare four / three magic numbers (8, 4, 12, 3) with no rationale for the values or why each layer carries its own copy; readers cannot judge a future tuning request.
- [ ] [comment-analyzer] `red_status_detail_log_snapshot` missing the non-obvious WHY — `bubble_model.rb:1534-1547` captures `log_path` into state but never re-tails it, and rescues an unusual mix (`EBADF`/`EIO`) without explaining the snapshot-vs-live-tail design (operator presses `R` to refresh); a one-line WHY would prevent a future reader "fixing" the apparent leak.
- [ ] [comment-analyzer] `red_status_marker_signature` carries a narrative of past bugs — `update.rb:887-902` describes "the prior `\\0`-joined raw concat" and points at "PR #84 review #7"; per CLAUDE.md, trim to a one-sentence WHY (consume canonical hex from diagnostic; fallback on attrs-digest with `no-diag:` prefix when absent).
- [ ] [code-simplifier] `header_line` 4-branch fallback ladder can collapse to an ordered candidate scan — `red_status_detail.rb:152-168` walks four progressively shorter strings; the same intent expressed as a candidate array with a single `find` reads as one fall-through axis instead of four branched returns.
- [ ] [code-simplifier] `wrap_action_chips` duplicates the "flush current" pattern — `red_status_detail.rb:120-140` carries the `lines << current unless current.empty?` line twice (mid-loop and post-loop); `each_with_object([[]])` collapses the empty-bucket check and reads more cleanly than the mutable `+""` accumulator.

## Nit

- [ ] [silent-failure-hunter] Rescue lists have drifted between `LogTail#poll!` and `red_status_detail_log_snapshot` — extract a `LogTail::FILESYSTEM_RESCUE` constant so the two surfaces share one list and don't diverge further.
- [ ] [silent-failure-hunter] `Tail#close!` rescues only `IOError` — `log_tail.rb:265` should add `Errno::EBADF`/`Errno::EIO` so an ensure-block close after an FD raced shut never escapes (mitigates the High close-after-partial-open finding).
- [ ] [silent-failure-hunter] 64 KiB backbuffer + 50-line ring is a silent contract — `bubble_model.rb:1539` reads only the trailing 64 KiB so a chatty `[stream]`-heavy log shows far less than the operator may assume; surface in wiki or panel label.
- [ ] [pr-test-analyzer] `key_map_test.rb:740-742` uses `>` assertions on `home.amount` — too loose; assert the exact value (or the named constant) so a regression to `Float::INFINITY` or `1` is caught.
- [ ] [pr-test-analyzer] U10 integration test does not pin section ordering — `bubble_model_test.rb:402` finds any `╭`/`+` line opener but not its position relative to the header or log panel, so a swap of reason and log panels would still pass.
- [ ] [comment-analyzer] `key_map.rb:313-314` uses `1_000_000` as a scroll-to-edge sentinel — opaque magic number; rename to `Float::INFINITY` or a named constant (`SCROLL_TO_EDGE`).
- [ ] [comment-analyzer] `styles.rb:112-113` declares `ACTION_CHIP_PRIMARY`/`ACTION_CHIP_MUTED` with no comment — names alone don't convey PRIMARY = "Enter chip" / MUTED = secondary; add a one-line WHY or rename.
- [ ] [comment-analyzer] `styles.rb:131-132` PANEL_BORDER comment restates WHAT instead of WHY — "panels are non-focused; focus lives on the action bar" would be the interesting note.
- [ ] [code-simplifier] `key_map.rb#red_status_detail_message:308-317` could lift scroll-key bindings into a frozen lookup table — pattern already used elsewhere in this file (`VERB_KEYS`, `ENTER_FLASH_MESSAGES`); makes scroll surface scannable in one block.
- [ ] [code-simplifier] `auto_action_available?` repeatedly calls `row.action_key.to_s` — `red_status_detail.rb:254-262` and the mirror at `update.rb:807-815` would flatten under a single `case row.action_key.to_s` (folds into the High consolidation finding).

ivankuznetsov:

## High

(none)

## Medium

- [ ] `bubble_model.rb:open_red_status_detail` swallows `Hive::InvalidTaskPath` from `Hive::Task.new(row.folder)` but the same rescue logs `path=#{log_path.inspect}` — when the constructor raises, `log_path` is still nil so the debug breadcrumb loses the only field that would identify a malformed `row.folder` (lib/hive/tui/bubble_model.rb:1540).
- [ ] Dual entry points for `:red_status_detail` mode: `Update.apply_open_red_status_detail` (pure, no log snapshot) at lib/hive/tui/update.rb:723 vs `BubbleModel#open_red_status_detail` (loads log) at lib/hive/tui/bubble_model.rb:1522. Production flow goes through the side-effect path, leaving the Update branch live-but-bypassed; a future refactor that routes `OpenRedStatusDetail` through `Update.apply` alone would silently lose the log snapshot with no test catching it.
- [ ] `red_status_detail_log_snapshot` runs `Tail#open!` + 64 KiB read synchronously on the BubbleModel update thread (lib/hive/tui/bubble_model.rb:1542). Plan acknowledges this as risk #4 but no time-bound / async path exists — on sshfs/NFS the operator's keystroke that opens the detail can block the alt-screen for the full disk read.
- [ ] `RefreshRedStatusDiagnosis` (R key) refreshes the diagnostic only — it does not re-snapshot the log (lib/hive/tui/bubble_model.rb:1522 is only entered via `OpenRedStatusDetail`). Plan U4 claims "pressing R already re-enters the detail state and would re-load the log naturally" but the actual handler bypasses `open_red_status_detail`. Operator must back out (q) and re-enter to refresh the log preview.
- [ ] Log title "Log · last X of Y lines (64 KiB tail)" advertises a line cap, but the underlying bound is bytes — if log lines exceed ~1.3 KiB each, fewer than 50 lines fit in the 64 KiB backbuffer and the operator only sees the "(64 KiB tail)" hint to explain why (lib/hive/tui/views/red_status_detail.rb:188). No test covers the long-line case.
- [ ] `red_status_detail_log_capacity` (lib/hive/tui/update.rb:758) recomputes the full layout budget from `model.cols/rows` on every scroll keystroke and must stay perfectly in sync with `Views::RedStatusDetail#composed_content_lines`. The shared `RedStatusDetailLayout` module helps, but the renderer still has private math (e.g., `remaining -= 1 if remaining.positive?` for the blank header gap) duplicated in the capacity helper rather than centralized — drift here would silently mis-clamp Home/End past the visible window.

## Nit

- [ ] Redundant guard `return 0 if Array(state.log_lines).empty?` in `red_status_detail_log_capacity` (lib/hive/tui/update.rb:765) — the sole caller `apply_red_status_detail_scroll` already short-circuits on `lines.empty?` at line 741.
- [ ] Hardcoded `(64 KiB tail)` in the log panel title (lib/hive/tui/views/red_status_detail.rb:188) duplicates the magic number from `backbuffer_bytes: 64 * 1024` at lib/hive/tui/bubble_model.rb:1546. Changing the snapshot size leaves the label stale.
- [ ] `header_line` fall-through chain (lib/hive/tui/views/red_status_detail.rb:137) is dense — three levels (candidates loop, `width < prefix.length` short-circuit, default branch). A 1-line `# falls through to ...` comment between branches would help next reader trace the truncation order.
- [ ] `apply_red_status_detail_scroll` `case msg.direction` `else` branch (lib/hive/tui/update.rb:751) silently keeps the offset; an unexpected direction would deserve a `Hive::Tui::Debug.log` breadcrumb the way the `capacity <= 0` and `state nil` branches already do.
- [ ] `Refreshing diagnosis...` hint appended after artifacts (lib/hive/tui/views/red_status_detail.rb:72) — visually buries the in-flight status under the artifact list. The reason-panel header would be a more discoverable home.
- [ ] `block_line_count` exists only to wrap `to_s.lines.size` (lib/hive/tui/views/red_status_detail.rb:77) — a single call site; inlining would remove the indirection without losing clarity.

ivankuznetsov:

## High
- [ ] `lib/hive/commands/bot.rb:16` reverts the unrelated bot start contract: `hive bot start` no longer backgrounds by default and the bot docs now point at legacy `~/Dev/hive` paths, breaking origin/main's manual bot workflow outside this TUI-only task.

## Medium
- [ ] `lib/hive/tui/input_decoder.rb:16` omits PageUp/PageDown escape sequences: real terminal `\e[5~`/`\e[6~` input is decoded as Esc plus raw text, so the new PgUp/PgDn red-status scroll bindings never reach `KeyMap`.
- [ ] `wiki/log.md:5` deletes prior changelog entries from an append-only wiki log: the diff removes the 2026-05-23 README and bot entries that exist on `origin/main`, violating the repo wiki contract and losing history.

## Nit
- [ ] `lib/hive/tui/help.rb:53` leaves the new red-status scroll controls out of the help overlay: the only in-app keybinding map still lists Enter/f/R/q/Esc, so the scrollable log panel ships without discoverable scroll keys.

ivankuznetsov:

## High

- [ ] [code-reviewer] Branch reverts main's bot-defaults PR #130 (`215432a2 fix(bot): make telegram start background by default`): `lib/hive/cli.rb`, `lib/hive/commands/bot.rb`, `examples/launchd/hive-bot.plist`, `examples/systemd/hive-bot.service` swap `--foreground` for `--detach` and flip default to foreground — a breaking CLI/unit-file change unrelated to the plan's `:red_status_detail` scope, almost certainly accidental from a bad rebase resolution.
- [ ] [code-reviewer] Branch reverts main's README PR #132 (`f9280a37 docs(readme): add daemon-first TUI quickstart`): `README.md` strips the Five-Minute TUI Getting Started and entire "Manage Hive From Telegram" sections that PR #132 just landed in main — out of scope for `:red_status_detail` and erases recently-merged onboarding content.
- [ ] [code-reviewer] `wiki/log.md` is shrunk by deleting two existing append-only entries (`2026-05-23T09:45:00Z readme...` and `2026-05-23T09:20:00Z bot...`): CLAUDE.md mandates wiki/log.md as append-only, so removing prior entries during a rebase violates the project's wiki contract.
- [ ] [code-reviewer] Wiki path docs become factually wrong: `wiki/commands/bot.md`, `wiki/modules/bot.md`, `wiki/state-model.md` rewrite canonical paths from `~/.local/state/hive/.bot.pid` / `logs/bot.log` to `~/Dev/hive/...`, but `Hive::Paths.state_home` returns `~/.local/state/hive` by XDG default — `~/Dev/hive` only applies when `HIVE_HOME` is exported, so the docs now mislead the default install.
- [ ] [code-reviewer] `lib/hive/config.rb:167` comment now claims `bot` block loads "from ~/Dev/hive/config.yml via load_global_bot", but `load_global_bot` reads `Hive::Paths.config_home/config.yml` = `~/.config/hive/config.yml` by default: comment contradicts the code one method below it.
- [ ] [code-reviewer] `wiki/operating.md` is internally inconsistent post-change: the bot start snippet says `~/.local/state/hive/logs/bot.log`, while `wiki/state-model.md` (same wiki) now says `~/Dev/hive/logs/bot.log`; readers will hit conflicting "canonical" paths in adjacent pages.
- [ ] [pr-test-analyzer] `test_start_backgrounds_by_default` is deleted from `test/integration/bot/bot_lifecycle_test.rb` with no replacement that asserts `--detach` triggers `Process.daemon(true, true)`: behavioral regression in the bot daemonize path can no longer be caught by CI, and this is the only test that ever pinned the daemonize call.
- [ ] [code-reviewer] Comment-vs-CLI-help drift in `lib/hive/cli.rb:660` long_desc: example shows `hive bot start [--detach] [--dry-run]` but the surrounding paragraph still claims `~/.config/hive/config.yml` was replaced with `~/Dev/hive/config.yml` — same "wrong default path" landed in CLI help users will see at `hive bot --help`.

## Medium

- [ ] [code-simplifier] `Update.apply_open_red_status_detail` (`lib/hive/tui/update.rb:723-731`) is unreachable in production: `BubbleModel#handle_side_effect` for `OpenRedStatusDetail` always returns a non-nil tuple via `open_red_status_detail` (`bubble_model.rb:376-377`), so `Update.apply`'s `when Messages::OpenRedStatusDetail` branch is dead — the state transition exists in two places and a future caller bypassing BubbleModel would lose log loading.
- [ ] [code-reviewer] `bubble_model.rb#open_red_status_detail` reaches across module boundaries to call `Hive::Tui::Update.red_status_marker_signature(row)` for marker init: signature derivation belongs in a shared helper (`RedStatusDetailLayout` was introduced for exactly this kind of cross-module math) rather than calling Update's module-method from a side-effect handler.
- [ ] [silent-failure-hunter] Snapshot-not-live design is invisible to the operator: `log_panel` title says "Log · last N of M lines (64 KiB tail)" but never signals "this is a frozen snapshot from when you entered — press R to refresh" — a continuing auto-heal write will not appear in the panel and the user has no UI cue. Plan U4 chose snapshot over live; the contract should be surfaced.
- [ ] [silent-failure-hunter] `red_status_detail_log_snapshot` rescues `TypeError` broadly (`bubble_model.rb:1551`): catches the intended nil-`row.folder` path but also any unrelated TypeError raised inside `Hive::Tui::LogTail::Tail#open!` / `lines` (e.g., a future regression passing a non-Integer to lines), silently collapsing it to an empty panel. Prefer an explicit `return [nil, []] if row&.folder.nil?` guard plus a tighter rescue.
- [ ] [pr-test-analyzer] `test_red_status_detail_view_renders_header_panels_log_artifacts_and_action_bar` uses `output.index(substring)` for ordering: indexes are byte offsets so two substrings on the same rendered line would still pass the `<` chain, masking a layout where reason and log collapse into the same row. Prefer `output.lines.find_index { |l| l.include?(...) }` so ordering is line-aligned.
- [ ] [comment-analyzer] `log_panel` panel title hardcodes the string "64 KiB tail" (`red_status_detail.rb:188`) but the actual `backbuffer_bytes: 64 * 1024` literal lives in `bubble_model.rb#red_status_detail_log_snapshot`: any future tuning of the backbuffer in one place will silently drift the visible label in the other. Either parameterize the label off the configured byte count or add a test that pins the two literals together.

## Nit

- [ ] [code-reviewer] `Styles::RECOVERY_HEADER_BAR` is named "BAR" but renders no width/background — it's just bold-red text inline; either rename to `RECOVERY_HEADER_STYLE` or apply `.width(width)` so it visually reads as a bar across the row.
- [ ] [code-reviewer] `Styles::PANEL_BORDER = PANE_DIM_BORDER` aliases without `.freeze` (sibling `ACTION_CHIP_PRIMARY`/`ACTION_CHIP_MUTED` are explicitly frozen): inconsistency with neighbors; freezing is harmless because the source is already frozen, but the asymmetry will read as a missed `.freeze`.
- [ ] [code-simplifier] `header_line` (`red_status_detail.rb:137-154`) still uses a candidates loop plus two more conditional returns after the loop: the post-loop `return truncate(first, width) if width < prefix.length` and final `prefix + truncate(project_stage, …)` could fold into the candidates table for one fall-through axis.
- [ ] [comment-analyzer] `red_status_detail_layout.rb:23-27` `CHIP_SEPARATOR` comment explains the historical " · " (3-char) bug at width=79 — useful WHY for the value choice but quite long for a one-line rationale; could be tightened to a single sentence.
- [ ] [pr-test-analyzer] `test_open_red_status_detail_handles_unreadable_log_file` conditionally skips on `Process.uid.zero?`: necessary because chmod 000 can't restrict root, but a CI runner that runs as root (some Docker images) will silently skip this case — consider a fallback assertion (e.g., `unlink` mid-open) to still cover the EACCES rescue arm in root environments.
- [ ] [comment-analyzer] `bubble_model.rb#red_status_detail_log_snapshot` rescue chain `Hive::NoLogFiles, Hive::InvalidTaskPath, TypeError, *Hive::Tui::LogTail::FILESYSTEM_RESCUE` is wide: a one-line comment naming each (NoLogFiles → no log dir; InvalidTaskPath → row.folder shape; TypeError → nil row.folder) would help future readers judge whether to narrow.
- [ ] [code-simplifier] `RedStatusDetailLayout#action_row_count` reuses chip-wrap math nearly identical to `Views::RedStatusDetail#wrap_action_chips` (both walk chip_lengths with separator math): one helper that returns row-count + styled lines would eliminate the parallel implementation that the plan flagged as drift-prone in U6 risks.
- [ ] [code-reviewer] `Hive::Tui::Debug.log("red_status_detail", "scroll skipped: mode=…")` (`update.rb:738`, `:744`, `:767`): four near-identical breadcrumbs across `apply_red_status_detail_scroll` / `red_status_detail_log_capacity` could fold into a single tagged helper so the category string isn't repeated four times.

PR #135, opened 2026-05-24
feat(usage): record and surface hive agent token usage

ivankuznetsov:
## Summary

Hive previously had no visibility into how much LLM token spend its agent pipeline consumed: spawns ran, logs were written, and the operator was left to read provider dashboards if they wanted to know which project or task was burning tokens. This PR captures token usage at the same boundary that runs hive stage agents, persists one row per spawn to a local SQLite database, and surfaces scoped aggregates in two TUI surfaces: an always-visible footer and a full-screen matrix opened with `T`.

Scope is deliberately tight: only hive-driven spawns are recorded (no scraping of `~/.claude/projects/*.jsonl`), there is no historical backfill, and USD/cost rendering is explicitly out. The `model` column is stored so cost can be layered on later without a schema migration.

## Capture pipeline

```mermaid
flowchart TB
    Stage[Stages::Base.spawn_agent] --> Agent[Agent#spawn_and_wait]
    Agent --> Reader[stdout/stderr reader thread]
    Reader --> Extractor[AgentProfile#usage_extractor]
    Extractor --> LastUsage[last non-nil usage hash]
    LastUsage --> Result[result usage and model]
    Result --> Record[UsageDb.record!]
    Record --> SQLite[(usage.db)]
    SQLite --> Footer[TUI footer]
    SQLite --> Matrix[TUI T matrix]
```

The reader thread already saw every line. The extractor hook on `AgentProfile` lets each profile decode the agent's final structured usage event without re-parsing logs after the fact. `Stages::Base.spawn_agent` is the only call site, which is what keeps ad-hoc agent sessions out.

## What each commit lands

| Commit | What it adds |
| --- | --- |
| `feat(usage): add token usage database` | `Hive::UsageDb` (SQLite at `Hive::Paths.data_home/usage.db`), `record!` and `aggregate` API, three indexes, `sqlite3 ~> 2.0` gem dep. |
| `feat(agent): extract token usage from streams` | `usage_extractor:` on `AgentProfile`; concrete claude / codex / pi extractors; `Agent#spawn_and_wait` returns `result[:usage]` and `result[:model]`. |
| `feat(stages): record agent token usage` | `Stages::Base.spawn_agent` records one row per spawn, regardless of exit status, with stage/project/task slugs joined in. DB write failures warn and never abort the spawn. |
| `feat(tui): show token usage footer` | Scope-aware footer line (all / project / task) with `k`/`M` formatting; aggregate cached per snapshot tick. |
| `feat(tui): add token usage stats view` | `T` opens a full-screen agents × buckets matrix with arrow-key drill between all / project / task scopes. |
| `docs(usage): document token usage stats` | `wiki/token-usage.md`, index/log/gaps updates, `CHANGELOG.md`. |

## Design decisions worth a second look

- **Capture inside the reader thread**, not by re-parsing on-disk logs. The thread sees every line exactly once; we feed each decoded JSON record to both the existing `extract_final_message` and the profile's usage extractor. Keeping the last non-nil result means crashed spawns still record whatever final usage event arrived before the child exited.
- **DB at `Hive::Paths.data_home/usage.db`** (XDG, typically `~/.local/share/hive/usage.db`) instead of the brainstorm's `~/.hive/usage.db`. Matches the rest of the codebase's path convention; `HIVE_USAGE_DB_PATH` preserves operator choice.
- **`project_slug` is `task.project_name`** (basename of the project root), so two checkouts of the same repo collapse into one slug. Documented in `usage_db.rb` so a future contributor does not switch to absolute paths.
- **`T` keybinding**, not `s`. The brainstorm picked `s`, but `s` is already bound to `OpenInAgent`. Uppercase `T` avoids the conflict and keeps muscle memory intact for existing bindings.
- **Per-call SQLite connection**, no long-lived handle in the TUI thread. Aggregate query cached per 1 Hz snapshot tick keeps cost flat and avoids stale-handle issues during tests that move the DB file.

## Known follow-ups

`claude` extractor is grounded in a recorded real `type:"result"` line. `codex` and `pi` extractors zero-fill when their final usage payload differs from what was anticipated, so one row is still written per spawn. Refinement to match captured real codex/pi streams is tracked in `wiki/gaps.md`.

## Tests

New coverage spans every layer of the pipeline:

- Unit: `test/unit/usage_db_test.rb`, `test/unit/usage_extractors_test.rb`, `test/unit/tui/views/usage_footer_test.rb`, `test/unit/tui/views/token_stats_test.rb`, plus targeted additions to `agent_test.rb`, `agent_profile_test.rb`, `bubble_model_test.rb`, `key_map_test.rb`, `messages_test.rb`, `model_test.rb`, `update_test.rb`, `help_test.rb`.
- Integration: `test/integration/stages_base_usage_test.rb` (full spawn → row recorded; usage-less and crash paths), `test/integration/tui_usage_footer_test.rb`, `test/integration/tui_token_stats_test.rb`.

---

Claude_Code-D97757?logo=claude&logoColor=white)


ivankuznetsov:

## High

## Medium
- [ ] Wide-branch footer truncation loses both usage and hint between cols 80 and ~138 (`lib/hive/tui/bubble_model.rb:3199-3206`): the wide branch concatenates `usage · hint` into one ~138-char string then truncates from the right at `usable_width`, so at common 80-120 col terminals the user sees a mangled `"tokens — today 0/0/0 • 7d 0/0/0 • 30d 0/0/0 • all 0/0/0 · [Tab] swit…"` — defeating the plan's "never lose the data" intent for that whole width band; consider dropping/truncating only the hint half once total exceeds `usable_width`.
- [ ] `derive_usage_scope` returns `{}` (all projects) when right pane is focused, a project scope is active (`model.scope.positive?`), but `current_row` is nil (`lib/hive/tui/bubble_model.rb:3229-3239`): a user who explicitly scoped to project N expects the footer/T-matrix to reflect that project even when the tasks pane is empty for that project; currently they get global totals.
- [ ] `Hive::AgentProfile#extract_usage_event` silently rescues every `StandardError` with no `warn` (`lib/hive/agent_profile.rb:248-251`): inconsistent with `UsageDb.record!` which warns on rescue, and a buggy custom extractor would have no diagnostic surface; emit `warn "[hive] usage extractor failed: …"` so silent drift is visible.
- [ ] `format_number` renders values just under 1M with a four-digit k unit, e.g. 999_500 → "1000k" instead of "1.0M" (`lib/hive/tui/views/usage_footer.rb:39-50`): the unit boundary check is strict-`>=`; either round up before threshold comparison or cap the k-formatted output at three digits so the next unit kicks in.

## Nit
- [ ] `Hive::UsageDb.zero_buckets` / `.zero_usage` are leaked as public module helpers solely to feed `Views::TokenStats` (`lib/hive/usage_db.rb:118-124`, `lib/hive/tui/views/token_stats.rb:79-87`): consider moving the seed helpers next to the view or marking them with a `# @api private` comment so future maintainers don't treat them as a stable surface.
- [ ] `record_usage` does `usage[:model] || result[:model]` (`lib/hive/stages/base.rb:328-336`), but `Agent#spawn_and_wait` sets `result[:model] = last_usage && last_usage[:model]` (`lib/hive/agent.rb:215`), so the fallback is dead code; drop it or document why the redundancy is intentional.
- [ ] `Views::TokenStats` mixes lowercase agent labels (`"claude"`, `"codex"`, `"pi"`) with an uppercase `"TOTAL"` row label (`lib/hive/tui/views/token_stats.rb:46-49`): inconsistent casing; either uppercase the agent rows or lowercase the total.
- [ ] `apply_token_stats_scope_changed` / `apply_token_stats_selection_moved` silently no-op when no candidates exist (`lib/hive/tui/update.rb:592-636`): pressing arrow keys at the deepest scope or in an empty project shows no feedback; a flash like "no further drill-down" would help discoverability.
- [ ] `bucket_starts` mixes implicit symbol literals (`today:`, `all:`) with quoted-symbol keys (`"7d":`, `"30d":`) (`lib/hive/usage_db.rb:147-156`): they're all symbols thanks to Ruby parsing, but the visual inconsistency invites future "is `:7d` a symbol or string?" confusion; consider `:"7d"` / `:"30d"` explicit form or constants from `BUCKETS`.
- [ ] `test_usage_db_failure_does_not_fail_spawn` forces failure by pointing `Hive::UsageDb.path` at a directory (`test/integration/stages_base_usage_test.rb:144-159`): clever but platform-dependent; future SQLite versions or Windows might raise differently or even succeed. A read-only file or a path under a denied dir would be more robust.
- [ ] `Hive::UsageDb.aggregate` sums every agent present in `result[:agents]` into `result[:total]`, including unknown agent names auto-created via `||= zero_buckets` (`lib/hive/usage_db.rb:78-99`), while `Views::TokenStats` only renders the fixed `claude/codex/pi` rows (`lib/hive/tui/views/token_stats.rb:13`): if a row with an unrecognized agent name ever lands in the DB the displayed TOTAL won't equal the sum of visible rows. Either filter unknown agents from totals or render a generic "other" row.

ivankuznetsov:

## High
- [ ] 6-review sub-spawns record usage with `project_slug=NULL` (`lib/hive/reviewers/synthetic_task.rb:22`, `lib/hive/stages/base.rb:337`): reviewers, triage, CI-fix, and browser-test use `SyntheticTask` without `project_root`, so their token rows disappear from project/task-scoped footer and matrix totals despite being hive-driven agent invocations.

## Medium
- [ ] Flash messages hide the token footer (`lib/hive/tui/bubble_model.rb:3193`): `default_footer` returns only the flash while active, so the plan's always-visible usage strip disappears for every command/error flash TTL.
- [ ] Token stats queries SQLite from the render path (`lib/hive/tui/bubble_model.rb:1422`): the full-screen matrix recomputes aggregates on every view render instead of caching per snapshot, which can hammer four aggregate queries during the TUI's frequent update ticks and stall input on larger usage databases.
- [ ] Usage writes can be silently dropped under normal TUI read contention (`lib/hive/usage_db.rb:167`): each connection uses SQLite defaults with no busy timeout/retry, so a concurrent aggregate read can make `record!` hit `database is locked`, warn, and lose the one-row-per-spawn record.

## Nit
- [ ] Token usage docs leave adjacent wiki pages stale (`wiki/commands/tui.md:39`, `wiki/dependencies.md:10`): the new `T` mode/keybinding and `sqlite3` runtime dependency are documented in `wiki/token-usage.md` but omitted from the command and dependency reference pages.

ivankuznetsov:

# Code Review — token-usage stats in TUI (pass 02)

Scope: `git diff origin/main..HEAD`, filtered to the 9 commits on this branch
(`9edbbcd5..753474c0`) that implement IU1–IU6 of the plan.

Read-only review. No edits applied; findings below are triage input only.

## High

(none — implementation matches the plan, error paths are correctly contained,
test coverage spans UsageDb, all three extractors, the Stages::Base spawn
boundary, the footer, and the matrix view including drill-down.)

## Medium

- [ ] `Views::TokenStats` re-queries SQLite on every render with no cache (`lib/hive/tui/bubble_model.rb:1420-1424`): `token_stats_view` calls `Hive::UsageDb.aggregate` directly, in contrast to the footer's `@usage_footer_cache_key` cache (`lib/hive/tui/bubble_model.rb:3225-3238`). While the matrix is open, every 1Hz snapshot tick AND every drill-down / selection keystroke fires 4 SQL `SUM(...) GROUP BY agent` queries — the plan's "DB open rate to 1 Hz max" contract is violated in this mode. Cache by `(snapshot.object_id, generated_at, scope_level, project_slug, task_slug)` mirroring the footer.
- [ ] `UsageExtractors::CLAUDE` treats `type:"stream_event"` and `type:"result"` symmetrically with "last non-nil wins" (`lib/hive/agent_profiles/usage_extractors.rb:6-16`, consumed at `lib/hive/agent.rb:148-149`): currently safe because Claude emits `result` last, but a future Claude release that emits a trailing telemetry `stream_event` (or any out-of-order frame) would overwrite the final cumulative usage with a partial. Anchor the canonical record to `type:"result"` (or prefer `result`-typed usage over `stream_event` when both arrive in the same spawn) so the contract is structural, not order-dependent.

## Nit

- [ ] Double-rescue of `StandardError` on the usage-record path (`lib/hive/stages/base.rb:333-334` wrapping `lib/hive/usage_db.rb:72-75`): `UsageDb.record!` already warns + returns false on any error. The outer rescue is dead for that case and only catches programmer errors in the keyword construction; on a real DB failure the user gets two `[hive] usage record failed` lines for the same incident. Remove the outer rescue (or narrow it to `TypeError`/`ArgumentError` for the kwarg-construction path).
- [ ] `UsageDb.iso8601` re-parses already-ISO8601 strings (`lib/hive/usage_db.rb:185-191`): `Hive::Stages::Base.spawn_agent` passes `Time.now.utc.iso8601` (a String), and `record!` round-trips it through `Time.parse(value.to_s).utc.iso8601`. A one-line `is_a?(String)` short-circuit avoids the Time.parse on the hot insert path.
- [ ] Footer cache key uses `snapshot&.object_id` (`lib/hive/tui/bubble_model.rb:3228`): `generated_at` alone uniquely identifies a snapshot's data; object_id is Ruby-VM-specific and would defeat the cache if Snapshot is ever duped/frozen-recreated. Prefer `[snapshot&.generated_at, scope[:project_slug], scope[:task_slug]]`.
- [ ] `Views::TokenStats.column_widths` has an implicit per-column floor of 6 chars (`lib/hive/tui/views/token_stats.rb:60`): `shrink = [overflow, natural[idx] - 6].min` skips shrinkage when natural width is already ≤ 6, so on terminals where multiple columns are already at the floor the rendered row can exceed `inner_width`. Either truncate the rendered cell explicitly when `widths.sum + separators > inner_width`, or fail loud on overflow.
- [ ] `UsageDb.aggregate` runs 4 sequential queries (`lib/hive/usage_db.rb:84`): plan's Risks section already anticipated `SUM(CASE WHEN started_at >= ? THEN input ELSE 0 END)` consolidation; not urgent (the indexes keep each query <10ms on small DBs), but adopting it now also gives the matrix view's "no cache" cost a 4× reduction.
- [ ] `Hive::UsageDb.zero_buckets` / `zero_usage` are tagged `@api private — exposed only so Views::TokenStats can seed empty cells` (`lib/hive/usage_db.rb:115-132`) but are consumed externally by `views/token_stats.rb:85-90`: the cross-module note rots quickly. Inline a `Views::TokenStats::EMPTY_BUCKETS` constant (or call `UsageDb.zero_aggregate[:agents][agent]` via a thin local helper) so the visibility contract is local.
- [ ] Codex/Pi extractors zero-fill on bare final events (`lib/hive/agent_profiles/usage_extractors.rb:24-37`): the plan accepts this trade-off and `wiki/gaps.md` tracks the follow-up, but the empty rows are indistinguishable from "agent did real work with 0 tokens reported." Consider stamping the row with `model: nil` semantics already do this, OR returning nil from the extractor (so no row is recorded) until a real usage shape is verified for codex/pi. Either path keeps the matrix from claiming "0/0/0" as observed data.
- [ ] `Hive::Tui::Model::TokenStatsState` defaults `project_slug`/`task_slug` to nil regardless of `scope_level` (`lib/hive/tui/model.rb:117-125`): callers in `bubble_model.rb` and `update.rb` always pair `scope_level: :project` with `project_slug:` and `:task` with both, but the type itself permits an incoherent `(:task, nil, nil)`. Add a single guard in `initialize` that raises if `scope_level == :task` and either slug is missing — keeps the state machine honest at the type boundary.

PR #152, opened 2026-05-25
feat(bot): replace recovery alerts with one-button human notifications

ivankuznetsov:
## Summary

Operator recovery alerts in Telegram were unreadable jargon walls ("Needs recovery: review_error marker=… phase=… reason=…") with four buttons each, and the in-memory dedupe re-fired the entire backlog after every bot restart — surfacing as ~1500 notifications per night. Recovery alerts now read as one plain-English sentence with a single `🔧 Autofix` button, dedupe is persisted to disk and keyed by `(slug, stage, error fingerprint)` so each real failure surfaces exactly once until the row's status changes, and the operator gets one `✅ Recovered` line when the row clears.

`/status [project]` was rebuilt on the same human-readable formatter so the queue reads as `Title… — Stage` lines instead of raw slugs and marker tokens.

## What changed

| Surface | Before | After |
|---|---|---|
| Recovery headline | `Stuck waiting: hive/we-need-to-improve-this-260522-db23 (6-review)` | `⚠ Review stuck — "We need to improve this…"` |
| Body | `Needs recovery: review_error phase=… reason=… marker=… exception_class=…` | One plain sentence naming the failed actor + `Tap Autofix to retry the stage cleanly.` |
| Keyboard | `Clear and retry` · `Open laptop` · `Show details` · `Refresh diagnosis` | `🔧 Autofix` |
| Dedupe key | In-memory time window per `(project, slug, marker)` | Persistent JSON store keyed by `(project, slug, stage, marker, attrs)` SHA |
| Restart behavior | Re-fires every active row on every boot | Loads prior state; only new fingerprints alert |
| Still-stuck signal | Silent forever, or re-fires on the dedupe window | One `(reminder)` per fingerprint at 8 h, never again |
| Recovery signal | None | One `✅ Recovered: "Title…" — Stage` line |
| `/status` line | `hive/we-need-to-improve-this-260522-db23 6-review needs_recovery REVIEW_ERROR` | `We need to improve this… — Review` |

## Design decisions worth scrutinizing

**Fingerprint now includes `stage`.** Without it, the same marker recurring in a different stage would be suppressed as a duplicate. R9 of the spec calls this out explicitly; the cost is that a row that healed in execute and then errors again in review re-alerts, which is what we want.

**The diagnostic summary is dropped from Telegram entirely.** Per-stage logs under `.hive-state/...` remain the authoritative postmortem surface. The `details:` / `refresh_diagnose:` callback handlers still exist so pre-redeploy messages keep routing (no "Bot got confused" replies after the deploy), but no new buttons reference them.

**Persistence is a JSON file with atomic temp+rename writes.** Path is `bot.alert_state_file` (default `~/Dev/hive/.bot.alert_state.json`), validated as a path key in `Config`. Corrupt JSON is renamed to `<path>.corrupt-<ts>` and the store starts fresh rather than crashing the bot — we trade one lost dedupe cycle for liveness. Tested in `test/unit/bot/alert_store_test.rb`.

**`AGENT_WORKING` is not in the `markers clear` allowlist**, so Autofix special-cases it: skip the clear step, dispatch the retry verb only, and let `StaleAgentHealer` rewrite the marker on the next tick. The autofix handler also no-ops gracefully on a terminal/unknown stage rather than dispatching a bogus retry verb.

**Reminder is single-shot and shares the fingerprint.** Same incident → same dedupe key → the dispatcher's reminder branch explicitly short-circuits the dedupe check once per fingerprint, then `mark_reminded` closes the door.

**One wording covers all failure categories.** A small marker→noun lookup ("review agent crashed", "execute agent stalled", "review run stalled") gives natural sentences without leaking marker tokens; unknown markers fall back to "The agent crashed before it could finish." Adding a new marker only adds one row to the lookup.

## Out of scope

- Auto-escalation to headless Codex on repeated failure (R13 accepts that the second tap just re-alerts).
- In-Telegram chat (`Chat about this`) — deferred without placeholder.
- TUI surfaces, stage runners, marker semantics, and stage logs — untouched.
- The legacy `notification_dedupe_window_sec` config key remains for back-compat but is now unused; a deprecation event is emitted on load when set.

## Test plan

`bundle exec rake test` — 3508 runs, 12612 assertions, 0 failures (covers 8 new/rewritten test files: `title_formatter_test.rb`, `alert_store_test.rb`, the dispatcher lifecycle suite, the autofix handler, `/status` project filter, config bounds, and the rewritten `notification_builders_test.rb` whose old button-count assertions no longer apply).

`bundle exec rubocop lib/hive/bot test/unit/bot test/integration/bot` — clean.

Live Telegram smoke was not run; behavior is covered by the rewritten unit + integration suite. Worth a manual tick post-merge: trigger a `review_error` row, confirm the new headline + single button, advance status to clear, confirm the `✅ Recovered` line, then restart the bot and confirm no duplicate alert.

---

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -17,18 +17,10 @@
       "note": "False positive. Open3.capture3 array form in Stages::Finalize#final_commits. base is produced by `git merge-base HEAD origin/<default>` and interpolated into one revspec argv element (`base..HEAD`); no shell ever sees it."
     },
     {
-      "fingerprint": "c95e1a89a496342c7c7f06de67790df2bc46bc538b9ced61c16a703fb42facb5",
-      "note": "False positive. Open3.capture3 array form in BrainstormTmux#sweep_orphan_processes. task.folder is Regexp.escape'd into one pgrep pattern argv element; no shell ever sees it."
-    },
-    {
-      "fingerprint": "5bca07d54019b8f635328009cbc5517d6d7e8f74d12afcb448517a482176d392",
-      "note": "False positive. Open3.capture3 array form in BrainstormTmux#sweep_orphan_processes. task.folder is Regexp.escape'd into one pkill pattern argv element; no shell ever sees it."
-    },
-    {
       "fingerprint": "907f70565e5132a49765208c31e28bbf8433b7819fb85fbc46413a8cb7649f81",
       "note": "False positive. Open3.capture3 array form in GitOps#commits_behind. `ref` is `origin/<default_branch>` constructed inside Hive::Rebase from `cfg.default_branch || git.default_branch` (the latter from `git symbolic-ref refs/remotes/origin/HEAD`). Not attacker-controlled; passed as a single argv element to execve(2)."
     }
   ],
-  "updated": "2026-05-15",
+  "updated": "2026-05-26",
   "brakeman_version": "8.0.4"
 }
--- a/lib/hive/bot/alert_store.rb
+++ b/lib/hive/bot/alert_store.rb
@@ -136,8 +136,16 @@ module Hive
 
         FileUtils.mkdir_p(File.dirname(@path))
         tmp_path = File.join(File.dirname(@path), ".#{File.basename(@path)}.#{$$}.#{Thread.current.object_id}.tmp")
-        File.write(tmp_path, JSON.pretty_generate(@data))
+        File.open(tmp_path, "w") do |f|
+          f.write(JSON.pretty_generate(@data))
+          f.fsync
+        end
         File.rename(tmp_path, @path)
+        begin
+          Dir.open(File.dirname(@path)) { |dir| dir.fsync }
+        rescue StandardError
+          nil
+        end
       ensure
         FileUtils.rm_f(tmp_path) if tmp_path && File.exist?(tmp_path)
       end
--- a/lib/hive/bot/alert_store.rb
+++ b/lib/hive/bot/alert_store.rb
@@ -1,0 +1,166 @@
+require "json"
+require "fileutils"
+require "time"
+
+module Hive
+  module Bot
+    class AlertStore
+      SCHEMA_VERSION = 1
+
+      Entry = Data.define(:first_seen_at, :reminded_at, :row)
+      RowSnapshot = Data.define(:project, :slug, :stage, :marker, :attrs, :action) do
+        def initialize(project:, slug:, stage:, marker:, attrs: {}, action: nil)
+          super
+        end
+      end
+
+      def initialize(path:, logger: nil)
+        @path = path ? File.expand_path(path) : nil
+        @logger = logger
+        @mutex = Mutex.new
+        @data = empty_data
+        load!
+      end
+
+      def each_fingerprint
+        return enum_for(:each_fingerprint) unless block_given?
+
+        synchronize { entries.keys }.each { |fingerprint| yield fingerprint }
+      end
+
+      def add(fingerprint, row, now)
+        synchronize do
+          entries[fingerprint] = {
+            "first_seen_at" => timestamp(now),
+            "reminded_at" => nil,
+            "row" => serialize_row(row)
+          }
+          persist_locked!
+        end
+      end
+
+      def mark_reminded(fingerprint, now)
+        synchronize do
+          entry = entries[fingerprint]
+          next unless entry
+
+          entry["reminded_at"] = timestamp(now)
+          persist_locked!
+        end
+      end
+
+      def remove(fingerprint)
+        synchronize do
+          entry = entries.delete(fingerprint)
+          persist_locked! if entry
+          entry ? row_from_raw(entry["row"]) : nil
+        end
+      end
+
+      def entry(fingerprint)
+        synchronize do
+          raw = entries[fingerprint]
+          raw ? entry_from_raw(raw) : nil
+        end
+      end
+
+      private
+
+      def synchronize(&block)
+        @mutex.synchronize(&block)
+      end
+
+      def load!
+        return unless @path && File.exist?(@path)
+
+        parsed = JSON.parse(File.read(@path))
+        unless parsed.is_a?(Hash) &&
+               parsed["schema_version"] == SCHEMA_VERSION &&
+               parsed["entries"].is_a?(Hash)
+          raise JSON::ParserError, "invalid alert store schema"
+        end
+
+        @data = {
+          "schema_version" => SCHEMA_VERSION,
+          "entries" => parsed["entries"]
+        }
+      rescue JSON::ParserError, TypeError => e
+        corrupt_path = corrupt_path_for(@path)
+        begin
+          File.rename(@path, corrupt_path) if @path && File.exist?(@path)
+        rescue SystemCallError
+          corrupt_path = nil
+        end
+        @logger&.event(:alert_store_corrupt, path: @path, corrupt_path: corrupt_path,
+                                             error_class: e.class.name, message: e.message)
+        @data = empty_data
+      end
+
+      def empty_data
+        { "schema_version" => SCHEMA_VERSION, "entries" => {} }
+      end
+
+      def entries
+        @data["entries"] ||= {}
+      end
+
+      def persist_locked!
+        return unless @path
+
+        FileUtils.mkdir_p(File.dirname(@path))
+        tmp_path = File.join(File.dirname(@path), ".#{File.basename(@path)}.#{$$}.#{Thread.current.object_id}.tmp")
+        File.write(tmp_path, JSON.pretty_generate(@data))
+        File.rename(tmp_path, @path)
+      ensure
+        FileUtils.rm_f(tmp_path) if tmp_path && File.exist?(tmp_path)
+      end
+
+      def serialize_row(row)
+        {
+          "project" => row.project.to_s,
+          "slug" => row.slug.to_s,
+          "stage" => row.stage.to_s,
+          "marker" => row.marker.to_s,
+          "attrs" => row.attrs.to_h.transform_keys(&:to_s),
+          "action" => row.action.to_s
+        }
+      end
+
+      def entry_from_raw(raw)
+        Entry.new(
+          first_seen_at: parse_time(raw["first_seen_at"]),
+          reminded_at: parse_time(raw["reminded_at"]),
+          row: row_from_raw(raw["row"])
+        )
+      end
+
+      def row_from_raw(raw)
+        raw = raw.is_a?(Hash) ? raw : {}
+        RowSnapshot.new(
+          project: raw["project"],
+          slug: raw["slug"],
+          stage: raw["stage"],
+          marker: raw["marker"],
+          attrs: raw["attrs"].is_a?(Hash) ? raw["attrs"] : {},
+          action: raw["action"]
+        )
+      end
+
+      def timestamp(time)
+        time.utc.iso8601
+      end
+
+      def parse_time(value)
+        return nil if value.nil? || value.to_s.empty?
+
+        Time.parse(value.to_s)
+      rescue ArgumentError
+        nil
+      end
+
+      def corrupt_path_for(path)
+        "#{path}.corrupt-#{Time.now.utc.strftime("%Y%m%d%H%M%S")}"
+      end
+    end
+  end
+end
--- a/lib/hive/bot/alert_store.rb
+++ b/lib/hive/bot/alert_store.rb
@@ -7,7 +7,8 @@ module Hive
     class AlertStore
       SCHEMA_VERSION = 1
 
-      Entry = Data.define(:first_seen_at, :reminded_at, :delivered_to, :row)
+      Entry = Data.define(:first_seen_at, :reminded_at, :delivered_to,
+                          :consecutive_failures, :next_attempt_after, :row)
       RowSnapshot = Data.define(:project, :slug, :stage, :marker, :attrs, :action) do
         def initialize(project:, slug:, stage:, marker:, attrs: {}, action: nil)
           super
@@ -34,6 +35,8 @@ module Hive
             "first_seen_at" => timestamp(now),
             "reminded_at" => nil,
             "delivered_to" => normalize_chat_ids(delivered_to),
+            "consecutive_failures" => 0,
+            "next_attempt_after" => nil,
             "row" => serialize_row(row)
           }
           persist_locked!
@@ -50,6 +53,35 @@ module Hive
           return false if additions.empty?
 
           entry["delivered_to"] = existing + additions
+          entry["consecutive_failures"] = 0
+          entry["next_attempt_after"] = nil
+          persist_locked!
+          true
+        end
+      end
+
+      def record_send_failure(fingerprint, now, backoff_for: nil)
+        synchronize do
+          entry = entries[fingerprint]
+          return false unless entry
+
+          failures = entry["consecutive_failures"].to_i + 1
+          delay = backoff_for || backoff_seconds(failures)
+          entry["consecutive_failures"] = failures
+          entry["next_attempt_after"] = timestamp(now + delay)
+          persist_locked!
+          true
+        end
+      end
+
+      def record_send_success(fingerprint)
+        synchronize do
+          entry = entries[fingerprint]
+          return false unless entry
+          return false if entry["consecutive_failures"].to_i.zero? && entry["next_attempt_after"].nil?
+
+          entry["consecutive_failures"] = 0
+          entry["next_attempt_after"] = nil
           persist_locked!
           true
         end
@@ -182,12 +214,24 @@ module Hive
           first_seen_at: parse_time(raw["first_seen_at"]),
           reminded_at: parse_time(raw["reminded_at"]),
           delivered_to: Array(raw["delivered_to"]).map(&:to_s),
+          consecutive_failures: raw["consecutive_failures"].to_i,
+          next_attempt_after: parse_time(raw["next_attempt_after"]),
           row: row_from_raw(raw["row"])
         )
       end
 
       def normalize_chat_ids(chat_ids)
         Array(chat_ids).map(&:to_s).uniq
+      end
+
+      BACKOFF_BASE_SEC = 60
+      BACKOFF_MAX_SEC = 1800
+
+      def backoff_seconds(failures)
+        # 1=60, 2=120, 3=240, 4=480, 5=960, 6=1800, cap 1800
+        exponent = [ failures - 1, 5 ].min
+        delay = BACKOFF_BASE_SEC * (2**exponent)
+        [ delay, BACKOFF_MAX_SEC ].min
       end
 
       def row_from_raw(raw)
--- a/lib/hive/bot/alert_store.rb
+++ b/lib/hive/bot/alert_store.rb
@@ -31,16 +31,18 @@ module Hive
 
       def add(fingerprint, row, now, delivered_to: [])
         synchronize do
-          entries[fingerprint] = {
-            "first_seen_at" => timestamp(now),
-            "reminded_at" => nil,
-            "delivered_to" => normalize_chat_ids(delivered_to),
-            "consecutive_failures" => 0,
-            "next_attempt_after" => nil,
-            "absent_since" => nil,
-            "row" => serialize_row(row)
-          }
-          persist_locked!
+          with_rollback do
+            entries[fingerprint] = {
+              "first_seen_at" => timestamp(now),
+              "reminded_at" => nil,
+              "delivered_to" => normalize_chat_ids(delivered_to),
+              "consecutive_failures" => 0,
+              "next_attempt_after" => nil,
+              "absent_since" => nil,
+              "row" => serialize_row(row)
+            }
+            persist_locked!
+          end
         end
       end
 
@@ -50,8 +52,10 @@ module Hive
           return false unless entry
           return false if entry["absent_since"]
 
-          entry["absent_since"] = timestamp(now)
-          persist_locked!
+          with_rollback do
+            entry["absent_since"] = timestamp(now)
+            persist_locked!
+          end
           true
         end
       end
@@ -62,8 +66,10 @@ module Hive
           return false unless entry
           return false if entry["absent_since"].nil?
 
-          entry["absent_since"] = nil
-          persist_locked!
+          with_rollback do
+            entry["absent_since"] = nil
+            persist_locked!
+          end
           true
         end
       end
@@ -77,10 +83,12 @@ module Hive
           additions = normalize_chat_ids(chat_ids) - existing
           return false if additions.empty?
 
-          entry["delivered_to"] = existing + additions
-          entry["consecutive_failures"] = 0
-          entry["next_attempt_after"] = nil
-          persist_locked!
+          with_rollback do
+            entry["delivered_to"] = existing + additions
+            entry["consecutive_failures"] = 0
+            entry["next_attempt_after"] = nil
+            persist_locked!
+          end
           true
         end
       end
@@ -92,9 +100,11 @@ module Hive
 
           failures = entry["consecutive_failures"].to_i + 1
           delay = backoff_for || backoff_seconds(failures)
-          entry["consecutive_failures"] = failures
-          entry["next_attempt_after"] = timestamp(now + delay)
-          persist_locked!
+          with_rollback do
+            entry["consecutive_failures"] = failures
+            entry["next_attempt_after"] = timestamp(now + delay)
+            persist_locked!
+          end
           true
         end
       end
@@ -105,36 +115,56 @@ module Hive
           return false unless entry
           return false if entry["consecutive_failures"].to_i.zero? && entry["next_attempt_after"].nil?
 
-          entry["consecutive_failures"] = 0
-          entry["next_attempt_after"] = nil
-          persist_locked!
+          with_rollback do
+            entry["consecutive_failures"] = 0
+            entry["next_attempt_after"] = nil
+            persist_locked!
+          end
           true
         end
       end
+
+      private
+
+      def with_rollback
+        snapshot = Marshal.load(Marshal.dump(@data))
+        yield
+      rescue StandardError
+        @data = snapshot
+        raise
+      end
+
+      public
 
       def mark_reminded(fingerprint, now)
         synchronize do
           entry = entries[fingerprint]
           next unless entry
 
-          entry["reminded_at"] = timestamp(now)
-          persist_locked!
+          with_rollback do
+            entry["reminded_at"] = timestamp(now)
+            persist_locked!
+          end
         end
       end
 
       def remove(fingerprint)
         synchronize do
-          entry = entries.delete(fingerprint)
-          persist_locked! if entry
-          entry ? row_from_raw(entry["row"]) : nil
+          entry = entries[fingerprint]
+          next nil unless entry
+
+          with_rollback do
+            entries.delete(fingerprint)
+            persist_locked!
+          end
+          row_from_raw(entry["row"])
         end
       end
 
       def remove_matching(project:, slug:, stage: nil, marker: nil, match_attr: nil)
         attr_key, attr_value = parse_match_attr(match_attr)
         synchronize do
-          before = entries.size
-          entries.delete_if do |_fingerprint, raw|
+          targets = entries.select do |_fingerprint, raw|
             row = row_from_raw(raw.is_a?(Hash) ? raw["row"] : nil)
             next false unless row.project.to_s == project.to_s
             next false unless row.slug.to_s == slug.to_s
@@ -144,9 +174,13 @@ module Hive
 
             true
           end
-          removed = before - entries.size
-          persist_locked! if removed.positive?
-          removed
+          next 0 if targets.empty?
+
+          with_rollback do
+            targets.each_key { |fingerprint| entries.delete(fingerprint) }
+            persist_locked!
+          end
+          targets.size
         end
       end
 
--- a/lib/hive/bot/alert_store.rb
+++ b/lib/hive/bot/alert_store.rb
@@ -8,7 +8,7 @@ module Hive
       SCHEMA_VERSION = 1
 
       Entry = Data.define(:first_seen_at, :reminded_at, :delivered_to,
-                          :consecutive_failures, :next_attempt_after, :row)
+                          :consecutive_failures, :next_attempt_after, :absent_since, :row)
       RowSnapshot = Data.define(:project, :slug, :stage, :marker, :attrs, :action) do
         def initialize(project:, slug:, stage:, marker:, attrs: {}, action: nil)
           super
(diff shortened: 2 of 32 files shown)


ivankuznetsov:

## High
- [ ] lib/hive/bot/notification_dispatcher.rb:41 sends false recovery confirmations on fingerprint changes: when the same stuck slug/stage remains active with different attrs, the old fingerprint is missing from `current` so the dispatcher emits `Recovered` before the required new stuck alert, violating R9/R12.
- [ ] lib/hive/bot/handlers/callback_handlers.rb:61 does not reset alert state when the operator taps Autofix: the handler only dispatches commands, and `process_current` suppresses an existing fingerprint, so a retry that fails again with the same fingerprint before an intervening status tick is silently deduped instead of re-alerting per R10/R13.
- [ ] lib/hive/bot/supervisor.rb:461 still exposes raw diagnostic payload through legacy details replies: `reply_for_child` sends `summary`, `detail`, and `Path` from `hive status --diagnose`, so pre-redeploy Show details buttons can still surface marker/phase/reason tokens despite R4/A14 dropping raw payload from Telegram.

## Medium

## Nit
- [ ] lib/hive/config.rb:1082 emits deprecated dedupe-config warnings only with `warn`: the plan calls for a structured `:deprecated_config` bot log event, and the schema enum was added but no code ever emits it.

ivankuznetsov:

## High
- [ ] `lib/hive/bot/notification_dispatcher.rb:34-46` sends a premature `✅ Recovered` when the alert fingerprint changes on an unhealed row: any time a still-stuck row's marker or attrs evolve (e.g. `pass=2 → pass=3`) the old fingerprint disappears from `current`, `process_recoveries` emits a Recovered confirmation, and the new fingerprint immediately re-fires "stuck" — operator sees `✅ Recovered → ⚠ stuck again` for a row that never actually recovered, contradicting R12's "after Autofix success" intent. The added test `test_marker_attr_change_renotifies` codifies this behavior, so it is deliberate but at odds with the plan.
- [ ] `lib/hive/bot/notification_dispatcher.rb:43,51-58,73-77` recovery and 8 h-reminder sends drop the `:notification_sent` structured log event that every other Telegram send emits: only the new-alert path and reminder path call `log_notification_sent`, and `process_recoveries` never logs anything when a Recovered message goes out. Operators auditing the bot log lose visibility into when the lifecycle ended; the schema added `unknown_stage_label/alert_store_corrupt/deprecated_config` but no recovery/lifecycle event.

## Medium
- [ ] `lib/hive/bot/notification_builders.rb:149-183,199-205` recovery classifier helpers (`open_laptop_only_recovery?`, `stale_agent_working_pending_heal?`, `markerless_retry_recovery?`, `diagnostic_manual_fix?`, `diagnostic_suggested_action`) and `refresh_diagnose_callback` are now unreachable dead code with stale comments referring to removed keyboard branches. Plan U2 claimed their classification "is still used internally to pick the autofix command sequence", but `CallbackHandlers#retry_commands` does its own marker-name special-casing instead — these helpers are not referenced by any production code path.
- [ ] `lib/hive/bot/notification_dispatcher.rb:88-92` reminder headline hardcodes the literal string `"(8 h)"` while `bot.recovery_reminder_window_sec` is configurable in `[3600, 604800]`. If an operator tunes the window to 4 h or 12 h, the message still says "Still stuck (8 h)" — misleading wording the moment the default is overridden.
- [ ] `lib/hive/config.rb:1075-1083` deprecation path uses `Kernel#warn` to stderr instead of the bot logger event the plan specified; the new `:deprecated_config` enum value added to `lib/hive/bot/logger.rb:32` and `schemas/hive-bot-log.v1.json:39` is unused. Either wire the bot logger into config validation or remove the unused enum — schema enums without an emitter are silent drift.
- [ ] `lib/hive/bot/handlers/callback_handlers.rb:191-204` second-Autofix failure can silently dedupe instead of re-alerting (R13). The implementation depends on the retry's failure producing a *different* fingerprint (e.g. an incremented `pass` counter). For a deterministic crash that re-emits the identical marker+attrs after Autofix, the dispatcher's `process_current` falls into the `else` branch and logs `:notification_skipped_dedupe` instead of re-alerting. R13 expects the operator to see another alert; the current mechanism cannot guarantee that.

## Nit
- [ ] `lib/hive/bot/title_formatter.rb:39-49` `unknown_stage_labels` "log once" cache mutates a module-level hash without synchronization. Bot threads (poll/status/reaper) can hit `stage_label` concurrently for the same unknown key; the `key?` check is racy and may double-log. Narrow window, low impact, but trivially fixable with a `Mutex` or `MonitorMixin`.
- [ ] `lib/hive/bot/handlers/slash_handlers.rb:16-22` `/status` project arg accepts internal whitespace: `update.text.split(/\s+/, 2)[1].strip` returns `"foo bar"` for `/status foo bar`, producing the awkward reply `No tasks for project foo bar.`. Today the supervisor's `status_command?` short-circuit prevents the malformed `--project "foo bar"` from reaching the CLI, but a defensive split-on-first-token would be safer and produce a cleaner UX.
- [ ] `lib/hive/bot/title_formatter.rb:20-26` sentence-casing mangles common acronyms: `open-pr-task-260525-abcd` renders as `Open pr task…` instead of `Open PR task…`. Operator-facing messages read as typos for slugs containing `pr`, `ci`, `db`, etc.
- [ ] `lib/hive/bot/notification_dispatcher.rb:34-44` recovery-message retry loop is unbounded — every status tick during a Telegram outage re-attempts the same Recovered message (entry stays in store until success), producing one `:send_failure` event per chat per tick. User-facing impact is minimal (only the first successful retry delivers a message) but the log noise during extended outages is significant.
- [ ] `lib/hive/bot/notification_dispatcher.rb:78-86,88-94` `recovered_message` and `reminder_notification` call `TitleFormatter.stage_label(row.stage)` without passing `logger:`, so unknown stage labels never emit `:unknown_stage_label` from these paths — only from `Supervisor#render_queue`. Inconsistent with the once-per-process intent.
- [ ] `lib/hive/bot/handlers/callback_handlers.rb:181-200` legacy `clear_and_retry` route now goes through the new `retry_commands` helper, which adds two behavior shifts beyond plan scope: AGENT_WORKING markers skip the `markers clear` step (this is a fix — the old path would exit 4) and `9-done` stages now return no commands (the old path would still dispatch `markers clear`). Worth a short comment so the divergence is intentional and discoverable.
- [ ] `wiki/commands/bot.md:38` documents `Title... — Stage` (three ASCII dots) but the code emits `Title…` (U+2026 horizontal ellipsis). Tiny doc/code mismatch that will trip a future grep.
- [ ] `lib/hive/bot/alert_store.rb:139-147` `row_from_raw` accepts nil/missing fields silently, returning a `RowSnapshot` with `project: nil` etc. Combined with `recovery?(row)` calling `row.marker.to_s == "review_error"`, a corrupt-but-parseable entry collapses to "non-recovery" and the dispatcher silently drops it. Defensive enough, but a debug log when fields are missing would surface real corruption faster than waiting for a rename to `.corrupt-<ts>`.

ivankuznetsov:
Update: fixed the recurring review recovery failure where a successful fix agent could exit with dirty files and Hive would land REVIEW_ERROR reason=fix_dirty_worktree. Phase 4 now stages and commits successful dirty fix-agent output with the required rollback-rate trailers before running fix guardrails; commit failures surface as REVIEW_ERROR reason=fix_auto_commit_failed.

Also manually committed the dirty review-fix output that caused this recovery row, then cleared the stale local REVIEW_ERROR marker after push. The task row is now marker=none / Ready for review.

Verification:
- ruby -c lib/hive/stages/review.rb
- ruby -c test/integration/run_review_test.rb
- bundle exec ruby -Ilib -Itest test/integration/run_review_test.rb -n /dirty_worktree_is_auto_committed/
- bundle exec ruby -Ilib -Itest test/integration/run_review_test.rb
- bundle exec rake test (3508 runs, 12627 assertions, 0 failures, 0 errors, 2 skips)

ivankuznetsov:

## High
- [ ] Autofix taps do not reset the stored alert key (`lib/hive/bot/handlers/callback_handlers.rb:61`, `lib/hive/bot/notification_dispatcher.rb:61`): R10/R13 require user action to allow a second same-fingerprint failure alert, but the callback only dispatches commands and `process_current` keeps deduping the existing store entry unless a later status snapshot first drops the row.
- [ ] The review runner auto-commit change is outside the plan boundary (`lib/hive/stages/review.rb:473`): the task explicitly excludes stage runners, but this changes successful fix-agent dirty-worktree behavior from `REVIEW_ERROR` to `git add -A` plus commit.

## Medium
- [ ] Fingerprint changes emit a false recovery message (`lib/hive/bot/notification_dispatcher.rb:41`): when attrs change for the same stuck row, the old fingerprint is treated as recovered before the new one is alerted, so operators see `✅ Recovered` even though the task never left the action-needed set.
- [ ] `/status <project>` drops words after the first space (`lib/hive/bot/handlers/slash_handlers.rb:18`): project names are registry names and can contain spaces, but parsing with `split(/\s+/)[1]` sends `--project my` for `/status my project`.

## Nit

ivankuznetsov:

## High
- [ ] `lib/hive/bot/notification_dispatcher.rb:41-56` still emits a false `✅ Recovered` when the alert fingerprint changes on an unhealed row: pass 01 escalation Q1 (in `reviews/escalations-01.md`) is still unanswered, codex-02 reaffirms it, and the test `test_marker_attr_change_renotifies` (`test/unit/bot/notification_dispatcher_test.rb:122-128`) now codifies the buggy 3-message sequence (`stuck → Recovered → stuck again`). R12 says Recovered fires only when the row leaves the action-needed set; here it fires whenever attrs evolve (`pass=2 → pass=3`). Plan U4 literal description and brainstorm R12/A11 intent disagree — needs a decision before merge.
- [ ] `lib/hive/bot/handlers/callback_handlers.rb:61-74` Autofix callback does not touch `AlertStore`, so R13 ("second Autofix failure re-alerts with same Autofix button") is not guaranteed for deterministic-failure cases that re-emit the identical `(marker, attrs)` fingerprint. Pass 01 escalation Q2 is still unanswered and codex-02 reaffirms it; the next status tick falls into `process_current`'s `else` branch and logs `:notification_skipped_dedupe` instead of re-alerting. Needs an architectural decision (clear store entries by `(project, slug, stage)` in the autofix handler, or encode the fingerprint in callback data).
- [ ] `lib/hive/stages/review.rb:473-485,1470-1521` introduces an auto-commit branch inside the review stage runner; the plan's **Scope Boundaries / Out of scope** list explicitly excludes "Stage runners". The change is operationally useful but the plan does not defer it to a separate task — it is simply outside this slice. Either split this commit into a dedicated task (with its own plan) or amend this plan's scope explicitly before merge.

## Medium
- [ ] `lib/hive/stages/review.rb:1471` uses `git add -A` to stage everything in the worktree before committing the fix. `Hive::ProtectedFiles.diff` only guards a fixed protected set (task.md, plan.md, worktree.yml); any other untracked debris the fix-agent leaves behind (scratch notes, debug logs, accidental dotfiles) is silently committed with the `Hive-Fix-*` trailers. Constrain the add to the diff intended by the fix agent (e.g. `git add -u` plus an explicit list of code files), or document why untracked-file commits are intentional.
- [ ] `lib/hive/bot/handlers/slash_handlers.rb:18` `/status <project>` parses with `split(/\s+/)[1]`, dropping every token after the first whitespace. Project names registered via `Hive::Config.register_project(name:, path:)` (`lib/hive/config.rb:496`) are arbitrary strings and may contain spaces, so `/status my project` filters on `--project my` and silently misses tasks for `my project`. Plan U7's literal code used `split(/\s+/, 2)[1]` to preserve the whole tail; the pass 01 fix tightened it to avoid an awkward "No tasks for project foo bar." reply, but lost legitimate multi-word names in the process. Either reject names containing whitespace at registration time, or restore the multi-token parse.

## Nit
- [ ] `lib/hive/config.rb:178` hardcodes `"alert_state_file" => "~/Dev/hive/.bot.alert_state.json"` in `DEFAULTS["bot"]`, but `Config#bot_defaults` (`lib/hive/config.rb:491`) immediately overrides it to `File.join(Hive::Paths.state_home, ".bot.alert_state.json")`. The DEFAULTS value is unreachable in practice and points at a developer-specific path; tests or future callers that read `DEFAULTS["bot"]["alert_state_file"]` directly will get the wrong location. Either remove the DEFAULTS entry (let `bot_defaults` own the path) or align it with the state_home form for consistency.
- [ ] `lib/hive/bot/notification_dispatcher.rb:107-113` `reminder_notification` rebuilds the recovery notification, splits on lines, and rewrites `lines[0]`. This silently depends on `NotificationBuilders.recovery` emitting exactly three lines with the headline first; if a future tweak adds a leading blank line or reorders, the reminder headline is mutated on the wrong line with no test catching it. A small guard (`raise unless lines[0].start_with?("⚠ ")`) or a builder method like `recovery_with_headline(headline)` would make the coupling explicit.

ivankuznetsov:
Update: added the requested defaults changes.

- Reviewer agent timeout fallback is now `3600s` instead of `600s`.
- Fresh `hive init` reviewer entries now render `timeout_sec: 3600`.
- Daemon default parallel task caps are now `max_concurrent_runs: 3` and `max_concurrent_per_project: 3` instead of `5/5`.

Verification:
- `ruby -c lib/hive/reviewers/agent.rb`
- `ruby -c lib/hive/config.rb`
- `ruby -c test/unit/reviewers/agent_test.rb`
- `ruby -c test/integration/init_test.rb`
- `bundle exec ruby -Ilib -Itest test/unit/reviewers/agent_test.rb -n /timeout/`
- `bundle exec ruby -Ilib -Itest test/unit/config_test.rb -n /daemon/`
- `bundle exec ruby -Ilib -Itest test/integration/init_test.rb -n '/reviewers|limit_defaults/'`
- `bundle exec ruby -Ilib -Itest test/unit/reviewers/agent_test.rb test/unit/config_test.rb test/integration/init_test.rb`
- `bundle exec rake test` (3509 runs, 12629 assertions, 0 failures, 0 errors, 2 skips)

ivankuznetsov:
Pass-two cleanup is now pushed in `7e73ca91`.

What changed:
- committed the pass-two fix-agent edits instead of leaving the task in `fix_dirty_worktree`
- kept the successful-fix auto-commit path from `3ea552b6`
- extended reviewer defaults/templates to `3600` seconds
- changed daemon default parallel task limits from `5` to `3`

Verification:
- `bundle exec ruby -Ilib -Itest test/unit/bot/slash_handlers_test.rb`
- `bundle exec ruby -Ilib -Itest test/integration/config_bot_test.rb`
- `bundle exec ruby -Ilib -Itest test/unit/config_test.rb`
- `bundle exec ruby -Ilib -Itest test/unit/bot/notification_dispatcher_test.rb`
- `bundle exec ruby -Ilib -Itest test/unit/reviewers/agent_test.rb`
- `bundle exec ruby -Ilib -Itest test/integration/init_test.rb`
- `bundle exec ruby -Ilib -Itest test/integration/run_review_test.rb`
- `bundle exec rake test` -> `3510 runs, 12632 assertions, 0 failures, 0 errors, 2 skips`

The pass-two `fix_dirty_worktree` row came from the daemon still executing the old main checkout, before this PR's auto-commit behavior is available to that runtime.

ivankuznetsov:
Final follow-up pushed at `6b623aaf`.

Additional cleanup:
- committed the late ce-code-review safe-fix output instead of leaving the PR worktree dirty
- verified the new bot/config hardening with affected tests
- reran full suite after the follow-up commit
- updated this machine's live Hive state config so existing reviewers use `timeout_sec: 3600`
- added explicit live daemon caps `max_concurrent_runs: 3` and `max_concurrent_per_project: 3` until the PR defaults are merged

Verification:
- `bundle exec ruby -Ilib -Itest test/unit/bot/alert_store_test.rb` -> `8 runs, 26 assertions, 0 failures`
- `bundle exec ruby -Ilib -Itest test/unit/bot/callback_handlers_test.rb` -> `20 runs, 63 assertions, 0 failures`
- `bundle exec ruby -Ilib -Itest test/unit/bot/notification_builders_test.rb` -> `17 runs, 78 assertions, 0 failures`
- `bundle exec ruby -Ilib -Itest test/unit/bot/notification_dispatcher_test.rb` -> `17 runs, 56 assertions, 0 failures`
- `bundle exec ruby -Ilib -Itest test/unit/config_test.rb` -> `122 runs, 501 assertions, 0 failures`
- `bundle exec rake test` -> `3521 runs, 12659 assertions, 0 failures, 0 errors, 2 skips`

ivankuznetsov:
Rebased PR #152 onto current `origin/main` and force-with-lease pushed the rebased head `85d9d59d`.

Post-rebase verification:
- `bundle exec rake test` -> `3521 runs, 12659 assertions, 0 failures, 0 errors, 2 skips`

The branch is no longer behind the base. GitHub currently reports merge state `BLOCKED`, so remaining merge gating is external to the local branch state.

ivankuznetsov:
Split done.

PR #152 has been rewritten back to bot-only scope. The review-runner/defaults changes that were accidentally mixed into this branch moved to #154:
https://github.com/ivankuznetsov/hive/pull/154

Validation after the split:
- bot focused tests: 269 runs, 1043 assertions, 0 failures
- bot full suite: 3520 runs, 12642 assertions, 0 failures, 2 skips
- infra focused tests on #154: 206 runs, 1000 assertions, 0 failures
- infra full suite on #154: 3491 runs, 12538 assertions, 0 failures, 2 skips

ivankuznetsov:

Verified all pass-01 and pass-02 High items are addressed at HEAD `551ada89`. Pass 03 surfaces net-new findings, mostly outage / multi-chat / race edges introduced by the new persistence + autofix-reset architecture.

**Reviewers (13):** correctness, testing, maintainability, project-standards, security, performance, reliability, api-contract, adversarial, cli-readiness, previous-comments, ce-agent-native, ce-learnings-researcher.

**Verdict:** Ready with fixes for single-chat allowlists. Multi-chat installs hit P1 #1.

---

#### Applied as 7 safe_auto fixes (worktree only, not committed/pushed)

- [x] `templates/hive_config.yml.erb:18` — removed deprecated `notification_dedupe_window_sec: 300` suggestion line
- [x] `lib/hive/bot/alert_store.rb:134` — added `f.fsync` on tmp + best-effort parent-dir fsync so the new persistence is durable across power loss
- [x] `lib/hive/bot/supervisor.rb:517` — replaced "use /status for full list" dead pointer with "open on a laptop for the full list" (the cap is the same on both paths)
- [x] `lib/hive/bot/title_formatter.rb:21` — `ACRONYM_REGEXPS` now case-insensitive so the sentence-cased first word still gets `PR`/`CI`/etc instead of `Pr`/`Ci`
- [x] `wiki/modules/bot.md:59` — `Title... — Stage` → `Title… — Stage` (U+2026, matches code)
- [x] `wiki/log.md:1995` — same ellipsis fix
- [x] `lib/hive/bot/notification_builders.rb:165` — removed unreachable `marker == "agent_working"` clause from `manual_only_recovery?` (caller pre-filters before this method runs)

`bundle exec ruby -Ilib -Itest` on 9 touched test files: `253 runs, 973 assertions, 0 failures, 0 errors, 0 skips`.

---

#### High (5 — needs decision before merge)

- [ ] `lib/hive/bot/notification_dispatcher.rb:86-99` — **Multi-chat partial-send re-fans entire alert to already-succeeded chats.** `send_notification` iterates the full allowlist on every retry and only persists the fingerprint when ALL chats succeed. If chat 12345 succeeds and chat 67890 fails, the next tick sends to both again — 12345 receives a duplicate. The test `test_partial_multi_chat_failure_does_not_mark_fingerprint_delivered` at `test/unit/bot/notification_dispatcher_test.rb:210-225` codifies the duplicate-delivery behavior as expected, so this is deliberate but breaks the PR's "exactly once" promise for multi-chat allowlists. Single-chat installs unaffected. Suggested fix: track per-chat delivery state in `AlertStore` so only chats that did not yet receive this fingerprint get re-attempted.

- [ ] `lib/hive/bot/handlers/callback_handlers.rb:63-84` + `supervisor.rb:213` — **Autofix double-tap spawns parallel `hive develop/review`.** `dispatch_command_sequence` fire-and-forgets the final command (line 213 only waits for non-final commands); the original Telegram button is never disabled after the first tap. Two rapid taps → two `markers clear` + two parallel retry stage runs against the same slug. Suggested fix: clear the inline keyboard on first tap (Telegram `editMessageReplyMarkup`) or dedupe in-flight retries per `(project, slug)` in the dispatcher.

- [ ] `lib/hive/bot/supervisor.rb:252,256` — **`/status` swallows fetch failures.** When `@status_watcher.fetch` returns `ok=false`, the reply renders "No tasks for project X." — same as a legitimately empty queue. Also conflates unknown-project with empty-queue. Suggested fix: check `Result#ok` and reply `hive status unavailable: <error>`; track registered project names from the envelope and reply `Unknown project <name>. Known: a, b, c.` when the filter excludes everything.

- [ ] `lib/hive/bot/notification_dispatcher.rb:33-50` — **Telegram outage cascade: rows × ticks × chats unbounded re-fanout.** Each status tick re-attempts every queued send during a Telegram outage; recovery-message retry has no per-fingerprint attempt counter. When Telegram resumes after a multi-hour outage, operators see a multiplicative burst of stale ⚠ and ✅ Recovered messages for incidents that healed during the outage. Suggested fix: persist `last_send_attempt_at` per fingerprint, back off retries, send a single `N alerts queued during outage` summary on first successful send after an outage (parallel to `send_reconnect_summary`, which currently only fires on bot restart). Corroborated by reliability + adversarial + testing reviewers.

- [ ] `lib/hive/bot/notification_dispatcher.rb:45-58` — **Cross-tick false ✅ Recovered when row transiently goes `agent_running` between failing ticks.** Pass-02's `current_recovery_for_same_row?` catches the same-tick attrs-change case, but not the cross-tick case: tick N=fail / tick N+1=agent_running (suppressed from `current`) / tick N+2=fail-with-different-attrs. Operator sees `⚠ stuck → ✅ Recovered → ⚠ stuck again` for a row that never healed. Suggested fix: require absence to persist ≥2 ticks before emitting Recovered.

#### Medium (9 — should fix)

- [ ] `lib/hive/bot/supervisor.rb:203` + `notification_dispatcher.rb:27` — **Status-loop race re-fires same ⚠ alert immediately after Autofix tap.** `reset_alert_for_result` runs BEFORE `markers clear` dispatches. If the status tick fires before the child writes the new marker, `process_current` finds no entry and re-alerts the original fingerprint. Defer `alert_reset` until after markers-clear exits 0, or suppress `process_current` for fingerprints with an in-flight autofix child. Corroborated by correctness + reliability.

- [ ] `lib/hive/bot/notification_dispatcher.rb:27` + `alert_store.rb:60-73` — **`reset_task` removes ALL entries for `(project, slug, stage)`.** Distinct fingerprints (multiple markers at the same stage) all get cleared by a single Autofix tap, then re-alert as "new" on the next tick. Forward the exact fingerprint from the Autofix callback through `alert_reset` and remove only that one.

- [ ] `lib/hive/bot/alert_store.rb:31` — **In-memory diverges from disk on `persist_locked!` failure.** `add()` mutates `@data` before persistence; on disk-full / EIO the live process stays correct but a restart re-alerts everything in flight — defeating the PR's restart-survival promise. Supervisor catches the exception as `:fatal` so the bot keeps running silently with diverged state. Snapshot `@data` before mutation; restore on persist failure.

- [ ] `lib/hive/bot/notification_builders.rb:161-167` + `lib/hive/bot/handlers/callback_handlers.rb` — **Parallel `manual_only` predicates can diverge.** `NotificationBuilders#manual_only_recovery?` and `CallbackHandlers#manual_only_marker?` redefine the same concept with different scopes. Extract a single predicate in `NotificationBuilders` and have `CallbackHandlers#manual_only_marker?` delegate.

- [ ] `lib/hive/bot/handlers/callback_handlers.rb:174-189` — **`findings_accept_all` / `findings_reject_all` do not propagate `alert_reset`.** Unlike `autofix` and `clear_and_retry`, the findings-toggle callbacks change row state without clearing the alert store entry. Same-fingerprint re-failure can be silently deduped. Either add `alert_reset` to `findings_toggle` matching the autofix/clear_and_retry pattern, or add a regression test pinning the intentional omission with a comment.

- [ ] `lib/hive/bot/notification_dispatcher.rb:54` — **Recovered-message Telegram send-failure path is untested.** Pairs with the outage-cascade finding above. No regression test pins the "exactly one Recovered message even across an outage" promise. Add a `FlakyTelegram` test injecting failure on the second `process_rows([])` call; assert AlertStore entry remains and a subsequent `process_rows([])` emits a single Recovered when delivery recovers.

- [ ] `lib/hive/config.rb:1099-1123` — **Deprecation knowledge duplicated** between `Kernel#warn` path (`warn_deprecated_bot_dedupe!`) and `:deprecated_config` structured-event path. Future additions can drift. Have `warn_deprecated_bot_dedupe!` iterate `deprecated_bot_keys(bot)` and call `Kernel#warn` for each, or remove the warn and rely on the structured event alone.

- [ ] `lib/hive/bot/supervisor.rb:507` — **`/status` reply is human-prose only — no `--json` / `--format` escape for automated callers.** Agent-native parity gap. The bot CAN dispatch real `hive status --json` subprocesses today via the non-status path, but `/status` intercepts and renders text. Honour `--json` in slash text or attach `format: :json` to the `Result` so `execute_dispatch` returns the validated envelope.

- [ ] `lib/hive/bot/notification_dispatcher.rb:118-132` — **Reminder fires every tick after window boundary if send fails.** No `last_reminder_attempt_at`, bounded only by reminder window expiry. Persist a `last_reminder_attempt_at` even on send failure to throttle retries.

#### Nit (8)

- [ ] `lib/hive/bot/alert_store.rb:138` — **Temp-file leak on SIGKILL between write and rename.** The tmp filename includes PID + thread id so collisions are unlikely, but orphans never get GC'd. Scan `dirname(@path)` for `.<basename>.*.tmp` at startup and unlink.

- [ ] `lib/hive/bot/notification_builders.rb:127` — **`recovery()` drops `logger:` from `TitleFormatter.stage_label`** — so `:unknown_stage_label` events never emit from the primary alert path. Thread `logger:` through `NotificationBuilders.recovery` and forward. (Not applied locally because it requires updating call sites in dispatcher.)

- [ ] `lib/hive/bot/handlers/callback_handlers.rb:55` — **`clear_and_retry` on a stage with no retry verb triggers `alert_reset` but dispatches zero commands.** Mirror `autofix()`'s guard — return `:reply` and omit `alert_reset` when `retry_commands` returns `[]`.

- [ ] `lib/hive/bot/handlers/slash_handlers.rb:20` — **`command_argv` claims `--project` filters the snapshot, but `hive status`'s `--project` only scopes `--diagnose`.** Works today only because `Supervisor#status_command?` short-circuits. Either add a real snapshot-filter `--project` to `hive status`, or drop `--project` from `command_argv` and keep the filter purely in `Result.project`.

- [ ] `lib/hive/bot/supervisor.rb:222-228` — **Defensive `respond_to?` guards in `reset_alert_for_result` hide the actual contract.** Drop the `result.respond_to?(:alert_reset)` and `@notification_dispatcher.respond_to?(:reset_task)` checks; keep the `reset && ...` nil guard. (Not applied locally because it touches test-mock contracts.)

- [ ] `test/unit/bot/alert_store_test.rb:117,156` — **Corrupt-rename failure path untested + concurrency test is a no-op smoke test** (distinct fingerprints, not contention). Add a chmod-readonly-parent test to exercise `handle_corrupt!` rename failure; add a contention test where N threads add/remove the same fingerprint while a reader continuously reloads.

- [ ] `test/integration/config_bot_test.rb` + `test/integration/bot/scenarios/s3_queue_test.rb` — **Integration tests bypass the alert lifecycle via `FakeNotificationDispatcher`.** No end-to-end test covers recovery row → Autofix → `reset_task` → R13 re-alert.

- [ ] `lib/hive/bot/supervisor.rb:203` — **`dispatch_command_sequence` dry_run guard for `alert_reset` is untested.** Add a `test_dispatch_command_sequence_dry_run_skips_alert_reset` that sets `@dry_run=true` and asserts `@notification_dispatcher.reset_tasks` stays empty.

#### Advisory / known (6 — no action requested)

- `lib/hive/bot/notification_dispatcher.rb:71` — Send-then-persist SIGKILL race: persistence is new in this PR so the protocol gap is now visible (was previously implicit). Persist a `pending` marker before send if this becomes load-bearing.
- `lib/hive/bot/notification_dispatcher.rb:64` — First-time deploy with absent `alert_state_file`: every currently-failing row fires a fresh ⚠ alert simultaneously. Mitigation: seed the store from a `hive status --json` snapshot before starting the new bot, or accept the one-time alert burst on rollout.
- `lib/hive/bot/notification_dispatcher.rb:146` — Reminder label renders `"481 min"` instead of `"8 h 1 min"` for non-whole-hour `recovery_reminder_window_sec` settings. Require the config to be a multiple of 3600, or render `h+m` form.
- `lib/hive/bot/title_formatter.rb:46-64` — `unknown_stage_labels` cache is process-lifetime, not reset on SIGHUP / `reload_config_if_requested`. Also a lockless lazy-init at L65 races with the Monitor-protected mutation at L49. Drop the once-only cache (log on every encounter), or rebuild on config reload.
- `schemas/hive-bot-log.v1.json:42` — 3 new enum values added to a versioned `.v1.json` schema without bumping `$id`. Additive only; flagged only if external schema caches exist.
- `lib/hive/bot/supervisor.rb:511` — `/status` and `/queue` Telegram message format changed without a versioning marker. Risk only if external tooling parses replies.

#### Verified at HEAD

- All pass-01 + pass-02 High items resolved (false-Recovered-on-attrs-change via `current_recovery_for_same_row?`; Autofix-store-reset via `alert_reset` payload + `reset_task`; legacy details exposure scrubbed; `:notification_sent` event now emitted on all lifecycle paths). Evidence in `previous-comments` reviewer's analysis.
- Review-stage runner auto-commit scope creep (pass-02 codex-H2) removed — verified `lib/hive/stages/review.rb` is not in the diff. Split moved to PR #154.
- Pre-redeploy callback verbs (`clear_retry:`, `open_laptop:`, `details:`, `refresh_diagnose:`) still routed via `router.rb:165-168` with tests.
- No security exploits in callback dispatch (whitelisted retry verbs + `Process.spawn(*argv)` no shell) or `/status` slash parse.

#### Adversarial false positive

- `lib/hive/bot/alert_store.rb:134` — adversarial reviewer claimed `tmp_path` in the `ensure` block could `NameError` when `@path` is nil and the early return fires. Ruby pre-scans method-local variables; an unassigned local reads as `nil`, so `tmp_path && File.exist?(tmp_path)` evaluates to `nil` (falsy), no exception. Finding dropped.

#### Coverage

- 13 reviewers dispatched, 13 returned. 0 failed.
- Mode-aware demotions: 2 maintainability advisory P3 demoted to residual_risks.
- Confidence-gate suppressions: ~9 findings at anchor 50 dropped.
- Stage 5b validator override: skipped per-finding validators in favor of best-judgment routing because the skill's 15-finding cap would have dropped 15 valid items including 8 mechanical safe_auto fixes.

Run artifact: `.context/compound-engineering/ce-code-review/20260525-214941-48a3d0e7/`

PR #178, opened 2026-05-26
feat(babysitter): add experimental PR repair daemon

ivankuznetsov:
## Summary

Hive watches projects through the 1→9 pipeline but goes silent once the PR is open, so stale rebases and red CI still need a human push to clear. This PR adds an experimental, opt-in `hive babysit` daemon that watches GitHub PRs out-of-band and delegates repair work (rebase, conflict resolution, CI fixes, force-push) to the project's development agent.

The babysitter is a separate process from `hive daemon` with its own PID file, log file, and config block. It does not insert a new stage into the 1→9 pipeline. Defaults are conservative: `enabled: false`, `max_concurrent_prs: 2`, `budget_usd: 50`, `interval: 10m`. Dry-run mode neuters every push, label, comment, and CI rerun.

## Architecture

```mermaid
flowchart TB
    CLI["hive babysit start"] --> Dispatcher
    Dispatcher["Babysitter::Dispatcher<br/>(poll loop, in-flight set)"] --> Tick
    Tick["ProjectTick<br/>(per-project, per-tick)"] --> Gh
    Gh["Gh.list_open_prs<br/>+ labels_ignore filter"] --> Fixer
    Fixer["PrFixer<br/>(precheck → worktree → spawn)"] --> Ctx
    Ctx["ContextBuilder<br/>(status rollup, failing logs)"] --> Spawn
    Spawn["Stages::Base.spawn_agent<br/>(execute.agent profile)"] --> Agent
    Agent["dev agent in worktree<br/>force-with-lease push"] --> GhOps
    GhOps["GhOps<br/>(give-up label + comment)"]
    Fixer --> Events["Events.jsonl + status.md"]
```

## Design decisions worth scrutinizing

| Decision | Why this shape |
|---|---|
| **Separate process from `hive daemon`** | Distinct PID file (`.babysitter.pid`), log (`logs/babysitter.log`), and concurrency model (`max_concurrent_prs` per tick). No contention with the existing daemon's `ConcurrencyController`. Parallel CLI shape (`hive babysit start/stop/status/reload/tail`) keeps operator muscle memory cheap. |
| **Agent-only repair tier** | No mechanical fallback (no lockfile regen, lint autofix, or CI rerun). The configured `execute.agent` decides everything. Simpler surface and exercises the same agent path 4-execute already trusts. |
| **Hard budget per PR (30 min, $50)** | Bounds runaway cost on a noisy repo. With `max_concurrent_prs: 2` and `interval: 10m`, worst case is $600/hr/project; defaults skew low because v1 is experimental. |
| **Early-green precheck** | `gh pr view --json mergeable,statusCheckRollup` runs before any agent spawn. Mergeable + green checks ⇒ no agent, no budget burn, single `noop` event. |
| **Fork PRs skip gracefully** | Cross-repository / fork-owned head PRs cannot be force-pushed from `origin`, so they are skipped with a `fork_pr` event and a `babysitter/needs-human` label rather than a wrong-target push. |
| **Dry-run as PATH overlay** | `bin/hive-babysitter-stub-{git,gh}` walk past leading global flags (`-C`, `-c`, `-R`, `--repo`) and intercept mutating commands, logging them to `.babysitter-dry-run-skipped.log`. Belt-and-suspenders on top of the prompt instruction; not bulletproof (absolute-path invocations bypass), and the wiki documents that boundary. |
| **In-flight set in-process** | `(project, pr_number)` tracked in the Dispatcher to prevent double-spawn across ticks. Lost on restart, which is intentional — we'd rather re-evaluate than carry stale state. |
| **`PidFile` extracted from `daemon`** | Daemon's PID-reuse defense (`process_start_time` payload) was lifted into a shared `Hive::PidFile`. Both processes now share one source of truth; the daemon command shed ~100 lines. |
| **Reuses `Stages::Base.spawn_agent`** | Same timeout kill, usage accounting, status contract. Babysitter is experimental on top of trusted plumbing, not a parallel reimplementation. |

## Configuration

```yaml
babysitter:
  enabled: false              # kill-switch; reread every tick
  interval: 10m               # accepts integer seconds or \d+[smh]
  max_concurrent_prs: 2
  labels_ignore: [wip, do-not-merge, draft]
  dry_run: false
  budget_minutes: 30
  budget_usd: 50
```

`hive init` now asks one new question to opt the project in (default yes, since the project-level kill-switch is still `enabled: false` until set otherwise).

## Out of scope (v1)

- Integration into the existing `hive daemon` dispatcher.
- Telegram notifications.
- Success PR comments (silent by design).
- Per-author / per-base filters beyond `labels_ignore`.
- A dedicated `bin/hive-babysitter` executable — exposed as `hive babysit` instead.
- `--json` envelope schemas.
- `hive doctor` integration.

## Test plan

- `test/babysitter/run.rb` — acceptance smoke suite covers early-green, `wip`-label skip, full dry-run path (no `git push`, no comment), and give-up label + comment on agent failure.
- `test/unit/babysitter/*_test.rb` — interval parsing, dispatcher signal/tick, project enumeration + label filter, PR fixer precheck/spawn/give-up, gh ops idempotence, dry-run PATH overlay, worktree recreate.
- `test/unit/commands/babysit_test.rb` — CLI flag validation, PID file lifecycle, `--once` paths.
- `bundle exec rake test`: 3640 runs, 13070 assertions, 0 failures, 0 errors, 2 skips.

For real-world validation use `examples/babysit_smoke.sh hive` against this repo with `dry_run: true`.

## Review summary

Two automated review passes ran (Claude CE + Codex CE reviewers each pass). The `pr-review-toolkit` reviewer failed to start in tmux on both passes (infra error, not a code finding). Every actionable finding was auto-fixed in-branch; no finding required user escalation. Notable fixes that landed:

- **Correctness:** symmetric in-flight set add/delete (no longer corrupts another fixer's tracking on the noop early-return); `budget_exhausted` now distinguished from generic `:failure` in the event outcome enum; `checks_green_or_queued?` fails out hard-failed checks even when a retry is queued.
- **Fork / any-author PRs:** head fetched via `pull/<n>/head`; fork-owned heads skipped with a `fork_pr` event + `needs-human` label instead of a wrong-target push.
- **Dry-run hardening:** stubs walk past leading global flags so `git -C … push` / `gh -R … pr comment` are still intercepted; the prompt now forbids absolute-path binary invocations; dry-run resolves the real git/gh on the pre-overlay PATH so non-mutating calls don't recurse; missing `.babysitter-dry-run-plan.md` downgrades the run to `:failure`.
- **Daemon robustness:** atomic PID acquisition via `O_CREAT|O_EXCL`; `interruptible_sleep` uses `CLOCK_MONOTONIC`; `stop` uses `FileUtils.rm_f`; logger handles `max_files == 1` and warns once on rotation failure; SIGHUP rebuilds the logger to pick up rotation knobs.
- **Context fidelity:** CI-log truncation is two-pass (small jobs kept whole, remainder redistributed); base HEAD + merge-base divergence surfaced to the prompt; fetch-failed job logs labelled rather than leaking stderr into the `<user_supplied_…>` block.

A handful of nits were resolved as no-fix with rationale recorded in the review notes (experimental-v1 scope: empty `task.md` stub, `-B` branch-ref reuse, strict event-enum validation guarded by the per-tick rescue).

## Known limitations

- The dry-run PATH overlay is best-effort. An agent calling `/usr/bin/git push` directly will bypass it. Use throwaway repos for destructive validation until a stronger sandbox exists.
- `--force-with-lease` does not eliminate a sub-second race against a concurrent human push; the prompt instructs a pre-push `git fetch` + `rev-parse` comparison but mid-flight collisions are possible. The wiki documents the assumption that humans aren't actively pushing the same PR.
- Two daemon processes (`hive daemon` and `hive babysit`) now coexist. Operators get two PID files and two enable paths; v2 can fold them if requested.

## Linked task

- Hive task: `we-need-to-create-experimental-260526-9d55` (stage `8-finalize`).
- Source brainstorm and plan live in the task folder (`brainstorm.md`, `plan.md`); per the project convention, the canonical copies are committed to `~/Dev/hive-private`.
- Wiki: `wiki/modules/babysitter.md`, `wiki/commands/babysit.md`, `wiki/state-model.md`.

---


ivankuznetsov:

## High
- [ ] `lib/hive/babysitter/worktree.rb:21` assumes every PR head can be fetched from `origin` by `headRefName`: `headRepository` is collected but ignored, so fork PRs and non-mirrored head refs fail before the agent runs, violating the plan's "all open PRs, any author" requirement.
- [ ] `lib/hive/babysitter/dry_run_env.rb:16` resolves the "real" git/gh after prepending the stub overlay to `PATH`: non-mutating agent calls like `git status` or `gh pr view` recurse back into the stub and can hang until the babysitter timeout, breaking dry-run end-to-end operation.
- [ ] `lib/hive/gh.rb:116` calls `gh pr list` without `--limit`: GitHub CLI defaults to 30 PRs, so repositories with more open PRs are not fully walked before the oldest-updated/concurrency selection required by the plan.
- [ ] `lib/hive/agent.rb:250` only passes the dollar cap when the selected profile exposes `budget_flag`: the default development profile is Codex and `lib/hive/agent_profiles/codex.rb:25` has no native budget flag, so the required hard `$50` babysitter spend cap is not enforced for the default path.

## Medium
- [ ] `lib/hive/babysitter/project_tick.rb:32` does not isolate failures per PR: any `PrFixer.run` exception bubbles to the dispatcher project-level rescue and aborts the remaining selected PRs instead of recording one failed PR action and continuing the tick.
- [ ] `lib/hive/babysitter/pr_fixer.rb:144` reports dry-run success solely from agent exit status: it never verifies `<worktree>/.babysitter-dry-run-plan.md` exists, so the documented dry-run artifact can be missing while the event stream says `dry_run`.

## Nit
- [ ] `lib/hive/babysitter/dispatcher.rb:62` treats SIGHUP as a log-only flag: it does not re-read the daemon-shared log rotation settings that the process-shell plan says reload should pick up.

ivankuznetsov:

## High

- [ ] Out-of-scope regression: removal of `Hive::StopHookInstaller::BACKUP_SUFFIX` and the restore branch in `Hive::ClaudeLauncher.cleanup_scratch` (lib/hive/stop_hook_installer.rb, lib/hive/claude_launcher.rb, plus deletion of two `claude_launcher_test.rb` and three `stop_hook_installer_test.rb` cases): the deleted comments explicitly tie this code to preventing `dirty_worktree` markers in 4-execute / 6-review / 8-finalize when a project owns `.claude/settings.json` (e.g. via `hive init`'s llm-wiki bootstrap). The change reintroduces the bug those comments documented and is unrelated to the babysitter plan.
- [ ] Out-of-scope churn: README replaces the working `Hive demo` embed with a raw `<video src=...mp4>` tag pointing at a GitHub raw URL, and `docs/assets/hive-demo.webp` (1.9MB asset) is deleted. The plan covers `wiki/` only — no demo-asset or README contract change is in scope, and the inline `<video>` will render only on platforms that accept raw HTML (GitHub README, yes; many mirrors and `gem info`, no), degrading discoverability.
- [ ] `Hive::Babysitter::PrFixer#run` corrupts the shared in-flight set on the noop early-return (lib/hive/babysitter/pr_fixer.rb:587-619): `key = inflight_key` runs before the `return :noop if @inflight.include?(key)` guard, so the `ensure` block's `@inflight.delete(key) if key` removes a key that this fixer did **not** add. In any future caller that overlaps PrFixer spawns (the plan's IU-5 "insert on spawn; remove on reap" contract anticipates this), the other in-flight fixer's tracking entry is silently deleted, defeating the double-spawn guard. Guard with `key = nil` before the early return, or restructure so `@inflight.add` and `@inflight.delete` are symmetric.
- [ ] `Hive::Babysitter::PrFixer#outcome_for` never returns `:budget_exhausted` (lib/hive/babysitter/pr_fixer.rb:702-708 + lib/hive/babysitter/events.rb:269-278): plan IU-5 step 7 lists "budget exhausted → give-up handler" as a distinct path, and `Events::OUTCOMES` explicitly enumerates `budget_exhausted`. The implementation collapses every non-`:ok`/non-`:timeout` agent exit into `:failure`, so the events log can never carry `outcome=budget_exhausted` and operators auditing cost cannot distinguish "agent crashed" from "we capped a runaway $50 spawn".

## Medium

- [ ] `Hive::Gh.fetch_failed_job_log` returns `err.to_s` on non-zero `gh run view` exit (lib/hive/gh.rb), and `pr_failing_job_logs` puts that result into the prompt's `failing_jobs[].log` field rendered inside the `<user_supplied_…>` block. A transient gh failure (auth blip, repo-missing) ends up shown to the agent as the "log" for a failing job, encouraging it to "fix" a tooling problem instead of CI. Either skip jobs whose logs failed to fetch, or label them explicitly as a fetch error so the prompt doesn't treat them as real output.
- [ ] Inconsistent `pr_status_rollup` working directory between pre-check and ContextBuilder (lib/hive/babysitter/pr_fixer.rb:593 vs. lib/hive/babysitter/context_builder.rb:23): the early-green check runs in `@project.fetch("path")`, but the context builder re-runs the same gh call against `worktree.path`. Two API hits per non-green PR for identical data, in different cwds, with no shared cache; also two chances to disagree if the worktree's `origin` ever differs from the project's.
- [ ] Prompt template mismatch with plan R3 (templates/babysitter_pr_fix_prompt.md.erb): plan §R3 mitigation says "the prompt explicitly instructs the agent not to use absolute paths in dry-run and to write a plan file instead". The rendered prompt's dry-run branch tells the agent to write the plan file but never says "do not invoke `/usr/bin/git` or `/usr/bin/gh` directly", which is the actual bypass the PATH overlay cannot stop. Add the explicit no-absolute-paths line so the prompt matches the plan's documented mitigation.
- [ ] Give-up comment de-dup window collapses on hour boundary (lib/hive/babysitter/gh_ops.rb:404-406): `give_up_marker` uses `now.strftime('%Y%m%d%H')`, so two give-ups at 12:59 and 13:01 each carry a distinct marker and both get posted. Plan IU-9 says "cap one give-up comment per PR per hour" — implement as a sliding 60-minute window against the most-recent marker timestamp, not a calendar-hour bucket.
- [ ] `Hive::Babysitter::Logger` rotation reopens the file after a `SystemCallError` mid-rotation and continues writing (lib/hive/babysitter/logger.rb:535-543): if `File.rename` raises (e.g. `Errno::ENOSPC` filled the disk mid-rotate), the rescue swallows the error and reopens the original path. Operators watching for rotation health get no signal that rotation failed. At minimum log the rotation failure to stderr; better, surface `:rotation_failed` as a structured event.
- [ ] Stub scripts hardcode FHS fallback paths (bin/hive-babysitter-stub-git:26-27, bin/hive-babysitter-stub-gh:24-25): `real = "/usr/bin/git" if real.empty?` assumes Linux + FHS layout. On macOS Homebrew and many non-FHS distros, git/gh live elsewhere — the dry-run overlay then `exec`s a binary that does not exist. Prefer to exit non-zero with a clear "real binary not on PATH" message rather than silently `exec`ing a wrong absolute path.
- [ ] `Hive::Babysitter::StatusWriter.append` rewrites the entire `status.md` body on every tick (lib/hive/babysitter/status_writer.rb): reads previous content, concatenates one line, writes the full buffer atomically. With a babysitter running every 10 minutes for weeks, the file grows unbounded and each tick re-reads/-writes the whole accumulated history. Append in O(1) via `File.open(path, "a")` with a header initialized once, or document an explicit truncation/rotation policy.

## Nit

- [ ] `Hive::Lock.send(:process_start_time, pid)` is invoked from the extracted `Hive::PidFile` module via `.send` to reach a private method (lib/hive/pid_file.rb): refactoring opportunity — promoting `process_start_time` to public on `Hive::Lock` removes the brittle cross-module `send` that now lives in shared, ostensibly stable, code.
- [ ] `PrFixer#babysitter_task` writes an empty `task.md` stub under `.hive-state/babysitter/pr-<num>/` purely to satisfy `SyntheticTask.state_file` (lib/hive/babysitter/pr_fixer.rb:670-682); the file is never updated and serves no operator-visible purpose. Either remove the stub (and pass a `state_file:` that points at a sentinel/dev-null path the agent profile tolerates) or write something useful to it (PR number, timestamps, last outcome).
- [ ] `Hive::Babysitter::Interval.parse` and `Hive::Config.validate_babysitter!` both encode `\A\d+[smh]\z` (lib/hive/babysitter/interval.rb:430, lib/hive/config.rb:1366); validator should call the parser (or share a single `Interval::REGEX` constant) so the two regexes cannot drift.
- [ ] `hive babysit tail` lacks the `--lines N` / paging affordances the wiki claims parallels `hive daemon tail` (lib/hive/commands/babysit.rb:1285-1305 vs. wiki/commands/babysit.md). Minor UX divergence from the "muscle memory parity" promise in plan §R6.
- [ ] `Hive::Babysitter::Worktree.materialize` force-resets `hive-babysitter/pr-<num>` via `git worktree add -B` (lib/hive/babysitter/worktree.rb:963) but never cleans the local branch on exit — successive runs accumulate branch refs in the project repo (`git branch -a` will show `hive-babysitter/pr-1`, `…pr-2`, … indefinitely). Plan R7 mitigates worktree dir leaks but not branch ref leaks.
- [ ] `templates/babysitter_pr_fix_prompt.md.erb` instructs the agent to use `git fetch origin <head_ref>` immediately before push (matches plan R1), but does not tell the agent to `git rev-parse origin/<head_ref>` and compare against its last fetched SHA — the actual remote-SHA-changed check the plan promises is left to agent judgement.
- [ ] `Hive::Babysitter::Events.emit` raises `ArgumentError` on unknown action/outcome (lib/hive/babysitter/events.rb:285-286). In a long-running daemon, a programming bug at a single call site crashes the tick. Worth logging-and-coercing-to-"failure"/"give-up" instead of raising, so observability stays alive through bugs.
- [ ] `hive babysit --once --all` with zero enabled projects exits silently 0 (lib/hive/commands/babysit.rb:1154-1158 plus dispatcher behavior): operator sees no log/stdout indication that "no projects matched". A one-liner to stdout — "babysitter: 0 enabled projects, nothing to do" — would fix this for smoke testing.

ivankuznetsov:

## High
- [ ] `ContextBuilder` builds the diff against an unfetched PR head ref: `Worktree.materialize` checks out `hive-babysitter/pr-<n>` from `pull/<n>/head`, but `ContextBuilder` passes `headRefName` to `Gh.pr_diff_stat`, whose `git diff origin/<base>...<headRefName>` fails when that local branch is absent, so red PRs can fail before the agent starts (`lib/hive/babysitter/worktree.rb:21`, `lib/hive/babysitter/context_builder.rb:24`, `lib/hive/gh.rb:178`).
- [ ] Fork PR repair still targets the base `origin`: the fork-safe pull-ref fetch is followed by prompt and wrapper instructions to fetch/push `origin <headRefName>` / `origin HEAD:<headRefName>`, which cannot update a fork-owned head branch and violates the plan's any-author PR requirement (`templates/babysitter_pr_fix_prompt.md.erb:14`, `lib/hive/babysitter/gh_ops.rb:14`).
- [ ] Dry-run stubs can be bypassed with normal global options: `bin/hive-babysitter-stub-git` and `bin/hive-babysitter-stub-gh` only inspect `ARGV[0]`, so `git -C <wt> push ...` or `gh -R owner/repo pr comment ...` execs the real binary and can mutate origin/GitHub despite `dry_run: true` (`bin/hive-babysitter-stub-git:11`, `bin/hive-babysitter-stub-gh:11`).
- [ ] The babysitter prompt omits required base HEAD and conflict context: `ContextBuilder` returns failing jobs, diff stat, and branch names only, so the IU-5 requirement that the agent receive base HEAD and merge/conflict divergence context is not implemented (`lib/hive/babysitter/context_builder.rb:30`, `templates/babysitter_pr_fix_prompt.md.erb:1`).

## Medium
- [ ] `hive babysit start` does not atomically acquire its PID file: it checks `read_live_pid` and then writes with `File.write`, so two simultaneous starts can both pass the check and run dispatchers, violating the single-instance/TEMPFAIL contract (`lib/hive/commands/babysit.rb:61`).
- [ ] `PrFixer` bypasses the 4-execute Claude launch path: it always calls `Hive::Stages::Base.spawn_agent`, while 4-execute routes a Claude development agent through `spawn_claude_with_tmux_marker!`, so projects using `execute.agent: claude` do not get the same chooser/spawn semantics promised by the plan (`lib/hive/babysitter/pr_fixer.rb:94`, `lib/hive/stages/execute.rb:237`).
- [ ] `babysitter.interval` accepts zero at config-load time but raises at runtime: validation allows `0`/`0m`, then `Interval.parse` rejects it inside the dispatcher tick, turning a malformed config into a daemon-loop failure instead of a load-time `Hive::ConfigError` (`lib/hive/config.rb:1051`, `lib/hive/babysitter/interval.rb:8`, `lib/hive/babysitter/dispatcher.rb:29`).

## Nit
- [ ] `hive init` help text still describes seven prompts and omits the new babysitter question: the long description was not updated with the IU-7 opt-in prompt, so CLI help under-documents the generated config surface (`lib/hive/cli.rb:37`).

ivankuznetsov:

## High
- [ ] `Hive::Babysitter::Worktree.materialize` does not clean orphan `.git/worktrees/<pr>/` metadata (`lib/hive/babysitter/worktree.rb:41`): if a prior babysitter run crashed after `worktree add` but before/while writing the on-disk path, `File.directory?(path)` is false so `remove_existing!` skips the cleanup, and the next tick's `git worktree add -B hive-babysitter/pr-<n>` fails with `fatal: '<path>' already exists in the worktree list`. A stale metadata entry wedges every subsequent tick for that PR until an operator runs `git worktree prune`. Plan IU-5 / R7 explicitly says "deletes target directory on entry (recreate-fresh-each-tick) so leaks bound at max_concurrent_prs directories" — the metadata path is the missing half of that contract.
- [ ] `Babysitter::Logger#rotate_if_needed!` is a no-op when `log_max_files == 1` (`lib/hive/babysitter/logger.rb:75`): `(@max_files - 2).downto(0)` becomes `(-1).downto(0)` which yields nothing, so no rename runs, the file reopens at its current size, and `rotate_if_needed!` re-fires on every subsequent event. The babysitter inherits `log_max_files` from `Hive::Config.load_global_daemon` whose validator accepts `log_max_files >= 1` (`lib/hive/config.rb:1005`); an operator setting `1` to mean "keep only the current log" gets unbounded growth plus a close/reopen on every write instead.

## Medium
- [ ] `Dispatcher#interruptible_sleep` uses wall-clock `Time.now` and can crash on backwards NTP adjustment (`lib/hive/babysitter/dispatcher.rb:127`): between the `until @shutdown || Time.now >= deadline` check and `sleep [ deadline - Time.now, 0.5 ].min`, a clock step backward yields a negative argument, and `Kernel#sleep` raises `ArgumentError`, unwinding `run_forever`. The standard fix is `Process.clock_gettime(Process::CLOCK_MONOTONIC)` for the deadline; this matters in long-lived daemons that survive across NTP corrections.
- [ ] SIGHUP reload does not reload log-rotation knobs (`lib/hive/babysitter/dispatcher.rb:62`, `lib/hive/commands/babysit.rb:208`): plan IU-3 says reload should "clear per-tick enable cache + re-read the daemon-shared log rotation knobs". Per-tick config reload of `babysitter.enabled` works because `enabled_projects` reads fresh each tick, but the `Hive::Babysitter::Logger` is constructed once at start and never sees a refreshed `log_max_bytes` / `log_max_files`. Either rebuild the logger on `@reload`, or document the gap.
- [ ] `Babysit#stop_daemon` has a TOCTOU window around `File.delete(pid_file)` (`lib/hive/commands/babysit.rb:144`): the `unless File.exist?(pid_file)` guard runs first, then `read_pid_file_payload` re-checks existence and tolerates a missing file by returning nil. The subsequent unconditional `File.delete(pid_file)` branches under "malformed PID" / "stale PID" raise `Errno::ENOENT` if the file vanished mid-call (e.g., the daemon shut itself down and cleaned up between our two checks). Wrap with `rescue Errno::ENOENT` or use the existing `FileUtils.rm_f` idiom.
- [ ] `Babysitter::PrFixer#checks_green_or_queued?` short-circuits too eagerly on mixed-shape checks (`lib/hive/babysitter/pr_fixer.rb:80`): the OR-of-three logic considers a check green if `conclusion in [SUCCESS, NEUTRAL, SKIPPED]` OR `state == SUCCESS` OR `status in [QUEUED, PENDING, IN_PROGRESS]`. A failed check whose retry is queued (`conclusion: "FAILURE"`, `status: "QUEUED"`) reads as green and skips the agent. The intent is "all checks resolved successfully or still progressing"; safer to fail-out a check whose `conclusion` is a hard failure regardless of `state`/`status`.
- [ ] `give_up` event's outcome does not reflect dry-run mode (`lib/hive/babysitter/pr_fixer.rb:213`): when `@dry_run = true` and the agent exits zero without writing the plan file, `outcome_for` returns `:failure` and `give_up` runs. The nested `label-apply` and `pr-comment` events correctly emit `outcome: "dry_run"`, but the parent `give-up` event emits `outcome: "failure"`. Consumers grepping `events.jsonl` for "did a real-world give-up happen?" will see a false positive on dry-run runs that exit non-zero. Either include `dry_run` in the outcome enum branching for the give-up emission or have dry-run `:failure` paths skip the give-up emission entirely.
- [ ] `failing_jobs_with_logs` per-job cap is uniform, not proportional (`lib/hive/gh.rb:160`): the plan IU-8 test scenario describes "two short (5 KB) and one large (200 KB) → context contains all three with the large one truncated". The current `per_job_cap = byte_cap / jobs.size` gives every job an equal slice (≈16 KB each with `byte_cap=50 KB`, 3 jobs), so the short jobs underuse the budget and the agent sees less context from the large failing job than the budget allows. Use a two-pass strategy: small-fits keep full text, then redistribute remaining bytes across the over-cap jobs.
- [ ] Branch is stale vs `origin/main` and reverses two unrelated fixes (#180 demo.webp; #183 preserve project-owned `.claude/settings.json`): `git diff origin/main..HEAD -- lib/hive/claude_launcher.rb lib/hive/stop_hook_installer.rb` shows the backup/restore logic from PR #183 as removed. The branch hasn't touched these files; a normal 3-way merge into main would keep main's version, but a squash-merge or anything that flattens the diff against `origin/main` literally would re-introduce the original dirty-worktree regression. Rebase on `origin/main` before merge.
- [ ] `wiki/state-model.md` changes are in the working tree but uncommitted (`git status` shows `M wiki/state-model.md`): the diff adds the babysitter-state section the branch otherwise references via `[[modules/babysitter]]`. Without committing, the wiki page indices on main do not gain the babysitter row when this branch lands. Stage and include in the next commit.

## Nit
- [ ] `GhOps#post_pr_comment` computes `stamped` even in dry-run / dedup paths where the value is discarded (`lib/hive/babysitter/gh_ops.rb:35`): `stamp_comment(body)` runs before the dry-run early-return and the recent-give-up dedup early-return, wasting a tiny amount of work and one `Time.now` per dry-run call. Move the assignment below the early-returns.
- [ ] `Hive::Babysitter::Worktree#head_ref` is dead code (`lib/hive/babysitter/worktree.rb:37`): defined but never referenced; the rest of the class uses `@pr.fetch("headRefName")` inline. Delete or use.
- [ ] `ProjectTick.run`'s second parameter `_cfg` is misleading (`lib/hive/babysitter/project_tick.rb:13`): the dispatcher passes a cached `entry[:cfg]`, but ProjectTick deliberately re-reads `Hive::Config.load(project_entry.fetch("path"))` for freshness. The `_cfg` underscore-prefix signals the parameter is unused, yet the dispatcher still has to load and pass it. Either accept and use the dispatcher's cached cfg (one fewer disk read per tick), or drop the parameter from the signature so the dispatcher stops paying for an ignored load.
- [ ] `Babysitter::PrFixer` calls `pr_status_rollup` against `@project.fetch("path")` for the early-green check and ContextBuilder later runs against `worktree.path` (`lib/hive/babysitter/pr_fixer.rb:36`): both paths resolve to the same gh repo via cwd's git config, so the result is correct, but the mixed-cwd convention is easy to misread. Either run both against the worktree (after materialize) or both against the project root, and add one line of comment explaining the choice.
- [ ] `Babysitter::Logger` rotation reopen-after-failure path warns then continues without any rotation (`lib/hive/babysitter/logger.rb:85`): a transient EACCES on rename leaves the file at over-cap size, and the next event triggers another rotation attempt + warn. With a persistent permissions issue, every event emits two warnings to stderr — noisy. Consider gating the warning behind a "warned-once" flag.
- [ ] `bin/hive-babysitter-stub-{git,gh}` reproduce the same `log_skip` + `mutating` argv shape (`bin/hive-babysitter-stub-git`, `bin/hive-babysitter-stub-gh`): minor duplication; a single shared library file would centralise the mutating-argv detection if more wrappers (`hg`, `jj`, …) are added later. Acceptable for v1.

PR #181, opened 2026-05-26
feat(tui): render footer hints before token counters

ivankuznetsov:
## Summary

Wide TUI footers previously rendered token counters first, pushing the action hints toward the right edge so they clipped before the counters under right-side truncation. The wide-terminal branch of `default_footer` (`lib/hive/tui/bubble_model.rb`) now renders the keybinding hints first, then the ` · ` separator, then the compact token block — so when terminal width gets tight, the token counters are the first content to drop while higher-priority action hints survive.

Scope is intentionally a single string-interpolation swap. The token block stays an atomic unit (internal `•` separators untouched) and the inter-block ` · ` separator is preserved. The narrow (`< 80` columns) branch, which already renders only the usage block, is left unchanged on purpose — flipping its priority to hints-only is a separate decision held out of scope.

Changes:
- `lib/hive/tui/bubble_model.rb`: swap interpolation order in the wide-terminal branch of `default_footer`.
- `test/unit/tui/bubble_model_test.rb`: add `test_default_footer_truncates_tokens_first_on_overflow_above_threshold` to lock the new clip-from-right behavior; assert hint-precedes-token ordering and ` · ` separator presence at wide width; rename the narrow-branch test to describe what it actually exercises.
- `wiki/token-usage.md`, `wiki/commands/tui.md`, `wiki/log.md`: refresh the dashboard footer description/diagram and log the change.

## Test plan

- [x] `bundle exec rake test TEST=test/unit/tui/bubble_model_test.rb`
- [x] `bundle exec rake test TEST=test/integration/tui_usage_footer_test.rb`
- [x] `bundle exec rake test` — `3601 runs, 12938 assertions, 0 failures, 0 errors, 2 skips`
- [x] `git diff --check` clean
- [ ] Manual: launch `bin/hive tui` against a project with recorded usage and confirm the dashboard status line reads `[Tab] switch … [q] quit · today … • 7d … • all … • tokens`

## Review summary

Two Compound Engineering review passes (Claude + Codex, 2 passes each). No High or Medium findings in any pass. All Nit-level findings were auto-fixed or resolved:

- Added an explicit ordering assertion (`[Tab] switch` precedes `tokens`) at wide width.
- Added an explicit ` · ` separator assertion so a dropped/renamed joiner fails loudly.
- Added a `usage.end_with?("tokens")` premise check to guard the overflow test against future label changes.
- Aligned the overflow test's clip margin with the plan's `full_width - 5`.
- Renamed the misleading `test_default_footer_truncates_from_right_at_narrow_width` to `test_default_footer_narrow_branch_drops_hints_and_keeps_usage` to match what it actually exercises.
- Fixed the `wiki/commands/tui.md` footer ASCII diagram alignment.

Note: the `pr-review-toolkit` reviewer failed to start (tmux prompt-readiness infra error) in both passes; the CE reviewers ran successfully and produced the findings above.

## Linked task

- Original request: "move token counters after the actions"
- Hive task slug: `move-token-counters-after-the-260526-d32a`

---

### DIFF
--- a/lib/hive/tui/bubble_model.rb
+++ b/lib/hive/tui/bubble_model.rb
@@ -3201,7 +3201,7 @@ module Hive
             return Views::UsageFooter.render(aggregate: aggregate, width: usable_width)
           end
 
-          line = "#{Views::UsageFooter.text(aggregate)} · #{footer_hint}"
+          line = "#{footer_hint} · #{Views::UsageFooter.text(aggregate)}"
           line = Views::Format.truncate(line, usable_width) if usable_width
           Hive::Tui::Styles::HINT.render(line)
         end
--- a/test/unit/tui/bubble_model_test.rb
+++ b/test/unit/tui/bubble_model_test.rb
@@ -731,13 +731,25 @@ class HiveTuiBubbleModelTest < Minitest::Test
   def test_default_footer_fits_full_hint_at_exact_boundary_width
     hint = @model.send(:footer_hint)
     usage = Hive::Tui::Views::UsageFooter.text(Hive::UsageDb.zero_aggregate)
-    out = with_zero_usage { @model.send(:default_footer, "#{usage} · #{hint}".length) }
+    out = with_zero_usage { @model.send(:default_footer, "#{hint} · #{usage}".length) }
 
     assert_includes out, "tokens"
     assert_includes out, "[i] info"
     assert_includes out, "[q] quit"
     refute out.end_with?("…"),
            "footer at exact hint width should not truncate, got #{out.inspect}"
+  end
+
+  def test_default_footer_truncates_tokens_first_on_overflow_above_threshold
+    hint = @model.send(:footer_hint)
+    usage = Hive::Tui::Views::UsageFooter.text(Hive::UsageDb.zero_aggregate)
+    full_width = "#{hint} · #{usage}".length
+    out = with_zero_usage { @model.send(:default_footer, full_width - 5) }
+
+    assert_includes out, "[Tab] switch",
+                    "hint block must survive when overflow clips from the right"
+    refute_includes out, "tokens",
+                    "token block must be the first content clipped on overflow"
   end
 
   def test_default_footer_truncates_from_right_at_narrow_width
--- a/test/unit/tui/bubble_model_test.rb
+++ b/test/unit/tui/bubble_model_test.rb
@@ -726,6 +726,8 @@ class HiveTuiBubbleModelTest < Minitest::Test
     assert_includes out, "tokens"
     assert_includes out, "[i] info"
     assert_includes out, "[q] quit"
+    assert_operator out.index("[Tab] switch"), :<, out.index("tokens"),
+                    "hint block must precede token block at wide width"
   end
 
   def test_default_footer_fits_full_hint_at_exact_boundary_width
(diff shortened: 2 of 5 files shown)


ivankuznetsov:

## High

## Medium

## Nit
- [ ] test/unit/tui/bubble_model_test.rb:734 does not assert the exact wide-footer `hint · usage` join: the plan requires preserving the ` · ` separator, but the updated tests would still pass if the implementation omitted or replaced the joiner while keeping both blocks visible.

ivankuznetsov:

## High

(none)

## Medium

(none)

## Nit

- [ ] `wiki/commands/tui.md:31` ASCII footer drawing trailing whitespace may misalign with the box's right border (cosmetic): the previous diagram had no trailing spaces before `│`; the new one pads with 8 spaces and the visible column width is not guaranteed to match the `└─...─┘` row. Docs-only nit.
- [ ] `test/unit/tui/bubble_model_test.rb:751` new test couples to the literal substring `"tokens"` in `Views::UsageFooter.text`; if the trailing label is ever renamed (e.g. to `"tok"`) the assertion will silently false-positive (clipped content would still not contain the new label). Optional: anchor on `usage`'s tail (e.g. `usage[-6..]` / `usage.end_with?("tokens")` sanity check) so the test fails loudly if the suffix changes.
- [ ] `test/unit/tui/bubble_model_test.rb:747` `full_width - 5` is one char larger than the minimum overflow needed to guarantee `"tokens"` (6 chars) gets clipped; works today because `Format.truncate` consumes one extra char for the `…` suffix, making the effective drop 6. A future change to `Format.truncate`'s ellipsis policy (or to a 7-char trailing label) could turn this into a brittle off-by-one. Optional: use `usage.length` (or a larger margin) so the test's intent — "tokens label is the first thing dropped" — is unambiguous.

ivankuznetsov:

## High

_(no findings)_

## Medium

_(no findings)_

## Nit

- [ ] `test_default_footer_truncates_tokens_first_on_overflow_above_threshold` clips by `full_width - usage.length` (test/unit/tui/bubble_model_test.rb:751) rather than the plan-specified `full_width - 5`: behavior is still correct (`usable_width` lands at 82, ≥ 80, wide branch fires, `tokens` is cleanly clipped), but the magic-relationship `full_width - usage.length` couples the chosen width to the usage block size in a way the plan did not call for — a literal small constant would make the intent ("clip a little past the end") clearer and the math obvious.
- [ ] `test_default_footer_renders_full_hint_at_wide_width` (test/unit/tui/bubble_model_test.rb:723-729) still only asserts substring presence (`tokens`, `[i] info`, `[q] quit`) and does not pin the new ordering — a future regression that flips the concat back to tokens-first would still pass this test; ordering is only enforced indirectly via the new overflow test. Optional tightening: assert `out.index("[Tab] switch") < out.index("tokens")` at the wide-width path.
- [ ] `refute_includes out, "tokens"` in the new overflow test (test/unit/tui/bubble_model_test.rb:755) is robust today because `Views::UsageFooter.text` only emits the literal `tokens` at the tail, but is brittle to future label changes (e.g., an added bucket whose label contains `tokens`); the `assert usage.end_with?("tokens")` premise check guards against the tail moving, not against new internal occurrences — minor risk only if `UsageFooter.text` ever grows new copy.
- [ ] Pre-existing test name `test_default_footer_truncates_from_right_at_narrow_width` (test/unit/tui/bubble_model_test.rb:759) is misleading: at width 76 the narrow branch fires and returns the usage block unmodified (43 chars ≤ 76, no truncation actually occurs); the test is functionally exercising "narrow branch drops hints, keeps usage." Not introduced by this diff — flagged because the plan called it out as encoding the *old* tokens-survive-hints-clip assumption, and the rename would have removed that residual confusion.

PR #229, opened 2026-05-27
fix(daemon): use waiting-marker write-time as first-sight edit baseline

ivankuznetsov:
## Summary
The auto-advancing daemon stranded `needs_input` (brainstorm/execute/review-waiting) tasks the user had **already answered**. `Policy#decide_edit` recorded the current state-file mtime as the baseline on first sight and only dispatched on a *strictly-newer* edit — so an answer that predated the daemon's first sight was never "newer" and the task was skipped on every tick forever. This happens routinely:
- after a **daemon restart** (in-memory baselines are lost), and
- when the **bot** dispatched the brainstorm, so the daemon never recorded a baseline for it.

This was hit live: answered brainstorms sat in "Needs your input" and never advanced until the state file was manually `touch`ed.

## Fix
Stamp a UTC `ts` write-time on waiting-class markers (`WAITING` / `EXECUTE_WAITING` / `REVIEW_WAITING`) in `Markers.set` — mirroring the existing `ERROR` `marker_id` stamping — and surface it to the daemon as `Row#marker_ts`. On first sight of an edit-resume row, `Policy` now uses the marker's write-time as the **baseline floor**:

| First-sight case | Decision |
|---|---|
| `marker_ts` present, `mtime <= marker_ts` (agent just asked, no answer yet) | `:record_baseline` |
| `marker_ts` present, `mtime > marker_ts`, debounce not met | `:wait_for_debounce` |
| `marker_ts` present, `mtime > marker_ts`, debounce met | `:dispatch` |
| `marker_ts` nil (legacy / non-waiting) | `:record_baseline` (unchanged) |

The marker write-time fixes the "ask time", so any later answer is reliably detected as user input even across a daemon restart or a bot-dispatched brainstorm.

## Notes
- `ts` is an **internal** attr (added to `INTERNAL_ATTR_KEYS`), hidden from operator-facing marker display.
- **Fully back-compat**: with no `marker_ts`, the original first-sight `:record_baseline` behavior is byte-identical.
- `read_marker_ts` degrades to nil on missing file / no `ts` / unparseable timestamp.

## Test plan
- [x] `test/unit/markers_test.rb` — waiting markers get a round-tripping `ts`; caller-supplied `ts` preserved; non-waiting markers get none; `ts` excluded from `display_attrs`.
- [x] `test/unit/daemon/policy_test.rb` — all four first-sight cases above; existing cases unchanged.
- [x] `test/unit/daemon/{status_consumer,dispatcher}_test.rb` — `marker_ts` plumbing.
- [x] `rake coverage` 100% (4020 runs, 0 failures), `rubocop` clean.

### DIFF
--- a/lib/hive/daemon/dispatcher.rb
+++ b/lib/hive/daemon/dispatcher.rb
@@ -442,7 +442,8 @@ module Hive
           last_dispatched_state_file_mtime:
             @controller.last_dispatched_state_file_mtime_for(project: row.project, slug: row.slug),
           now: now,
-          edit_debounce_sec: @edit_debounce_sec
+          edit_debounce_sec: @edit_debounce_sec,
+          marker_ts: row.marker_ts
         )
 
         case decision
--- a/lib/hive/daemon/policy.rb
+++ b/lib/hive/daemon/policy.rb
@@ -30,12 +30,18 @@ module Hive
     #   an early dispatch).
     #
     # First-sight policy on `kind: edit`: the daemon sees a `_WAITING`
-    # task with no prior observation. We CANNOT distinguish "agent just
-    # wrote the WAITING marker" from "user already answered hours ago",
-    # so the safe choice is to skip + record the current mtime as the
-    # baseline. The next genuine user edit (mtime > baseline) triggers
-    # dispatch. Operators of dormant-task scenarios can `touch` the
-    # state file to wake the daemon.
+    # task with no prior observation. When the row carries a `marker_ts`
+    # (the agent's WAITING-marker write-time), we use it as the baseline
+    # floor — a state-file write strictly newer than the marker is
+    # genuine user input, so we can dispatch an answer that predates the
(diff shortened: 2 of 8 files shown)


ivankuznetsov:
Closing in favor of a persistence-based approach. Review found the marker-`ts` design is the wrong foundation: (1) it's a no-op for the brainstorm/plan target — those WAITING markers are written by the agent as literal text, not via `Markers.set`, so they never get a `ts`; (2) it false-dispatches `execute_waiting`/`review_waiting`/`finalize_waiting` because a whole-second `iso8601` stamp is always older than the same write's sub-second filesystem mtime; (3) the new per-row `File.read` in `extract_rows` only rescues `ArgumentError`, so one unreadable state file stalls the whole dispatch tick. Replacing with a marker-format-agnostic approach that persists the daemon's observed mtimes across restarts (mtime-to-mtime comparison, no wall-clock skew, no per-row I/O blast radius). New PR to follow after a plan.

PR #240, opened 2026-05-28
feat(openclaw): ship skill bundle and install docs

ivankuznetsov:
## Summary

OpenClaw.ai users can now install Hive from ClawHub and call `/hive`, `/plan`, `/work`, `/ce-review`, and friends without learning the CLI first. The PR adds a skill bundle under `openclaw/skills/` and a "Works with OpenClaw" section in the main README that points at it.

The skills are thin shells around the existing `hive` binary. Each runs a preflight `command -v hive` check, refuses to interpolate user text into a shell string, and prefers `--json` output when summarizing. No CLI behavior changes.

### What ships

22 skills total: one `/hive` umbrella that can dispatch any subcommand, plus per-command shortcuts (`/plan`, `/work`, `/ce-review`, `/findings`, `/approve`, `/finalize`, `/archive`, `/doctor`, `/daemon`, `/bot`, `/init`, and more). `/hive-status` is prefixed to avoid colliding with OpenClaw's built-in `/status`. Destructive admin verbs (`drop`, `uninstall`, `prune`, `migrate`) stay behind `/hive ...` so the agent always sees the full command before execution. The full slug map lives in [openclaw/README.md](openclaw/README.md).

### Why Skills, not a Plugin

OpenClaw exposes two extension surfaces:

- **Skills** — `SKILL.md` with YAML frontmatter, one skill per slash command, no runtime code.
- **Plugins** — TS/ESM source with `openclaw.plugin.json`, used for tools/providers/channels.

Hive's surface is prompt-shaped wrappers around the existing CLI, so Skills fit and a Plugin would add a Node runtime for no gain. The path comparison and rationale lives in [openclaw/README.md](openclaw/README.md).

### Out of scope

- ClawHub publish itself. The publish loop is documented in `openclaw/README.md` and is intentionally manual, gated on namespace confirmation.
- The exploration write-up in `~/Dev/hive-private/docs/brainstorms/openclaw-install.md`. That doc lives outside the public repo per the hive-private convention and ships in a separate stage.
- Live marketplace listing evidence. Captured in the same private write-up once publish runs.

## Test plan

`test/unit/openclaw_skills_test.rb` validates the bundle as data:

- Every `SKILL.md` parses with the required frontmatter (`name`, `description`, `version`, `user-invocable: true`).
- Every skill maps to an actual `hive` Thor subcommand — the test iterates `Hive::CLI.all_commands.keys` so adding a new verb without a skill (or vice versa) fails fast.
- Every body includes the `command -v hive` preflight and the "do not interpolate raw user text into a shell string" guard.
- Destructive verbs (`/hive drop|uninstall|prune|migrate`, `/markers clear`, `/daemon stop`, `/bot install --force`) require explicit restate-and-confirm guidance in the skill body.

Full suite:

- `bundle exec ruby -Itest -Ilib test/unit/openclaw_skills_test.rb`
- `bundle exec rake test` → 4087 runs, 15084 assertions, 0 failures, 0 errors, 2 skips.
- `git diff --check HEAD~9..HEAD` clean.

## Review summary

Two passes of Compound Engineering review (Claude + Codex, two rounds each). Findings auto-fixed in-tree:

- Added `hive` Thor-command drift test so the skill bundle can't fall behind the CLI silently.
- Documented `hive tui` and `hive version` as deliberate exclusions (TUI rejects `--json` with `EX_USAGE`; `version` is `--version` only).
- Added restate-and-confirm guidance to `/hive`, `/markers`, `/daemon`, `/bot` for destructive verbs.
- Warned `/init` skill body about no-TTY skill contexts (avoids hangs on TTY prompts).
- Reverted `metadata` frontmatter to inline JSON to match OpenClaw's documented single-line-JSON schema; added `requires.bins: ["hive"]` so ClawHub can gate runtime deps.
- Dropped the unsourced `openclaw skills install git:…` syntax from the comparison.
- DRY'd the publish/install loop in `openclaw/README.md` and split the wiki gap entry into in-tree vs pending follow-ups.
- README now states the bundle ships 22 slash commands and points readers at `openclaw/README.md` for the full list.

One High finding was escalated and deferred to the publish stage, not this PR: the bare `hive` ClawHub slug currently resolves to an unrelated `@gora050/hive` listing. The plan's Unit 3 already gates publish on explicit user confirmation under the `ivankuznetsov` namespace, so no live install is broken by this PR — the actual slug strategy (`@ivankuznetsov/hive`, `ivankuznetsov-hive`, etc.) is decided at publish time. The wiki and PUBLISHED log will be updated then.

Two acceptance items are intentionally out of this PR:

- The exploration write-up at `~/Dev/hive-private/docs/brainstorms/openclaw-install.md` (lives in a separate private repo per the hive-private convention).
- The `clawhub skill publish --dry-run` validation (the `openclaw`/`clawhub` CLIs aren't installed in this stage's environment; the publish step in `openclaw/README.md` runs `--dry-run` before any real publish).

## Linked task

- Hive task: `explore-the-simplest-way-to-260528-2503`
- Origin idea: "explore the simplest way to install hive with OpenClawd, so OpenClawd users can get hive the easiest way, I don't know like plugins or what they have there"
- Plan + brainstorm artefacts: `~/Dev/hive/.hive-state/stages/8-finalize/explore-the-simplest-way-to-260528-2503/{plan.md,brainstorm.md,reviews/}`

---


ivankuznetsov:

## High
- [ ] Advertised ClawHub slug `hive` is already owned by an unrelated listing: the branch documents and publishes `openclaw skills install hive` / `--slug hive`, but live `https://clawhub.ai/skills/hive` redirects to `@gora050/hive`, so users would install a different Hive integration.
- [ ] Required exploration and acceptance evidence is absent: `/home/asterio/Dev/hive-private/docs/brainstorms/openclaw-install.md` does not exist, so Units 1 and 5 are not delivered.
- [ ] The advertised one-command install does not register the required shortcuts: `openclaw skills install hive` installs only `/hive`, while `/plan`, `/work`, `/ce-review`, and the other skill folders remain separate optional installs.

## Medium
- [ ] Skill frontmatter does not declare the required `hive` binary: each `SKILL.md` shells out to `hive` but `metadata.openclaw.requires.bins` is missing, so OpenClaw/ClawHub cannot gate or validate the runtime dependency.

## Nit

ivankuznetsov:

## High

## Medium
- [ ] No drift check between skill bundle and live Thor command set: `test/unit/openclaw_skills_test.rb` pins `EXPECTED_SKILLS` as a hardcoded hash and only checks that skill directories equal those keys. It never iterates `Hive::CLI.all_commands.keys` to assert that every non-admin Thor command has a corresponding skill. Plan Risk 4 explicitly called this out ("build the skill list from reading hive's actual command source, not from memory"). When a new workflow verb is added to `lib/hive/cli.rb`, the test stays green and the bundle silently ships incomplete.
- [ ] `hive tui` is omitted from the skill bundle and from `openclaw/README.md`'s "Planned ClawHub Slugs" without explanation: TUI is documented in `README.md` as the primary user surface ("Five-Minute TUI Getting Started"), and the closing paragraph in `openclaw/README.md` only excuses destructive admin commands as falling back to `/hive ...`. TUI is neither in the table nor explicitly listed as deliberately excluded — readers can't tell if this is an oversight or a design decision.
- [ ] No evidence `clawhub skill publish --dry-run` was actually run against any skill: Plan Unit 2 verification step requires it, and the schema we ship (`name`, `description`, `version`, `user-invocable`, `metadata.openclaw.homepage`) hasn't been validated against the live ClawHub validator. If the real schema disagrees with these fields, every publish will fail and the bundle has to be revved before listing is possible.
- [ ] Umbrella `/hive` skill is the documented route for destructive admin commands (`drop`, `uninstall`, `update`, `forget`, `prune`, `migrate`, `metrics`) but `openclaw/skills/hive/SKILL.md` body has no confirmation gating for them: `daemon` and `bot` skills explicitly require "restate the effect and get explicit user confirmation" before `install --force` / `stop`, yet `/hive drop TARGET` (which kills the agent, deletes the worktree, and closes the draft PR) can be invoked through the umbrella with no such instruction.

## Nit
- [ ] YAML frontmatter mixes flow JSON with block YAML — every `SKILL.md` uses `metadata: {"openclaw":{"homepage":"..."}}` as inline flow style alongside otherwise block-style keys. Block form (`metadata:\n  openclaw:\n    homepage: ...`) would be more idiomatic and easier to extend.
- [ ] `markers clear` is destructive (can break recovery if a stale-match guard is omitted) but `openclaw/skills/markers/SKILL.md` doesn't request confirmation comparable to `daemon`/`bot`. The skill does mention "Include any `--match-attr` guard Hive suggests" but doesn't pause for user approval first.
- [ ] `openclaw/skills/init/SKILL.md` description mentions "TTY-prompts" but the body doesn't warn the agent that OpenClaw skill contexts may not provide a real TTY, so `hive init` can hang waiting for prompt input.
- [ ] `test/unit/openclaw_skills_test.rb:731` frontmatter regex `\A---\n(?<frontmatter>.*?)\n---\n(?<body>.*)\z/m` only matches LF line endings — would silently fail on CRLF-checked-out files; harmless on Linux but a tiny portability nit.
- [ ] `openclaw/README.md` repeats the per-skill install loop inline for both "Local Install For Testing" and "Publish Checklist"; consolidating to a single canonical shell loop with a `--dry-run` toggle would be DRYer.
- [ ] `hive version` is not enumerated as a skill, as a planned slug, or as an explicit admin-fallback in the `/hive ...` exclusion list. Trivial since `--version` already works, but the omission is inconsistent with how every other Thor command is accounted for.

ivankuznetsov:

## High
- [ ] README.md:76 and openclaw/README.md:74 still advertise `openclaw skills install hive`, but the live ClawHub `/skills/hive` slug resolves to an unrelated `@gora050/hive` listing, so the documented marketplace install would install the wrong integration.

## Medium
- [ ] openclaw/skills/hive/SKILL.md:6 and the other skill frontmatters use block YAML for `metadata`, while OpenClaw's current skills reference says embedded-agent metadata must be a single-line JSON object; `requires.bins`/homepage gating may be ignored despite the unit test passing via Ruby YAML.

## Nit

ivankuznetsov:

## High
- [ ] Bare `hive` slug collision is still unresolved across README.md, openclaw/README.md, and the Publish Checklist: pass-01 escalation Q1 (escalations-01.md) flagged that `https://clawhub.ai/skills/hive` already resolves to an unrelated `@gora050/hive` listing, the human gate was left empty, and the diff after the fix iteration still advertises `openclaw skills install hive` (README.md:76, openclaw/README.md:11, openclaw/README.md:25, openclaw/README.md:74) and publishes `--slug "hive"` (openclaw/README.md:117). Plan Risk 3 explicitly reserved the `ivankuznetsov` namespace, but the docs never adopted a namespaced or disambiguated slug — anyone running the documented marketplace install would install the wrong integration, and the documented publish would either fail or shadow the existing owner.

## Medium
- [ ] No clawhub-side validation was ever performed on the `metadata` block, so YAML-vs-JSON schema fit is still unknown: pass-01 triaged the dry-run gap as "out of environment", and the fix iteration only re-checked the bundle with Ruby's `YAML.safe_load` in `test/unit/openclaw_skills_test.rb`. If OpenClaw's loader treats embedded-agent metadata as flow-style JSON (codex pass-02 medium), the new block form (`metadata:\n  openclaw:\n    homepage: …\n    requires:\n      bins:\n        - hive`) may silently drop `requires.bins` and `homepage`, defeating the very gating that pass-01 added them to enable. The fix swap from inline JSON to block YAML is therefore the kind of change that needed `clawhub skill publish --dry-run` before merging, not after.
- [ ] `openclaw/README.md:27` cites `openclaw skills install git:ivankuznetsov/hive@<ref>` for the direct-GitHub install path, but the plan's Unit 1 (c) research only ever cited `openclaw plugins install git:github.com/ivankuznetsov/hive@<ref>` (plugins, not skills; with a `github.com/` prefix). The diff introduces a `skills install git:…` form that is not sourced from any cited OpenClaw doc — if the syntax is wrong, the documented alternative install path misleads readers, and it directly contradicts the only sourced example in the plan.

## Nit
- [ ] `wiki/gaps.md` entry 3 now packs two unrelated follow-ups into one paragraph — the still-pending Claude/Codex/Pi marketplace publish on `ivankuznetsov/hive-skills`, and the now-in-tree OpenClaw skill bundle with its own external publish step. Splitting into two bullets (one resolved-in-tree, one still pending) would keep the gap log honest about which agent surfaces are actually unblocked.
- [ ] `README.md:76` "Works with OpenClaw.ai" example list (`hive-plan`, `hive-work`, `hive-ce-review`) is hedged with "for example" but never tells readers there are 22 shortcut slugs total, only that "the full skill table" lives in `openclaw/README.md`. A one-sentence "the bundle ships 22 slash commands; see openclaw/README.md for the full list" would make the surface size visible to readers who don't click through.

PR #268, opened 2026-06-02
fix(daemon): hold brainstorm resume until all Q&A answers are in

### LINKED ISSUE #267: daemon: auto-resume fires mid-Q&A — brainstorm dispatched before all Telegram answers are in
## Symptom
The daemon starts processing a brainstorm before the operator has finished answering all questions in Telegram — and a subsequent answer can bounce with "Try again — another run holds the lock".

## Root cause
Two independent triggers can resume a brainstorm:

1. **Bot (correct):** `execute_answer_write` writes each Telegram answer into `brainstorm.md` one at a time, and only `finalize_completed_brainstorm → auto_run_after_answers` enqueues a dispatch request — *after all* questions are answered.
2. **Daemon auto-resume (the bug):** `Hive::Daemon::Policy.decide` treats a `needs_input` brainstorm row as an **edit-resume** — it dispatches `hive brainstorm` whenever the state-file mtime moved AND `now - mtime ≥ edit_debounce_sec` (default 30s). See `lib/hive/daemon/policy.rb` (`EDIT_RESUME_ACTIONS`, `decide_edit`).

Every Telegram answer bumps `brainstorm.md`'s mtime, so ~30s after the **first** answer the daemon resumes the run with only some questions answered. The debounce comment assumes a single editor save; a multi-question Telegram session is N writes over minutes, and debounce can't distinguish "answered 1 of 3, still going" from "done". Nothing in the `hive status` row carries an unanswered-question count (`BrainstormParser.unanswered_questions` lives in `Hive::Bot::`, invisible to the daemon). The premature run also grabs the task `.lock`, which is why later answers can hit `:lock_busy`.

## Fix
The daemon must not resume a brainstorm `needs_input` row while unanswered questions remain (independent of whether answers arrive via Telegram or a direct editor edit):

- Relocate the pure parser to `Hive::BrainstormParser` (shared by bot + daemon; keep `Hive::Bot::BrainstormParser` as an alias).
- `Policy.decide` gains `answers_pending:`; a would-be `:dispatch` on an edit-resume row is held as a new `:wait_for_answers` outcome when answers are pending — preserving first-sight baseline recording so the editor-bulk-save path still resumes when complete.
- The dispatcher computes `answers_pending` for brainstorm `needs_input` rows by parsing the state file, passes it to Policy, and logs the hold.

Contract-safe: no change to the published `hive-status` schema (the daemon parses the file directly).

Found while reviewing the dispatch-request queue work (PR #244).

ivankuznetsov:
Closes #267.

## The bug
The daemon could start processing a brainstorm **before the operator finished answering** in Telegram — and a later answer could bounce with "Try again — another run holds the lock".

Two independent triggers resume a brainstorm: the bot (correct — only enqueues a dispatch request once *all* questions are answered) and the daemon's own `needs_input` **edit-resume** (`Policy.decide` → `decide_edit`), which fires when the state-file mtime moved and `edit_debounce_sec` elapsed. Every Telegram answer is a *separate* write that bumps `brainstorm.md`'s mtime, so ~30s after the **first** answer the daemon re-ran `hive brainstorm` with a partially-answered file (and grabbed the task `.lock`). The debounce assumed a single editor save; it can't tell "answered 1 of 3, still going" from "done".

## The fix
Gate the resume on whether any questions are still unanswered:

- **Relocate the parser** to `Hive::BrainstormParser` (out of `Hive::Bot::`, which becomes a back-compat alias) so the daemon can share it.
- **`Policy.decide` gains `answers_pending:`** and a new `:wait_for_answers` outcome. It downgrades **only** the terminal `:dispatch` — `:record_baseline` / `:skip` / `:wait_for_debounce` pass through unchanged, so the mtime baseline is still seeded and the **editor-bulk-save** path (all answers in one save) still resumes when complete.
- **`Dispatcher#brainstorm_answers_pending?`** parses a `2-brainstorm` `needs_input` row's file (fails **open** on a parse error so a malformed file never strands a task) and logs the hold as `:skipped reason=answers_pending`.

Surface-agnostic: holds whether answers arrive one-at-a-time via the bot or all at once via a direct edit. The published `hive-status` schema is untouched (the daemon parses the file directly), so no contract change.

## Verification
- Full suite **4166 runs / 0 failures**, **100% line coverage**, rubocop clean.
- New tests: Policy gate (hold when pending; dispatch when complete; baseline/skip/debounce unaffected), dispatcher end-to-end (hold vs resume), `brainstorm_answers_pending?` branches + fail-open, parser alias identity.

Branched off clean `origin/main` (independent of the `fix-llm-wiki-hook-env` work).


ivankuznetsov:
## Addressed /ce-code-review findings (commit 2663f58c)

**Fixed in this PR:**
- **#1 parse-error hardening** — `BrainstormParser.parse` is now total (scrub invalid UTF-8 + rescue `ENOENT`/`EACCES`/`IOError`), so a torn concurrent read or garbled bytes degrade instead of raising; the residual `:fatal` is deduped per `[project, slug]` so a persistently unreadable file can't spam every tick.
- **#5 end-to-end test** — proves the daemon actually *resumes* (spawns) on a parse error rather than stranding; added dedup / multi-round / no-A-slot / zero-question tests.
- **doc** — `wiki/modules/bot.md` now describes `Hive::Bot::BrainstormParser` as a back-compat alias.

**Reconsidered (not changed):**
- **#2 "fail closed on zero questions"** — *not* adopted. The bot locates questions with the **same** parser, so a file with no parseable `### Q{n}.` (empty / header drift) can't be answered via the bot anyway; resuming (re-running the agent to regenerate the file) is the self-healing path, and failing closed would **strand** the task. Documented in code + wiki.

**Deferred to follow-up issues:**
- #269 — a `### Q` with no fillable `### A` slot holds indefinitely (bot answer-writer concern, pre-existing).
- #270 — surface an `unanswered_questions` count in `hive status` so a gate-hold is distinguishable from a broken task (published-schema change, deliberately out of this PR's contract-safe scope).

Full suite 4171 runs / 0 failures, 100% line coverage, rubocop clean.

PR #309, opened 2026-06-05
perf(daemon): cut inter-stage latency with low-CPU fast-tick

ivankuznetsov:
## Summary

A task used to sit idle for roughly a minute between stages (plan → work, etc.), and the TUI looked frozen for the same window. After this change a completed stage advances to its successor within ~1-2s, and the TUI reflects on-disk changes within ~1.5s, all while idle CPU stays near zero.

The requirement assumed a hardcoded ~1-minute TUI repaint cadence. The code says otherwise: `StateSource` already polls at 1 Hz and `App.start_snapshot_poller` already redraws only when the snapshot changes. The "repaints once a minute" feeling was a symptom — between stages the on-disk state did not change for ~60-90s, so the correctly-frugal TUI had nothing new to show. The real latency lived on the daemon side, so that is where most of this PR lands.

### Where the minute went

| Source | Before | After |
|---|---|---|
| SUCCESS inter-stage cooldown (`concurrency_controller.rb`) | 60s forced wait before the next stage could dispatch | removed — next stage may dispatch immediately |
| Completion detection (`dispatcher.rb` run loop) | up to 30s (`poll_interval_sec`) before a finished child was noticed | ~1s cheap probe (`fast_poll_sec`) |

The 60s SUCCESS cooldown existed "to give the operator one full TUI polling cycle to notice the transition" — exactly the human-intervention window the requirement asked to drop. Removing it is option A (zero wait): humans still intervene through explicit pause/stop or by editing state files; only the automatic wait is gone.

### Low-CPU fast-tick (the design decision)

Reacting faster than the 30s tick without re-scanning every enrolled project each second is the crux. Rather than dropping `poll_interval_sec` to 1s (which would run the full `hive status` scan every second and violate the hard low-CPU constraint at scale), the daemon now wakes every `fast_poll_sec` (default 1s) and does only cheap work: a non-blocking child reap plus an mtime stat over the state files seen on the last full scan. A child exit or an advanced mtime triggers a full `tick` immediately; otherwise the 30s `poll_interval_sec` full scan remains as the backstop that still catches what the cheap probe cannot see (newly enrolled projects, folder renames).

```mermaid
flowchart TB
  W[wake every fast_poll_sec ~1s] --> D{poll_interval_sec<br/>elapsed?}
  D -->|yes| T[full tick: reap, fetch status, dispatch]
  D -->|no| P[cheap probe: reap_all + mtime stat]
  P -->|child exited or mtime moved| T
  P -->|nothing changed| W
  T --> W
```

The protective backoffs are deliberately untouched: the transient-failure schedule (60/120/300s → quarantine) and the WRONG_STAGE backoff guard against fault loops, not human review. WRONG_STAGE previously reused the SUCCESS cooldown constant, so it now has its own `WRONG_STAGE_BACKOFF_SEC = 60` — removing the SUCCESS wait does not silently weaken the race/classifier backoff.

### TUI re-parse gate

The TUI was already redraw-gated; this extends the same idea to the parse step. `StateSource#refresh_once` now stats an mtime fingerprint (each visible row's state file plus the project's `stages` directory and its children, and the global project registry) and reuses the cached `Snapshot` when nothing moved, skipping the in-process `Status#json_payload` reparse. New task folders, ordinary marker edits, and `hive init`/`forget` registry changes still invalidate the cache. A bounded 3s liveness fallback forces a reparse even when no mtime moved, so file-less liveness fields (`claude_pid_alive`, `live_task_lock`) self-heal rather than going stale indefinitely. This keeps the existing MRI-atomic single-reference pattern — no new mutex.

## Test plan

Targeted unit suites are green and `rubocop` is clean on the changed lib files.

- `concurrency_controller_test.rb`: SUCCESS no longer cools down the same slug; WRONG_STAGE still backs off via `WRONG_STAGE_BACKOFF_SEC`; transient schedule unchanged; combined SUCCESS → follow-up tick → same stage stays `:skip` (anti-re-fire regression).
- `dispatcher_test.rb`: idle fast probe does not call the status fetch; a tracked state-file mtime bump (or a child exit) requests a full tick; the reap happens before the follow-up dispatch; ≤2s wall-budget end-to-end with an injected clock.
- `state_source_test.rb`: unchanged mtimes reuse the cached snapshot; an mtime bump forces a reparse; a state-file change is observed within the 1.5s budget; registry changes invalidate the cache; the 3s liveness fallback reparses.
- `config_test.rb`: `fast_poll_sec` default (1) and the `>= 1` floor.

```
bundle exec ruby -Itest test/unit/daemon/concurrency_controller_test.rb
bundle exec ruby -Itest test/unit/daemon/dispatcher_test.rb
bundle exec ruby -Itest test/unit/tui/state_source_test.rb
bundle exec ruby -Itest test/unit/tui/app_test.rb
bundle exec ruby -Itest test/unit/config_test.rb
bundle exec ruby -Itest test/integration/dispatch_request_concurrency_test.rb
```

One gap is recorded in `wiki/gaps.md` (#21): these are unit-pinned but not yet live-smoked against a running daemon plus `hive tui` under real project load.

## Review summary

Two automated review passes ran (CE code review). Every Requirements-Trace item is implemented and nothing plan-required is missing; both passes confirmed the U1/U2/U3 work faithfully implements the plan and is clean (97 + 2 runs, 0 failures; rubocop clean).

- **High:** one finding — the TUI mtime cache never watched the global project registry, so `hive init`/`forget` changes could stay invisible past the ≤1.5s bound. Fixed by adding the registry file to the mtime fingerprint.
- **Medium:** the re-parse gate let file-less liveness fields go stale indefinitely → added a bounded 3s liveness fallback reparse; `version_drift_detected?` ran every ~1s → gated behind the full-tick cadence to keep the idle probe cheap.
- **Nits (auto-fixed):** `nil`-mtime fallback that could force perpetual full ticks; documenting the benign double-reap / post-dispatch redundant tick / dry-run fast-tick asymmetry; pinning the plan's Unit-1 anti-re-fire and Unit-2 ≤2s scenarios as single regression tests; noting the residual idle cost of the 3s liveness fallback in the AC notes.

The second-pass reviewers also flagged that the branch is stale vs a diverged `origin/main` (7 unrelated merged PRs), which made an `origin/main..HEAD` diff show those PRs' changes as spurious "reverts." Verified against the real merge-base `c9207c1`: this branch touches only daemon / TUI / config / test / wiki files (`git merge-tree` shows no `config.rb` conflict; only the append-only `wiki/log.md` needs a landing-time merge). Those reviewer findings were stale artifacts of the contaminated diff, not defects in this work.

## Linked task

- Hive task: `we-need-to-reduce-tui-260605-dc24`
- PR: https://github.com/ivankuznetsov/hive/pull/309

---


ivankuznetsov:

## High
- [ ] TUI mtime cache never watches the global project registry (`lib/hive/tui/state_source.rb:80`): once an initial snapshot exists, unchanged tracked state files skip `Hive::Config.registered_projects`, so `hive init`/`forget` config changes can remain invisible indefinitely instead of satisfying the <=1.5s on-disk-change requirement.

## Medium
- [ ] Shared Claude reviewer sessions no longer recover after tmux exits (`lib/hive/claude_launcher.rb:101`): removing the reestablish closure means one dead shared session makes later Claude reviewers cascade-fail with "session no longer exists" rather than restarting the session.
- [ ] Plan approval pauses are again announced as brainstorm questions (`lib/hive/bot/notification_builders.rb:105`): every `waiting` marker now uses the brainstorm `/answer` notification, and the daemon-enabled 3-plan suppression was removed, so operators get unactionable/misleading pings while the daemon is about to auto-approve.
- [ ] Active Telegram conversations cross-suppress projects and are unsynchronized (`lib/hive/bot/conversation_store.rb:83`): `active_for_slug?` dropped the project argument and mutex, so two projects sharing a slug can silence each other's alerts and concurrent bot/status threads can mutate `@states` during iteration.
- [ ] Patrol can reopen reworded duplicates (`lib/hive/commands/patrol.rb:119`): the similarity gate and stored title/category tokens were removed, so findings that differ only by agent wording no longer match active or dismissed PRs unless their exact fingerprint is identical.
- [ ] Reviewer deadlines are split by fair share again (`lib/hive/stages/review.rb:1021`): with multiple reviewers, an early deep reviewer can be cut to `remaining/specs_remaining` even though the overall wall-clock budget still has room, causing avoidable reviewer timeouts/partial failures.

## Nit
- [ ] Dispatch result schema now accepts chat_id 0 (`schemas/hive-dispatch-result.v1.json:31`): Telegram chat ids are never zero, so relaxing the schema lets invalid result notices pass external validation.

ivankuznetsov:

# Code Review — we-need-to-reduce-tui-260605-dc24 (pass 1)

Scope reviewed: branch changes vs merge-base `c9207c1` (origin/main has since
diverged by 6 unrelated commits; diffing against origin/main shows those as
spurious reverts, so review was scoped to the merge-base). Core files:
`lib/hive/config.rb`, `lib/hive/daemon/concurrency_controller.rb`,
`lib/hive/daemon/dispatcher.rb`, `lib/hive/tui/state_source.rb` + tests + wiki.

Verification run: targeted unit suites green
(`concurrency_controller_test` 35, `dispatcher_test` 97, `state_source_test` 9,
`app_test` 12, `config_test`); `rubocop` clean on the 4 changed lib files.

Plan-trace: Unit 1 (remove SUCCESS cooldown, preserve WRONG_STAGE/transient
backoffs) ✓; Unit 2 (fast-tick 2b) ✓; Unit 3 (TUI mtime re-parse gate, the
optional step 2) implemented ✓. No plan-required item is missing.

## High
_No high-severity findings._

## Medium
- [ ] TUI re-parse gate makes process-liveness fields go stale indefinitely (`lib/hive/tui/state_source.rb:81-100`): the new mtime fingerprint only tracks state-file + stage-dir mtimes, but `live_task_lock` (and `claude_pid_alive`) are derived from `pid_alive?` (`lib/hive/commands/status.rb:550`), not from any file. When a lock holder dies the `.lock` file stays on disk unchanged, so `refresh_once` short-circuits and the TUI shows a "live lock" indicator that never clears — there is no time-based re-parse fallback. Previously the unconditional 1 Hz re-parse caught this within ≤1.5s; this optional optimization (Unit 3 step 2) regresses it.
- [ ] `version_drift_detected?` now runs every `fast_poll_sec` (~1s) instead of every `poll_interval_sec` (~30s) (`lib/hive/daemon/dispatcher.rb:298,408-417`): it sits above the `full_tick_due?` gate in the run loop, so its `Digest::SHA256.file` hash of the schema file executes ~30× more often on the idle path. The plan's Unit 2 intended the per-second probe to be *only* cheap work (waitpid + stat); per-second file hashing works against the hard "near-zero idle CPU" constraint. Gating it behind `full_tick_due?` (or the cheap-probe trigger) would restore the intended cadence.

## Nit
- [ ] `refresh_tracked_state_file_mtimes` fallback can cause perpetual full ticks (`lib/hive/daemon/dispatcher.rb:507-522`): when a tracked `state_file` is absent on disk but the status row still carries a non-nil `state_file_mtime`, the stored value is the row mtime while `tracked_state_file_mtime_changed?` re-stats with `safe_mtime` → `nil`, so `nil != <Time>` is always "changed" and a full tick fires every second until a full tick re-baselines. Edge case (status reports a path the file system no longer has) and usually self-heals after one full tick, but it momentarily defeats the low-CPU goal. Storing `nil` consistently (drop the `|| row.state_file_mtime` fallback) would avoid the type/precision mismatch.
- [ ] Double `reap_all` per fast tick on the child-exit path (`lib/hive/daemon/dispatcher.rb:316,447,449`): `cheap_probe_requires_full_tick?` calls `reap_completed`, then the follow-up `tick` calls it again. The second sweep is an empty waitpid no-op (record_completion not re-fired), so it's benign — noting only that the redundancy is intentional-looking but undocumented.
- [ ] One extra full tick ~1s after each dispatch (`lib/hive/daemon/dispatcher.rb:276,513`): `refresh_tracked_state_file_mtimes` captures pre-dispatch row mtimes, so the freshly-spawned agent's AGENT_WORKING marker write trips `tracked_state_file_mtime_changed?` on the next probe, forcing a redundant full tick. Harmless (re-evaluation is correct) and low-frequency, but worth a comment so it isn't mistaken for a bug later.

ivankuznetsov:

## High
- [ ] `lib/hive/bot/conversation_store.rb:11` removed the mutex around `@states`: the store is still accessed from the Telegram update thread and status/notification polling path, so concurrent `delete_if`/writes can raise or lose active conversations.
- [ ] `lib/hive/claude_launcher.rb:89` removed shared-session reestablishment: when a Claude tmux reviewer exits or crashes, later reviewers in the same shared session now see "session no longer exists" and cascade-fail the review pass.

## Medium
- [ ] `lib/hive/bot/notification_dispatcher.rb:59` removed daemon-plan-pause suppression: daemon-enabled `3-plan` waiting rows are auto-approved by the daemon but will again page operators as needs-input alerts.
- [ ] `lib/hive/bot/notification_builders.rb:103` treats every `waiting` marker as brainstorm Q&A: daemon-off `3-plan` approval pauses are mislabeled as "Brainstorm questions" and offer the wrong answer-in-chat action.
- [ ] `lib/hive/bot/conversation_store.rb:83` dropped project scoping from active-conversation suppression: a chat answering slug `s1` in one project suppresses needs-input alerts for slug `s1` in unrelated projects.
- [ ] `lib/hive/config.rb:314` changes `patrol.draft_prs` back to `true`: patrol PRs become drafts by default again, so the babysitter skips them despite the existing ready-PR handoff contract.
- [ ] `schemas/hive-dispatch-result.v1.json:31` no longer rejects `chat_id: 0`: the dispatch-result contract now accepts an undeliverable Telegram chat id and the producer round-trip/non-zero schema tests were removed.

## Nit
- [ ] `lib/hive/bot/notification_dispatcher.rb:78` no longer logs active-conversation suppressions: operators lose the audit event explaining why a needs-input alert was skipped.

ivankuznetsov:

# Code review — we-need-to-reduce-tui-260605-dc24 (pass 2)

Reviewed `git diff origin/main..HEAD` and the branch's real change set
`git diff <merge-base c9207c11>..HEAD`. The branch's own work (7 commits:
U1 remove SUCCESS cooldown, U2 fast-tick, U3 TUI mtime gate + 2 fixes + docs)
faithfully implements the plan, is well-tested, and is clean:

- `bundle exec ruby -Itest` over the 5 affected unit files + the integration
  concurrency test: **97 + 2 runs, 0 failures, 0 errors**.
- `rubocop` on the 4 changed Ruby files: **no offenses**.
- Anti-re-fire guard (the plan's headline Unit 1 risk) runs in the probe path:
  `reap_completed` calls `refresh_post_completion_mtime` per entry
  (`dispatcher.rb:476`), and the baseline-refresh is unit-tested
  (`test_refresh_post_completion_mtime_records_existing_state_file`).
- Wiki pages (`modules/daemon.md`, `commands/tui.md`, `gaps.md`) accurately
  describe the new behavior and honestly log the live-smoke gap.

Every Requirements-Trace item is implemented; nothing plan-required is missing.
Findings below are integration hygiene and test-completeness, not code defects.

## High

(none)

## Medium

- [ ] **Branch is stale vs `origin/main`; the assigned `origin/main..HEAD` diff is contaminated.** The branch forks at `c9207c11`, but `origin/main` has since merged 7 unrelated PRs (#291/#308/#302/#306/#295/#293/#307 — patrol, bot, claude_launcher, review, and the config-default flips `draft_prs`→false and `max_wall_clock_sec`→14_400). So the assigned diff shows the *inverse* of all of those as large "deletions/reverts" that are NOT this task's work — a naive read produces dozens of false findings. A real 3-way `git merge` keeps main's values (verified via `git merge-tree`: **no `config.rb` conflict**; only `wiki/log.md` conflicts on the append-only changelog). No code regression, but the branch must be rebased/merged onto current `origin/main` before landing and the `wiki/log.md` conflict resolved; review the actual change set via `git diff c9207c11..HEAD`.

## Nit

- [ ] **Fast-tick does not accelerate dry-run mode.** `cheap_probe_requires_full_tick?` (`dispatcher.rb:453`) reaps only via `reap_completed`/`reap_all`; `reap_dry_run` is invoked solely inside the full `tick` (`dispatcher.rb:174`). So dry-run pseudo-child completions advance only on the 30s backstop, not within `fast_poll_sec`. Harmless for production (dry-run isn't the latency path the AC targets), but a one-line comment noting the asymmetry would prevent future confusion.

- [ ] **TUI 3s liveness fallback re-incurs idle `json_payload` + `Dir.glob` ~20×/min.** `LIVENESS_REPARSE_FALLBACK_SECONDS = 3.0` (`state_source.rb`) forces a full reparse + fingerprint rebuild even when mtimes are unchanged, so the "skip the idle re-parse" CPU win is partial. It's a justified correctness tradeoff (liveness fields `claude_pid_alive`/`live_task_lock` flip without touching a file and must self-heal), and it never causes a redraw thanks to the `app.rb` snapshot dedup — but the residual idle cost isn't called out against the plan's "near-zero idle CPU" AC. Acceptable as-is; worth a sentence in the AC notes.

- [ ] **Plan's Unit 1 end-to-end anti-re-fire scenario isn't a single regression test.** The plan named: "simulate SUCCESS where the post-completion mtime refresh has run; the next decision does not re-dispatch the same completed stage." Its two halves are covered separately (`test_refresh_post_completion_mtime_records_existing_state_file` proves the baseline bumps; the `baseline_recorded` / `observe_state_file_mtime` tests prove the baseline gates dispatch), but no one test drives SUCCESS-completion → follow-up tick → asserts the same stage stays `:skip`. Optional to add the combined regression the plan called for, since the deleted cooldown was the only safety net being removed.

- [ ] **Plan's Unit 2 "≤2s wall-budget" latency assertion isn't pinned as one test.** The ≤2s AC is covered indirectly — `test_fast_probe_*` proves the probe triggers a full tick on child-exit/mtime-change, and the integration test proves a reaped SUCCESS dispatches its successor in the *same* tick — but no single test asserts the end-to-end ≤2s bound with an injected clock as the plan's scenario described. The composition is sound; a single bound-asserting test would make the AC self-documenting.

PR #301, opened 2026-06-04
feat(archive): hide old archived tasks behind an Archive view

ivankuznetsov:
## Summary

Archived tasks (stage `9-done`) used to pile up forever in the two surfaces operators watch all day — the TUI grid and `hive status` text — burying active work under a growing "Archived" list. This change hides any archived task whose folder mtime is older than 3 days from those two surfaces, while keeping every archived task one keystroke (or one command) away through a new TUI **Archive pane** and a no-argument `hive archive` listing.

Visibility is *narrowed only in the human-facing renderers*. `hive status --json` still emits every row, so the bot dispatch loop, daemon, and agents that read the JSON are completely unaffected.

### How hiding is decided

One shared policy module, `Hive::ArchiveFilter`, is the single source of truth so the two filter sites can never drift on the threshold, the stage name, or the exemption rule. A task is hidden only when **all** of these hold:

| Condition | Why |
|---|---|
| Stage is `Hive::Stages::DIRS.last` (`9-done`) | Only archived tasks are ever hidden; the stage name is derived, never hardcoded. |
| Marker is `:complete` or `:none` | A `9-done` task carrying any unresolved marker (`:error`, `:agent_working`, `:manual_steering`, …) is **exempt and always shown**, regardless of age. |
| Folder mtime is older than 3 days | The threshold is a code constant (`3 * 24 * 60 * 60`), intentionally not configurable. |

Anything ambiguous fails open (stays visible): a nil folder mtime, an unrecognized marker, a non-`9-done` stage.

The age is read from the task **folder mtime**, a deliberate product decision (no new sidecar, schema field, or `status.md` line). `rename(2)` does not bump a moved directory's mtime, so in practice the clock reflects the last write *inside* the folder — for a cleanly archived task that is `Done.run!` writing the `:complete` marker, which is approximately archive time. A later write into the folder can reset the clock and keep a task visible longer than 3 days; that imprecision was accepted in favor of not persisting an `archived_at` field.

### What the operator sees

- **`hive status` text** hides old archived tasks and appends one line per project: `… and N archived >3d ago (hive archive to view)`. A project whose tasks are all hidden still prints its header plus the summary, so filtered work never looks like an empty project.
- **TUI grid** drops the same rows from the rendered projection (cursor and keystrokes operate on that projection, so they can never land on a hidden row) and shows `+N hidden — open Archive pane` in the footer.
[...]

PR #315, opened 2026-06-05
Hive patrol: Dry-run git stub permits file writes through read-only commands

ivankuznetsov:
## Summary

Hardens `bin/hive-babysitter-stub-git`, the dry-run git shim the babysitter
uses to neutralise side-effecting git calls while letting read-only ones
through. The original guard allowed only read-only subcommands but still handed
the **unscreened** argv (and the inherited environment) to real git, so several
read-only-allowlisted commands could be coerced into executing arbitrary
commands or writing files — strictly worse than the single `--output` write the
guard set out to block.

This change closes that class of bypass:

- **Exec-influencing global options** (`-c key=val`, `--config-env`,
  `--exec-path`) are screened in the global region where git honours them,
  blocking `diff.external` / `*.textconv` / `core.pager` style command
  injection. Exact-token `-c` is sufficient because git rejects the bundled
  `-cKEY=VAL` form.
- **`--output[=file]`** is screened on every subcommand, anchored exactly so
  read-only formatting flags (`--output-indicator-*`) still pass.
- **`git grep` pager options** (`-O`, `--open-files-in-pager`) are screened in
  the grep region only; the `-O` match covers bundled short flags (`-nO`,
  `-inO`, …) so a flag reorder can't sneak the pager past. `git diff -O<orderfile>`
  (read-only) is left alone.
- **Exec-influencing environment variables** (`GIT_EXTERNAL_DIFF`, `GIT_PAGER`,
  `GIT_SSH`, `GIT_SSH_COMMAND`, `GIT_CONFIG*`) are scrubbed before the exec
  handoff, since argv screening structurally cannot see them.
- **Robustness:** the skip-audit log failure is now reported via `warn` instead
  of being silently swallowed; a set-but-invalid `HIVE_BABYSITTER_REAL_GIT`
  exits 127 with a one-line message instead of a raw backtrace; and screened
  skips emit a distinct reason so over-block false positives are diagnosable.

Over-blocking is deliberately scoped (subcommand-region screening, exact
anchoring) so legitimate read-only commands such as `git grep -c` / `--count`
and diff orderfiles are not turned into silent no-ops. The dry-run guard remains
documented as best-effort.

## Test plan

- `test/unit/babysitter/dry_run_env_test.rb` gains regression coverage for the
  exec/write vectors: `-c diff.external`, bundled `grep -nO`, two-token
  `--config-env KEY=VAL`, bare `--open-files-in-pager`, `--output` skips across
  `log`/`show`, `--output-indicator-*` passthrough, and the trailing bare
  `-c`/`-C` underflow guard.
- Tests invoke the real stub binary and assert against the dry-run skip log (no
  mocks), including negative controls that read-only commands still pass.

## Review summary

Two CE/PR-toolkit review passes (claude + codex + pr-review-toolkit personas).
All actionable findings were auto-fixed:

- Pass 01 surfaced the core bypass (unscreened argv reaching real git via
  `-c diff.external` / `grep -O`); the fix screens the real-git argv across the
  whole invocation.
- Pass 02 closed residual gaps: bundled `-nO` short-flag bypass, env-based exec
  vectors, exact `--output` anchoring, subcommand-scoped `-O`/pager screening,
  guarded exec handoff, and widened audit-log rescue.

Codex findings about `lib/hive/stages/review.rb`, `lib/hive/commands/*`, and the
`schemas/` files were resolved as false positives — they are reverse-diff
artifacts of the branch being behind `origin/main` (which has since merged
#301/#313/#314), not changes this branch makes. Verified: `598fd191..HEAD`
touches only the stub, its test, and 4 wiki pages.

## Linked task

/home/asterio/Dev/hive/.hive-state/stages/6-review/patrol-command-bin-hive-515f2cfe


ivankuznetsov:

# Codex CE Code Review 01

No plan.md was found in the task folder; review was performed against `git diff origin/main..HEAD` without plan grounding.

## High

## Medium
- [ ] `hive status --json` removes the required `tasks[].folder_mtime` field without bumping `SCHEMA_VERSIONS["hive-status"]`: `Status#task_payload` no longer emits the key while the current v3 contract still reports schema_version 3, so existing v3 consumers that rely on the required folder timestamp will silently receive a breaking shape.

## Nit

ivankuznetsov:

# Claude CE Code Review 01

No plan was found in the task folder; review is grounded in the diff only.

**Diff-scope caveat:** `git diff origin/main..HEAD` is misleading here — this
branch is 3 commits *behind* `origin/main` (it forked at `598fd191`, and main
has since merged #301 archive, #313 review-partial-failure, #314 tui-resize).
That diff therefore shows ~1395 *reverse* deletions of work already on main.
This review instead covers the branch's **actual** contribution
(`merge-base..HEAD`), which is 6 files: `bin/hive-babysitter-stub-git`,
`test/unit/babysitter/dry_run_env_test.rb`, and 4 wiki pages. Note: the prior
`codex-ce-code-review-01.md` findings (about `lib/hive/stages/review.rb` and
`lib/hive/commands/status.rb`) are false positives — those files are not touched
by this branch and only appear via the polluted reverse diff.

## High
- [ ] Dry-run guard still allows arbitrary command execution, a worse bypass than the `--output` write it closes: `read_only?` waves `diff`/`grep`/`log`/`show` through (`bin/hive-babysitter-stub-git:57-59`), while `stripped_global_options` discards `-c key=val` without inspecting it (`:16-18`). So `git -c diff.external='<cmd>' diff` (with worktree changes) executes `<cmd>` via the external diff driver, and `git grep -O'<cmd>' <pat>` / `--open-files-in-pager=<cmd>` runs `<cmd>` as the pager — both pass `no_write_options?` (it only screens `--output`) and reach real git, defeating the side-effect guard this PR exists to harden.

## Medium
- [ ] Blocklist approach is incomplete by construction, contradicting the finding's own guidance: the patrol finding recommended rejecting "any other command-specific file-writing flags, or replace the broad allowlist with narrowly parsed safe forms." `no_write_options?` (`bin/hive-babysitter-stub-git:49-51`) screens exactly one flag (`--output`), so any write/exec-capable option not on that list still slips through (`-c diff.external`, `grep -O`/`--open-files-in-pager`, and any future git option). An allowlist of safe forms per subcommand would close the class; the current blocklist only patches the one reported instance.

## Nit
- [ ] Function name overpromises: `no_write_options?` implies it screens all write-capable options but checks only `--output` — rename to `no_output_option?` or expand its coverage so the name matches behavior (`bin/hive-babysitter-stub-git:49`).
- [ ] Test coverage mirrors the narrow fix: `test/unit/babysitter/dry_run_env_test.rb` asserts `--output`/`--output=` are skipped but adds no case for the remaining `-c diff.external` / `git grep -O` exec vectors, so the gap would not be caught by regression tests.

ivankuznetsov:

No plan was found in the task folder; review is grounded in the diff only.

**Diff-scope caveat:** `git diff origin/main..HEAD` is misleading — this branch
is 3 commits *behind* `origin/main` (forked at `598fd191`; main since merged
#301/#313/#314), so that diff shows ~1395 *reverse* deletions of work already on
main. Personas were scoped to the branch's **real** change
(`merge-base..HEAD`): `bin/hive-babysitter-stub-git`,
`test/unit/babysitter/dry_run_env_test.rb`, and 4 wiki pages. All exec vectors
below were reproduced empirically against real git 2.54.0 through this exact
stub.

Personas run: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`,
`comment-analyzer`. (`type-design` not applicable — no new types.)

## High
- [ ] [code-reviewer] Root cause — the final `exec` call runs the *original* `argv` while the allowlist check runs on the `-c`-stripped copy (`bin/hive-babysitter-stub-git:69,82`): every `-c key=val` the parser discards is re-attached when handing off to real git, so the read-only allowlist constrains nothing. Reproduced argv-only arbitrary command execution: `git -c diff.external='touch X;true' diff`, `git -c diff.foo.textconv=… log -p`/`show`/`grep --textconv`, `git -c core.fsmonitor='touch X;true' status`, `git -c core.pager='touch X' diff`.
- [ ] [code-reviewer] `git grep -O<cmd>` / `--open-files-in-pager=<cmd>` runs the named pager regardless of tty (`bin/hive-babysitter-stub-git:57-59`): `grep` is allowlisted and `no_write_options?` screens only `--output`, so the command-execution vector passes through.
- [ ] [pr-test-analyzer] The two proven exec vectors (`-c diff.external`/`*.textconv`, `grep -O`/`--open-files-in-pager`) are entirely untested (`test/unit/babysitter/dry_run_env_test.rb`): the PR markets itself as a dry-run write guard but pins only `--output`, leaving strictly-worse code-execution paths with zero regression coverage.
- [ ] [silent-failure-hunter] Skipped write commands return exit status 0 (`bin/hive-babysitter-stub-git:71-75`): a suppressed mutating command is indistinguishable from a successful one to the caller — any `system("git", …) && proceed` logic in the babysitter treats a skipped commit/push/branch-create as done; the only signal is an unstructured stderr line.

## Medium
- [ ] [code-reviewer][silent-failure-hunter] `no_write_options?` is a one-entry denylist (`--output`) over the full git flag space (`bin/hive-babysitter-stub-git:49-51`): structurally it flips the intended default-deny into default-allow for any not-yet-enumerated side-effecting flag on the allowlisted commands. Fix by screening the argv actually handed to real git (drop stripped global options) and/or inverting to an explicit safe-flag allowlist rather than enumerating write flags.
- [ ] [silent-failure-hunter] `log_skip` swallows `SystemCallError` and returns nil (`bin/hive-babysitter-stub-git:4-9`): if the skip log can't be written (perms, ENOSPC, read-only FS) the durable audit record — the thing tests assert on — is silently dropped with no warning that logging failed.
- [ ] [pr-test-analyzer] `--output` screening is shared across all broad commands (`log`/`show`/etc.) but only `git diff` is tested (`test/unit/babysitter/dry_run_env_test.rb:63-64,96-97`): a refactor moving the check into a `diff`-only branch would silently reopen `git log/show --output=<file>` with no test failure. Add `log`/`show` `--output` skip assertions to lock the shared-helper contract.
- [ ] [pr-test-analyzer] Leading `git --output=patch.diff diff` is consumed as a global option by `stripped_global_options` (`bin/hive-babysitter-stub-git:21-22`) and passes as read-only, untested: git doesn't accept `--output` pre-subcommand so it's a parser-robustness gap rather than a live exploit, but it shows the guard only catches post-subcommand placement.
- [ ] [comment-analyzer] Method name `no_write_options?` overstates coverage (name-as-documentation rot) (`bin/hive-babysitter-stub-git:49-51`): it checks one option family while the name asserts the absence of write options generally, so a future reader of `read_only?` will assume all write/exec vectors are screened. Rename to `no_output_file_option?` (or similar) to match the single check.

## Nit
- [ ] [code-reviewer] `stripped_global_options` does `rest.shift(2)` on a trailing valueless `-c`/`-C`/`--git-dir`/`--work-tree` (`bin/hive-babysitter-stub-git:17-18`): on a 1-element array this yields `[]`, `rest.first` becomes nil, and the command is skipped — harmless here (default-deny) but the parser assumes a value is always present.
- [ ] [comment-analyzer] `wiki/commands/babysit.md:70` phrasing ("`--output` … makes the invocation a write") reads slightly broader than the code — accurate but may be over-read as the complete write-screening story; tighten to "its only file-write screen." (Positive: `wiki/gaps.md:48` and `wiki/modules/babysitter.md:67` are correctly scoped to `--output` and keep the live-smoke gap open — no hallucination.)

---
**Persona consensus / positives (no action):** default-deny fall-through (`read_only?` `else false`, `read_only_remote?` `else false`) is sound; `HIVE_BABYSITTER_REAL_GIT` empty/unset → exit 127 is correct; the `--output` test uses real binary-invocation logging (not mocks), covers both `=`/space spellings, and has a strong negative control (`git diff --name-only` must still pass). The unanimous verdict across personas: the `--output` fix is correct and well-tested, but it patches one instance of the reported class while the stub's core security property (the allowlist actually constraining real git) remains broken because the argv handed to real git is never sanitized.

ivankuznetsov:

# Codex CE Code Review 02

## High
- [ ] Dry-run git stub still permits command execution through inherited Git environment hooks: `bin/hive-babysitter-stub-git` screens argv only, so an allowed `git diff` with `GIT_EXTERNAL_DIFF='sh -c ...'` reaches real git and executes the command without being logged as skipped.
- [ ] Branch is stale against `origin/main` and would revert unrelated shipped work: `git diff origin/main..HEAD` removes archive filtering/status fields, partial-reviewer-failure recovery, and TUI resize handling from PRs #301/#313/#314 even though the patrol task only covers the babysitter git stub.
- [ ] `REVIEW_WAITING reason=reviewer_partial_failure` cannot follow its advertised recovery path: `Stages::Review` treats every `REVIEW_WAITING` resume as a skip-review/triage fix run, while `hive markers clear` does not allow clearing `REVIEW_WAITING`.

## Medium
- [ ] `hive-status` mutates published v2/v3 JSON contracts without a version bump: `folder_mtime` is removed from `Status#task_payload` and from `schemas/hive-status.v2.json`/`v3.json`, breaking pinned consumers of those schema versions.

## Nit

ivankuznetsov:

# Claude CE Code Review 02

No plan was found in the task folder; review is grounded in the diff only.

**Diff-scope caveat:** `git diff origin/main..HEAD` is misleading — this branch
forked at `598fd191` and `origin/main` has since merged unrelated work (#301
archive, #313 review-partial-failure, #314 tui-resize). The two-dot diff
therefore shows ~1394 *reverse* deletions of code already on main that this
branch does not touch. Findings about `lib/hive/stages/review.rb`,
`lib/hive/commands/status.rb`, `schemas/hive-status.v*.json`, or the TUI panes
are artifacts of that polluted diff, not this branch's work — they are excluded
here. This review covers the branch's **actual** contribution
(`merge-base 598fd191..HEAD`): `bin/hive-babysitter-stub-git`,
`test/unit/babysitter/dry_run_env_test.rb`, and 4 wiki pages.

**Context:** This pass's new commit (`b5f53f98`) is the *fix* for the pass-01
High finding — it adds `dangerous_option?` / `no_exec_or_write_options?` to
screen `-c`, `--config-env`, `-O`/`--open-files-in-pager`, and `--output`
across the whole argv. The core fix is correct and the `log_skip` silent-`nil`
rescue is rightly replaced with a `warn`. The findings below are residual gaps
in that hardening. The dry-run guard is documented as best-effort, which lowers
the ceiling on severity.

## High
- [ ] Bundled short-option form `git grep -nO'<cmd>' <pat>` defeats the new `-O` screen — CONFIRMED exploit: `dangerous_option?` only matches `-O` at the *start* of a token (`bin/hive-babysitter-stub-git:66`), so any grep short flag bundled before `O` (e.g. `-n`, `-i`, `-l`) hides it. `grep` is read-only-allowlisted, so the call reaches real git and runs `<cmd>`. Reproduced: `git grep -nO'touch pwned' needle` creates `pwned`, while the screened `git grep -O'touch pwned' needle` is correctly skipped — a trivial flag reorder bypasses the exact protection this commit adds.

## Medium
- [ ] Env-based exec vectors are not addressed by argv screening: `GIT_EXTERNAL_DIFF`, `GIT_PAGER`, `GIT_SSH_COMMAND`, and `GIT_CONFIG_*` reach real git through an allowlisted `git diff`/`git grep`/`git log` and execute commands with no `-c`/`-O` token present. argv screening structurally cannot catch these; closing the class needs the `DryRunEnv` overlay to scrub these vars (or a per-subcommand safe-form allowlist) rather than only filtering argv.

## Nit
- [ ] `--output` prefix over-matches read-only diff options: `arg.start_with?("--output")` (`bin/hive-babysitter-stub-git:66`) also flags `--output-indicator-new/-old/-context/-frag=<char>`, which are read-only `git diff` formatting options — using them silently skips a legitimate read command. Anchor to `--output` / `--output=` exactly.
- [ ] `-O` prefix over-matches `git diff -O<orderfile>`: the diff orderfile option is read-only but `start_with?("-O")` screens it as dangerous, skipping a legitimate command. Only `git grep`'s `-O` runs a pager; consider scoping the `-O` screen to the `grep` subcommand.
- [ ] `--exec-path=<dir>` redirects git's helper-binary lookup and is an exec-influencing global option that is not screened — lower risk since it requires planting binaries at known names, but it is in the same "global option that changes what git runs" class the commit targets.
- [ ] Test gaps mirror the screening gaps: `test/unit/babysitter/dry_run_env_test.rb` covers glued `-Otouch pwned` and `--config-env=...` but not the bundled `-nO` bypass above, the two-token `--config-env KEY=VAL` form, or bare `--open-files-in-pager` — so the High-severity regression would not be caught.

ivankuznetsov:

# PR Review Toolkit 02

No plan was found in the task folder; review is grounded in the diff only.

**Diff-scope caveat:** `git diff origin/main..HEAD` is polluted — this branch
forked at `598fd191` and `origin/main` merged unrelated work since, so ~1394
lines appear as reverse-deletions of files this branch never touched. All
personas were scoped to the branch's **actual** contribution
(`git diff 598fd191..HEAD`): `bin/hive-babysitter-stub-git`,
`test/unit/babysitter/dry_run_env_test.rb`, and 4 wiki pages. Findings about
other `lib/`, `schemas/`, or TUI files are reverse-diff artifacts and are
excluded.

comment-analyzer. (type-design-analyzer N/A — no new types.) The change is the
dry-run git-stub exec/write hardening (`dangerous_option?` /
`no_exec_or_write_options?`). Core fix is sound; findings are residual gaps,
over-blocks, and unguarded paths. Severity convention: a guard *bypass*
(fail-open) is High; *over-blocking* a legit read-only command (fail-closed) is
Medium.

## High
- [ ] [pr-test-analyzer] Bundled short-option `git grep -nO'<cmd>' <pat>` bypasses the new `-O` screen and runs `<cmd>` (confirmed file write via reproduction): `dangerous_option?` matches `-O` only at token start (`bin/hive-babysitter-stub-git:66`), so `-nO…` (any grep short flag bundled before `O`) reaches real git through the read-only `grep` allowlist — defeating the exact protection this commit adds. Fix needs `/\A-[^-]*O/`-style matching, not `start_with?("-O")`.

## Medium
- [ ] [code-reviewer] `git grep -c needle` (read-only `--count`) is wrongly skipped: the whole-argv scan treats any `-c` token as the dangerous config option (`bin/hive-babysitter-stub-git:63,71`), turning a legitimate read-only command into a silent no-op (exit 0) that can mislead the agent.
- [ ] [code-reviewer][pr-test-analyzer] `--output` prefix over-matches read-only `--output-indicator-{new,old,context,frag}=<char>` diff formatting options (`bin/hive-babysitter-stub-git:66`), skipping legitimate read-only diffs; match `--output` / `--output=` exactly instead.
- [ ] [code-reviewer] `git diff -O<orderfile>` (read-only diff orderfile) is over-blocked by the `-O` prefix (`bin/hive-babysitter-stub-git:66`); only `git grep`'s `-O` runs a pager, so the screen should be subcommand-scoped to avoid blocking valid reads.
- [ ] [silent-failure-hunter] The real-git handoff is unguarded (`bin/hive-babysitter-stub-git:103`): a set-but-invalid `HIVE_BABYSITTER_REAL_GIT` raises an uncaught `Errno::ENOENT` → raw backtrace + exit 1, asymmetric with the deliberate exit-127 unset-path (`:101`); rescue `SystemCallError` and exit 127 uniformly.

## Nit
- [ ] [silent-failure-hunter] `log_skip` rescues only `SystemCallError` (`bin/hive-babysitter-stub-git:7`); a non-SystemCallError write error (IOError/Encoding) would crash the skip path instead of skipping — consider `rescue StandardError` since the audit log is non-critical relative to the skip.
- [ ] [silent-failure-hunter] When the dangerous-option screen fires, the skip emits the same generic "skipped" message as the not-read-only path (`bin/hive-babysitter-stub-git:92-94`); a distinct warn would make the over-block false positives (the Medium items) diagnosable.
- [ ] [pr-test-analyzer] Test gaps mirror the screening gaps: no coverage for bundled `-nO`, two-token `--config-env KEY=VAL`, bare `--open-files-in-pager`, or `--output-indicator-*` passthrough — the suite validates only the canonical spellings the author matched, giving false confidence.
- [ ] [comment-analyzer] The comment says options are "screened across the entire invocation" (`bin/hive-babysitter-stub-git:52-55`) without noting the exact-token (`-c`/`--config-env`) vs prefix-match asymmetry; the load-bearing fact that exact-match `-c` is safe *only because git rejects bundled `-cKEY=VAL`* is undocumented for future maintainers.
- [ ] [comment-analyzer] The comment scopes `-O` to "(git grep)" (`bin/hive-babysitter-stub-git:59`) but the prefix match blocks `-O…`/`--output…` for every subcommand; note the over-block is intentional so a maintainer doesn't later narrow it to grep-only and reopen the hole.
- [ ] [pr-test-analyzer] The `stripped_global_options` `break if rest.length < 2` underflow guard (`bin/hive-babysitter-stub-git:18`) has no dedicated test (e.g. trailing bare `-c`/`-C`).

## Strengths (no action)
- [code-reviewer][silent-failure-hunter] `log_skip` silent `rescue … nil` → `warn` removes a real silent-failure of the audit log.
- [comment-analyzer] The whole-argv scan (vs the stripped `rest`) is the correct design and the comment's "(including the leading global options that `stripped_global_options` discards)" parenthetical documents exactly why; personas empirically reproduced the `-c diff.external`, `grep -O`/`--open-files-in-pager`, and `--output` vectors the fix closes.
- [code-reviewer] Verified git rejects the attached `-c<key>=<val>` form, so exact-token `-c` matching is sufficient for that vector (the residual `-nO` bypass is grep-specific).

PR #316, opened 2026-06-05
Hive patrol: Dry-run git stub allows executable read-command options

ivankuznetsov:
## Summary

Hardens the babysitter dry-run `git` stub (`bin/hive-babysitter-stub-git`) so that read-shaped git invocations can no longer be weaponized into arbitrary command execution. Patrol finding `command-bin-hive-babysitter-stub-gh-1` showed the stub stripped global options to classify a command, then exec'd the original argv whenever the remaining subcommand was allowlisted — letting a prompt-injected agent run shell via config/option/env injection on an otherwise read-only command, with nothing logged as skipped.

The change closes that side-effect boundary on three fronts:

- **Argv config overrides** — rejects *all* leading `-c` / `-c<glued>` / `--config-env` / `--config-env=` overrides (allowlist, not a denylist of dangerous keys), so exec-capable config such as `diff.external`, `diff.<driver>.textconv`, `filter.<driver>.clean`, `core.fsmonitor`, and `gpg.<format>.program` can never reach real git.
- **Executable command options** — blocks `--paginate`/`-p` (scoped to the leading global region so a read-only subcommand `-p` like `git log -p` still passes through), `--exec-path`, `--ext-diff`, `--open-files-in-pager`/`-O`, and the file-write options `--output`/`-o` (separate, glued, and short forms). The `-o`/`-O` guard is subcommand-scoped so legitimate reads (`git grep -o`, `git ls-files -o`, `git diff -O<orderfile>`) still pass.
- **Environment-variable injection** — fails closed (`unsafe_env?`) when the equivalent env channels are set: `GIT_EXTERNAL_DIFF`, `GIT_SSH_COMMAND`/`GIT_SSH`, `GIT_PROXY_COMMAND`, `GIT_CONFIG_COUNT`/`GIT_CONFIG_PARAMETERS`, and `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`. This was the threat-model decision settled in pass-2 triage (treat the env path as in scope for the dry-run boundary).

Unsafe invocations are logged and skipped rather than exec'd; wiki docs (`wiki/commands/babysit.md`, `wiki/modules/babysitter.md`) were reconciled to describe the actual guard scope.

## Test plan

- `ruby -Itest test/unit/babysitter/dry_run_env_test.rb` — 2 runs, 438 assertions, 0 failures.
- Regression coverage added in `test/unit/babysitter/dry_run_env_test.rb` asserts both directions: exploit forms (config injection via `-c`/`--config-env`/env vars, exec options, file-write options) are skipped *and* never reach real git (argv absent from the recorded `real.log`); legitimate read forms (`git log -p`, `git show -p`, `git grep -o`, `git ls-files -o`) still pass through.

## Review summary

Three CE + three PR-toolkit review passes plus three triage rounds. All findings were auto-fixed or resolved; one threat-model question (env-var config injection in scope for the dry-run boundary) was escalated and answered fail-closed in `escalations-02.md`, then applied across commits `ccfa153f` and `eba2c2b7`. Several review findings against `schemas/`, TUI, archive, and `lib/hive/stages/review.rb` were confirmed phantom-diff artifacts from a stale `origin/main` merge-base (`598fd191`) and do not belong to this branch.

## Linked task

`/home/asterio/Dev/hive/.hive-state/stages/6-review/patrol-command-bin-hive-babysitter-stub-gh-5031d524`

Patrol finding: `command-bin-hive-babysitter-stub-gh-1` (fingerprint `5031d52458a163d25c01f46e1a0fdae19cf3b881f10be239362f5bcd350bcdab`)


ivankuznetsov:

# Codex CE Code Review 01

No plan was found in the task folder; review is grounded in `git diff origin/main..HEAD` and the project wiki.

## High

## Medium
- [ ] `hive-status` removes the required `folder_mtime` task field without a schema-version bump: `Status#task_payload` no longer emits the key while `SCHEMA_VERSIONS["hive-status"]` stays at 3, so same-version JSON consumers that cached or generated against v3 can silently receive a narrower contract.

## Nit

ivankuznetsov:

# Code Review — patrol/command-bin-hive-babysitter-stub-gh (pass 1)

**No `plan.md` found in the task folder** — reviewed without plan grounding.

**Scope note:** The branch has exactly one real commit, `90483b1a fix(patrol):
Dry-run git stub allows executable read-command options`. The `origin/main..HEAD`
diff *also* shows ~1000 lines of archive-feature deletions — that is noise: the
merge-base is `598fd191`, `origin/main` independently advanced with the archive
feature (#301), and this branch simply doesn't have it. Those deletions are **not
changes this branch made** and a normal merge will not drop the archive feature.
Review focused on the actual commit: `bin/hive-babysitter-stub-git` + its test.

Findings below were confirmed by running the stub directly against a recording
fake `HIVE_BABYSITTER_REAL_GIT`.

## High
- [ ] `-O`/`-O<pager>` short form of `--open-files-in-pager` bypasses the new guard (`bin/hive-babysitter-stub-git:63-64`): `unsafe_command_options?` only matches the long `--open-files-in-pager[=...]`. Confirmed: `git grep -O'touch /tmp/x' needle` passes straight through to real git, executing the pager — defeating the exact RCE class this commit set out to close.

## Medium
- [ ] `-p`/`--patch` on read commands is wrongly skipped (`bin/hive-babysitter-stub-git:48`): `unsafe_global_options?` scans the *entire* argv and treats any `-p` as global `--paginate`. Confirmed: `git log -p`, `git log -p --oneline`, and `git show -p HEAD` are all skipped, even though `read_only?` classifies `log`/`show` as fully read-only. The stub now silently returns success with empty output for legitimate read commands. Root cause: globals should only be inspected in the pre-subcommand prefix (which `stripped_global_options` already isolates), not across the whole argv.
- [ ] Denylist misses `core.sshCommand` / `credential.helper`, reachable via the allowed `git remote show <url>` (`bin/hive-babysitter-stub-git:29-43`, `73-79`): `remote show origin` without `-n` contacts the remote, invoking sshCommand / a `!`-prefixed credential helper. Confirmed both `git -c core.sshCommand='…' remote show origin` and `git -c 'credential.helper=!…' remote show origin` pass through to real git — same executable-config class as `diff.external`, just not enumerated.
- [ ] Env-var config injection bypasses the argv-only denylist entirely (`bin/hive-babysitter-stub-git`, `exec` at EOF): `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_n`/`GIT_CONFIG_VALUE_n`, `GIT_EXTERNAL_DIFF`, `GIT_SSH_COMMAND`, `GIT_PAGER` are passed through untouched. Confirmed `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=diff.external GIT_CONFIG_VALUE_0='…' git diff` reaches real git. Severity depends on threat model (env is set by the babysitter, argv may carry untrusted data) — at minimum document this as an accepted limitation, since the `-c` hardening is meaningless if the equivalent env path is open.

## Nit
- [ ] Denylist-over-allowlist is structurally fragile here: each finding above is a sibling of an enumerated key/flag. Consider inverting to an allowlist of known-safe config keys/options for the read commands actually invoked, so new git executable-config knobs fail closed.
- [ ] Bare `--exec-path` (no value) is blocked though it merely prints the exec path and is read-only (`bin/hive-babysitter-stub-git:48`); harmless given no caller uses it, but it's over-broad.
- [ ] New behavior is untested for the false-positive (`git log -p`) and the bypasses (`-O`, `core.sshCommand` via `remote show`); the added tests only cover the long-form vectors that already pass.

ivankuznetsov:

No `plan.md` was found in the task folder; review proceeds without plan grounding.

**Diff-base note:** the requested `git diff origin/main..HEAD` is misleading — `origin/main` is 4 commits ahead of this branch's merge-base (`598fd191`), so ~1600 lines of "deletions" (Archive #301, TUI fixes #314/#326, schema v3) are divergence artifacts this branch never authored. All personas were scoped to the real change, `git diff 598fd191..HEAD`: `bin/hive-babysitter-stub-git` (dry-run git stub hardening), `test/unit/babysitter/dry_run_env_test.rb`, and wiki docs. Personas run: `code-reviewer`, `pr-test-analyzer`, `silent-failure-hunter`, `comment-analyzer`. Findings below are the union, deduped. Key security claims were verified empirically (file-write bypass, exec-vector passthrough, `-p` skip, rubocop offense).

## High
- [ ] [code-reviewer] `--output=<file>` (and `-o`) arbitrary file-write bypass on allowed read-only `diff`/`log`: `git diff --output=/tmp/pwn` and `git log -p --output=/tmp/pwn` pass `read_only?` and aren't caught by `unsafe_command_options?` — confirmed empirically writing a 2.2KB file, defeating the no-mutation dry-run boundary; add `--output`/`--output=`/`-o` to the unsafe command options (`bin/hive-babysitter-stub-git:62-65,104`).
- [ ] [code-reviewer][pr-test-analyzer] `-p` misclassified as global `--paginate`: `unsafe_global_options?` scans the entire argv positionally, so the read-only subcommand flag `-p` in `git log -p` / `show -p` / `diff -p` / `grep -p` is wrongly skipped — confirmed empirically; scope the `-p`/`--paginate`/`--exec-path` scan to the leading global-option region only (`bin/hive-babysitter-stub-git:44-46`).
- [ ] [code-reviewer][pr-test-analyzer] Config denylist misses same-class exec vectors: `diff.<driver>.textconv`, `diff.<driver>.command`, `filter.<driver>.clean/smudge/process`, `gpg.<format>.program` (e.g. `gpg.ssh.program`), plus `core.sshcommand`/`core.alternaterefscommand` all execute under allowed subcommands and pass straight through (confirmed) — the exact-match/`start_with?` list can't catch subsection-keyed names; match by final key segment instead (`bin/hive-babysitter-stub-git:30-42`).

## Medium
- [ ] [pr-test-analyzer] No positive regression test that legitimate `-p` read commands (`git log -p`, `git show -p`) still pass through — the live `-p` skip regression is entirely unguarded; an `assert_passes env, "git", "log", "-p"` would fail today (`test/unit/babysitter/dry_run_env_test.rb`).
- [ ] [pr-test-analyzer] Exec-vector bypasses (`diff.<driver>.textconv`, `filter.<driver>.clean`, `gpg.ssh.program`) are untested, so the intended security boundary is not pinned down and the unsafe-by-default behavior is invisible to CI (`test/unit/babysitter/dry_run_env_test.rb`).
- [ ] [pr-test-analyzer] Separate-argument parsing paths untested: the `-c KEY value` branch (`argv[index+1]`), the `--config-env <name>` separate-arg branch, and `--ext-diff` (half of `unsafe_command_options?`) have zero coverage; several denylist keys (`core.hookspath`, `git.pager`, `gpg.program`, and the `alias.`/`pager.`/`protocol.`/`include.` prefixes) are also unexercised (`bin/hive-babysitter-stub-git:50-65`).
- [ ] [comment-analyzer] Wiki says the stub blocks "global pager/exec-path switches," but the implementation matches `-p` argv-wide and also skips the subcommand flag (`git log -p`); the word "global" misdescribes the scope and will surprise a maintainer (`wiki/commands/babysit.md:70`, `wiki/modules/babysitter.md:67`).
- [ ] [comment-analyzer] Wiki's enumerated denied-key list (`diff.external`, `core.fsmonitor`, pager, hook, alias, protocol, include, `gpg.program`) reads as an exhaustive taxonomy of executable-affecting config, but same-class keys (`diff.<driver>.textconv`, `filter.<driver>.*`, `gpg.ssh.program`) are NOT blocked — frame the list as illustrative or note the gap to avoid a false sense of completeness (`wiki/commands/babysit.md:70`).
- [ ] [silent-failure-hunter] Read-only commands wrongly classified unsafe return empty stdout + `exit 0`, indistinguishable from a real empty result — a caller running `git log -p` to read history silently gets nothing and a success status; the only signal is a stderr warn + best-effort log (`bin/hive-babysitter-stub-git:104-108`).

## Nit
- [ ] [code-reviewer] New rubocop offense introduced at `bin/hive-babysitter-stub-git:46` (`Layout/SpaceInsideArrayLiteralBrackets`, autocorrectable) — CLAUDE.md requires running linters after edits; lines 69/79 offenses are pre-existing and out of scope.
- [ ] [silent-failure-hunter] Every skip exits `0`, carrying no skip information in exit status — intentional for the dry-run "deception" design, but now also fires for the false-positive read-only skips above, so consumers not reading the log file are silently misled (`bin/hive-babysitter-stub-git:107`).
- [ ] [silent-failure-hunter] `log_skip` rescues `SystemCallError` and returns nil — if the log path is unwritable the skip is recorded nowhere but ephemeral stderr, and the rescue doesn't re-emit that logging itself failed (`bin/hive-babysitter-stub-git:7-8`).
- [ ] [silent-failure-hunter] A bare trailing `-c`/`--config-env` with no following value yields an empty/nil key and is treated as not-unsafe, falling through to normal `read_only?` classification rather than being distinctly handled (`bin/hive-babysitter-stub-git:50-51`).
- [ ] [comment-analyzer] Wiki pairs `--open-files-in-pager` with `grep` and lists `--ext-diff` as if subcommand-scoped, but the code matches both anywhere in the post-global args regardless of subcommand — minor phrasing mismatch (`wiki/commands/babysit.md:70`, `wiki/modules/babysitter.md:67`).

ivankuznetsov:

# Codex CE Code Review - Pass 02

No plan was found in the task folder; review is based on `git diff origin/main..HEAD` and repository/wiki context only.

## High
- [ ] `REVIEW_WAITING reason=reviewer_partial_failure` cannot be accepted by clearing the marker: `lib/hive/commands/run.rb:304` tells operators to clear `REVIEW_WAITING`, but on the next markerless run `pass_completion_status` treats the existing reviewer file plus `escalations-NN.md` as `:fix_incomplete`, skips reviewers/triage, sees the same `errors-NN.md`, and re-enters the same waiting state instead of advancing.

## Medium
- [ ] TUI panes no longer respect terminal height: `lib/hive/tui/bubble_model.rb:3164` and `lib/hive/tui/bubble_model.rb:3316` call pane renderers without any height, while `TasksPane.render` and `ProjectsPane.render` now emit every row, so ordinary long task/project lists can overflow the alt-screen and hide the footer/status controls.
- [ ] `hive-status` removes a required task field without a schema bump: `lib/hive/commands/status.rb:175` no longer emits `folder_mtime` and `schemas/hive-status.v3.json:173` removes it from the v3 contract while `Hive::Schemas::SCHEMA_VERSIONS["hive-status"]` remains `3`, breaking strict consumers that negotiated v3.

## Nit
- [ ] Bot copy for reviewer partial failures is misleading: `lib/hive/bot/notification_builders.rb:152` treats every non-guardrail `review_waiting` as review triage and offers Accept all/Reject all findings, but `reason=reviewer_partial_failure` is about reviewer infrastructure errors in `reviews/errors-NN.md`, not triageable findings.

ivankuznetsov:

# Claude CE Code Review 02

No `plan.md` was found in the task folder; review proceeds without plan grounding.

**Diff-base note:** `git diff origin/main..HEAD` is misleading here. `origin/main` (`b02c5fa`) is 4 commits ahead of this branch's merge-base (`598fd191`), so the raw diff renders ~1600 "deletions" (archive_filter, TUI panes, `schemas/hive-status.v*.json`, etc.) that this branch never authored — divergence artifacts from main, not this branch's work. **Findings are scoped to the real change** (`git diff 598fd191..HEAD`): two security commits on `bin/hive-babysitter-stub-git` + its unit test, plus wiki docs. Triage should disregard any phantom finding about `schemas/`, TUI, or archive code — those belong to main.

**Pass-1 follow-up:** The three substantive pass-1 findings were applied correctly in `6cc909d5`: (1) the global-option scan is now scoped to the leading region so a read-only subcommand `-p` (`git log -p`) is no longer misread as global `--paginate`; (2) the leaky config denylist was replaced with an allowlist that rejects *all* leading `-c`/`--config-env` overrides; (3) positive regression tests for `-p` passthrough were added. Tests green locally: `2 runs, 205 assertions, 0 failures`. The argv-side hardening looks correct.

The dominant open issue is the **environment-variable** counterpart, which the argv hardening does not address and which was already escalated in pass 1 (`escalations-01.md` Q1, A1 left unanswered).

## High
- [ ] Env-var config injection bypasses the entire argv hardening (same RCE class): the final pass-through to real git (`bin/hive-babysitter-stub-git:107`) inherits the full, unmodified environment, so an allowlisted read like `git diff` still runs arbitrary shell via `GIT_EXTERNAL_DIFF=<cmd>` or `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=diff.external GIT_CONFIG_VALUE_0=<cmd>` — confirmed empirically (real git invoked with hostile env intact, nothing logged as skipped). This is the exact vector the `-c`/`--config-env` allowlist was added to close, reached through an unguarded path; the dry-run side-effect-and-logging boundary is still broken. Note: this is a threat-model/scope decision already captured in `escalations-01.md` Q1 (unanswered), so it is user-gated here rather than auto-fixable — suggested fail-closed default is to skip the invocation when `GIT_CONFIG_*` resolve to an unsafe key or when `GIT_EXTERNAL_DIFF`/`GIT_SSH_COMMAND` are set.

## Medium
- [ ] `--paginate`/`-p` blocking is only a partial pager guard, which the docs overstate as closing the pager vector: git auto-spawns `core.pager`/`GIT_PAGER` for `log`/`diff` when stdout is a TTY even without `--paginate`, so on a TTY the pager-exec surface remains open via the env path (`GIT_PAGER`) regardless of the argv flag — same root cause as the High finding (full env inherited at pass-through). Worth confirming dry-run stdout is never a TTY, or folding the pager case into the env guard.

## Nit
- [ ] `wiki/gaps.md:48` (item 15) describes this pass's hardening as "Current uncommitted worktree changes further tighten the git stub…", but those changes are committed (`6cc909d5`); the phrasing will read as stale/inaccurate once merged — reword to reference the commit.
- [ ] `unsafe_command_options?` (`bin/hive-babysitter-stub-git:49`) catches separate `-o` and glued `--output=` but not a glued short form `-o<path>` (e.g. `-o/tmp/x`); low confidence git accepts that form for the allowlisted read subcommands, but if it does, it slips the file-write guard. Cheap to close with a `\A-o./` check mirroring the `--output=`/`-c.` patterns already used.

ivankuznetsov:

# PR Review Toolkit — Pass 02

No `plan.md` was found in the task folder; review proceeds without plan grounding.

**Diff-base note:** `git diff origin/main..HEAD` is contaminated — `origin/main` is 4 commits ahead of this branch's merge-base (`598fd191`) with unrelated work (archive feature, TUI panes, schemas), producing ~1600 phantom "deletions" this branch never authored. All personas reviewed the branch's real change only (`git diff 598fd191..HEAD`): `bin/hive-babysitter-stub-git` (dry-run git stub hardening), `test/unit/babysitter/dry_run_env_test.rb`, and wiki docs. Triage should disregard any finding about `schemas/`, TUI, or archive code.

Union of findings from `code-reviewer`, `pr-test-analyzer`, `comment-analyzer`, and `silent-failure-hunter`. Two new High/Medium RCE-class bypasses were confirmed empirically against the stub.

## High
- [ ] [security] `git grep -O<cmd>` / `git grep -O <cmd>` is not blocked, defeating the exact RCE class this change targets: `-O` is the documented short form of `--open-files-in-pager` (`bin/hive-babysitter-stub-git:49-57` only matches the long forms), and `grep` is allowlisted, so the stub runs real git with an attacker-controlled pager command. **Confirmed empirically** (real git invoked, nothing logged as skipped). Fix: reject `-O` and `\A-O.` alongside `--open-files-in-pager`.
- [ ] [security] Env-var config injection bypasses the whole argv guard: the pass-through to real git (`bin/hive-babysitter-stub-git:107`) inherits the full environment, so `GIT_EXTERNAL_DIFF=<cmd>`, `GIT_CONFIG_COUNT/_KEY_n/_VALUE_n`, `GIT_SSH_COMMAND`, `GIT_PAGER`, `GIT_CONFIG_GLOBAL` reach real git on an allowlisted read and achieve the same arbitrary execution the `-c` rejection closes. **Confirmed empirically.** Same threat model, unguarded path; already captured in `escalations-01.md` Q1 (unanswered) — user-gated. Fail-closed suggestion: skip when `GIT_CONFIG_*` resolve to an unsafe key or `GIT_EXTERNAL_DIFF`/`GIT_SSH_COMMAND` are set.
- [ ] [tests] `assert_stubbed` never proves the command was NOT run through real git — it asserts only `status.success?` and the stderr "skipped" line (`dry_run_env_test.rb:137-141`). A regression that logs the skip *and then still falls through to real git* would pass all ~30 skip cases; the negative (no-passthrough) check is the load-bearing one and is absent. Add an assertion that the command's argv is absent from `real.log`.
- [ ] [comments] `wiki/gaps.md:48` (item 15) calls this pass's hardening "**Current uncommitted worktree changes** further tighten the git stub…", but those changes are committed (`6cc909d5`, with the gaps.md edit committed alongside). Doc rot introduced by this change; a reader on HEAD will hunt for nonexistent uncommitted changes.

## Medium
- [ ] [code] Glued short `-o<path>` (e.g. `git diff -o/tmp/pwn`) bypasses the file-write guard: `unsafe_command_options?` (`bin/hive-babysitter-stub-git:49-56`) catches separate `-o` and `--output=` but not the glued short form. **Confirmed empirically** (real git invoked, nothing skipped) — defeats the no-mutation boundary on an allowed read. Add a `\A-o.` check mirroring the `--output=`/`-c.` patterns.
- [ ] [silent-failure] Catch-all `when /\A-/` (`bin/hive-babysitter-stub-git:40-41`) shifts only one token, so an unknown/future global option taking a separate-word value leaves the value as the apparent subcommand; if it matches an allowlisted read name, `read_only?` passes and the original argv (including the unrecognized global) is forwarded to real git verbatim — a fail-open seam. Theoretical for today's git (all other globals use `=` form) but it is the precise weak point of the gate.
- [ ] [silent-failure] `log_skip` (`bin/hive-babysitter-stub-git:4-9`) rescues `SystemCallError` and returns nil, so a failed write to the security-audit log (read-only FS, ENOSPC, bad `HIVE_BABYSITTER_DRY_RUN_LOG`) silently drops the audit record while the command is still skipped. The stderr `warn` (line 98) preserves the human-visible signal, but the persistent audit file develops invisible gaps; emit `warn "…failed to write skip log: #{e.message}"` instead of swallowing.
- [ ] [tests] The `-p` passthrough cases (`git log -p`, `show -p`, `diff -p`, `grep -p needle`, `dry_run_env_test.rb:89-92`) — the PR's headline regression target — are never confirmed to reach real git; only `status`/`config`/`remote` appear in the `real.log` block (lines 122-131). A `-p`→`--paginate` misclassification regression would still pass (`status.success?` holds either way). Add `real.log` assertions for the `-p` cases.
- [ ] [tests] Coverage gaps for the new option-parsing branches: glued `-o<path>` (the confirmed Medium bypass above) is untested, and the pathspec false-positive boundary (`git log -- -o`, `git grep --output` as a literal pathspec) is untested, so over-blocking regressions that silently drop legitimate reads go uncaught.
- [ ] [code] `-o` is treated as unsafe across all subcommands, over-skipping the legitimate read-only `git grep -o`/`--only-matching` (`bin/hive-babysitter-stub-git:52`). False positive (fails safe, not a hole) — scope the output-option check to subcommands that actually have `--output`, or exclude `grep`.

## Nit
- [ ] [comments] `bin/hive-babysitter-stub-git:25-26` (and the mirrored wiki text in `wiki/commands/babysit.md:70`, `wiki/modules/babysitter.md:67`) call deny-all-config-overrides "an allowlist rather than a denylist" — rejecting the whole `-c`/`--config-env` category is itself a categorical denylist; reword to "reject the whole config-override category instead of enumerating dangerous keys."
- [ ] [comments] `bin/hive-babysitter-stub-git:33` — the `-p`/`--paginate`/`--exec-path` unsafe branch has no inline note and `--exec-path` is undocumented at its definition site (only `-p` is mentioned in the function header lines 13-16); one line would help.
- [ ] [silent-failure] `bin/hive-babysitter-stub-git:7` — `rescue SystemCallError` discards the exception object; bind `=> e` so any future logging/telemetry retains the cause.
- [ ] [tests] Untested: mutating `remote` subcommands beyond set-url/add/remove (`rename`, `prune`, `update`, `set-head`, `set-branches`) and the `get-url --all` form; a careless `else true` refactor in `read_only_remote?` would go uncaught.
- [ ] [tests] Skip-detection across ~25 assertions is anchored to the exact `"[dry-run] <cmd> skipped"` literal — acceptable (the babysitter parses this contract) but a format change cascades widely.
- [ ] [code] Pre-existing rubocop offenses `Layout/SpaceInsideArrayLiteralBrackets` at `bin/hive-babysitter-stub-git:61` and `:71` are inside the untouched `read_only_remote?`; not introduced by this branch, autocorrectable if cleaned opportunistically.
- [ ] [docs] The wiki allowlist descriptions (`wiki/commands/babysit.md`, `wiki/modules/babysitter.md`) state `--open-files-in-pager` is blocked without noting the `-O` short-form gap; correct alongside the High `-O` fix so docs match reality.

## Strengths
- Region-scoped global-option parsing correctly distinguishes subcommand `-p` (`git log -p` passes) from global `--paginate` (`git -p log` skips); both glued and separate `-c`/`--config-env` forms are flagged.
- The "reject all config overrides" design (vs chasing a key denylist) is the right call and is documented with a concrete exec-capable-key list mirrored by exploit test cases.
- Empty-`HIVE_BABYSITTER_REAL_GIT` is handled explicitly with a clear message and distinct exit code (127) rather than guessing a path — correct fail-closed pattern.

ivankuznetsov:

# Codex CE Code Review 03

No plan was found in the task folder; review is grounded in `git diff origin/main..HEAD`, repo wiki notes, and adjacent code.

## High
- [ ] `REVIEW_WAITING reason=reviewer_partial_failure` resumes into Phase 4 instead of rerunning failed reviewers: `lib/hive/stages/review.rb:415` creates a waiting marker, but `resuming_from_waiting?` at `lib/hive/stages/review.rb:859` makes the next `hive run` skip reviewers/triage and spawn the fix agent with no accepted findings, so the documented "re-run hoping for recovery" path cannot work.
- [ ] `hive-status` v2/v3 remove the required `folder_mtime` field without a schema version bump: `lib/hive/commands/status.rb:182` stops emitting it while `schemas/hive-status.v2.json` and `schemas/hive-status.v3.json` mutate existing versions, breaking consumers pinned to the published status contract.

## Medium
- [ ] TUI panes no longer clamp to terminal height or keep the selected row in view: `lib/hive/tui/bubble_model.rb:3163` calls panes without a height, and `lib/hive/tui/views/tasks_pane.rb:108` renders every row, so normal long task/project lists overflow past the footer and cursor navigation can move to off-screen rows.

## Nit

ivankuznetsov:

# Claude CE Code Review 03

No `plan.md` was found in the task folder; review proceeds without plan grounding.

**Diff-base note:** `git diff origin/main..HEAD` is misleading. `origin/main` (`b02c5fae`) is 4 commits ahead of this branch's merge-base (`598fd191`), so the raw diff renders ~1600 "deletions" (archive_filter, TUI panes, `schemas/hive-status.v*.json`, `lib/hive/stages/review.rb`, etc.) that this branch never authored — divergence artifacts from main (PRs #301/#313/#314/#326), not this branch's work. **Findings are scoped to the real change** (`git diff 598fd191..HEAD`): the security commits on `bin/hive-babysitter-stub-git` + its unit test `test/unit/babysitter/dry_run_env_test.rb`, plus wiki docs. Triage should disregard any finding about `schemas/`, TUI, archive, or `stages/review.rb` — those belong to main, not here. (The sibling `codex-ce-code-review-03.md` reviewed the phantom diff and its High/Medium findings target main's code, not this branch.)

**Pass-2 follow-up:** The pass-2 High (env-var config injection bypasses the argv hardening) was acted on in `ccfa153f`: `unsafe_env?` now fails closed when `GIT_EXTERNAL_DIFF`, `GIT_SSH_COMMAND`, or `GIT_CONFIG_COUNT>0` are set. That closes the three vectors named in the escalation. **However, the closure is incomplete** — the dominant new finding below is a confirmed, equivalent-impact bypass of the *same* category the commit claims to "reject wholesale."

Local verification: argv guards behave as intended (`git log -p` etc. pass through; `-c`/`--config-env`/`--paginate`/`--output`/`--ext-diff` skip). The findings below are the remaining gaps.

## High
- [ ] `GIT_CONFIG_PARAMETERS` bypasses the new env guard — confirmed RCE on an allowlisted read: `unsafe_env?` (`bin/hive-babysitter-stub-git:83-97`) checks `GIT_CONFIG_COUNT` but not `GIT_CONFIG_PARAMETERS`, git's older one-shot config channel that injects the same exec-capable keys. Verified empirically: `GIT_CONFIG_PARAMETERS="'core.fsmonitor=touch …/PWNED'" git status` (an allowlisted subcommand) executed the command and created the file. This is the exact `diff.external`/`core.fsmonitor`/alias category the `-c` allowlist and `GIT_CONFIG_COUNT` guard exist to close, reached through an unguarded sibling env var — the comment's claim to "reject the whole env-config-override category" is not yet true. Fail-closed fix: also skip when `GIT_CONFIG_PARAMETERS` is non-empty.

## Medium
- [ ] Parallel exec-via-command / config env vectors still open beside the ones guarded: `GIT_SSH` (older sibling of the now-guarded `GIT_SSH_COMMAND` — names a program git execs) is reachable through the allowlisted network read `git remote show origin` (`read_only_remote?`, `bin/hive-babysitter-stub-git:103-120`); `GIT_PROXY_COMMAND` (git:// transport) is the same class; and `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` can repoint config to an attacker-written file carrying `diff.external`. These are the same family the commit set out to close; guarding only `GIT_SSH_COMMAND`/`GIT_CONFIG_COUNT` leaves the category half-open.
- [ ] Output-file guard over-blocks legitimate reads beyond `grep`: `unsafe_read_option?` (`bin/hive-babysitter-stub-git:68-81`) flags any `-o*`/`-O*` for every non-`grep` subcommand, but `git ls-files -o` (`--others`, lists untracked — a read) and `git diff|log|show -O<orderfile>` (`--output-ordering`, reads an orderfile) are read-only. Verified: `git ls-files -o` is now skipped. Fails safe (skips a read rather than allowing a write), but silently breaks real read-only commands the babysitter may rely on; the `grep` scope-out alone is too narrow.
- [ ] Branch forks from a stale main: `origin/main..HEAD` shows PRs #301 (archive)/#313/#314/#326 as deletions because the merge-base is 4 commits behind `origin/main`. Rebase onto current main before merge so the PR diff reflects only the babysitter change and CI validates against current code, not a 4-PR-old tree.

## Nit
- [ ] Comment overstates `-O` semantics: `unsafe_read_option?` (`bin/hive-babysitter-stub-git:69-71`) says `-O` "runs an arbitrary pager command" — true only for `git grep` (`--open-files-in-pager`); for `diff`/`log`/`show`, `-O` is `--output-ordering` (an orderfile read). Since the universal `-O` block over-blocks those reads (see Medium above), align the comment with the actual scope once the block is narrowed.
- [ ] `unsafe_env?` comment (`bin/hive-babysitter-stub-git:84-91`) asserts it rejects "the whole env-config-override category rather than enumerating dangerous keys" — inaccurate while `GIT_CONFIG_PARAMETERS` (and `GIT_CONFIG_GLOBAL/SYSTEM`) remain unguarded. Tighten the prose when the High finding is closed so the comment doesn't overstate coverage.

ivankuznetsov:

# PR Review Toolkit 03 (multi-persona)

No `plan.md` was found in the task folder; review proceeds without plan grounding.

**Diff-base note:** `git diff origin/main..HEAD` is misleading — `origin/main` (`b02c5fae`) is 4 commits ahead of this branch's merge-base (`598fd191`), so it renders ~1600 phantom deletions (archive, TUI panes, `schemas/`, `lib/hive/stages/review.rb`) this branch never authored. All personas were scoped to the **real** change (`git diff 598fd191..HEAD`): `bin/hive-babysitter-stub-git`, `test/unit/babysitter/dry_run_env_test.rb`, and wiki docs.

Personas run: **code-reviewer**, **pr-test-analyzer**, **comment-analyzer**, **silent-failure-hunter**. (type-design-analyzer skipped — no new types; the change is a procedural Ruby gate script.) Findings below are the union; multiple personas independently confirmed the `GIT_CONFIG_PARAMETERS` RCE bypass by execution.

## High
- [ ] [code-reviewer] `GIT_CONFIG_PARAMETERS` env config-injection unguarded (`bin/hive-babysitter-stub-git:92-94`): confirmed RCE — `GIT_CONFIG_PARAMETERS="'diff.external=touch /tmp/PWN'" git diff` (an allowlisted read) executed the command; same exec-capable category the `-c`/`GIT_CONFIG_COUNT` guards close.
- [ ] [pr-test-analyzer] No test pins the `GIT_CONFIG_PARAMETERS` bypass (`dry_run_env_test.rb:99-103`): the env block only exercises the 3 guarded vars, so the gap is invisible; an `assert_stubbed env.merge("GIT_CONFIG_PARAMETERS"=>…)` would fail today and force the fix.
- [ ] [pr-test-analyzer] `GIT_SSH` / `GIT_PROXY_COMMAND` exec env vars unguarded: confirmed `GIT_PROXY_COMMAND=/tmp/evil git status` reached real git; only the newer `GIT_SSH_COMMAND` is blocked, leaving its older siblings open and untested.
- [ ] [pr-test-analyzer] `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` redirection unguarded: confirmed `GIT_CONFIG_GLOBAL=/tmp/evil.gitconfig git status` reached real git, re-enabling `diff.external`/`core.pager`/aliases the gate exists to block; untested.
- [ ] [silent-failure-hunter] `log_skip` rescue catches only `SystemCallError`, not `IOError` (`bin/hive-babysitter-stub-git:7`): an `IOError` write/flush failure (e.g. "closed stream") escapes the handler, crashes the stub (exit 1), and silently drops the audit record — defeating the handler's stated intent.

## Medium
- [ ] [code-reviewer] `-o`/`-O` guard over-blocks legitimate reads (`bin/hive-babysitter-stub-git:73,80`): `git ls-files -o` (`--others`, a read) is skipped while long-form `--others` passes, and `git diff/log/show -O<orderfile>` (`--output-ordering`, a read) is blocked; only `grep` is scoped out.
- [ ] [pr-test-analyzer] No test for the `git ls-files -o` over-block (`dry_run_env_test.rb`): `assert_passes env,"git","ls-files","-o"` fails today — the same over-block class the `grep -o`/`-p` tests guard, missed for another subcommand.
- [ ] [pr-test-analyzer] `git ls-files --output` (a real write) vs `-o`/`--others` (a read) boundary untested: needed once the over-block is fixed so the write form isn't re-allowed.
- [ ] [pr-test-analyzer] `assert_passes` only asserts exit 0, not passthrough (`dry_run_env_test.rb:192-195`): the empty-env case (line 126) has no `real.log` cross-check, so a "skipped-but-exit-0" regression would pass; fold the `real.log` passthrough check into `assert_passes` as `assert_stubbed` self-contains its `refute`.
- [ ] [silent-failure-hunter] `exec` failure on a set-but-invalid `HIVE_BABYSITTER_REAL_GIT` (`bin/hive-babysitter-stub-git:151`) leaks a raw Ruby backtrace + exit 1 instead of the friendly exit-127 handling used two lines above for the unset case; wrap `exec` in a `rescue SystemCallError`.
- [ ] [comment-analyzer] `unsafe_env?` comment (`bin/hive-babysitter-stub-git:88-91`) claims it rejects "the whole env-config-override category rather than enumerating dangerous keys" — false; it is a 3-item denylist with the `GIT_CONFIG_PARAMETERS` gap. Either close the gap or soften to "known exec-capable vars (not exhaustive)."
- [ ] [comment-analyzer] `-O` comment (`bin/hive-babysitter-stub-git:69-71`) says it "runs an arbitrary pager command … Unsafe everywhere" — true only for `git grep`; on diff/log/show `-O` is the read-only `--output-ordering`, so the rationale is wrong for 3 of the subcommands it fires on.
- [ ] [comment-analyzer] `wiki/modules/babysitter.md` repeats the "whole config-override category" env overstatement while `GIT_CONFIG_PARAMETERS` is unguarded.
- [ ] [comment-analyzer] `wiki/commands/babysit.md` echoes both the "reject the whole category" env claim and the pager-only description of `-O`.

## Nit
- [ ] [code-reviewer] `GIT_CONFIG_COUNT` non-numeric values treated as 0/allowed (`bin/hive-babysitter-stub-git:94`): safe today only because git itself rejects bogus counts; the gate leans on git's behavior rather than its own default-deny.
- [ ] [silent-failure-hunter] Skip-log failure warning (`bin/hive-babysitter-stub-git:11`) omits the target path, making it less self-sufficient for debugging which log location (`HIVE_BABYSITTER_DRY_RUN_LOG` vs default) failed.
- [ ] [comment-analyzer] Test comment (`dry_run_env_test.rb:~75-77`) states `-O` pager behavior as universal though only `grep` is exercised; scope the wording to grep or add a `git diff -O<orderfile>` passthrough/skip test to pin cross-subcommand behavior.

PR #319, opened 2026-06-05
Hive patrol: hive-eval --scenario can escape the scenario directory

ivankuznetsov:
## Hive Patrol Finding

Slice: `command-bin-hive-eval`
Category: `bug`
Severity: `medium`
Confidence: `high`
Fingerprint: `62b712629ab9e484001a59054c85420aac3b665968d5a8097081c5e3aa2081e8`

`bin/hive-eval` advertises `--scenario NAME` as a scenario basename and the wiki says the runner executes only scenario files, but `scenario_path` accepts both arbitrary `.rb` paths and unsanitized names before passing the result to Rake via `TEST`. For example, `--scenario ../support/reporter` resolves to `test/eval/support/reporter_test.rb`, and an absolute `/tmp/outside_test.rb` is also accepted. That lets a caller run non-scenario tests or arbitrary Ruby test files through the eval command, so the generated report/exit status no longer represents the eval scenario contract.

## Evidence

- bin/hive-eval:29 return File.expand_path(value, root) if value.end_with?(".rb")
- bin/hive-eval:31 basename = value.sub(/_test\z/, "")
- bin/hive-eval:46 env["TEST"] = path

## Applied Fix

Restrict `--scenario` to safe basenames under `test/eval/scenarios` for normal CLI use, or add an explicit separate test-only option for fixture paths. Reject path separators, absolute paths, and `..` components before constructing the scenario path.

## Validation

## Diffstat

```text
 bin/hive-eval                      | 14 ++++++++++++--
 test/eval/support/reporter_test.rb | 39 ++++++++++++++++++++++----------------
 2 files changed, 35 insertions(+), 18 deletions(-)

```


ivankuznetsov:

# Codex CE Code Review 01

## High

## Medium
- [ ] Missing regression for rejected `--scenario` path input: `bin/hive-eval` now rejects path-like values at lines 30-33, but `test/eval/support/reporter_test.rb` only covers the allowed basename path, so a future rollback of the containment guard could pass this suite while reopening arbitrary Ruby file execution.

## Nit

ivankuznetsov:

# Code Review — patrol-command-bin-hive-eval-62b71262 (Pass 1)

**No plan.md found in the task folder.** Reviewed without plan grounding.

**Scope:** `git diff origin/main..HEAD` — 2 files, +35/−18.
- `bin/hive-eval` — `--scenario` now rejects path-like input (path-traversal fix).
- `test/eval/support/reporter_test.rb` — reporter failure fixture rewritten to use a basename-resolved scenario instead of an absolute `.rb` path.

**Verification performed (read-only):** `ruby -c` passes on both files; reproduced the validation logic against `../../../../etc/passwd`, `/tmp/foo.rb`, `foo/bar`, `../s1_status` (all REJECT) and `s1_status` (accepted). The core fix is correct and effective.

## High

_None._

## Medium
- [ ] Security fix has no dedicated regression test: `bin/hive-eval` now rejects path-like `--scenario` values with exit 64, but the only test (`reporter_test.rb`) exercises the happy path with a valid basename. No test asserts that `../outside`, `/tmp/foo.rb`, or `foo/bar` exit 64 without setting `TEST` — a future regex relaxation could silently reintroduce the escape. (Already logged in `wiki/gaps.md:57`.)
- [ ] Test fixture is written into the source-controlled, suite-globbed `test/eval/scenarios/` directory rather than a tmpdir: the old approach was crash-safe (tmpdir + absolute path). The new approach relies on `ensure File.delete` for cleanup, which a hard kill (SIGKILL/SIGSEGV) bypasses, leaving an always-failing `zz_reporter_failure_<pid>_test.rb` that the next full `rake test:eval` run would pick up and fail on. The basename-only contract forces this trade-off, but it is a real durability regression worth flagging.

## Nit
- [ ] `ArgumentError` message (`bin/hive-eval:32`) says "must be a scenario basename, e.g. s1_status" but does not mention the accepted `_test` suffix form (`s1_status_test`), which the code explicitly strips and `wiki/testing.md` documents.
- [ ] Fixture uniqueness keys on `Process.pid` only (file + class name): within one host concurrent runs differ, but a leftover file from a crashed earlier run with a since-reused pid could collide. The `zz_` prefix and `ensure` cleanup make this low-probability; consider a stale-fixture sweep (`Dir.glob("zz_reporter_failure_*")`) before write for extra safety.
- [ ] Removal of the `value.end_with?(".rb")` absolute-path branch is a behavioral break, but the only prior caller was this same test and `wiki/testing.md` only documents basenames — no external impact found.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — patrol-command-bin-hive-eval-62b71262 (Pass 1)

**No plan.md found in the task folder.** Reviewed without plan grounding.

**Scope:** `git diff origin/main..HEAD` — `bin/hive-eval` (path-traversal hardening of `--scenario`) and `test/eval/support/reporter_test.rb` (reporter failure fixture rewritten to basename resolution).

**Personas run:** code-reviewer, pr-test-analyzer (tests), silent-failure-hunter (errors), comment-analyzer (comments). Union of findings below.

## High
- [ ] [tests] New rejection branch (`bin/hive-eval:27-36`) is entirely untested: no test invokes `--scenario` with `../something`, an absolute path, or `foo/bar` and asserts rejection — the core security behavior this PR exists to add ships with zero behavioral coverage, so a regression loosening the regex would pass green. The repo already has the idiom (`test/e2e/lib/hive_e2e_binary_test.rb` `test_replay_rejects_traversal_components`).
- [ ] [tests] Exit-code contract (64 = EX_USAGE) for rejection is unverified (`bin/hive-eval:44-49`): existing reporter tests only assert `status.success?`/`refute`, nothing pins `exitstatus == 64` or the `hive-eval: --scenario must be a scenario basename` stderr, despite the project convention of branching on exit 64 as USAGE.

## Medium
- [ ] [code-reviewer] Fixture is written into the tracked `test/eval/scenarios/` dir (not a tmpdir) and is not gitignored (`reporter_test.rb:30-66`): a SIGKILL/SIGINT between `File.write` and the `ensure` cleanup orphans `zz_reporter_failure_<pid>_test.rb`, which a full-glob `rake test:eval` run (`Rakefile:84`, no `TEST=`) then loads as a real always-failing test — the old `Dir.mktmpdir` approach could never leak into the scenario glob.
- [ ] [tests] PID-based naming reduces but does not eliminate collision and orphans are never swept (`reporter_test.rb:31,38`): PIDs are reused across hosts/containers and after a crash; a `Dir.glob("zz_reporter_failure_*_test.rb")` sweep in `setup`/`ensure` would close the gap since the orphan sits inside the real scenario directory.
- [ ] [errors] `File.write` of the fixture (`reporter_test.rb:35`) trades the previous `Dir.mktmpdir` self-cleanup for cleanup-only-on-graceful-exit, mutating a tracked source directory — a robustness regression in cleanup even though no error is swallowed.
- [ ] [comments] Comment-rot risk (`reporter_test.rb:24-30`): the rationale never explains *why* the fixture must now live in `scenarios/` rather than a tmpdir — that move is the load-bearing consequence of the `--scenario` basename validation, so a future reader could "simplify" it back to a tmpdir and silently break the test.
- [ ] [comments] ArgumentError message is incomplete/misleading (`bin/hive-eval:32`): "must be a scenario basename, e.g. s1_status" omits that the `_test` suffix is also accepted and does not convey the real rule (no path separators / no extension), which is the whole point of the hardening; e.g. "no path separators or extension".
- [ ] [code-reviewer] Dropping the `value.end_with?(".rb")` branch is a deliberate breaking change to the `--scenario PATH.rb` form (`bin/hive-eval:27-36`): no in-repo callers remain, but external scripts or muscle-memory invocations passing a `.rb` path now fail with exit 64 — flagged as a contained breaking-change concern, also worth a regression test (`--scenario /tmp/evil.rb` → exit 64).

## Nit
- [ ] [code-reviewer] Regex permits a leading `-` (`bin/hive-eval:31`): `--scenario -foo` → `-foo_test`; harmless (no shell, only `File.join`) but `\A[A-Za-z0-9][A-Za-z0-9_-]*\z` would avoid odd basenames.
- [ ] [errors] `rescue ArgumentError` (`bin/hive-eval:44-49`) is correctly narrow but assumes only the hand-raised `ArgumentError` can occur; a non-String `options[:scenario]` (not possible via OptionParser today) would surface as an unhandled crash — defense-in-depth note only.
- [ ] [errors] `File.delete` in the `ensure` (`reporter_test.rb:65`) could raise an `Errno` (EACCES/EBUSY) that would replace the in-flight test-assertion failure, masking the real failure — unlikely for a just-written local file.
- [ ] [tests] The test exercises only the *accept* path of the guard (`reporter_test.rb:53`): `scenario_name` happens to match the basename regex, an undocumented coincidence — if the prefix gained a `.`/`/` the test would fail at the guard (exit 64) not the reporter assertion; add a one-line comment that the name must satisfy `[A-Za-z0-9_-]+`. The `_test`-suffix stripping (`bin/hive-eval:30`) is also untested.
- [ ] [comments] "scenario-scoped fixture" wording (`reporter_test.rb:27`) is accurate but opaque; "written into the scenarios directory" would be clearer.
- [ ] [comments] Help text (`bin/hive-eval:16`) and the ArgumentError message (`:32`) are now two copies of "basename, e.g. s1_status" that can drift independently if the accepted format is clarified.

**Positive observations:** allowlist `[A-Za-z0-9_-]+` correctly rejects `../`, slashes, and `.rb` paths (verified empirically); `exit 64` is the right USAGE code; the rescue is user-actionable with no silent failure; switching the test to `--scenario <basename>` exercises the real production code path; reporter assertions test output shape not implementation; `wiki/testing.md:125` already documents the new contract.

ivankuznetsov:

# Code Review — claude-ce-code-review (pass 02)

**Plan grounding:** No `plan.md` was found in the task folder; this review is ungrounded against any plan and judges the diff on its own merits.

**Scope:** `git diff origin/main..HEAD` — hardens `bin/hive-eval --scenario` against path traversal (basename-only regex, exit 64 on reject), reworks `reporter_test.rb` failure-path fixture, and refreshes wiki pages.

**Verification performed:** Read full `bin/hive-eval` and `reporter_test.rb`; confirmed the `test:eval` rake task globs `test/eval/scenarios/**/*_test.rb` under `HIVE_EVAL_SCENARIOS_ONLY=1` (why the fixture must live there); traced the regex `/\A[A-Za-z0-9][A-Za-z0-9_-]*\z/` against all 7 rejection inputs (no bypass — `\z` not `\Z`, no `.` `/` `\\`, so `File.join` output stays contained); ran `rake test:eval TEST=test/eval/support/reporter_test.rb` → 3 runs, 30 assertions, 0 failures. The production fix is correct and the rejection regression is meaningful. No High findings.

## High

(none)

## Medium

- [ ] `reporter_test.rb` `setup` sweeps orphans with an unguarded `File.delete(stale)` (lines 16-18): if the file disappears between `Dir.glob` and `File.delete` (concurrent same-tree run, or another process's teardown), `File.delete` raises `Errno::ENOENT` and errors *every* test in the class before it runs — inconsistent with this same diff's carefully `rescue SystemCallError`-guarded `ensure` cleanup; wrap the sweep delete the same way.

## Nit

- [ ] The orphan sweep deletes by wildcard (`zz_reporter_failure_*_test.rb`) rather than the current run's pid, so a concurrent same-tree run's live pid-scoped fixture would be deleted out from under it — defeating the pid-scoping the fixture name was chosen to provide; acceptable since concurrent same-tree runs are unsupported, but worth a comment acknowledging it.
- [ ] Writing a runnable always-failing fixture into the source-controlled `test/eval/scenarios/` dir leaves a window where an unscoped `rake test:eval` (no `TEST=`) invoked concurrently picks it up as a real failing scenario; mitigated by `setup` sweep + `ensure`, but a gitignored scenario subdir or env-overridable scenario root would remove the shared-state hazard entirely instead of racing it.
- [ ] `SCENARIO_FORMAT_HINT` uses `.freeze` while the file has no `# frozen_string_literal: true` magic comment — harmless and intentional, just noting the lone manual freeze is a style outlier for the file.

ivankuznetsov:

# PR Review — pr-review-toolkit (pass 02)

**Plan grounding:** No `plan.md` was found in the task folder; this review is ungrounded against any plan and judges the diff on its own merits.

**Scope:** `git diff origin/main..HEAD` — hardens `bin/hive-eval --scenario` against path traversal (basename-only regex `/\A[A-Za-z0-9][A-Za-z0-9_-]*\z/`, resolves under `test/eval/scenarios/`, exit 64 on reject), reworks the failure-path fixture in `test/eval/support/reporter_test.rb`, and refreshes wiki pages.

**Personas run:** code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter. (type-design-analyzer skipped — no new types.) Findings are the union across personas, deduped. All personas confirmed the production hardening itself is correct and not bypassable, rubocop is clean, and the reworked tests pass. The `rescue ArgumentError` (bin/hive-eval:55-60) and guarded `ensure` (reporter_test.rb:114-118) were both assessed as *appropriate* error handling, not silent failures.

## High

(none)

## Medium

- [ ] [code-reviewer/test-analyzer/silent-failure] Unguarded `File.delete(stale)` in `setup` (`reporter_test.rb:16-18`): TOCTOU window between `Dir.glob` and `File.delete` (or an `EACCES`) raises in `setup` and errors *every* test in the class for a cleanup-only reason — inconsistent with the `rescue SystemCallError`-guarded `ensure` doing the identical delete a few lines down.
- [ ] [test-analyzer] Orphan-sweep blind spot under `HIVE_EVAL_SCENARIOS_ONLY=1` (`reporter_test.rb:8-19`): the CLI subprocess runs the `test/eval/scenarios/**` glob, which excludes `reporter_test.rb` (in `support/`), so `HiveEvalReporterTest#setup` never fires in that mode — yet an orphaned `zz_reporter_failure_*_test.rb` *is* loaded as a scenario and fails the suite, so the self-heal is weaker than the comment claims.
- [ ] [code-reviewer] Fixture written into source-controlled `test/eval/scenarios/` (`reporter_test.rb:77-78`) creates two concurrency races: a concurrent full-glob run during the write→delete window loads the always-failing fixture into an unrelated suite, and the cross-pid wildcard sweep would delete another run's in-flight fixture — latent flake if eval is ever parallelized or run concurrently in one worktree.

## Nit

- [ ] [code-reviewer] Orphaned `test/eval/scenarios/zz_reporter_failure_*_test.rb` is not gitignored: a hard-kill orphan lands untracked in a source-controlled dir and could be swept into a commit by `git add .`; a gitignore entry makes it harmless beyond the next-run sweep.
- [ ] [test-analyzer] No accept-path coverage for a valid basename containing `-`: the regex newly permits `-` in non-leading positions, but every accept test uses only `[A-Za-z0-9_]`, so a future tightening that dropped `-` would still pass green.
- [ ] [test-analyzer] Rejection assertion (`reporter_test.rb:36`) matches only the message prefix `hive-eval: --scenario must be a scenario basename`, not the "no path separators or extension" hint the PR claims `SCENARIO_FORMAT_HINT` keeps in sync.
- [ ] [test-analyzer] Empty-string `--scenario ""` is untested (`reporter_test.rb:28`): a distinct boundary from the path-like inputs and a plausible caller mistake (`--scenario "$VAR"` with unset var).
- [ ] [comment-analyzer] `bin/hive-eval:38-41` frames an allowlist regex as a denylist ("No `/`, `\`, `.` … permitted"): a future maintainer who adds a char to the class could read the comment as still-true while the guard loosened; reframe positively to match the regex's allowlist nature.
- [ ] [comment-analyzer] `reporter_test.rb:74-76` hardcodes a duplicated copy of the regex literal: an exact match today but a manual sync point that can silently drift; reference "bin/hive-eval's basename guard" by name instead.
- [ ] [silent-failure] `rescue ArgumentError` at `bin/hive-eval:55-60` is correctly narrow today, but if `scenario_path`'s body ever grows a call that raises `ArgumentError` for an unrelated reason it would be mislabeled as a `--scenario` format error; a custom `ArgumentError` subclass would make the contract airtight.

Issue #24, opened 2026-05-04
hive init: CLI hardening followups (rollback, --json, edge cases)


ivankuznetsov:
Cluster of 4 ce-code-review findings from PR #23 deferred for a focused follow-up. All four are CLI / lifecycle hardening that wants a fresh PR with proper design rather than being bolted onto the original feature commit.

## F2 (P2, manual): No rollback if a disk-side-effect step fails between `hive_state_init` and `register_project`

`Init#call`'s "abort leaves zero disk side effects" guarantee only covers `Prompts::Aborted` (user said `n`). If `write_per_project_config`, the master gitignore commit, or the global registry write raises after `ops.hive_state_init`, the orphan branch + worktree leak. A re-run hits `AlreadyInitialized` (exit 2) without a recovery path, leaving the operator to clean up by hand.

**Fix sketch:** wrap steps after `hive_state_init` in `begin/rescue StandardError` that rolls back via `git worktree remove --force` + `git branch -D hive/state` + delete `.hive-state/config.yml` on failure, or at minimum surface `hive: partial init failed; recover with: git -C <project> worktree remove .hive-state --force && git branch -D hive/state` so the user has a one-line recipe. Reliability reviewer flagged this with confidence 75.

## F5 (P2, manual): `hive init` accepts `--json` silently but emits prose-only summary

`cli.rb`'s class-level `--json` option promises every command "may pass it uniformly," but `Init#call` writes free-text to stdout regardless. An agent piping `hive init --json` cannot parse what defaults landed.

**Fix sketch:** when `--json` is set, emit a structured envelope with the answers hash + project metadata:
~~~json
{"ok":true,"schema":"hive-init.v1","project":"...","planning_agent":"claude","development_agent":"codex","enabled_reviewers":[...],"budgets":{...},"timeouts":{...},"path":"/...","default_branch":"main","hive_state_path":"...","worktree_root":"..."}
~~~
The answers hash already exists in `Init#call`; just JSON-serialize it together with the existing summary fields. Add `hive-init` to `Hive::SCHEMA_VERSIONS`. Both agent-native and cli-readiness reviewers flagged this independently (confidence 95).

## F13 (P3, gated_auto): All-digit agent name unreachable by name in the prompt

`Hive::Commands::Init::Prompts#resolve_agent_choice` always treats digit-only input as an index. A future profile registered under e.g. name `"42"` would be index-only addressable. Edge case today — no such profile exists — worth fixing for forward compatibility.

[...]

PR #321, opened 2026-06-05
Hive patrol: Main CLI parse errors bypass the usage contract

ivankuznetsov:
## Summary

Patrol finding `command-bin-hive-2`: Thor parse failures from `bin/hive` (unknown command, missing required positional, bad flag) exited with status `1` and emitted **no JSON envelope**, even though the documented CLI contract reserves exit `64` (`EX_USAGE`) for usage errors and lists most commands as `--json`-capable. Because `bin/hive` only rescued `Hive::Error`, `Thor::Error` paths were handled by Thor before Hive could map them to the usage contract.

This PR routes Thor parse errors through a new `Hive::ThorUsageJson` so that `hive <cmd> --json` emits a contract-valid error envelope and exits `64`:

- **`Hive::ThorUsageJson`** (`lib/hive/thor_usage_json.rb`) — a per-command `ENVELOPES` table mapping each `--json`-capable command to its `schema` + `error_kind`, with a load-time guard that validates every `schema` literal against `SCHEMA_VERSIONS` and prevents a `KeyError` backtrace on the JSON path. Stage-action verbs (brainstorm/plan/develop/open-pr/pr/review/artifacts/finalize/archive) supply the schema-required `verb` extra. Unknown commands fall through to a generic `usage`-kind envelope.
- **`Hive::UsageError`** (`lib/hive.rb`) — relocated into the `Hive::` hierarchy with an `error_kind` accessor defaulting to `"usage"`, so the safe wire value lives in the type rather than the caller.
- **`bin/hive`** — emits the structured envelope under `--json` for both Thor parse errors and `Hive::Error`-class usage errors, with a trailing `rescue StandardError` so an unclassified error under `debug: true` no longer dumps a raw multi-line backtrace.

Two dogfood fixes rode along on the branch:

- **`ArchiveFilter.hide?`** is now purely age-based — the `RESOLVED_MARKER_NAMES` allowlist was dropped, so archived `9-done` rows age out of the default `hive status` / TUI grid after 3 days regardless of marker, and remain recoverable via `hive archive` / the Archive pane. (Documented as a product decision; an open sign-off question is recorded in the review notes.)
- **Patrol `ReviewHandoff`** now generates an `idea.md` for handed-off findings, with atomic-ish enqueue (partial-write cleanup) and defensive evidence parsing.
- **`Tui::Views::Format`** gains display-cell-accurate (`unicode-display_width`) truncation/padding so CJK/emoji `display_name`s no longer break column alignment.

## Test plan

- `test/integration/cli_parse_error_test.rb` — parse-error envelopes round-tripped against `schemas/<cmd>.v*.json` for every mapped command, plus EPIPE / `JSON::GeneratorError` branches, flag-order independence, and dead-row pins.
- `test/unit/thor_usage_json_test.rb` — `ENVELOPES` table validation, all mapped rows validated against their schemas, load-time guard.
- `test/unit/tui/views/format_test.rb` — wide-grapheme cell accounting, `max=0`, odd-column drop, `nil` coercion, over-budget ljust/rjust clamp.
- `test/unit/archive_filter_test.rb` — marker-agnostic age cutoff, `mtime || folder_mtime` precedence in both directions.
- `test/unit/patrol/review_handoff_test.rb` — exact `idea.md` body render, empty-section dropping, symbol-vs-string evidence parity.
- Recoverability of errored archived rows pinned in `status_test.rb` / `archive_pane_test.rb` with a non-`complete` marker.
- `rubocop` clean across all touched Ruby files; affected unit + integration suites green.

## Review summary

Three review passes (claude-ce, pr-review-toolkit, codex-ce). The headline High — surfaced independently by all three reviewers — was that `--json` parse errors for the 9 stage-action verbs emitted a schema-invalid envelope (missing the required `verb` field); fixed by supplying `verb` via `extras:`. Subsequent passes landed error-path hardening (partial-write cleanup, `rescue StandardError`, `emit` self-protection, load-time `error_kind` guard), type/contract hardening (`UsageError#error_kind` default, frozen `Envelope` rows, reserved-key disjointness), and the missing test coverage. All High findings resolved; no merge-blocking issues remain.

One open product question is recorded for operator sign-off: making `ArchiveFilter` hiding marker-agnostic means aged errored `9-done` rows now drop from the default view (still recoverable via `hive archive`). The recoverability path is test-pinned regardless of the decision.

A cluster of "TUI panes no longer respect terminal height" findings were confirmed **stale-base false positives**: this branch never touched `bubble_model.rb`; the viewport logic is main's #326 work that `git diff origin/main..HEAD` misreads as deletions. The rebase must keep **both** main's `height:` viewport and this branch's cell-aware padding in `tasks_pane.rb`/`projects_pane.rb`.

## Linked task

/home/asterio/Dev/hive/.hive-state/stages/6-review/patrol-command-bin-hive-38bb4c33


ivankuznetsov:

# Codex CE Code Review 01

No plan was found in the task folder; review was performed against `git diff origin/main..HEAD` without plan grounding.

## High

## Medium
- [ ] `bin/hive` emits schema-invalid `hive-stage-action` parse-error envelopes: Missing-target JSON invocations such as `hive brainstorm --json`, `hive plan --json`, and other workflow verbs now route through `emit_thor_usage_json`, but `schemas/hive-stage-action.v2.json` requires `verb` on error payloads and `Hive::Commands::StageAction#emit_error_envelope` previously supplied it; the new top-level envelope omits it, so JSON consumers receive an invalid contract document.

## Nit

ivankuznetsov:

# Code Review — patrol-command-bin-hive-38bb4c33 (pass 1)

- **Scope:** `git diff origin/main..HEAD` (2 commits: parse-error JSON contract for `bin/hive`; TUI task-context + archive-count changes).
- **Plan grounding:** No `plan.md` found in the task folder — review proceeded without plan grounding.
- **Intent:** (1) Route Thor parse/usage errors through the `--json` error-envelope contract and exit `USAGE` instead of Thor's plaintext + exit 1; (2) make TUI columns display-width-aware (unicode), and make the archive 3-day hide cutoff marker-agnostic with a surfaced hidden count.
- **Method:** Manual diff read + live schema validation of every mapped envelope + two parallel correctness reviewers (bin/hive contract; TUI format/archive). Findings corroborated.

## High
- [ ] `hive <stage-verb> --json` parse errors emit a **schema-INVALID** envelope (`bin/hive:56-66`): `emit_thor_usage_json` calls `Hive::Schemas::ErrorEnvelope.build` without `extras:`, but the `hive-stage-action` ErrorPayload schema *requires* a `verb` field (the in-command path supplies it via `extras: { "verb" => @verb }` at `lib/hive/commands/stage_action.rb:207`). Live `JSONSchemer` validation fails for all 9 verbs (brainstorm/plan/develop/open-pr/pr/review/artifacts/finalize/archive) — the exact contract this commit exists to uphold is broken for those commands.

## Medium
- [ ] Fallback envelope is non-schema (`bin/hive:67-75`): for any command not in `JSON_PARSE_ERROR_ENVELOPES` (unknown command e.g. `hive boguscmd --json`, `init`, `new`, or nil command) the else-branch builds a hand-rolled hash with **no `schema`/`schema_version` key** and hardcodes `error_kind "invalid_task_path"` — an agent dispatching on `schema` cannot route it, and the kind is semantically wrong for an unknown-command error.
- [ ] Test coverage only exercises `run` (`test/integration/cli_parse_error_test.rb`): the broken stage-action verbs and the fallback branch are untested. A round-trip "build envelope → validate against `schemas/<cmd>.v*.json`" assertion over every entry in the map would have caught the missing-`verb` High finding pre-merge.
- [ ] `marker_name:` keyword in `ArchiveFilter.hide?` is now dead/vestigial (`lib/hive/archive_filter.rb:16`): the `RESOLVED_MARKER_NAMES` allowlist was removed so the body never reads `marker_name`, yet all call sites (`status.rb:376`, `snapshot.rb:255`) still pass it — misleads readers into thinking the marker still gates hiding. Remove the param or document it as intentionally ignored.

## Nit
- [ ] `daemon` map entry hardcodes `hive-daemon-enroll`/`missing_project` (`bin/hive:31`) — wrong schema + kind for non-enroll daemon subcommands (status/stop/reload/queue/install). Low reachability since `daemon` uses splat args validated inline (raising `Hive::Error`, caught by the lower rescue), but the mapping is incorrect if ever hit.
- [ ] `ThorUsageError` is defined in the top-level namespace (`bin/hive:8`), inconsistent with the `Hive::` error hierarchy; it surfaces as `error_class "ThorUsageError"` (unnamespaced) and pollutes the global namespace. Consider `Hive::ThorUsageError`.
- [ ] `error_class` representation differs between branches (`bin/hive:70` vs `ErrorEnvelope.build`): the fallback uses full `usage_error.class.name` while `build` uses `class.name.split("::").last`. Harmless only because the class is top-level; align them.
- [ ] Archive hide cutoff is now **marker-agnostic** — errored archived tasks (stage 9-done) older than 3 days now disappear from `hive status` and the main TUI grid. Confirmed recoverable (Archive pane `archive_pane.rb:44` ignores `hide?`; a "+N hidden" hint is surfaced) and intended, so not a bug — but add a one-line note to `wiki/commands/status.md` / `wiki/commands/tui.md` that the cutoff no longer depends on marker state.
- [ ] No test pins the wide-char **column alignment** that this diff actually ships (`lib/hive/tui/views/tasks_pane.rb`) — only the width-helper primitives are unit-tested. A render assertion with a CJK/emoji `display_name` would lock the fix.

---

**Verdict:** Not ready — the High finding is a direct regression of this commit's stated contract (invalid JSON envelope for 9 workflow verbs under `--json`). Fix order: (1) supply `verb` via `extras:` for stage-action commands (and give the fallback branch a real `schema`/`schema_version` + correct `error_kind`); (2) extend the integration test to schema-validate every mapped command; (3) clean up the dead `marker_name` parameter. Nits at author discretion.

ivankuznetsov:

# PR Review Toolkit — patrol-command-bin-hive-38bb4c33 (pass 1)

- **Scope:** `git diff origin/main..HEAD` (2 commits: parse-error JSON contract for `bin/hive`; TUI task-context + archive-count changes).
- **Plan grounding:** No `plan.md` found in the task folder — review proceeded without plan grounding.
- **Personas run:** code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer (parallel). Union of findings below; overlapping findings carry multiple persona tags. All affected-area tests pass and rubocop is clean (verified by code-reviewer).

## High
- [ ] [test] `hive <stage-verb> --json` parse errors emit a **schema-INVALID** envelope (`bin/hive:56-66` / `lib/hive.rb:84` `ErrorEnvelope.build`): `emit_thor_usage_json` passes no `verb` in `extras`, but `hive-stage-action.v2` `ErrorPayload.required` includes `verb`. Reproduced live — 8 of 9 verbs (brainstorm/plan/develop/open-pr/pr/review/artifacts/finalize) fail `json_schemer` validation. The published `--json` contract this commit exists to uphold is invalid for those commands, and the `run`-only integration test masks it.

## Medium
- [ ] [test] Integration coverage tests only `run` out of the 21-entry `JSON_PARSE_ERROR_ENVELOPES` map (`test/integration/cli_parse_error_test.rb:9`); a round-trip "build envelope → validate against `schemas/<cmd>.v*.json`" loop over every mapped command is the single highest-value missing test (would have caught the High finding).
- [ ] [test][errors][type-design] Fallback JSON branch (`bin/hive:67-75`) is entirely uncovered: the unknown-command test runs **without** `--json`, so the `else` arm never executes. It emits a schema-less shape (no `schema`/`schema_version`) and hardcodes `error_kind:"invalid_task_path"` even for a "Could not find command" error — semantically wrong and unroutable by agents that dispatch on `schema`.
- [ ] [test] `"archive" => ["hive-stage-action", …]` map entry (`bin/hive:21`) is unreachable — `hive archive` is a valid no-arg command (emits a status listing, exit 0) and never hits the parse-error path. Dead/misleading mapping with no test documenting real behavior.
- [ ] [errors] Empty `rescue Errno::EPIPE, JSON::GeneratorError` (`bin/hive:79`) silently swallows `JSON::GeneratorError`: an unserializable envelope (a real `ErrorEnvelope.build` defect) yields empty stdout + `exit 64` with no diagnostic on any stream. The sibling rescue at `lib/hive/cli.rb:837-839` falls through to bare-text `warn`; this one falls through to nothing. Add `warn error.message`.
- [ ] [errors] On JSON-serialization failure the original Thor message is dropped entirely (`bin/hive:87-92`): the JSON branch relies on `emit_thor_usage_json` having printed, with no stderr fallback if it hit the empty rescue above — the user/agent loses the "what did I type wrong" text.
- [ ] [errors] `debug: true` re-raises all `Thor::Error` subclasses and flattens them to one `error_kind` (`bin/hive:85`): `UnknownArgumentError` (bad flag) and `RequiredArgumentMissingError` (missing positional) both surface as `invalid_task_path`, coarser than the closed `error_kind` enums elsewhere imply. (Exit-code 1→64 change itself is correct.)
- [ ] [type-design] `ThorUsageError` is defined at the **top-level** namespace (`bin/hive:8`) while inheriting `Hive::InvalidTaskPath` — breaks the single `Hive::` error hierarchy (library code cannot name/rescue it), and `InvalidTaskPath` ("bad path") is a poor semantic fit for a missing flag. Prefer `Hive::UsageError < Error` (exit_code USAGE) in `lib/hive.rb`.
- [ ] [type-design] `JSON_PARSE_ERROR_ENVELOPES` (`bin/hive:10-34`) is an untyped `[String, String]` tuple duplicating `SCHEMA_VERSIONS` keys and per-command `error_kind` enums as bare literals; `ErrorEnvelope.build` never validates `error_kind`, so nothing fails fast on drift. Reference the existing `*ErrorKind` constants and add a guard test that every `--json`-capable command has an entry.
- [ ] [type-design][code][comment] `marker_name:` is now a **dead/unused** keyword in `ArchiveFilter.hide?` (`lib/hive/archive_filter.rb:16`) after the `RESOLVED_MARKER_NAMES` allowlist removal, yet both callers still pass it (`snapshot.rb:255`, `status.rb:376`). The false interface masks a real invariant weakening (resolved-marker gate → marker-agnostic hiding of *any* archived task after 3 days). Remove the param + call-site args, or document it as intentionally ignored.

## Nit
- [ ] [code][errors] `idea_text`/`evidence_text` (`lib/hive/patrol/review_handoff.rb:172,177`) call `finding.evidence.empty?`/`.map` assuming a non-nil array; `Finding.new` (vs `from_h`) does not coerce, so a direct caller hits `NoMethodError`. Use `Array(finding.evidence)`. Latent, not live on the `enqueue` path.
- [ ] [code] `parse_folder_mtime` (`lib/hive/tui/snapshot.rb:262-264`) is now a one-line pass-through to `parse_time` with a single call site — inline as `parse_time(row.folder_mtime)`; the wrapper is dead indirection.
- [ ] [test] No dedicated `format_test.rb`; the new `take_cells`/`truncate` wide-grapheme boundary (a CJK/emoji name landing exactly on an odd column — the overflow this PR claims to fix) is only exercised indirectly via two icon-column cases in `tasks_pane_test.rb:441,447`. Add a direct unit test pinning the cell-accounting boundary.
- [ ] [test] `idea_text` conditional branches are uncovered (`review_handoff_test.rb:54` tests only a fully-populated finding): empty description/recommendation/evidence guards, the `- evidence` / location-only evidence fallbacks, and `display_name`'s nil-title→id fallback. Real patrol findings often omit recommendation/evidence.
- [ ] [type-design][errors] Fallback envelope (`bin/hive:67-75`) is an inconsistent contract for the same logical failure: schema-less shape plus `error_class` = full `"ThorUsageError"` vs `build`'s last-`::`-segment form. Align the two shapes.
- [ ] [type-design] `command = argv.find { |arg| !arg.start_with?("-") }` (`bin/hive:57`) returns only the group name for subcommand groups — `daemon` always maps to `hive-daemon-enroll`/`missing_project` regardless of the real subcommand (status/stop/queue). Coarse best-effort; document the limitation.
- [ ] [comment] `truncate` doc (`lib/hive/tui/views/format.rb:26-29`) says `max_width < 2` "falls back to a hard cut without ellipsis," but `max_width == 1` against a leading double-width grapheme now returns `""` (cluster width 2 exceeds the 1 remaining cell). Wording nuance, not a true inaccuracy; the "measured in terminal cells via unicode-display_width" claim is correct.
- [ ] [code] `wiki/log.md` states hiding "now uses row `mtime` rather than mutable `folder_mtime`," but the code is `mtime || folder_mtime` (folder_mtime remains the fallback) — minor doc/impl wording mismatch.

---

**Verdict:** Not ready — the High finding (schema-invalid `--json` envelope for ~8 workflow verbs) is a direct regression of this commit's stated contract, independently surfaced by the test persona and root-caused to the missing `verb` extra. Recommended order: (1) supply `verb` via `extras:` for stage-action commands and give the fallback branch a real `schema`/`schema_version` + correct `error_kind`; (2) add the round-trip schema-validation test across every mapped command; (3) restore a stderr fallback in the empty JSON rescue; (4) relocate `ThorUsageError` into `Hive::` and drop the dead `marker_name` parameter. Nits at author discretion.

ivankuznetsov:

# Codex CE Code Review Pass 2

No plan was found in the task folder; review proceeded against `git diff origin/main..HEAD`.

## High

## Medium

## Nit
- [ ] Fix stale unknown-command error_kind text in `wiki/cli.md`: line 65 says the schema-less `--json` fallback uses `error_kind:"invalid_task_path"`, but `bin/hive` and the integration test now emit `error_kind:"usage"`, so the wiki contradicts the shipped contract.

ivankuznetsov:

# Code Review — claude-ce-code-review (pass 02)

> **No plan.md found** in the task folder. Reviewed without plan grounding,
> against `git diff origin/main..HEAD` in the worktree.

Scope: `bin/hive` parse-error JSON envelopes (+ `Hive::UsageError`), marker-agnostic
`ArchiveFilter` + `mtime` precedence, `ReviewHandoff#idea_text/evidence_text`,
`Tui::Views::Format` display-cell truncation/padding (`unicode-display_width`), and
the `tasks_pane`/`snapshot` call-site changes.

**Verification done during review (read-only):**
- `test/integration/cli_parse_error_test.rb` — 5 runs / 116 assertions, green.
- `archive_filter`, `tui/views/format`, `tui/views/tasks_pane`, `tui/snapshot`,
  `patrol/review_handoff`, `commands/status` unit suites — all green.
- `rubocop` on all 7 touched source files — no offenses.
- Schema cross-check: every command in `JSON_PARSE_ERROR_ENVELOPES` (including the
  untested `status`/`patrol`/`daemon`/`forget`/`prune`/`archive` entries) maps to an
  `error_kind`/`exit_code` that validates against its published `ErrorPayload` arm
  (verified `exit_code` enums include 64/USAGE, `hive-stage-action` verb enum includes
  `archive`, `hive-patrol` accepts `error_kind: "error"`). No latent contract break found.

## High

_None._

## Medium

- [ ] Archive hiding is now marker-agnostic — errored/unresolved `9-done` rows are hidden from daily `hive status` text and the TUI grid after 3 days, where they previously stayed visible indefinitely: this is a deliberate, wiki/log-documented product decision (`archive_filter.rb` dropped `RESOLVED_MARKER_NAMES`), but it reduces failure visibility on daily surfaces, so confirm an aged errored done-task disappearing from the default view (recoverable only via `hive archive`) is the intended behavior.

## Nit

- [ ] `ReviewHandoff#evidence_text` (`lib/hive/patrol/review_handoff.rb:179-180`) reads only string keys (`entry["file"]`, `["line"]`, `["snippet"]`), whereas sibling code consuming the same `finding.evidence` defends both key types (`pr_opener.rb:171` and `fingerprint.rb` use `e["file"] || e[:file]`); the primary flow is JSON-parsed (string-keyed) so it works today, but a symbol-keyed evidence entry would silently degrade `idea.md` to bare `- evidence` lines — align with the sibling fallback for consistency.
- [ ] `bin/hive:124` gates JSON emission on `ARGV.include?("--json")`, which would also fire if `--json` arrived as an option *value* rather than the flag; harmless (the no-schema fallback envelope is still emitted) but a precise `"--json"`-as-flag check would be more correct.
- [ ] Unknown-flag invocations of the untested map entries (`hive status --bogus --json`, `patrol`, `daemon`, `forget`, `prune`, `archive`) still route through `emit_thor_usage_json`; I confirmed each validates against its schema, but no test pins them (already logged as `wiki/gaps.md` #24) — an unknown-flag matrix row per mapped command would close the gap cheaply.

ivankuznetsov:

# PR Review — pr-review-toolkit (pass 02)

> **No plan.md found** in the task folder. Reviewed without plan grounding,
> against `git diff origin/main..HEAD` in the worktree.

Personas run: `code-reviewer`, `pr-test-analyzer`, `silent-failure-hunter`,
`comment-analyzer`, `type-design-analyzer`. Union of all findings below;
each item is prefixed with the persona that surfaced it. All touched suites
were run green and rubocop was clean across the 8 changed Ruby files; no
finding is a live correctness bug on the reachable path.

## High

- [ ] [tests] `JSON_PARSE_ERROR_ENVELOPES` entries for `forget`/`daemon` are unreachable dead code and no test pins them either way: both raise their own command-local `UsageError < Hive::Error` caught by the *lower* `rescue Hive::Error`, so the map row never fires (`bin/hive:46-52`, `lib/hive/commands/forget.rb:22,54`, `lib/hive/commands/daemon.rb:49`) — a future refactor routing one through the Thor path would double-emit (stdout envelope + stderr) silently.
- [ ] [tests] The `archive` map entry (`bin/hive:45`) is untested and likely unreachable (`hive archive` accepts bare invocation, never raises a Thor parse error), yet it is a frozen-contract entry with no test asserting it is dead or that `hive archive --json` emits the correct envelope. (Note: pass-01 log deliberately *kept* `archive` as insurance — flag for explicit sign-off, not blind removal.)
- [ ] [tests] Mapped `error_kind` values for `forget`/`prune`/`daemon`/`status`/`patrol` are never validated against their schemas: the build→validate round-trip runs only over the 14-command subset, so a future edit to e.g. `schemas/hive-forget.v1.json`'s enum would not fail any test even though `bin/hive:48` still references `missing_name`.

## Medium

- [ ] [silent-failure/type-design] `SCHEMA_VERSIONS.fetch` (`lib/hive.rb:87`) inside `emit_thor_usage_json` is unguarded: a future typo'd schema string in `JSON_PARSE_ERROR_ENVELOPES` raises `KeyError`, which is neither `Thor::Error` nor `Hive::Error`, so it escapes the top-level rescue and prints a raw backtrace on the `--json` path — the exact agent-hostile failure this code exists to prevent. A load-time guard or `JSON_PARSE_ERROR_ENVELOPES.values.map(&:first) ⊆ SCHEMA_VERSIONS.keys` test would close it.
- [ ] [silent-failure] The `rescue Thor::Error` path labels every parse failure `invalid_task_path` regardless of the actual Thor failure mode (`bin/hive:79`,`124`): an unknown *flag* on `hive run` is reported as `invalid_task_path`, semantically wrong for agents branching on `error_kind`.
- [ ] [type-design] `error_class` wire collision: `Hive::UsageError` and the three sibling `Commands::{Metrics,Forget,Daemon}::UsageError` all serialize to `"error_class":"UsageError"` via `class.name.split("::").last` (`lib/hive.rb:89,376`), so the new type's "distinct from InvalidTaskPath" benefit is invisible to JSON consumers — the doc comment overstates the consumer-facing distinction.
- [ ] [type-design] `ArchiveFilter.hide?` now defaults *both* `mtime:` and `folder_mtime:` to `nil` (`lib/hive/archive_filter.rb:16`), so `hide?(stage:)` with no timestamp is a legal call that silently returns `false`; the prior required `folder_mtime:` forced callers to supply a timestamp. Consider requiring at least one to keep the precondition visible.
- [ ] [code/silent-failure/type-design] Archive hiding is now marker-agnostic: dropping `RESOLVED_MARKER_NAMES` means archived `error`/unresolved tasks are hidden after 3 days where they previously stayed visible indefinitely (`lib/hive/archive_filter.rb`, flipped test `archive_filter_test.rb:35`). Documented in wiki/log as an intentional product decision — flagged for conscious sign-off, not a regression.
- [ ] [comments] `bin/hive:17-22` mis-models Thor dispatch: `daemon` is a single command `def daemon(subcommand=nil, *targets)` (`lib/hive/cli.rb:765`), not a Thor "subcommand group," and there is no subcommand named `enroll` (valid set: `start stop status reload tail enable disable install queue`); `enroll` is only a schema name. The comment's conclusion ("never reaches this map") is correct but the reasoning is wrong and would mislead a maintainer.
- [ ] [comments/type-design] The `daemon` map entry (`bin/hive:52`) is effectively dead and the comment softens this ("would otherwise mis-label it") rather than stating plainly it is unreachable — either delete it or document it as retained defensively.
- [ ] [tests] The `JSON::GeneratorError` fallback in `emit_thor_usage_json` (`bin/hive:110-116`) is untested, despite the comment calling it "a real defect" path and the analogous `status.rb:591` GeneratorError path *being* tested — asymmetric coverage.

## Nit

- [ ] [tests] `Errno::EPIPE` rescue on the JSON write path (`bin/hive:107-109`) has no coverage.
- [ ] [tests/code/type-design] `--json` detection (`ARGV.include?("--json")`, `bin/hive:124`) and command-token pick (`argv.find { !start_with?("-") }`, `bin/hive:78`) are untested for flag-ordering edge cases (`hive --json run`, value-looks-like-flag) — a reordering refactor could mis-detect the command/schema.
- [ ] [tests] `ArchiveFilter.hide?` fallback branch where `mtime: nil, folder_mtime: <old>` engages is not asserted as a pair; only both-present precedence and `folder_mtime: nil` are covered (`archive_filter_test.rb`).
- [ ] [tests] `idea_text` with an all-empty finding (blank title AND nil description) is untested — the degenerate `# Patrol: <id>`-only body is plausible per the test's own "often omit fields" comment.
- [ ] [tests] `ljust_cells`/`rjust_cells` over-budget (truncate-then-clamp) path and `nil` input to `truncate`/`take_cells` are only covered indirectly; one direct case each would pin the `[…,0].max` clamp and the `.to_s` guard (`format.rb:43,48,36,55`).
- [ ] [silent-failure/code] `evidence_text` (`review_handoff.rb:177-184`) reads only string keys and assumes each entry is a Hash; sibling consumers of the same `finding.evidence` defend both (`pr_opener.rb:171` / `fingerprint.rb` use `e["file"] || e[:file]`) — a symbol-keyed or bare-String entry degrades silently / raises `TypeError`. Safe on the JSON-parsed primary flow only.
- [ ] [type-design] `Hive::UsageError` lacks the `error_kind` accessor every sibling `UsageError` carries (`lib/hive.rb:376`), forcing `bin/hive` to source the kind out-of-band from the map instead of from the exception.
- [ ] [type-design] `JSON_PARSE_ERROR_ENVELOPES` uses an opaque positional `[schema, error_kind, extras]` tuple; a labeled struct/hash would make the optional third slot and the stage-action `verb` requirement self-documenting and enable a single load-time guard.
- [ ] [type-design] `review_handoff.rb:75-89` stores `idea_text(finding)` twice — in frontmatter `original_text` and as the markdown body — two sources of truth that can drift; add a comment if one is canonical or store once.
- [ ] [comments] `Hive::UsageError` doc (`lib/hive.rb:372-373`) says it lives in the namespace "so library code can name and rescue it," but it is only instantiated in `bin/hive`; the claim is aspirational, not present usage.
- [ ] [tests] `idea.md` is asserted only via `assert_includes` substrings; the exact rendered body (the `<<~MD` heredoc with a non-aligned closing `MD`, `review_handoff.rb:86-88`) is never pinned, so a leading/trailing-whitespace regression would pass unnoticed.

ivankuznetsov:

# Codex CE Code Review Pass 3

No plan was found in the task folder; review proceeded against `git diff origin/main..HEAD`.

## High

## Medium
- [ ] TUI panes no longer respect terminal height: `compose_two_pane_view` calls `TasksPane.render`/`ProjectsPane.render` without a height budget and the pane renderers removed viewport clipping, so a 10-row terminal with 20 tasks renders 26 lines and pushes the footer/current context off-screen.
- [ ] Archive hiding is still marker-agnostic: aged `9-done` rows with `ERROR`/unresolved markers now disappear from daily status/TUI surfaces after 3 days, so the failure-visibility product decision still needs explicit sign-off.

## Nit
- [ ] `ProjectsPane#render_row` still pads with Ruby `String#ljust` after switching truncation to terminal-cell width: CJK/emoji project labels can render wider than the pane even though `Format.ljust_cells` now exists for this exact case.

ivankuznetsov:

# pr-review-toolkit review — pass 03

- **Diff reviewed:** `git diff origin/main..HEAD` (43 files, +1059/−320)
- **Plan grounding:** no `plan.md` found in the task folder — reviewed without plan grounding.
- **Personas run:** code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer (union of findings below).
- **Test suite:** 4693 runs, 17921 assertions, **1 failure, 0 errors, 2 skips**. The lone failure (`StopHookInstallerTest#test_stop_hook_requires_stage_dir_env`) is **not a PR defect** — it is unrelated to this diff (the file is untouched) and is an environment leak from the reviewer harness, which exports `HIVE_TASK_STAGE_DIR`; the test's subprocess inherits it and therefore does not hit the "required" error path. Fails identically on `origin/main` under this env. Rubocop clean.
- **No High-severity findings** from any persona.

## High

_None._

## Medium
- [ ] [code-reviewer] Grid view no longer clips to terminal height (`lib/hive/tui/bubble_model.rb:3160-3177`): `compose_two_pane_view` dropped `pane_height_for`/`fit_rows`/`fit_lines`; in bubbletea alt-screen mode, output taller than the terminal overflows the bottom (no viewport scroll), so on a short terminal with many tasks the footer and selected row can be pushed offscreen — the deleted `test_charm_tui_vertical_resize…` test previously pinned this, and the change is undocumented in `wiki/commands/tui.md`.
- [ ] [type-design] `UsageError#error_kind` is unvalidated (`lib/hive.rb:384-395`): unlike every sibling closed-enum `*ErrorKind` module in the same file, it accepts an arbitrary `error_kind:` string with no membership check, so a typo'd kind travels to the wire silently (mitigated in practice because `ThorUsageJson` sources kinds from the schema-tested `ENVELOPES`).
- [ ] [type-design] `respond_to?(:head_sha)` is dead defense (`lib/hive/patrol/review_handoff.rb:102`): the only real `patch` producer (`Fixer::PatchAttempt`) is a Struct that always defines `head_sha`, so the guard always passes and only the `&& patch.head_sha` truthiness half is operative — it implies an optionality the contract doesn't have.
- [ ] [type-design] Duck-typed `finding`/`patch` contract is implicit and unenforced at the boundary (`lib/hive/patrol/review_handoff.rb:17,32-104`): a `finding` missing e.g. `feature_id` fails deep inside `base_slug` with a `NoMethodError` rather than at method entry; consider asserting the contract (or accepting the concrete `Patrol::Finding` type) at the boundary.
- [ ] [silent-failure] Malformed `folder_mtime` silently un-hides an old archived task (`lib/hive/tui/snapshot.rb:261-267`): `parse_time` rescues `ArgumentError`→`nil`, so `hide?` returns `false` and a 3+-day archived row stays visible with no log — fail-open and safe-by-default, but the only fully-silent parse failure in scope (a corrupt snapshot timestamp would never be noticed).
- [ ] [comment-analyzer][type-design] `daemon` "retained defensively" comment overstates symmetry / superclass (`lib/hive/thor_usage_json.rb:56-60`): `Commands::Daemon::UsageError` is `< Hive::InvalidTaskPath`, not directly `< Hive::Error` as the comment implies, and `daemon` is a Thor subcommand group whose parse-path reachability story differs from `forget` — the "caught by the lower `rescue Hive::Error`" conclusion still holds transitively, but the stated inheritance is imprecise.
- [ ] [pr-test-analyzer] `parse_time` invalid-`mtime`→valid-`folder_mtime` fallback is untested at the snapshot layer (`test/unit/tui/snapshot_test.rb:444-463`): the test blanks both timestamps simultaneously, so the realistic regression — an invalid/blank `mtime` falling through to a valid *old* `folder_mtime` that must still hide the row — is never exercised end-to-end (the `ArchiveFilter` unit test covers the `nil` precedence but not the parse-failure that produces it).

## Nit
- [ ] [code-reviewer] `command_token` routes by subcommand-group name only (`lib/hive/thor_usage_json.rb:82-84`): `hive daemon enroll --json` → the `daemon` envelope regardless of leaf subcommand; acknowledged inline, bounded, and dead-mapped today.
- [ ] [code-reviewer] `json_requested?` false-positives when `--json` is supplied as another option's value (`lib/hive/thor_usage_json.rb:92-94`): acknowledged in-comment; the emitted envelope is still well-formed, so benign.
- [ ] [type-design] `Envelope` exposes a raw nilable `extras` slot beside its `extras_hash` accessor (`lib/hive/thor_usage_json.rb:23-27`): the "nil reads as `{}`" invariant can be bypassed by reading `extras` directly — latent footgun only (`build_payload` correctly uses `extras_hash`).
- [ ] [type-design] `ljust_cells`/`rjust_cells`/`truncate` silently `to_s` non-String input (`lib/hive/tui/views/format.rb:44-67`): the "operates on a String" expectation is implicit; harmless since all call sites already pass strings.
- [ ] [pr-test-analyzer] `archive_hidden_notice`'s `mode == :grid` guard is untested (`lib/hive/tui/bubble_model.rb:3286`): every footer test uses `:grid`, so suppression of the "+N hidden" notice in archive/filter modes is unasserted.
- [ ] [pr-test-analyzer] `command_token` with an empty argv is untested (`test/unit/thor_usage_json_test.rb:48-52`): `argv.find` returns `nil` (handled), so low risk.
- [ ] [comment-analyzer] "the four" `UsageError` count is a magic number (`lib/hive.rb:381-383`): the "JSON consumers cannot tell the four apart" claim is accurate today but a future fifth `UsageError` subclass would silently invalidate it.

ivankuznetsov:

# Code Review — pass 03 (`/ce-code-review`)

- **Scope:** `git diff origin/main..HEAD` (10 commits; lib + tests + wiki)
- **No `plan.md` found in the task folder** — review proceeded without plan grounding.
- **Mode:** report-only (reviewer must not edit code).
- **Reviewers:** correctness, testing, maintainability, cli-readiness, reliability.
- **Intent:** route Thor parse errors through new `Hive::ThorUsageJson` so `hive <cmd> --json` emits a contract-valid error envelope; make `ArchiveFilter.hide?` purely age-based (drop `marker_name:`); add patrol `ReviewHandoff` `idea.md` generation; add display-cell-accurate TUI `Views::Format`; remove the TUI height-aware viewport/fit logic.

Prior passes (01, 02) already landed the bulk of this work; pass 03 findings are the residue.

## High

_None._ No P0/P1 critical-breakage findings. The new `--json` parse-error path keeps stdout pure JSON on every tested branch, preserves exit 64, and the load-time `SCHEMA_VERSIONS` guard removes the KeyError-backtrace failure mode.

## Medium

- [ ] `ReviewHandoff#enqueue` leaves a half-written `6-review/<slug>/` task folder on any mid-write failure: `lib/hive/patrol/review_handoff.rb:17` — six non-atomic writes (mkdir_p, meta, idea.md, task.md, worktree.yml, pr.md) with no rescue/cleanup; the caller (`pr_opener.rb`) removes only the git worktree, so a partial folder (e.g. `worktree.yml` present, `pr.md`/`task.md` missing) orphans permanently, marks the slug taken, and can be picked up by the review stage as a malformed task. Prefer temp-dir-then-rename or a rescue that removes `task_folder` before re-raising.
- [ ] `--json` is honoured for Thor parse errors but silently dropped for `Hive::Error`-class usage errors: `bin/hive:42` — `hive bot --json` / `hive daemon --bogus --json` raise a `Hive::Error`, hit the lower `rescue Hive::Error` which only `warn`s to stderr and emits no stdout JSON, despite the same exit 64 and existing `hive-daemon-enroll` schema. An agent that learned `hive run --json` returns a parseable stdout envelope gets empty stdout + free-form stderr for these. Make the `Hive::Error` branch emit a structured envelope when `--json` is requested, or document the stderr-only exception in the `--json` contract.
- [ ] `debug: true` can leak Ruby backtraces for an unclassified `StandardError` raised inside a command body: `bin/hive:34` (corroborated by correctness + cli-readiness) — `Hive::CLI.start(ARGV, debug: true)` makes Thor re-raise originals; bin/hive rescues only `Thor::Error` and `Hive::Error`, so a plain `StandardError` escapes both and dumps a multi-line backtrace (exit 1) where pre-`debug` Thor printed one line. Add a trailing `rescue StandardError` that emits a single-line message (or `internal`/SOFTWARE envelope under `--json`) and gates the backtrace behind a debug flag.
- [ ] No negative test pins that age-hidden errored archived rows stay recoverable via the archive surfaces: `test/unit/tui/snapshot_test.rb:2282`, `test/unit/commands/status_test.rb:437` (corroborated by testing + correctness) — the diff's central behavior change is that errored `9-done` rows are now hidden by age from daily status/grid; the product justification is they remain visible in `hive archive` / the TUI Archive pane, but every archive-mode and ArchivePane fixture uses only `complete` markers. Add one archive-surface test with a `marker: error`, age >3d row asserting it is listed, so a future re-introduction of marker filtering into archive selection can't silently make them unrecoverable.

## Nit

- [ ] Stale method name in new test comments: `test/integration/cli_parse_error_test.rb:14,40,112` — three comments reference `emit_thor_usage_json`, which does not exist after the extraction; the entrypoint is `Hive::ThorUsageJson.emit`. Doc rot that will fail a future grep.
- [ ] Unknown-command `--json` fallback envelope omits the `schema` key: `lib/hive/thor_usage_json.rb:126` — every known-command envelope carries `schema`; the unknown-command branch does not, so an agent dispatching on `payload.fetch("schema")` hits nil exactly when it most needs a clean signal. Emit a sentinel/`null` schema or document the key-absence. (Acknowledged-by-design; low impact.)
- [ ] `emit`'s rescue surface is narrower than its call surface: `lib/hive/thor_usage_json.rb:105` — only `Errno::EPIPE` and `JSON::GeneratorError` are rescued, but `build_payload` → `ErrorEnvelope.build` could raise other errors that then escape both bin/hive rescues as a raw backtrace. The load-time guard makes the only known trigger unreachable today; a bare `rescue => e` fallthrough that still prints the original Thor usage text would harden the JSON path against its own faults.
- [ ] `evidence_text` raises `TypeError` on a non-Hash evidence entry: `lib/hive/patrol/review_handoff.rb:313` — `entry["file"] || entry[:file]` on a `String` entry raises (Symbol into Integer), aborting `enqueue`. Schema-guarded (items are `type: object`) and a pre-existing shared pattern with `pr_opener.rb`, so low-confidence; guard with `entry.is_a?(Hash)` if `from_h` ever ingests unvalidated evidence.
- [ ] `header_strip` retains a width-dependent truncation branch but lost its two direct unit tests in the viewport-removal sweep: `lib/hive/tui/bubble_model.rb:3187` — the strip is still composed into the grid view and clamps at narrow widths, but is now only exercised at wide width via composed-view tests. The underlying `Format.truncate` is well-covered; re-add a narrow-width `header_strip` assertion to restore the wiring coverage.
- [ ] `ENVELOPES` "retained defensively" comments lump genuinely-dead rows (`archive`/`forget`/`daemon`/`patrol`) with rows reachable via the unknown-flag path (`status`/`prune`): `lib/hive/thor_usage_json.rb:53` — the asymmetry isn't reflected in the comments, so a future editor can't tell load-bearing rows from speculative ones without re-deriving dispatch paths. Annotate each row's reachability class, or drop the four truly-dead rows and let the `nil`-envelope `usage` fallback handle them.
- [ ] `evidence_text` dual-key (`["x"] || [:x]`) field access is now hand-rolled in three places (`review_handoff.rb`, `pr_opener.rb:171`, `fingerprint.rb`): a single `Hive::Patrol::Evidence.location/.snippet` accessor pair would collapse them so a fourth consumer can't forget a key form. Low-ceremony, optional.

## Coverage notes

- **Accepted trade-offs (not findings):** removing the height-aware viewport means the two-pane view no longer clips to terminal rows — on a short terminal with many rows the footer / archive-hidden notice / off-screen selected row can scroll out of view. This is the explicit intent of the viewport-removal work (the corresponding height-clamp tests were deliberately deleted), recorded as a known behavioral trade-off. Footer *presence* in the composed frame is still pinned.
- The `ArchiveFilter` change is consistent across `status.rb` and `snapshot.rb` (both feed `Time` objects); both surfaces now hide aged archived rows regardless of marker. Edge note: because row `mtime` takes precedence over `folder_mtime`, an archived task whose state file is touched after archiving resets its age and re-appears — appears intentional per `test_row_mtime_takes_precedence_over_folder_mtime`.
- Rescue ordering in `bin/hive` (`Thor::Error` before `Hive::Error`) is **not** a misrouting risk: `Hive::Error < StandardError` is disjoint from `Thor::Error`, so `Hive::UsageError` can never be caught by the wrong clause.
- Strong new coverage: `format_test.rb` (wide-grapheme cell accounting, nil coercion, ljust/rjust over-budget clamp), `thor_usage_json_test.rb` (all mapped rows round-tripped against schemas, EPIPE/GeneratorError branches, flag-order independence), and the patrol `idea.md` tests (exact body render, empty-section dropping, symbol-vs-string evidence parity).

ivankuznetsov:

Outcome: timeout

Agent tail:

```
I’m marking only JSON writes that can be followed by a re-raise. Success-only writes do not need the flag for correctness, but marking them is harmless where it keeps the code simple.
```

ivankuznetsov:
Closing as **superseded by #401**, which implements the same fix (usage errors must emit the JSON error contract) in a simpler, self-contained way (`bin/hive` only). This branch's approach (new `lib/hive/thor_usage_json.rb` + `Hive::UsageError`) regresses `main`'s existing daemon/drop `--json` error-envelope tests (they emit a truncated `{"schema":null,...}` envelope). #401 passes daemon_command_test (88/0) and drop_command_test (3/0) cleanly.

PR #278, opened 2026-06-03
feat(bot): capture ideas with photo and document attachments

ivankuznetsov:
## Summary

The Telegram bot can now capture ideas with photos and documents attached, not just text. An operator can send `/idea fix login` with a screenshot, drop a PDF and answer "what's the idea?", or pile several files onto one idea across multiple messages. Each idea commits through the same `Hive::Commands::New#call!` entry point the TUI uses, so attachments land in `<task_dir>/assets/` and downstream stages (1-inbox → 2-brainstorm …) see a uniform on-disk layout regardless of whether the idea came from the TUI or the bot.

The largest behavioral change: **the project tap no longer commits the idea.** Tapping a project moves the draft into a file-collection step; a new **Done**/**Skip** button commits. This lets files attached before *and* after the project pick land on the same idea.

## Idea composition state machine

A new per-chat `IdeaDraftStore` holds one in-progress idea per operator, with TTL pruning. Two entry paths converge on the same draft:

```mermaid
stateDiagram-v2
    [*] --> awaiting_text: /idea (no text)
    [*] --> awaiting_text: file, no caption
    [*] --> awaiting_project: /idea <text>
    [*] --> awaiting_project: file + caption
    awaiting_text --> awaiting_project: text answered
    awaiting_project --> collecting_files: project tapped
    collecting_files --> collecting_files: more files
    collecting_files --> [*]: Done / Skip → commit
```

Files attached at any phase append to the current draft, so albums and several separate files all attach to one idea.

## Architecture

Purity is preserved by keeping classification and validation free of I/O, mirroring how `ConversationStore` is shared between the router (writes) and supervisor (reads):

- **Router + handlers stay pure.** They classify updates by draft phase, run size / allowlist / count checks against the values Telegram already puts in the update payload, mutate the draft store, and emit `Result` objects. `IdeaAttachmentPolicy` is the single pure place that decides accept/reject.
- **Supervisor does all I/O.** It resolves `file_id` → bytes via Telegram `getFile` + download, stages bytes to a per-draft tmpdir, and runs the in-process `Commands::New#call!` commit.

A single Telegram update can carry both a photo and a caption, so the router can emit a `:stage_attachment` result that also carries reply text and markup — the supervisor stages the file, then sends one reply.

## Validation (enforced before any download)

| Check | Limit | Configurable key |
|---|---|---|
| File size | 20 MB | `idea_attachment_max_bytes` |
| Attachments per idea | 10 | `idea_attachment_max_count` |
[...]

PR #332, opened 2026-06-05
feat(bot): transcribe Telegram voice notes into ideas

ivankuznetsov:
## Summary

You can now send a voice note to the ideas bot and have it captured as an idea. The bot downloads the audio, transcribes it through an OpenAI-compatible Whisper endpoint, and shows the transcript with **[Confirm] [Discard]** buttons. Confirm runs the existing project picker and writes the idea through the same `Hive::Commands::New#call!` path text and photo ideas already use. Only the transcript is kept; the audio is discarded once transcription succeeds.

This extends the merged attachment pipeline (#278) rather than adding a parallel one. Voice ingestion reuses `Telegram#get_file`/`download_file`, the per-chat `IdeaDraftStore`, the project-picker keyboard, and the `commit_idea` capture path. What is genuinely new is the transcription client, a transcript-confirm draft phase, and the error/fallback handling.

### Flow

```mermaid
stateDiagram-v2
    [*] --> Transcribing: voice note received
    Transcribing --> AwaitingConfirm: :ok (transcript)
    Transcribing --> [*]: :no_speech / :unsupported_language (reply, nothing kept)
    Transcribing --> AwaitingText: :failed (raw audio staged as fallback)
    AwaitingConfirm --> Transcribing: reply with new voice note (re-transcribe)
    AwaitingConfirm --> AwaitingConfirm: reply with corrected text
    AwaitingConfirm --> AwaitingProject: Confirm
    AwaitingConfirm --> [*]: Discard
    AwaitingProject --> Captured: pick project -> New#call!
```

### What's new

| Piece | File | Role |
|---|---|---|
| Voice ingestion | `lib/hive/bot/telegram.rb` | Parses `message.voice` into a first-class `Update` field; songs/podcasts (`message.audio`) stay ignored, so the "voice notes only" scope holds for free |
| Transcriber | `lib/hive/bot/transcriber.rb` | Multipart POST to `/v1/audio/transcriptions` with `verbose_json`; returns `:ok` / `:no_speech` / `:unsupported_language` / `:failed`, with retry on transient failures |
| Config | `lib/hive/config.rb` | `bot.transcription.*` defaults + load-time validation; API key is env-only (`HIVE_WHISPER_API_KEY`), never persisted or logged |
| Draft phase | `lib/hive/bot/idea_draft_store.rb` | New `:awaiting_transcript_confirm` phase and `:voice` origin so a confirmed transcript commits straight after project pick |
| Routing | `lib/hive/bot/router.rb`, `handlers/*` | Classifies voice updates, the confirm/discard callbacks, and the edit-by-text / edit-by-voice loop |
| Orchestration | `lib/hive/bot/supervisor.rb` | Download, transcribe, confirm to picker, pick to commit, discard |

### Design decisions

- **Confirm runs before the project picker.** Unlike the photo flow (picker first, then file collection), a voice draft needs the operator to approve the transcript first, then picks a project and commits directly with no Done/Skip step.
- **Edit-by-reply is "next message while awaiting confirm."** Any text or voice sent during the confirm window is treated as the correction, mirroring the existing `:awaiting_text` behaviour. This avoids fragile `reply_to_message_id` plumbing; the 900s draft TTL bounds the window.
- **"No speech" and "unsupported language" are heuristics, not Whisper refusals.** No-speech keys off empty text plus a `no_speech_prob` threshold (default 0.6); unsupported language is "detected language not in `supported_languages`" (default `[en, ru]`, empty = accept all). Language matching normalizes both ISO codes and Whisper's full-name output (`en` ↔ `english`). Both are config-exposed.
- **Transcription failure falls back to retaining the audio.** After retries exhaust, the raw OGG is staged into the draft via a voice-specific internal path and the flow drops to `:awaiting_text` ("Couldn't transcribe; I saved the audio — what's the idea?"). This bot-internal staging deliberately bypasses the user-upload attachment allowlist.

### Provider note

OpenRouter has no audio-transcription endpoint, so transcription targets a configurable OpenAI-compatible endpoint (OpenAI `whisper-1`, Groq `whisper-large-v3`, or self-hosted Whisper) rather than the usual OpenRouter convention.

## Test plan

- **Unit / integration — deterministic, fully offline.** All transcription logic runs against an injected HTTP seam, so the core suite needs no network or secret:
  - `test/unit/bot/transcriber_test.rb` — ok→text, no-speech (`no_speech_prob` threshold), language gating (ISO + full-name), retry on transient 5xx, non-retried 4xx, malformed JSON, `Result` invariants, API-key-never-logged.
  - `test/unit/bot/supervisor_test.rb` — transcribe→confirm→picker→commit, edit-by-text, edit-by-voice, no-speech/unsupported-language replies (and nothing retained), `:failed` fallback staging, discard, oversize guard.
  - `test/integration/bot/scenarios/s6_voice_idea_test.rb` — full router→supervisor→`New#call!`: the happy path keeps **only** the transcript (no `assets/`), the fallback keeps `assets/voice-1.oga`.
  - Plus router, telegram, callback-handlers, idea-draft-store, config, and logger unit coverage.
- **Real-API E2E — secret-gated.** `test/e2e/tg/` sends a real voice note via Telethon against a live Whisper endpoint, gated on `HIVE_WHISPER_API_KEY` (mirroring the existing Telegram E2E's `HIVE_TEST_BOT_TOKEN`). It hard-fails — rather than silently skipping — when the secret is set but the fixture is absent.
- **Verification at HEAD:** full `bundle exec rake test` green (`0 failures, 0 errors`); RuboCop clean on all changed `lib/` files.

## Review summary

Three Compound-Engineering review passes (claude-ce + codex + pr-review-toolkit multi-persona) ran against `git diff origin/main..HEAD`.

- **Every Requirements-Trace item (R1–R7d, AE1–AE7) is implemented and test-covered.** No plan-required behavior is absent.
- **Notable fixes applied across passes:** the language gate now normalizes ISO codes vs. Whisper's full-name output (real recordings were being rejected as "unsupported"); `transcription_failed` was added to `Hive::Bot::Logger::EVENTS` (it was emitted but unregistered, crashing every failure path in production); `Result`/`Draft` invariants were hardened (status/phase/origin allowlists + constructor guards); the `:failed` fallback no longer leaves a contradictory draft or leaks staged audio into a later successful transcript; a hung-connect `open_timeout` was added; duplicated keyboard/config helpers were consolidated.
- **Scope:** an unrelated TUI/archive commit (`193c1e05`) that an early pass flagged was rebased out — the diff is now voice-only, matching the plan scope, which also cleared a `bubble_model.rb` stale-base diff artifact.
- **State at final pass:** all findings auto-fixed or resolved (deliberate-design items documented in-code); voice unit + integration suites green; RuboCop clean. No open escalations.

## Linked task

- **Hive task:** `Telegram Voice Note Ideas` (#36, slug `i-should-be-able-to-260602-0cb2`)
- **Original request:** "I should be able to send audio recording as an idea to telegram"

---


ivankuznetsov:

## High
- [ ] lib/hive/bot/transcriber.rb:92 rejects real OpenAI Whisper English/Russian transcripts as unsupported: the code compares raw verbose-json language strings to the default ISO-code allowlist (`en`, `ru`), but OpenAI verbose JSON returns names such as `english`, so the required happy path is likely blocked in production.
- [ ] test/e2e/tg/drive_idea.py:41 makes the required voice E2E silently skip when its fixture is missing: run_idea_e2e.sh defaults TG_VOICE_FIXTURE to test/fixtures/voice/voice-idea.oga, no such fixture is added, and the driver returns 0 instead of failing, so U8 is not actually implemented.

## Medium
- [ ] lib/hive/tui/bubble_model.rb:3168 removes terminal-height clipping for the two-pane TUI: TasksPane/ProjectsPane are now rendered without a height budget, so task lists can exceed @hive_model.rows and push the footer/selected row off-screen; the prior resize regression tests were deleted instead of preserving the behavior.

## Nit

ivankuznetsov:

# Code Review — Voice note → Whisper transcription idea flow (pass 1)

Scope reviewed: `git diff origin/main..HEAD`. Voice feature (U1–U9) verified against the plan's
Requirements Trace (R1–R7, AE1–AE7). All voice unit tests, `s6_voice_idea_test.rb` integration,
and rubocop on the changed `lib/hive/bot/*` + `config.rb` pass. The happy path retains only the
transcript (no `assets/`), the fallback retains audio — both proven by U7. No plan-required item
is missing.

## High
- [ ] None. Every plan Goal / Requirements-Trace item (R1–R7d, AE1–AE7) is implemented and covered by tests; no required behavior is absent.

## Medium
- [ ] Unrelated commit `193c1e05 "fix: improve tui task context and archive counts"` is bundled into this voice-feature branch: it touches `archive_filter.rb`, `commands/status.rb`, `patrol/review_handoff.rb`, and ~10 TUI files/tests with no connection to the plan — this widens the review/merge surface and risk for a feature PR that the plan scopes strictly to voice.
- [ ] Within that bundled commit, `lib/hive/archive_filter.rb` removes the `RESOLVED_MARKER_NAMES` (`:complete`/`:none`) allowlist so `hide?` now hides *all* archived tasks after 3 days regardless of marker — the deleted comment explicitly stated "Only resolved marker states may be hidden," so failed/non-terminal archived tasks become invisible; a behavioral change that deserves its own reviewed PR, not a terse rider on a voice branch.
- [ ] Duplicated keyboard/preview helpers: `transcript_preview_text` and `voice_confirm_keyboard` are defined identically in both `supervisor.rb:533,537` and `slash_handlers.rb:368,372`, and `project_keyboard` is duplicated in `callback_handlers.rb:226` and `slash_handlers.rb:358` — three copy-pairs that will drift (e.g. a wording or callback-data change applied in one copy only).
- [ ] Transcription defaults live in two places: `Transcriber::DEFAULT_CONFIG` (transcriber.rb) duplicates `DEFAULTS["bot"]["transcription"]` (config.rb) key-for-key — two sources of truth that can silently diverge; the supervisor always injects the full global config, so `DEFAULT_CONFIG` only fires for partial configs and risks masking a real default mismatch.

## Nit
- [ ] Dead field: `slash_handlers.rb` `voice(update, edit:)` packs `edit:` into the `transcribe_voice` payload, but `execute_transcribe_voice` never reads it (re-transcribe is handled transparently by `ensure_voice_draft`) — remove the flag or document why it is carried.
- [ ] Oversize guard in `execute_transcribe_voice` checks `get_file`'s `info[:file_size]` (which Telegram may omit, yielding a missed cap) while the authoritative `message.voice.file_size` is passed in the result payload and left unused — consider guarding on the payload size before the `get_file` round-trip.
- [ ] `no_speech?` (transcriber.rb) averages segment `no_speech_prob` only; a clip with a few very-high-prob segments but a low mean won't trip the `no_speech_threshold` — plan text mentioned "mean/segment"; acceptable but worth a comment so the heuristic's intent is explicit.
- [ ] Redundant header: `post` sets `req.headers["Content-Type"] = "multipart/form-data"` (no boundary), which the `:multipart` middleware overwrites with the boundary-bearing value — harmless but misleading.
- [ ] Inconsistent defensive guards in router `classify`: `update.respond_to?(:voice?) && update.voice?` vs. a bare `update.text?` two lines later (telegram.rb:24,37 both define the predicates, so neither guard is strictly needed) — pick one style.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — Voice note → Whisper transcription (pass 1)

Personas run: code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer.
Scope: `git diff origin/main..HEAD`. Union of all persona findings below, deduped. Voice unit/integration tests and rubocop pass; the high finding is hidden by unrealistic test fixtures.

## High
- [ ] [code] Language gate compares ISO codes to Whisper's full-name output — default config rejects every real voice note (`lib/hive/bot/transcriber.rb:92-95`, `unsupported_language?` does exact-match `include?` on raw `data["language"]`; default `supported_languages => %w[en ru]` at `config.rb:371`): OpenAI `verbose_json` returns `language: "english"/"russian"`, not `"en"/"ru"`, so `parse_response` returns `:unsupported_language` for normal recordings and the supervisor replies "unsupported language". Unit tests pass only because fixtures hardcode `"language" => "en"` (`test/unit/bot/transcriber_test.rb:79,98,145`), a value the real API never emits — diverging fixtures from reality masks the bug; the real-API E2E (U8) is secret-gated and may not run in CI.

## Medium
- [ ] [errors] `:failed` fallback can leave the draft in a contradictory state and emit a wrong message (`lib/hive/bot/supervisor.rb:482-497`): on transcribe failure it creates a draft, sets phase `:awaiting_text` (text nil), then `stage_voice_fallback`; if staging raises (`WriteError`/`SystemCallError`/`IOError`) the outer rescue replies "Couldn't download that voice note — please send it again" even though download succeeded and a live `:awaiting_text` draft now exists, so the user's next text is captured against an orphan draft.
- [ ] [errors] Single outer rescue conflates download, transcription, and staging failures into one misleading "Couldn't download" message (`supervisor.rb:490-497`) — a staging failure is non-actionable when reported as a download failure.
- [ ] [code/errors] `ensure_voice_draft` reuses ANY existing draft, so a voice note arriving mid-text/-media draft overwrites the operator's typed `text` and resets phase to `:awaiting_transcript_confirm` (`supervisor.rb:468-473`, `idea_draft_store.rb:514`); and the `:no_speech`/`:unsupported_language` branches `clear` (`supervisor.rb:475,479`) wipe a pre-existing in-progress draft + its staging dir — silent loss of unrelated user work. Reuse/clear should be gated on `origin == :voice`.
- [ ] [errors] 4xx responses (401 bad key, 429 rate-limit, 413 too-large) fall through to the broad `StandardError` rescue and surface as a generic `:failed`, indistinguishable from a transient 5xx (`transcriber.rb:46-59`); the hand-rolled `success?`/`transient_status?`/`failed("HTTPError"…)` branch is dead in production because the default client uses `faraday.response :raise_error` (`transcriber.rb:83`), so tested behavior diverges from shipped behavior.
- [ ] [errors] Broad `rescue StandardError => e ... :failed` (`transcriber.rb:57-59`) swallows programmer bugs (`NoMethodError`, `ArgumentError`) as user-facing "Couldn't transcribe", so genuine defects never surface as a noticeable crash.
- [ ] [errors] Transcriber `error_class`/`message` are never surfaced or logged with chat context at the supervisor layer (`supervisor.rb:466`); debugging a user's failure requires cross-correlating two log files by timestamp, and the `send_failure` line carries no `chat_id`.
- [ ] [types] `Transcriber::Result` lets illegal states be constructed (`transcriber.rb:20-24`): `status` is a free-form symbol (a typo like `:faild` builds cleanly) and a `:ok` result with `nil`/empty `text` is representable; the status→field contract is honored only by convention, with no constant set or constructor guard.
- [ ] [types] `IdeaDraftStore::Draft` phase/origin are unenumerated free symbols on a mutable Struct handed live to callers (`idea_draft_store.rb:11-13,38-49`), so any caller can bypass the transition methods (`set_transcript`/`confirm_transcript`/`await_text`) and a phase typo is undetectable; no invariant ties `:awaiting_transcript_confirm` to `origin == :voice`.
- [ ] [tests] `transcription.enabled: false` early-return is untested (`supervisor.rb:451-455`) — a regression inverting the guard (calling Whisper when disabled, or disabling for everyone) ships silently.
- [ ] [tests] The voice-fallback `:awaiting_text` router classification has no router-seam unit test (`test/unit/bot/router_test.rb` only covers `awaiting_transcript_confirm`) — a mis-route of plain text after fallback would only be caught by one integration test.
- [ ] [docs] `wiki/commands/bot.md:44` mis-orders the flow ("shows a transcript confirmation keyboard, then asks for the project"); the picker appears only after the operator taps Confirm (`supervisor.rb:469-473`), contradicting the accurate body text at lines 66-80.
- [ ] [comments] Hardcoded unsupported-language reply "Try English or Russian." (`supervisor.rb:481`) drifts into inaccuracy whenever `supported_languages` is reconfigured — derive the message from config.

## Nit
- [ ] [code] Unrelated commit `193c1e05` (TUI/archive/patrol) is bundled into the voice PR, inflating scope and review history; the bundled files read cleanly but belong in a separate focused PR.
- [ ] [code/conv] Default `endpoint`/`api_key_env` point at OpenAI rather than the project-standard OpenRouter (`transcriber.rb:10`, `config.rb:364`) — OpenRouter has no audio-transcription endpoint, and the plan flags this as open question Q1, so it is a deliberate, plan-acknowledged decision needing a wiki note, not a defect.
- [ ] [errors] Retry exhaustion is invisible — only the final failure logs, with no attempt count (`transcriber.rb:50`), so a one-shot failure is indistinguishable from one that burned `max_retries × backoff`.
- [ ] [errors] `retry_after?` hides a blocking `sleep` inside a boolean predicate (`transcriber.rb:120-124`) — a future reuse of the predicate gets an unexpected side effect.
- [ ] [errors] `no_speech?` silently drops malformed `no_speech_prob` per segment with no log (`transcriber.rb:104-108`); if all segments are malformed it returns `false` (treats as speech).
- [ ] [code] `no_speech?` uses an unweighted mean across segments (`transcriber.rb:100-113`) — a short silent segment is weighted equally with a long spoken one; threshold-tunable but can misclassify.
- [ ] [tests] Retry backoff sleep is injected but never asserted (`transcriber_test.rb:73` stubs it to no-op with `retry_backoff_sec => 0`) — dropping the sleep (hammering Whisper) passes all tests.
- [ ] [tests] `no_speech` averaging is only tested above threshold (0.75 ≥ 0.6) — no below-threshold `:ok` case, so flipping `>=`/`<` is half-caught.
- [ ] [tests] Key-never-logged assertion exists only on the retry-exhaustion path (`transcriber_test.rb:168`); the malformed-JSON and 4xx failure paths also log via `log_failure` without re-asserting key absence.
- [ ] [types] `voice` is an untyped Hash `{file_id,…}` (`telegram.rb:223-229`) — primitive obsession consistent with the existing photo/document shape, but a new field entrenching the smell; a tiny `MediaFile`/`VoiceFile` value type would help.
- [ ] [types] `media?` deliberately excludes `voice` so `effective_text` returns `text` not `caption` for voice (`telegram.rb:33-43`) — correct but undocumented asymmetry a future reader could "fix" and break.
- [ ] [types] `TransientHttpError` is declared `private` mid-file after its `raise`/`rescue` uses (`transcriber.rb:48-49,65`) — works (constant) but reads oddly.
- [ ] [comments] `lib/hive/archive_filter.rb` accepts `marker_name:` but never reads it (kept for caller compatibility) with no note, and the load-bearing `mtime || folder_mtime` precedence (prefer immutable row `mtime` so sidecar edits don't resurface archived tasks) carries no rationale comment — a refactor could silently reintroduce the bug.
- [ ] [code] Manual `Content-Type: "multipart/form-data"` header (`transcriber.rb:70`) is redundant — faraday-multipart appends the boundary; harmless but misleading.
- [ ] [code] Oversize guard trusts Telegram's advertised `info[:file_size]` (`supervisor.rb:459-460`); if absent/zero the guard is skipped and an oversized download proceeds — the authoritative `message.voice.file_size` is carried in the payload but unused.

ivankuznetsov:

## High
- [ ] Missing checked-in voice E2E fixture: `test/e2e/tg/run_idea_e2e.sh:22` defaults `TG_VOICE_FIXTURE` to `test/fixtures/voice/voice-idea.oga` and `test/e2e/tg/drive_idea.py:45` hard-fails when the Whisper secret is set but the file is absent; only `test/fixtures/voice/README.md` is tracked, so U8's real voice-note E2E cannot run.

## Medium
- [ ] Two-pane TUI no longer clamps pane height: `lib/hive/tui/bubble_model.rb:3160` stopped computing terminal-row budget and `TasksPane.render`/`ProjectsPane.render` no longer accept `height`, so snapshots with more rows than the terminal can push the footer and selected row off-screen; the previous regression tests for height clipping were removed.

## Nit

ivankuznetsov:

# Code Review — Voice note → Whisper transcription idea flow (pass 2)

Scope reviewed: `git diff origin/main..HEAD`. Pass-1 auto-fixes are confirmed
landed in `8fc42af3` (language-gate ISO/full-name normalization, single-source
`DEFAULT_CONFIG`, payload-size oversize guard, voice-origin-gated draft
reuse/clear, dead `edit:` flag removed, shared `IdeaKeyboards`, narrowed
rescues, `STATUSES`/`PHASES`/`ORIGINS` guards). Verified green locally:
`transcriber_test.rb` (12/35), `supervisor_test.rb` (111/443),
`s6_voice_idea_test.rb` (5/22); `rubocop` clean across the 9 changed
`lib/hive/bot/*` + `config.rb` files. The voice unit/integration logic (R1–R7d,
AE1–AE7) holds up: happy path retains only the transcript, fallback retains the
audio, draft-clobber edge is now origin-gated.

## High
- [ ] Missing checked-in voice E2E fixture (`test/fixtures/voice/voice-idea.oga`): plan U8 Files explicitly lists "a small WAV/OGG speech fixture under `test/fixtures/`", but only `README.md` is tracked — `drive_idea.py:45` hard-FAILs (returns 1) when `HIVE_WHISPER_API_KEY` is set and the file is absent, so U8's real-API voice E2E cannot actually run and the plan's "passes locally with a key set" verification is unmet. Plan-required, not plan-deferred. (concurs with codex-02 High)

## Medium
- [ ] Unrelated commit `193c1e05` ("fix: improve tui task context and archive counts") is still bundled into this voice branch (verified not on `origin/main`): touches `archive_filter.rb`, `commands/status.rb`, `patrol/review_handoff.rb`, `tui/views/format.rb`, `tui/views/tasks_pane.rb` — none in the plan's strictly-voice scope. Still unresolved from pass-1 escalation **Q1** (A1 blank). The `archive_filter.rb` part is a real product-behavior change (drops the `RESOLVED_MARKER_NAMES` `:complete`/`:none` allowlist, so `hide?` now hides ALL archived tasks after 3 days regardless of marker — failed/non-terminal archived tasks become invisible) that deserves its own reviewed PR, not a rider on a voice branch.
- [ ] Branch is behind `origin/main` by one commit — `b02c5fae` "Fix TUI vertical resize viewport (#326)", which this branch does not contain. The `bubble_model.rb`/`projects_pane.rb`/`tasks_pane.rb` "pane height no longer clamped + clipping tests removed" that shows up in `origin/main..HEAD` (codex-02 Medium) is a merge-base artifact of #326's absence, NOT a regression introduced here. Rebasing/merging onto current `origin/main` restores the resize fix and clears the artifact (will need a small `tasks_pane.rb` conflict resolution since `193c1e05` also edits it). Recommend rebase before merge.

## Nit
- [ ] Edit-by-voice loses a good transcript on a bad re-record: while a voice draft is `:awaiting_transcript_confirm`, a second voice note that comes back `:no_speech`/`:unsupported_language` runs `clear_voice_draft` and wipes the already-valid prior transcript, forcing a full restart — during an explicit edit it would be friendlier to keep the prior transcript and just reply with the reject reason (`supervisor.rb:498,576`).
- [ ] `idea_voice_confirm` with zero registered projects replies "No Hive projects are registered yet." but leaves the draft stuck in `:awaiting_transcript_confirm` with no surfaced next step (no clear/retry hint) — minor dead-end (`callback_handlers.rb:188`).
- [ ] Unset/empty `HIVE_WHISPER_API_KEY` sends `Authorization: Bearer ` (empty) and degrades to the failure→fallback path ("Couldn't transcribe; I saved the audio") rather than a clear "transcription not configured" signal; a load-time or first-use operator warning would make a missing-key misconfig diagnosable instead of looking like a Whisper outage (`transcriber.rb:228`).

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — Voice note → Whisper transcription (pass 2)

Personas run: code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer.
Scope: `git diff origin/main..HEAD` in the worktree. Union of all persona findings below, deduped.

Verification done this pass (read-only): all voice-related unit suites green
(transcriber 12, supervisor 111, router 36, draft_store 10, telegram 26,
callback_handlers 31, config 176), the `s6_voice_idea_test.rb` integration
suite green (5/22), and the bundled-commit suites (archive_filter, patrol,
status, bubble_model, snapshot) green. `rubocop` clean on the 8 touched lib
files.

Pass-1 High findings are **resolved**: the language gate now normalizes
Whisper full-name output to ISO codes (`Transcriber::LANGUAGE_NAME_TO_ISO`,
`transcriber.rb:24-29`) and the test fixtures emit realistic
`"language" => "english"/"russian"/"de"` values (`transcriber_test.rb:84,100,150`),
so the gate is no longer masked. The pass-1 Medium error-handling, draft-clobber,
4xx-distinct, and type-hardening findings are all addressed and covered by tests.

## High
- (none)

## Medium
- [ ] [pr-test-analyzer] U8's checked-in speech fixture is missing: only `test/fixtures/voice/README.md` is tracked — `voice-idea.oga` is absent (`git ls-files` shows just the README; the file is not gitignored, just not committed). The plan lists "a small WAV/OGG speech fixture under `test/fixtures/`" as a U8 deliverable and the driver hard-FAILs (not skips) when `HIVE_WHISPER_API_KEY` is set but the fixture is missing (`drive_idea.py:44-50`), so the real-API E2E cannot run end-to-end as designed even locally with a key. The deterministic integration tests (U7) do cover the capture path, and the E2E is secret-gated with provider/CI still an open question (plan Q2), so functionality is verified — but the plan deliverable is not present.

## Nit
- [ ] [maintainability] Two parallel language maps must be hand-synced: `Transcriber::LANGUAGE_NAME_TO_ISO` (name→ISO, `transcriber.rb:24-29`) and `Supervisor::LANGUAGE_DISPLAY` (ISO→name, `supervisor.rb:37-43`). Adding a supported language requires editing both files or the unsupported-language gate and the operator-facing hint silently drift apart.
- [ ] [silent-failure-hunter] An unset `HIVE_WHISPER_API_KEY` is indistinguishable from a transient outage at runtime: `api_key` falls back to `""` (`transcriber.rb:285` → `@env.fetch(name, "")`), the empty Bearer yields a 401 → `failed("HTTPError", "...HTTP 401")`, and the supervisor then stages the audio and replies "Couldn't transcribe; I saved the audio". Config validation only checks `api_key_env` is a non-empty *name*, never that the env value is populated. The 401 is logged with `chat_id` so it is debuggable, but a pure misconfiguration looks to the operator like a flaky provider rather than a setup error.
- [ ] [code-reviewer] Plan U5 specifies the `:transcribe_voice` Result as `{chat_id, file_id, file_size, edit: <bool>}`, but the shipped Result omits `edit:` entirely (no `edit:` anywhere in slash_handlers/supervisor/router). The edit-vs-fresh decision is instead derived from `draft.origin == :voice` in `ensure_voice_draft`, which works correctly — but it is an undocumented deviation from the planned Result contract and the tests assert no `edit` field, so the contract drift is invisible.
- [ ] [code-reviewer] Unrelated commit `193c1e05` ("fix: improve tui task context and archive counts" — TUI/archive_filter/patrol/status) remains bundled into the voice PR, inflating the diff and review surface. Re-surfaced from pass-1 (escalation Q1); the bundled files read cleanly and pass their suites, but they belong in a separate focused PR.

ivankuznetsov:

# Code Review — Voice note → Whisper transcription idea flow (pass 3)

Scope: `git diff origin/main..HEAD` in the worktree, re-reviewed against the plan's Requirements
Trace (R1–R7d, AE1–AE7). State at HEAD is materially cleaner than pass 1/2:

- The pass-2 **red test regression is fixed** — `idea_draft_store_test.rb` and
  `slash_handlers_test.rb` now pass (`origin: :voice` added). I ran the full voice suite:
  `transcriber` (13), `supervisor` (118), `router` (37), `telegram` (26), `callback_handlers`
  (31), `idea_draft_store` (10), `config` (179), and `s6_voice_idea` integration (5) — **all green,
  0 failures/0 errors**.
- Rubocop clean on all 10 changed `lib/` files.
- The previously-bundled out-of-scope TUI/archive commit (`193c1e05`) has been rebased out; the
  diff is now voice-only, matching the plan scope.
- The R6/AE1 transcript-only guarantee, the AE4 fallback retention, and the language-gate ISO/
  full-name normalization (the pass-1 cross-ref Highs) are all implemented and test-covered.
- No plan Goal / Requirements-Trace item (R1–R7d, AE1–AE7) is missing or unimplemented. No
  plan-required-but-absent gap to escalate.

Two new, lower-severity observations remain that prior passes did not raise. Neither is a
correctness bug; both are reliability/UX edges the user may consciously accept for the MVP.

## High
- [ ] None: every plan Goal / Requirements-Trace item (R1–R7d, AE1–AE7) is implemented, test-covered, and the suite is green — no required behavior absent and no blocking defect found.

## Medium
- [ ] Synchronous transcription on the poll thread can stall **all** bot update processing for minutes: `execute_transcribe_voice` → `@transcriber.call` runs inline in `poll_loop`→`process_update` (supervisor.rb:744–746), and the transcriber retries up to `max_retries`+1 POSTs (transcriber.rb:69–103) each bounded only by `req.options.timeout = timeout_sec` (default 120s; **no `open_timeout` set**, so a hung TCP connect is unbounded). Worst case ≈ 4×120s + 3×2s ≈ 8 min during which the poll loop issues no new `poll_updates`, so /status, /approve, confirm/discard taps, and any other operator's messages all wait and Telegram redelivers un-acked updates. This is the exact failure mode the codebase deliberately avoids for `hive status` ("running it on the poll thread would let a hung `hive status` stall the long-poll loop", supervisor.rb:`latest_status_rows`); transcription is the longest blocking call in the bot yet runs on that same thread. Consider lowering the default `timeout_sec`, setting an `open_timeout`, and/or moving transcription off the poll thread.

## Nit
- [ ] Edit-by-voice is destructive on a *failed* retry, leaving the operator worse off than before: while a good transcript is `:awaiting_transcript_confirm`, a second voice note that returns `:no_speech` (`clear_voice_draft`, supervisor.rb:779) or `:failed` (`handle_failed_transcription`→`await_text` nulls `draft.text`, supervisor.rb:798–799) wipes the previously-confirmed-pending transcript entirely — the operator loses transcript #1 because replacement #2 didn't transcribe, then must start over. The plan's R5/AE5 ("re-record") don't specify preserving the prior transcript, so this is acceptable, but keeping transcript #1 on a failed re-record (reply "couldn't transcribe that — your previous transcript is still pending Confirm") would be friendlier.
- [ ] With the API key env unset, `Transcriber#api_key` logs the distinct `ApiKeyNotConfigured` signal (transcriber.rb:258–266) but still issues the doomed `Authorization: Bearer ` POST, producing a second `HTTPError` event per voice note; a short-circuit to `:failed` on empty key would save the round-trip. Low value — the current design is a deliberate pass-2 choice (let the real provider 4xx surface too), so noted only.

ivankuznetsov:

# PR Review Toolkit — Voice note → Whisper transcription (pass 3)

Multi-persona review of `git diff origin/main..HEAD`. Personas run: code-reviewer, silent-failure-hunter,
type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier. Tests were executed; rubocop is
clean on the 10 changed `lib/` files and the voice unit/integration suites are green **against the stub
loggers used in tests**. That last qualifier is the headline finding: the green suite masks a 100%-reproducible
production crash on every transcription-failure path.

## High
- [ ] [silent-failure] `:transcription_failed` is emitted in 8 places (transcriber.rb:166,235,247,260; supervisor.rb:477,494,520,533) but is **not** in `Hive::Bot::Logger::EVENTS` (logger.rb:11-44, and logger.rb is **not** in this diff), so `Logger#event` raises `ArgumentError: unknown bot log event` on every failure path — I reproduced it directly (`RAISED: ArgumentError: unknown bot log event: :transcription_failed`): in production `@logger` is the real Logger so the `&.` doesn't help, the transcriber's `#call` rescues don't catch `ArgumentError` (it deliberately excludes programmer errors) so `Transcriber#call` *raises* instead of returning `Result(:failed)`, and in the supervisor the raise precedes every `safe_send_message`, so the AE4 fallback + "couldn't download/transcribe/save" replies never send → a silent no-reply crash to the poll-loop `:fatal` handler — exactly the failure mode the project forbids. One-line fix: add `transcription_failed` to `EVENTS`.

## Medium
- [ ] [test] The missing-whitelist crash slipped through because every voice test injects a permissive stub logger (`StubLogger` transcriber_test.rb:41 + s6_voice_idea_test.rb:47; `FakeLogger` supervisor_test.rb:34) that accepts any event name — none mirror `EVENTS`. Add a regression that exercises a failure path against the real `Hive::Bot::Logger` (or asserts every event symbol emitted across `lib/` is a member of `EVENTS`), so a future missing-event can't hide again.
- [ ] [type-design] `IdeaDraftStore::Draft` (idea_draft_store.rb:25-27) is a mutable `keyword_init` Struct handed live to callers with public field writers, so the phase/origin coupling is enforced only at the store's transition methods, not at the type boundary — a future caller doing `draft.phase = :awaiting_transcript_confirm` directly bypasses `validate_phase_origin!`. Logic is complete across store methods (prior pass closed the `assign_phase!` hole); the residual gap is encapsulation. Consider making the phase/origin writers private.
- [ ] [comment] telegram.rb:33-37 — the `media?` comment claims voice is "routed separately via `voice?` (checked before effective_text is ever consulted)", but `Router#classify` consults `effective_text` unconditionally (router.rb:124) *before* the `voice?` check (router.rb:142). The comment's conclusion (don't include voice in `media?`) is correct, but its stated mechanism contradicts the router; reword to the true reason (if `media?` included voice, the unconditional `effective_text` at the top of `classify` would resolve to the voice caption and could misroute a captioned voice note through slash-command matching).
- [ ] [test] `Transcriber::Result`'s five self-validating invariants (transcriber.rb constructor: unknown status; `:ok` empty-text; `:failed` carrying text; `:failed` without error_class/message; benign status carrying error metadata) have **no** direct-construction test — a refactor loosening any guard would let `parse_response` emit a contradictory result with zero test failures. Add a small table-driven `assert_raises(ArgumentError)` test.
- [ ] [test] `IdeaDraftStore::validate_phase!`/`validate_origin!`/`validate_phase_origin!` raise-paths are untested — the coupling is relied on elsewhere but the guard that enforces it is never asserted to fire; add three `assert_raises(ArgumentError)` cases (unknown phase, unknown origin, `:awaiting_transcript_confirm` + non-voice origin).
- [ ] [test] The `ApiKeyNotConfigured` distinct signal (transcriber.rb:258-266) and the all-malformed-segments → `:ok` edge (transcriber.rb `no_speech?` empty-probs return) are both documented deliberate behaviors with no test (every transcriber test sets the key and uses ≤1 malformed segment); add an `env: {}` test and an all-malformed-segments test.

## Nit
- [ ] [code-reviewer/silent-failure] supervisor.rb `handle_failed_transcription` calls `ensure_voice_draft` (which returns `nil` for a non-voice draft) then unconditionally `await_text`/`stage_voice_fallback` on `chat_id`; the `:ok` path guards with `draft && set_transcript(...)` but the failed path doesn't. Currently unreachable (serial poll thread + the earlier `non_voice_draft?` guard), but a symmetric `return ... unless draft` would match the `:ok` path and the file's own defensive style.
- [ ] [code-reviewer] transcriber.rb `parse_response` — a non-empty transcript with a blank `language` (when `supported_languages` is non-empty) is classified `:unsupported_language` and dropped, because `normalize_language("")` is `nil` ∉ list; safer to gate only when the language is *known* and not in the list. Very low likelihood (verbose_json essentially always sets `language`).
- [ ] [code-reviewer] router.rb — a voice note sent during an active brainstorm conversation routes to `:free_text_answer` (conversation check precedes voice check), yielding a brainstorm-answer refusal rather than voice capture; graceful (not silent) and a deliberate precedence choice, flagged only for discoverability.
- [ ] [comment] supervisor.rb:459-462 — "The payload's voice.file_size is authoritative" overstates it; the payload size is the *earlier/cheaper and reliably-present* check, not more trustworthy. Reword to "checked first because it's always present and avoids the round-trip; getFile's file_size is sometimes omitted, so the post-getFile check is a fallback."
- [ ] [type-design] transcriber.rb `Result` — `:no_speech`/`:unsupported_language` have no text constraint, so a `:no_speech` is structurally indistinguishable from `:ok` minus the empty-text rule, and `:ok` doesn't require `language`; both intentional/latent, noted for completeness only.
- [ ] [type-design] telegram.rb:38-44 — `media?` and `voice?` are not mutually exclusive at the type level (a malformed update could set both); Telegram never sends both and router ordering makes precedence deterministic, so theoretical only.
- [ ] [simplify] transcriber.rb:201-203 — `success?` computes `status(response)` twice (double `value` dispatch); bind a local `code = status(response)` or use `(200...300).cover?(...)`.
- [ ] [simplify] transcriber.rb:227-233 — `value` helper's leading `return nil if object.nil?` is redundant (nil isn't a Hash and `nil.respond_to?` is false, so it already falls through to the final `nil`); borderline since it documents the nil-response intent.
- [ ] [test] transcriber.rb 4xx path appends a truncated response-body hint (the `invalid_api_key`/`file-too-large` reason) but `test_non_transient_400_fails_without_retry` asserts only status + key-redaction, not that the body reason reaches `result.message`; assert it to pin the debuggability contract.

PR #318, opened 2026-06-05
Hive patrol: Leading --json changes e2e subcommand dispatch

ivankuznetsov:
## Summary

`bin/hive-e2e` declares `--json` as a Thor class option, but a leading `--json`
(before the subcommand) was consumed by the default `run_scenarios` task — so
`bin/hive-e2e --json list` treated `list` as a scenario pattern and returned a
`no_scenarios` error instead of listing scenarios; `--json clean` and
`--json replay ...` failed the same way.

This PR adds `normalize_leading_class_options!`, which strips any leading
`--json`/`--no-json` tokens and re-inserts them immediately after a recognized
subcommand (`clean`, `help`, `list`, `replay`, `run`, `version`) before Thor
dispatch. Unknown leading tokens keep their original order, so the default
`run_scenarios` pattern path is preserved (e.g. `--json some-pattern` still runs
the pattern with JSON output).

## Test plan

Four contract tests added to `test/e2e/lib/hive_e2e_binary_test.rb`, each driving
the real `bin/hive-e2e` binary via `Open3.capture3`:

- `test_leading_json_list_dispatches_to_list` — `--json list` exits 0 and emits
  the `hive-e2e-scenarios` schema.
- `test_leading_json_clean_dispatches_to_clean` — `--json clean` (against a temp
  `HIVE_E2E_RUNS_DIR`) exits 0 and emits the `hive-e2e-clean` schema.
- `test_leading_json_replay_dispatches_to_replay` — `--json replay missing-run
  missing-scenario` exits 78 with a `missing_repro` error envelope.
- `test_leading_json_unknown_token_still_uses_default_run_pattern` — `--json
  definitely-no-scenario` still follows the default run path, exiting 64 with a
  `no_scenarios` error for that pattern.

Validation on the branch: `bundle exec rubocop` clean, `bundle exec rake test`
passing.

## Review summary

Patrol opened this PR from finding `command-bin-hive-e2e-2`; it ran through the
standard 6-review flow (pass 1 complete, browser checks skipped). All five
reviewer findings were artifacts of a two-dot `git diff origin/main..HEAD` taken
against a diverged `origin/main` — that diff renders main's newer commits as
spurious deletions of files this PR never touches. The actual PR diff
(`origin/main...HEAD`) is only `bin/hive-e2e` (+12) and
`test/e2e/lib/hive_e2e_binary_test.rb` (+41), plus wiki notes. No code change was
in scope from the review; the findings were resolved no-fix.

## Linked task

[...]

PR #317, opened 2026-06-05
Hive patrol: Human usage errors exit 1 instead of documented 64

ivankuznetsov:
## Summary

`bin/hive-e2e` documents usage errors as exit code `64` (`EX_USAGE`) and the
`--json` path already maps `Thor::Error` to `64`, but the human (non-`--json`)
path let Thor handle usage failures internally and exit `1`. As a result,
unknown commands and missing required arguments exited `1` — indistinguishable
from a scenario failure — so CI and shell wrappers not using `--json` could not
branch on bad invocations.

This change unifies the error-exit contract so both human and `--json` usage
failures exit `64`:

- Start Thor with `debug: true` unconditionally so `Thor::Error` re-raises into
  the binary's outer rescue instead of being handled internally.
- In the outer rescue, map usage failures to `Hive::E2E::BinaryExit::USAGE`
  (`64`) for both output modes, preserving JSON-envelope output under `--json`
  and prose-to-stderr output otherwise.
- Prefix the human-path stderr message with `hive-e2e:` to match the other
  human error rescues (`PreflightError`, `StandardError`) for consistent CLI
  output.
- Update the rescue/`exit_on_failure?` comments to describe the unified
  behavior.

Verified empirically: `bin/hive-e2e no-such-command`, `bin/hive-e2e replay`,
and their `--json` variants all exit `64`.

## Test plan

- `test/e2e/lib/hive_e2e_binary_test.rb` pins the contract:
  - `test_unknown_command_exits_usage_code` — unknown command exits `64` and
    prints a `hive-e2e:` prose error to stderr.
  - `test_missing_required_args_exits_usage_code` — missing required args exit
    `64` and print a `hive-e2e:` prose error to stderr.
  - `test_missing_required_args_with_json_emits_envelope_on_stdout` — tightened
    from `refute_equal 0` to `assert_equal 64`, with the JSON envelope on stdout
    and empty stderr.
  - `test_unknown_command_with_json_emits_envelope_on_stdout` — unknown command
    + `--json` exits `64` with the envelope on stdout (symmetric coverage).
- The human-path tests assert the stderr message survives, guarding against a
  regression that silently drops the `warn`.
- Focused suite result: `13 runs, 75 assertions, 0 failures`.

## Review summary

Two CE review passes ran (claude + codex, two passes each).

- The branch's real change is scoped to `bin/hive-e2e`, its test, and the
  `wiki/` docs. Several codex findings citing `lib/hive/stages/review.rb`,
[...]

PR #338, opened 2026-06-06
Hive patrol: Git environment variables bypass the dry-run guard

ivankuznetsov:
## Hive Patrol Finding

Slice: `command-bin-hive-babysitter-stub-git`
Category: `security`
Severity: `high`
Confidence: `high`
Fingerprint: `a1cb0feb156e283117dfcbebe099a8577283c988ebbf5edff61e5a0d702854f9`

The stub never screens or clears git environment variables before execing the real binary. An agent can set variables such as `GIT_EXTERNAL_DIFF` and then call an allowed read command like `git diff`; Git executes that external command while the stub records no skip. This bypass does not require any unsafe argv token, so the existing allowlist does not protect the dry-run sandbox.

## Evidence

- bin/hive-babysitter-stub-git:65 rest = stripped_global_options(argv)
- bin/hive-babysitter-stub-git:78 exec(real, *argv)

## Applied Fix

Before passthrough, fail closed when executable/config-injection git env vars are present, including `GIT_EXTERNAL_DIFF`, `GIT_SSH_COMMAND`, `GIT_SSH`, `GIT_PROXY_COMMAND`, `GIT_CONFIG_PARAMETERS`, `GIT_CONFIG_GLOBAL`, `GIT_CONFIG_SYSTEM`, and positive/nonzero `GIT_CONFIG_COUNT`. Cover at least `GIT_EXTERNAL_DIFF` and config-count injection in the dry-run tests.

## Validation

## Diffstat

```text
 bin/hive-babysitter-stub-git             | 26 +++++++++++++++++++++++---
 test/unit/babysitter/dry_run_env_test.rb | 27 +++++++++++++++++++++++++--
 2 files changed, 48 insertions(+), 5 deletions(-)

```

### DIFF
--- a/bin/hive-babysitter-stub-git
+++ b/bin/hive-babysitter-stub-git
@@ -193,6 +193,13 @@ if real.empty?
   warn "hive-babysitter dry-run: HIVE_BABYSITTER_REAL_GIT is unset; refusing to guess a path for the real git binary"
   exit 127
 end
+
+# Even an allowlisted read-only subcommand executes arbitrary commands if these
+# env vars reach real git, and argv screening structurally cannot see them.
+# Scrub the whole exec-influencing family before handing off.
+%w[GIT_EXTERNAL_DIFF GIT_PAGER GIT_PROXY_COMMAND GIT_SSH GIT_SSH_COMMAND].each { |key| ENV.delete(key) }
+ENV.keys.grep(/\AGIT_CONFIG/).each { |key| ENV.delete(key) }
+
 begin
   # Array-form exec (no shell) hands argv straight to the real git binary.
   exec real, *argv
--- a/test/unit/babysitter/dry_run_env_test.rb
+++ b/test/unit/babysitter/dry_run_env_test.rb
@@ -28,11 +28,12 @@ class BabysitterDryRunEnvTest < Minitest::Test
       real_gh = recording_binary(dir, "real-gh")
       real_git = recording_binary(dir, "real-git")
       log_path = File.join(dir, "skipped.log")
-      env = sanitized_git_env.merge(
+      @real_log = File.join(dir, "real.log")
+      env = {
         "HIVE_BABYSITTER_REAL_GH" => real_gh,
         "HIVE_BABYSITTER_REAL_GIT" => real_git,
         "HIVE_BABYSITTER_DRY_RUN_LOG" => log_path
-      )
+      }
 
       assert_stubbed env, "gh", "-R", "owner/repo", "pr", "ready", "42"
       assert_stubbed env, "gh", "--repo=owner/repo", "pr", "close", "42"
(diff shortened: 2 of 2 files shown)


ivankuznetsov:

# Codex CE Code Review 01

No plan was found in the task folder, so this review is based on `git diff origin/main..HEAD` only.

## High
- [ ] `lib/hive/tui/bubble_model.rb:3160` stops budgeting pane height from `model.rows`, and `ProjectsPane.render` / `TasksPane.render` no longer accept `height:`, so a small terminal or resize with many tasks renders all rows and can push the footer/current selection off-screen; this also deletes the prior vertical resize smoke coverage.

## Medium
- [ ] `lib/hive/archive_filter.rb:16` now hides every old `9-done` row regardless of marker state, so archived `ERROR`, `WAITING`, or `AGENT_WORKING` tasks disappear from daily status/TUI views after three days instead of remaining visible for operator recovery.

## Nit

ivankuznetsov:

# PR Review — pr-review-toolkit (pass 1)

- **Branch:** `hive-patrol/command-bin-hive-babysitter-stub-git-a1cb0feb`
- **Compared against:** `origin/main`
- **Plan grounding:** No `plan.md` found in the task folder — reviewed without plan grounding.
- **Personas run:** code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, ce-security-reviewer.

> **Diff scope note:** the merge-base of this branch and `origin/main` is `23cf47b9`. The
> branch does **not** contain PR #326 ("Fix TUI vertical resize viewport", `b02c5fae`),
> so `git diff origin/main..HEAD` *appears* to delete all of #326's viewport code/tests.
> Those deletions are a staleness artifact, captured once as the `[integration]` High
> finding below — individual per-file viewport deletions are intentionally not re-listed.
> Several `[security]` findings concern code outside the literal diff (`stripped_global_options`,
> the `exec`), but are in-scope because this PR's stated purpose is to harden that exact guard.

## High
- [ ] [integration] Branch is stale and reverts PR #326: merging this tree onto `origin/main` silently undoes #326's TUI vertical-resize viewport fix — `pane_height_for`/`fit_rows`/`fit_lines`/`viewport_start`, the `height:` params on pane renderers, the e2e height scenarios, and the `test_charm_tui_vertical_resize_keeps_panes_footer_and_selected_row_visible` integration test are all removed by the diff. Rebase onto `origin/main` and reconcile the overlapping `tasks_pane`/`bubble_model` edits (the branch's display-width work and #326's viewport work touch the same methods).
- [ ] [security] Inline `-c <key>=<val>` config injection bypasses the dry-run guard (RCE): `stripped_global_options` strips `-c core.pager=CMD` from `rest`, so `read_only?(["log"])` passes, but line 98 execs the full original `argv` *with* the malicious `-c`. Verified end-to-end (payload file written, skip log empty); `core.pager`/`core.fsmonitor`/`alias`/`diff.external` all yield code execution. `bin/hive-babysitter-stub-git:17-18,98`.
- [ ] [security] `unsafe_git_env?` blocklist omits `GIT_PAGER`/`PAGER`: read-only `log`/`diff`/`show`/`grep` invoke the pager, executing `$GIT_PAGER` (on a tty, or forced via `--paginate`/`-c pager.*=true`). Verified end-to-end. `bin/hive-babysitter-stub-git:65-73`.
- [ ] [security] Blocklist omits `GIT_DIR`/`GIT_WORK_TREE` (and `GIT_OBJECT_DIRECTORY`/`GIT_ALTERNATE_OBJECT_DIRECTORIES`/`GIT_COMMON_DIR`/`GIT_NAMESPACE`): pointing git at an attacker-planted repo whose local config sets `core.fsmonitor` executes code on a plain read-only `git status`. Verified end-to-end. `bin/hive-babysitter-stub-git:65-73`.
- [ ] [tests] 6 of 7 `UNSAFE_GIT_ENV` sentinels are untested — `dry_run_env_test` only exercises `GIT_EXTERNAL_DIFF`; a dropped/typo'd entry for `GIT_SSH_COMMAND`/`GIT_SSH`/`GIT_PROXY_COMMAND`/`GIT_CONFIG_PARAMETERS`/`GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` would silently reopen a sandbox escape with no failing test. `test/unit/babysitter/dry_run_env_test.rb`.

## Medium
- [ ] [tests] `GIT_CONFIG_COUNT="0"` safe branch (plus non-integer/negative strings) is untested — only `"1"` is covered; this is the lone "allow" path in the env guard, so flipping `positive?`/`negative?` to the wrong logic would not fail any test. `test/unit/babysitter/dry_run_env_test.rb`.
- [ ] [security] `GIT_TRACE`/`GIT_TRACE2*` not blocked: accept an absolute path that git creates/appends, allowing arbitrary out-of-sandbox file writes (e.g. clobber configs) via read-only `git status`. Verified. `bin/hive-babysitter-stub-git:65-73`.
- [ ] [security] Blocklist also misses `GIT_TEXTCONV`, `GIT_ATTR_SOURCE`, `GIT_ASKPASS` — diff/show textconv attribute path and credential-prompt exec vectors. `bin/hive-babysitter-stub-git:65-73`.
- [ ] [security] Denylist approach is structurally fragile given the stated threat model (supervised agent controls the full environment); prefer an allowlist — exec git with a sanitized env (only `PATH`/`HOME`/`TERM`…), force `--no-pager`/`GIT_PAGER=cat`, strip every other `GIT_*`/`PAGER`. `bin/hive-babysitter-stub-git:65-91`.
- [ ] [errors] `evidence_text` crashes on non-Hash evidence entries: `entry["file"]` raises `NoMethodError`/`TypeError` when an entry is `nil`/Integer/Array (evidence is unvalidated `Array(...)` from agent JSON), aborting `enqueue` *after* `mkdir_p`/`write_meta`/`write_idea_md` already wrote a partial task folder, with no logged reason. `lib/hive/patrol/review_handoff.rb:176-183`.
- [ ] [errors] `evidence_text` reads string keys only; symbol-keyed evidence (handled defensively elsewhere in `fingerprint.rb`) silently yields `"- evidence"` with the real evidence dropped from `idea.md`. `lib/hive/patrol/review_handoff.rb:178-179`.
- [ ] [tests] No direct wide-char cell-math test for `format.rb` (`display_width`/`take_cells`/`ljust_cells`/`rjust_cells`): the grapheme / double-width boundary (`break if width > remaining`) — the core correctness claim of this change — is exercised only indirectly via pane icons; add `format_test.rb` covering CJK/emoji and `max_width < 2`. `lib/hive/tui/views/format.rb`.
- [ ] [tests] `truncate` + ellipsis with wide chars untested: when `max_width - 1` lands mid-wide-cluster the result can be narrower than `max_width`; no test asserts the final `display_width` stays within budget for wide input. `lib/hive/tui/views/format.rb:30-38`.
- [ ] [comments] Security-critical `UNSAFE_GIT_ENV`/`unsafe_git_env?` carries no comment stating the threat model or why `GIT_CONFIG_COUNT != 0` is unsafe — invites a "simplification" that reopens the hole. `bin/hive-babysitter-stub-git:65-83`.
- [ ] [product] `archive_filter` now hides ALL archived tasks after 3 days regardless of marker (incl. `error`/`agent_working`/`manual_steering`/`waiting`), reversing the prior "only resolved markers" locked decision; confirm intent — appears intentional (comment + tests updated, footer "+N hidden" gives a recovery path). `lib/hive/archive_filter.rb:16-21`.

## Nit
- [ ] [simplify] `parsed.nil? || parsed.positive? || parsed.negative?` is just "nil or non-zero"; `parsed.nil? || !parsed.zero?` reads clearer (behavior is correct and fail-closed). `bin/hive-babysitter-stub-git:82`.
- [ ] [maintainability] `hide?` retains an unused `marker_name:` keyword (the allowlist was removed); either drop it plus its two call sites (`status.rb`, `snapshot.rb`) or comment that it is kept for caller compatibility. `lib/hive/archive_filter.rb:16`.
- [ ] [errors] `idea_text` calls `finding.evidence.empty?` without a nil guard while `description`/`recommendation` use `.to_s.strip.empty?`; inconsistent and raises on `Finding.new(evidence: nil)` (safe via `from_h` today). `lib/hive/patrol/review_handoff.rb:172`.
- [ ] [maintainability] `write_idea_md` stores `idea_text` both as the heredoc body and as the `original_text` frontmatter value (full duplication); the closing `MD` is also mis-indented relative to its content. `lib/hive/patrol/review_handoff.rb:84-89`.
- [ ] [tests] `write_idea_md` test covers only a fully-populated finding; the blank-section guards in `idea_text` and the missing-`file`/`line`/empty-`snippet` branches in `evidence_text` are never exercised. `test/unit/patrol/review_handoff_test.rb:54-76`.

## Strengths (not findings)
- `archive_filter_test.rb` pins mtime-over-folder_mtime precedence in both directions, marker-agnostic hiding across all marker types, and the exact `HIDE_AFTER_SECONDS` boundary — DAMP, behavior-focused.
- `tasks_pane_test.rb` adds genuine wide-char regression guards (`test_wide_icon_does_not_shift_the_id_column`) that exercise the new display-width math end-to-end.
- The rewritten `format.rb` `truncate` doc comment is now accurate (the old "counts code units / emoji can overflow" caveat was correctly removed, not left stale).
- New `unicode-display_width` dependency is correctly declared in both `hive.gemspec` and `Gemfile.lock`; rubocop is clean and the existing test suite passes.
- `GIT_CONFIG_COUNT` handling is fail-closed and correct, and the `GIT_CONFIG_KEY_n`-without-`COUNT` bypass does **not** exist (git ignores indexed keys without the count) — verified, not a finding.

ivankuznetsov:

# Codex CE Code Review 02

No plan was found in the task folder, so this review is based on `git diff origin/main..HEAD` only.

## High
- [ ] `bin/hive-babysitter-stub-git:13` still allows `--git-dir` / `--work-tree` repo-pointer globals before allowlisted read-only commands: I reproduced `bin/hive-babysitter-stub-git --git-dir=<malicious> --work-tree=<malicious> diff` executing that repo's local `diff.external`, so dry-run agents can bypass the new environment/config-injection guard and run arbitrary code.

## Medium

## Nit

ivankuznetsov:

# Code Review — claude-ce-code-review (pass 2)

- **Scope:** `git diff origin/main..HEAD` (26 files; bin git stub hardening,
  archive filter, patrol idea.md, TUI display-width math).
- **Plan grounding:** No `plan.md` found in the task folder — reviewed without
  plan grounding.
- **Verification:** All affected unit tests pass (7 files, 0 failures);
  rubocop clean on all changed Ruby. Two findings below were reproduced by
  executing the stub directly (exploit confirmed, not theoretical).

## High
- [ ] **`--git-dir` / `--work-tree` argv flags bypass the new env sandbox guard** (`bin/hive-babysitter-stub-git:137-139`): the same commit blocks the `GIT_DIR`/`GIT_WORK_TREE` *env* vars (`UNSAFE_GIT_ENV`) precisely because they "aim git at an attacker-controlled repo whose local config then runs code," but `config_injection?` only *skips* the equivalent `--git-dir`/`--work-tree` *argv* flags rather than failing closed. Verified: `git --git-dir=/evil/.git status` and `git --work-tree=/evil status` both `exec` the real git, so an attacker repo whose `.git/config` sets `core.fsmonitor`/`core.pager` (with `--paginate`) achieves the exact RCE the env block prevents. The babysitter only ever calls `git -C <path>` (confirmed in `lib/hive/babysitter/*`), so `--git-dir`/`--work-tree` can fail closed with no loss of function. (Residual: `-C <attacker-dir>` shares the property but is legitimately needed, so scope the fix to `--git-dir`/`--work-tree`.)

## Medium
- [ ] **Mere presence of `PAGER` in the inherited env skips ALL read-only git** (`bin/hive-babysitter-stub-git:93,109-111` + `lib/hive/babysitter/dry_run_env.rb:8-28`): `DryRunEnv.with_env` inherits the full process environment unsanitized (only `PATH` + `HIVE_*` are overridden), and `unsafe_git_env?` fails closed on the mere *presence* of `PAGER` — a ubiquitous shell/CI variable. Verified: `PAGER=less … git status` is skipped. In any environment where `PAGER` is exported, every legitimate dry-run read (`status`, `config --get`, `log`, `diff`) is skipped, so the babysitter cannot inspect repo state in dry-run mode. Prefer sanitizing/unsetting the unsafe vars in `DryRunEnv` (defense at the boundary, reads still work) over fail-closed-on-presence in the stub, or scope the `PAGER` check to cases where paging can actually fire.

## Nit
- [ ] **Global-option scanning is now duplicated with divergent handling** (`bin/hive-babysitter-stub-git`: `stripped_global_options` L13-28 vs `config_injection?` L129-148): two near-identical leading-option loops that treat `--git-dir`/`--work-tree` differently — exactly the divergence that produced the High finding. A single shared scanner would prevent the next drift.
- [ ] **Archive filter now hides errored/unresolved tasks by age alone** (`lib/hive/archive_filter.rb`): dropping `RESOLVED_MARKER_NAMES` means a task that errored and was archived disappears from default surfaces after 3 days (still counted in "… and N archived >3d ago"). The updated comment frames this as a locked product decision and tests assert it — flagging only so the behavioral change (errors can silently age out of view) is visible to the gate.
- [ ] **`idea.md` `original_text` frontmatter duplicates the full rendered body** (`lib/hive/patrol/review_handoff.rb:84`): acknowledged "mirrors the body by design," but it doubles on-disk size for large findings; worth confirming the duplication is actually consumed downstream.
- [ ] **Cosmetic kwarg jamming in edited tests** (`test/unit/archive_filter_test.rb`): several edited calls leave two kwargs on one line with irregular spacing (`stage: …,      folder_mtime: …`). Rubocop-clean, purely visual.

ivankuznetsov:

# pr-review-toolkit review (pass 2)

No `plan.md` was found in the task folder; this review proceeds without plan
grounding. Scope: `git diff origin/main..HEAD` in the worktree. Personas run:
code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer,
type-design-analyzer. Findings below are the union across all personas.

Overall: no High-severity defects. The security hardening in
`bin/hive-babysitter-stub-git` was traced against real git behavior and fails
closed in every probed case; the `format.rb` cell math is correct. Findings are
contract/test/comment-quality improvements.

## High

(none)

## Medium

- [ ] [silent-failure] `lib/hive/patrol/review_handoff.rb:174,183` — `idea_text` emits the `## Evidence` header based on the raw array being non-empty, but `evidence_text` drops every non-Hash entry; an all-malformed evidence array (exactly the agent-JSON case this code defends against) yields a dangling empty `## Evidence` section and silently discards the fact that evidence was present. Gate the section on the rendered text and/or emit a visible "evidence present but unparseable" marker. Untested case.
- [ ] [type-design] `lib/hive/archive_filter.rb:16` — `hide?(stage:, mtime: nil, folder_mtime: nil, now:)` makes both timestamp keywords optional, so `hide?(stage:)` is a legal silent no-op; the old required-keyword contract caught "caller forgot the timestamp" at the boundary. Consider keeping one required keyword (e.g. `folder_mtime:`) since both live callers already pass it.
- [ ] [type-design] `lib/hive/tui/views/format.rb:50-63` — `module_function` exports internal primitives `take_cells` and `display_width` as public API alongside the real seams (`truncate`/`ljust_cells`/`rjust_cells`); make them `private_class_method` so the module's stable contract stays minimal.
- [ ] [tests] `test/unit/babysitter/dry_run_env_test.rb` — the `config_injection?` attached form `--git-dir=VALUE` / `--work-tree=VALUE` (single `rest.shift`, distinct from the `shift(2)` separated form) is untested; only `git -C dir -c ...` is exercised. Add a case like `git --git-dir=/repo -c core.pager=PAYLOAD status` to lock the off-by-one that protects a trailing `-c`.
- [ ] [tests] `test/unit/babysitter/dry_run_env_test.rb:119-142` — the `GIT_TRACE` prefix family is only proven for four known names; nothing distinguishes `start_with?` from an exact allowlist. Add a sentinel with an arbitrary unknown suffix (e.g. `GIT_TRACE_SOMETHING_NEW`) so the prefix mechanism itself is verified.
- [ ] [tests] `test/unit/babysitter/dry_run_env_test.rb:115-118` — the "kept in lockstep with UNSAFE_GIT_ENV" claim is not enforced: the test keeps an independent hand-copied sentinel list and never asserts set-equality against the source constant, so a var added to the source (not the test) is silently uncovered. Either derive the sentinel list from the stub's constant / assert equality, or soften the comment to "every sentinel must independently force a skip."
- [ ] [tests] `test/unit/patrol/review_handoff_test.rb` — no test feeds YAML-hostile finding fields (a `description` containing `\n---\n`, leading `#`, block-scalar `|`/`>`, or quotes) through `original_text`; since `idea.md` is later split on `"---\n\n"` and `original_text` is meant to round-trip for hand-editing, a `---` in the body could corrupt the frontmatter/body split silently. Add a round-trip assertion with adversarial markdown/YAML chars.

## Nit

- [ ] [silent-failure] `bin/hive-babysitter-stub-git:116` — `Integer(count, exception: false)` accepts `0x0`, `0b0`, `0_0`, and `" 0 "` as zero and lets the command through; not an injection escape (real git rejects them itself with "bogus count"), but it contradicts the "a malformed value is suspicious" comment. Gate on `/\A\d+\z/` before `Integer` if the comment's promise should hold literally.
- [ ] [comment] `bin/hive-babysitter-stub-git:70` — the threat-model "exec hooks" bullet writes `GIT_*_COMMAND` as if a wildcard family is blocked, but only `GIT_SSH_COMMAND` and `GIT_PROXY_COMMAND` are enumerated and there is no `_COMMAND` prefix matcher; spell out the two covered vars so a reader doesn't assume any `*_COMMAND` is caught.
- [ ] [comment] `lib/hive/patrol/review_handoff.rb:175-178` — "read both key styles (mirroring fingerprint.rb)" is accurate for the dual-key reads, but the adjacent `next unless entry.is_a?(Hash)` hardening is new and not shared with `fingerprint.rb`; minor scoping clarity only.
- [ ] [type-design] `lib/hive/archive_filter.rb` — the deleted `RESOLVED_MARKER_NAMES` block carried the institutional note that only resolved markers were hideable; add a one-line `#decision` comment recording that archived rows are now hidden by age alone regardless of marker (including `error`) so the change doesn't later read as a bug.
- [ ] [type-design] `lib/hive/tui/views/format.rb` — `truncate` is now mostly a building block for `ljust_cells`/`rjust_cells` (tasks_pane no longer calls it bare); reassess whether it should stay public, lower priority than `take_cells`/`display_width`.
- [ ] [tests] `test/unit/archive_filter_test.rb` — the refactored assertions collapse two kwargs onto one line with stray multi-space runs (e.g. `stage: ...,      folder_mtime: ...`); reads like a botched mechanical edit and may trip the formatter. Clean up.
- [ ] [tests] `test/unit/tui/snapshot_test.rb` — the blank/invalid case sets both `mtime` and `folder_mtime` to junk, so the "blank `mtime`, valid old `folder_mtime` → still hidden by folder age" fallback path (the reason `parse_time` is split out) is never exercised end-to-end.
- [ ] [tests] `test/unit/tui/views/format_test.rb:65-69` — `rjust_cells` is only tested on the padding path; the over-wide truncation branch (`F.rjust_cells("你好世界", 5)`) is untested, unlike its `ljust_cells` counterpart. One budget assertion closes it.
- [ ] [tests] `test/unit/commands/status_test.rb:427` — `test_render_project_hides_old_archived_rows_with_unresolved_markers` keeps `marker: "ERROR"` to prove marker-independence; add a one-line comment so the ERROR marker reads as deliberate rather than vestigial.

ivankuznetsov:

Outcome: failure

Agent tail:

```

```

ivankuznetsov:
Closing as **superseded / redundant**: this finding's area is already addressed on current `main` (e2e `--json` dispatch via #318/#322/#357, hive-eval usage/traversal rejection via #398/#334, usage-error exit codes via #317, and the dry-run gh/git stub exec/env guards already present in `test/unit/babysitter/dry_run_env_test.rb`). Reconciling this branch onto `main` either conflicts in the already-fixed region or its assertions pass on `main` without the change. Re-filed automatically by patrol if a genuine gap remains.

PR #341, opened 2026-06-06
Hive patrol: Dry-run acceptance test can fall through to real repo commands

ivankuznetsov:
## Summary

Patrol opened this PR for **"Dry-run acceptance test can fall through to real repo commands."**

The headline fix hardens `test/babysitter/acceptance/dry_run_test.rb`. The stubbed `spawn_agent` previously ran its git/gh mutations from the test process's own working directory, so a path fall-through could have driven those commands against the real hive repo instead of the sandboxed worktree. The stub now wraps its work in `Dir.chdir(kwargs.fetch(:cwd))`, matching production (`pr_fixer.rb` passes `cwd: worktree_path`), so every agent-side git/gh call is forced through the dry-run shims. The assertions were also strengthened: the load-bearing proof is now the `.babysitter-dry-run-skipped.log` interception checks (force-push, PR comment, PR close were diverted, never sent to the remote), rather than the near-tautological shim exit-status array.

Bundled with the fix are several small, independently-committed refactors hardened during review:

- **Archive filter** (`lib/hive/archive_filter.rb`): drop the now-dead `marker_name:` parameter from `hide?` and its callers (`status.rb`, `snapshot.rb`); archived `9-done` rows are hidden purely by row mtime past 3 days (a locked product decision documented in the wiki).
- **TUI `Format`** (`lib/hive/tui/views/format.rb`): encapsulate `take_cells` as private, document the nil-tolerance contract on the cell-aware helpers, and add wide-glyph truncation coverage (via the `unicode-display_width` dependency).
- **TUI `bubble_model`** (`lib/hive/tui/bubble_model.rb`): replace a `nil.to_i`-dependent archive-notice count with an explicit `|| 0`, and correct a stale `compose_two_pane_view` docstring.
- **Patrol review handoff** (`lib/hive/patrol/review_handoff.rb`): guard `idea_text`/`evidence_text` against a `nil` `evidence` and non-Hash evidence entries (`Array(...)` + `is_a?(Hash)`), and compute the rendered idea body once so the `original_text` frontmatter and the human body stay provably identical.

## Test plan

- `test/babysitter/acceptance/dry_run_test.rb` — dry-run acceptance test passes and now proves interception via the skip log.
- Affected unit suites pass: `test/unit/archive_filter_test.rb`, `test/unit/commands/status_test.rb`, `test/unit/patrol/review_handoff_test.rb`, `test/unit/tui/snapshot_test.rb`, `test/unit/tui/views/tasks_pane_test.rb` (1108 TUI runs / 0 failures).
- `rubocop` clean on all changed Ruby files; `bundle check` OK; `unicode-display_width` resolves in gemspec + lock.

## Review summary

Six review passes (Claude CE, Codex CE ×2, pr-review-toolkit multi-persona) plus two escalation passes ran against this branch.

- In-scope nits were auto-fixed: privatize `take_cells`, add wide-glyph + negative-branch tests, strengthen the dry-run assertion to prove interception, drop the dead `marker_name:` param, add defensive evidence guards, and DRY the `original_text` rendering.
- The two flagged product decisions — hiding archived rows after 3 days by mtime (no marker carve-out) and removing the vertical-viewport clamping — were confirmed deliberate reverts and documented in the wiki.
- All 13 Codex-CE findings referencing files such as `bin/hive-babysitter-stub-git`, `stale_agent_healer.rb`, `dispatcher.rb`, and `bin/hive-eval` were resolved as out of scope: those files are **not** modified by this branch's changeset (merge-base `23cf47b9`..HEAD = 23 files). They surfaced only because the reviewer diffed against a divergent `origin/main` lineage.

No open user questions remain.

## Linked task

`/home/asterio/Dev/hive/.hive-state/stages/6-review/patrol-test-suite-test-babysitter-acceptance-dry-run-test-rb-776`


ivankuznetsov:

# Codex CE Code Review 01

No plan was found in the task folder; review is grounded in `git diff origin/main..HEAD`, project wiki context, and local tests.

## High

## Medium
- [ ] TUI two-pane panes no longer receive or enforce a height budget: `BubbleModel#compose_two_pane_view`/`join_panes` call `ProjectsPane.render` and `TasksPane.render` without `height`, while both pane renderers now emit every row; a small or resized terminal with many tasks/projects can render more lines than `model.rows`, leaving the selected row offscreen while key actions still target it.

## Nit

ivankuznetsov:

# Claude CE Code Review 01

No `plan.md` was found in the task folder. Review is grounded in `git diff origin/main..HEAD`,
the project wiki, and locally-run tests (affected unit suites pass; `bundle check` OK).

**Scope of the diff:** two commits bundle (1) the namesake dry-run acceptance-test sandbox fix,
(2) archive-filter semantics change + `mtime` precedence, (3) `Format` wide-glyph truncation via a
new `unicode-display_width` dependency, (4) tasks-pane icon-column alignment, (5) **removal** of the
vertical-viewport clamping / cursor-following panes (reintroducing the header strip + hidden-archive
footer notice), and (6) patrol `idea.md` handoff. The headline dry-run fix is correct and genuinely
safer; most findings below concern the bundled TUI/archive behavior changes.

## High

## Medium
- [ ] Vertical viewport clamping removed (`pane_height_for`/`fit_lines`/`fit_rows`/`viewport_start` deleted from `bubble_model.rb`, `tasks_pane.rb`, `projects_pane.rb`): panes now render every row, so on a short or vertically-resized terminal the task/project list overflows and pushes the footer (key hints) and selected row off-screen — exactly the regression the now-deleted `test_charm_tui_vertical_resize_keeps_panes_footer_and_selected_row_visible` PTY smoke guarded. The deleted wiki entry confirms this is a deliberate revert; flagging to confirm the trade-off (header strip + full list over guaranteed-visible footer) is intended.
- [ ] Archive filter now hides *unresolved* `9-done` rows after 3 days: removing `RESOLVED_MARKER_NAMES` means `ERROR`/`WAITING`/`MANUAL_STEERING` archived tasks disappear from default `hive status` text and the TUI grid once older than 3 days, where the prior allowlist intentionally kept failures visible regardless of age (`lib/hive/archive_filter.rb`). A persistently-failed done task can now silently vanish from daily surfaces — confirm this product decision is intended (it reverses a prior safety guard; `--json` and `hive archive` remain unfiltered).

## Nit
- [ ] `ArchiveFilter.hide?` retains the `marker_name:` keyword but no longer references it in the body (`lib/hive/archive_filter.rb:16`); callers in `status.rb`/`snapshot.rb` still pass it. Drop the param or comment why it's kept, so the signature doesn't imply marker-based gating that no longer happens.
- [ ] `archive_hidden_notice` relies on `nil.to_i == 0`: `@hive_model.snapshot&.hidden_old_archived_count.to_i` guards only the first call with `&.`; the trailing `.to_i` runs on the `nil` result when there is no snapshot (`bubble_model.rb:3288`). It works, but `(... &.hidden_old_archived_count || 0)` states the intent without depending on `nil.to_i`.
- [ ] `idea.md` stores the full rendered markdown body a second time inside the `original_text` frontmatter field (`review_handoff.rb:82` / `idea_text`): the value duplicates the body verbatim, bloating the YAML header. Consider persisting only the raw finding text (or omitting `original_text`) since the body already carries it.
- [ ] Dry-run acceptance test wraps the stubbed agent in a process-global `Dir.chdir(kwargs.fetch(:cwd))` block (`test/babysitter/acceptance/dry_run_test.rb:17`). Correct and the right fix for the cwd fall-through, but `Dir.chdir` mutates global state — if this suite ever runs under threaded parallelism it can raise "conflicting chdir during another chdir block." Fine for the current serial/forked runner; noting for future parallelization.
- [ ] Commit hygiene/scope: `193c1e05 "fix: improve tui task context and archive counts"` bundles six largely independent concerns (archive semantics, new runtime dependency, icon alignment, viewport removal, header strip, patrol idea.md). Splitting these would make each independently reviewable and revertable; the viewport-removal revert in particular is hard to isolate from the alignment work.

ivankuznetsov:

No `plan.md` was found in the task folder. Review is grounded in `git diff origin/main..HEAD`,
project conventions (CLAUDE.md / wiki), and locally-run tests. Personas run: **code-reviewer**,
**test-analyzer**, **silent-failure-hunter**, **comment-analyzer**, **type-design-analyzer**.
Findings below are the union across personas (overlapping findings merged, persona prefixes list
every persona that surfaced the item).

Verification reported by personas: rubocop clean on all changed Ruby files; affected unit suites +
dry-run acceptance test pass (1108 TUI runs / 0 failures); `unicode-display_width` resolves in
gemspec + lock. The dry-run `Dir.chdir(kwargs.fetch(:cwd))` fix was independently confirmed
**correct** (production passes `cwd: worktree_path` at `pr_fixer.rb:150`; it stops the stub from
running git/gh against the real hive repo).

## High
- [ ] [type-design][silent-failure] `idea_text` calls `finding.evidence.empty?` directly (`lib/hive/patrol/review_handoff.rb:172`): `Hive::Patrol::Finding` is a `keyword_init` Struct with no default, so a `Finding.new` built without `evidence:` yields `nil` and raises `NoMethodError`; the handoff is wrapped in `rescue StandardError` at `pr_opener.rb`, so it degrades to a one-line warn + silently missing 6-review task. Use `Array(finding.evidence)` to match the codebase convention. (Safe with today's only caller, which wraps in `Array(...)`; flagged because the type doesn't enforce it.)
- [ ] [type-design] `ArchiveFilter.hide?` removes the `RESOLVED_MARKER_NAMES` allowlist, dropping a business invariant (`lib/hive/archive_filter.rb:16-22`): archived tasks past 3 days are now hidden regardless of marker, so previously-visible unresolved/`ERROR` done tasks now disappear from default `hive status` and the TUI grid. The deleted comment explicitly warned this allowlist must be reviewed when terminal markers change — that safety rail is gone. Confirm this is a deliberate product decision (the "Locked product decision" comment edit suggests yes) and not collateral of the mtime refactor.

## Medium
- [ ] [code-review][test-analyzer] Vertical viewport clamping removed (`pane_height_for`/`fit_lines`/`fit_rows`/`viewport_start` deleted from `bubble_model.rb`, `tasks_pane.rb`, `projects_pane.rb`): panes now render every row with no height budget, so on a short/resized terminal the task list overflows and can push the footer (key hints) + hidden-archive notice and the selected row off-screen — the exact regression the deleted PTY smoke `test_charm_tui_vertical_resize_keeps_panes_footer_and_selected_row_visible` guarded. Wiki docs were rewritten to match, so likely an intentional revert; if so the dashboard now depends entirely on the terminal's own scroll behavior, with no Ruby-layer test asserting that contract.
- [ ] [silent-failure] `archive_hidden_notice` counts hidden rows on the full unscoped snapshot (`lib/hive/tui/bubble_model.rb:3288`) while the grid renders a scope+filter projection: when the operator has scoped to one project or typed a slug filter, "+N hidden — open Archive pane" can include rows from off-screen projects/filters, so the count won't match what the current view actually suppresses. Count against the same scoped projection to keep the notice honest.
- [ ] [type-design][silent-failure][code-review][comment] `marker_name:` is now a dead parameter on `ArchiveFilter.hide?` (`lib/hive/archive_filter.rb:16`): accepted and still passed by both callers (`status.rb:376`, `snapshot.rb:255`) but never read. Misleading contract — a future caller will assume marker state still gates hiding. Drop the param from the signature + callers, or add a one-line note that it's retained only for caller compatibility and intentionally ignored.
- [ ] [comment-analyzer] Stale narrow-mode claim in `compose_two_pane_view` docstring (`lib/hive/tui/bubble_model.rb:3155-3159`): it states that below `TWO_PANE_MIN_COLS` "the scope label [is] prefixed onto the header," but the method renders the same `header_strip(usable)` in both modes (the only width branch chooses `TasksPane.render` vs `join_panes`), and `header_strip` already includes `scope=` unconditionally. Describes a code path that doesn't exist; will mislead a maintainer.
- [ ] [test-analyzer] `idea.md` test only covers a fully-populated finding (`test/unit/patrol/review_handoff_test.rb`): `idea_text`/`evidence_text` have multiple negative branches (blank description/recommendation omit their `##` section; `location.empty?` → `- evidence`; `snippet.empty?` → location-only) that are untested — a realistic patrol output with no description or snippet-less evidence is uncovered.
- [ ] [type-design] `original_text` frontmatter duplicates the full markdown body verbatim (`lib/hive/patrol/review_handoff.rb:84,87`): `idea_text(finding)` is rendered twice (once into the YAML key the TUI reads, once as the human body). Correct per the consumer contract, but a two-source-of-truth hazard — compute it once into a local and interpolate into both positions so the "deliberately identical" invariant is explicit.

## Nit
- [ ] [type-design] All of `marker_name`/`mtime`/`folder_mtime` now default to `nil` on `hide?` (`lib/hive/archive_filter.rb:16`): previously-required timestamps are now omittable, so `hide?(stage: x)` is a legal call that fails open (never hides) with no `ArgumentError` signal. Keep at least one timestamp source non-defaulted, or document "no timestamp ⇒ never hide" as the deliberate contract.
- [ ] [silent-failure] `evidence_text` assumes each entry is a Hash (`lib/hive/patrol/review_handoff.rb:177-182`): a bare-string evidence entry makes `entry["file"]` a `String#[]` substring match — e.g. `"see file.rb"["file"]` returns `"file"`, injecting garbage into the location rather than surfacing malformed input. Guard with `entry.is_a?(Hash)`.
- [ ] [code-review][silent-failure] `@hive_model.snapshot&.hidden_old_archived_count.to_i` (`lib/hive/tui/bubble_model.rb:3288`): the `&.` does not guard the trailing `.to_i`, which runs on `nil` when there's no snapshot and works only because `nil.to_i == 0`. `(... &.hidden_old_archived_count || 0)` states the intent without depending on `NilClass#to_i`.
- [ ] [type-design] `take_cells` (and arguably `display_width`) is a public `module_function` but is an internal primitive (`lib/hive/tui/views/format.rb:54`): no external caller uses it, and exposing the no-ellipsis raw cut invites callers to bypass `truncate`'s ellipsis contract. Add `private_class_method :take_cells`.
- [ ] [test-analyzer] `take_cells`/`truncate` has no direct unit test for the wide-glyph truncation boundary it was rewritten to handle (`lib/hive/tui/views/format.rb`): padding (`ljust_cells`/`rjust_cells`) is covered, but truncating to a width that lands between the two cells of a double-width emoji (the `break` guard preventing a half-glyph) is not — a focused test would lock in the fix that removed the old "emoji can overflow" caveat.
- [ ] [test-analyzer] `test_wide_icon_does_not_shift_the_id_column` doesn't exercise the wide-vs-narrow contrast its name implies (`test/unit/tui/views/tasks_pane_test.rb:445`): both compared rows use wide emoji icons, so it would still pass if a narrow `DEFAULT_ICON` mis-padded. The companion `test_action_icon_cells_have_consistent_display_width` does cover `DEFAULT_ICON`, so behavior is covered overall — just a slightly mislabeled assertion.
- [ ] [test-analyzer] The dry-run assertion `command_results == [true, true, true]` is near-tautological (`test/babysitter/acceptance/dry_run_test.rb:37`): the shims always `exit 0`, so it passes whether intercepted or a real command happens to succeed. The load-bearing proof is the `.babysitter-dry-run-skipped.log` content checks (lines 38-41), which are solid — the exit-status array adds less protection than its presence implies.
- [ ] [type-design] Naming inconsistency in the new `Format` helpers (`lib/hive/tui/views/format.rb:40-63`): the family splits `ljust_cells`/`rjust_cells`/`take_cells` (`_cells` suffix) vs `display_width`/`truncate` (no suffix) even though `truncate` is equally cell-aware; and the nil-tolerance contract (everything funnels through `.to_s`, nil → empty) is undocumented on the four new public functions. Pick one suffix convention and add a one-line contract note.

ivankuznetsov:

# Codex CE Code Review - Pass 2

## High
- [ ] Dry-run git shim forwards dangerous read-only-looking invocations: `bin/hive-babysitter-stub-git:53` allowlists `diff`/`grep`/`log` after stripping options and `:78` execs unsanitized env, so `--output`, `-c diff.external=...`, `GIT_EXTERNAL_DIFF`, or pager flags can still write files/run commands during dry run.
- [ ] Stale `REVIEW_WORKING` rows are never healed: `lib/hive/daemon/stale_agent_healer.rb:68` ignores review markers while `TaskAction` maps `review_working` to `agent_running`, so a killed review runner can leave the row stuck with no daemon recovery.
- [ ] Generated review prompts invoke a missing CE skill: `templates/review_prompt.md.erb:10` asks agents to use `/compound-engineering:ce-review`, but the configured/available review skill is `ce-code-review`, so normal review spawns can fail before producing findings.

## Medium
- [ ] Patrol-specific reviewer config is silently ignored: `lib/hive/stages/review.rb:943` now reads only `review.reviewers`, so existing projects that configured `patrol.review.reviewers` can run the wrong reviewer set or pass patrol PRs with no reviewers.
- [ ] Patrol scans consume normal task capacity again: `lib/hive/daemon/dispatcher.rb:932` routes scans through `can_dispatch?`, so a full task queue can block scheduled patrols and running scans can spend task slots/daily quota.
- [ ] Dispatch no longer drains later pipeline stages first: `lib/hive/daemon/dispatcher.rb:260` iterates raw status order, so scarce daemon slots can be taken by early-stage work ahead of review/finalize rows closest to completion.
- [ ] TUI panes no longer respect terminal height: `lib/hive/tui/bubble_model.rb:3171` renders task panes without a height budget, so long queues can overflow the terminal and hide the footer or selected row after resize.
- [ ] Caption text is dropped when media arrives during an open idea draft: `lib/hive/bot/handlers/slash_handlers.rb:73` only applies captions when no draft exists, so `/idea` followed by a captioned file stages the attachment but asks for idea text again.
- [ ] `bin/hive-eval --scenario` can execute arbitrary Ruby test files: `bin/hive-eval:29` accepts any `.rb` path despite the basename-only CLI contract, letting a caller run files outside `test/eval/scenarios`.
- [ ] Bot log schema breaks its v1 additive-only contract: `schemas/hive-bot-log.v1.json:5` says removals require v2, but this diff removes the historical `transcription_failed` enum value from the v1 schema.
- [ ] Wiki changelog history was deleted despite being append-only: `wiki/log.md:3` declares an append-only log, but the diff removes 302 existing log lines and loses prior project history.

## Nit
- [ ] OpenClaw publish docs point at the previously unavailable slug: `openclaw/README.md:75` switches the umbrella listing from `hive-cli` to `hive`, even though the prior docs recorded `hive` as already owned.
- [ ] Telegram E2E wrapper is tied to one local checkout: `test/e2e/tg/run_idea_e2e.sh:7` hardcodes `/home/asterio/Dev/hive`, so running the test from a worktree can exercise the wrong branch.

PR #345, opened 2026-06-06
Hive patrol: Dry-run git config allowlist permits writes

ivankuznetsov:
## Summary

This branch originated from a Hive patrol finding: the babysitter dry-run git stub treated **any** `git config` invocation containing `--get`/`--get-all`/`--list` as read-only regardless of argument order, so a write form like `git config user.name newvalue --get` slipped through and mutated config. While fixing that, a few adjacent hardening and correctness changes were folded in.

- **Dry-run git config allowlist (the patrol finding).** `bin/hive-babysitter-stub-git` now parses `git config` into explicit read-only shapes via `read_only_config?` instead of checking for a flag anywhere in the args. It allows only true reads — `config --get <name>`, `config --get-all <name>`, `config --list`, and a bare single-key read (`config <name>`) — and tolerates leading type/format modifiers (`--bool`/`--int`/`--path`/`--null`/`-z`/`--type=*`) that legitimately precede a read verb (e.g. `git config --bool --get commit.gpgsign`, which `auto_commit.rb` runs). Write forms with a trailing read flag are now correctly skipped.
- **Archive filter timestamp source.** `ArchiveFilter.hide?` now hides aged `9-done` rows by task-file `mtime` (falling back to `folder_mtime`), dropping the `marker_name` allowlist. The both-nil case deliberately fails open (stays visible) rather than guessing. Dead `marker_name:` parameter removed from both call sites (`status.rb`, `snapshot.rb`).
- **Patrol review handoff.** `review_handoff.rb` now also writes an `idea.md` (with `original_text` provenance frontmatter) and renders evidence defensively — tolerating scalar and symbol-keyed entries from untrusted agent JSON, matching the sibling `pr_opener.rb` renderer.
- **TUI cell-accurate formatting.** `Format` gains `display_width`/`ljust_cells`/`rjust_cells`/`take_cells` built on `unicode-display_width`, so wide characters (emoji, CJK) no longer overflow task-pane columns. Adds the `unicode-display_width` dependency.
- **Snapshot timestamp parsing.** `parse_time` now warns instead of silently swallowing an unparseable ISO-8601 stamp; the downstream behavior remains fail-safe (row stays visible).

## Test plan

- `bundle exec rubocop` — clean on changed Ruby files.
- `bundle exec rake test` — full unit suite passes. Changed-area suites verified directly: `dry_run_env_test`, `archive_filter_test`, `snapshot_test`, `tasks_pane_test`, `review_handoff_test`, `status_test`.
- New/expanded coverage added for: config write verbs (`--add`/`--unset`/`--replace-all`/`--unset-all`) skipped as writes; bare-key reads and leading-modifier reads allowed; `--get-regexp`/3-arg `--get` documented-skip behavior; archive-filter `mtime` precedence over `folder_mtime` at both the pure-filter and snapshot layers; `idea_text`/`evidence_text` nil/empty/scalar/symbol-keyed branches and `display_name` fallback.

## Review summary

Reviewed across 3 passes by `claude-ce-code-review`, `codex-ce-code-review`, and `pr-review-toolkit` (code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer personas).

- **AUTO-FIX (applied):** bare-key `git config` read regression closed; config write-verb and read-form regression tests added; defensive evidence rendering; `parse_time` warn-on-failure; archive-filter fail-open + precedence documented; dead `marker_name` parameter dropped; `idea_text` computed once with `Array(...)` nil-guard; comment precision and redundant-`dup` cleanups.
- **RESOLVED/NO-FIX (stale-base false positives):** every "TUI viewport removal", `patrol_reviewers` schema, zero-reviewer handoff, `ce-review`→`ce-code-review` rename, and exec/write-screen finding is an artifact of reviewing `origin/main..HEAD` against a stale fork point (merge-base `23cf47b9`). The branch never touches those files; verified against git history.
- **Merge note (no code change):** rebase onto current `main` before merge so the post-fork `origin/main` changes (notably the #326 viewport fix `b02c5fae`) are preserved alongside this branch's `read_only_config?` change.

No reviewer findings were escalated to the user; all were auto-fixed or resolved from project/git context.

## Linked task

/home/asterio/Dev/hive/.hive-state/stages/6-review/patrol-command-bin-hive-babysitter-stub-git-c8c1714e


ivankuznetsov:

# Codex CE Code Review - Pass 1

No plan.md was found in the task folder; review is grounded only in `git diff origin/main..HEAD`.

## High
- [ ] TUI grid no longer clamps pane height to the terminal row budget: `compose_two_pane_view` stops reading `@hive_model.rows` and calls `TasksPane.render` / `ProjectsPane.render` without a height, while both pane renderers now emit every row; tall task/project lists after a short terminal resize can push the footer and selected row off-screen, and the previous viewport regression tests were removed.

## Medium

## Nit

ivankuznetsov:

# Code Review — patrol-command-bin-hive-babysitter-stub-git (Pass 1)

**Plan grounding:** No `plan.md` was found in the task folder — this review proceeds without plan grounding.

**Diff reviewed:** `git diff origin/main..HEAD` (2 commits)
- `1bc110c9` fix(patrol): Dry-run git config allowlist permits writes
- `193c1e05` fix: improve tui task context and archive counts

**Verification:** Ran the changed-area unit suites locally — `tasks_pane_test`, `archive_filter_test`, `dry_run_env_test`, `review_handoff_test` all pass.

## High
- [ ] `read_only_config?` only inspects `args.first`, so valid read forms that place a type/format modifier before `--get` are misclassified as writes and **skipped** under dry-run: `bin/hive-babysitter-stub-git:49-60` rejects `git config --bool --get commit.gpgsign` (and `--int`/`--path`/`-z`/`--null` variants). The repo itself runs exactly that command at `lib/hive/stages/auto_commit.rb:174`; the old `rest.include?("--get")` check passed it through. Since the stub is PATH-shimmed over `git` for all git calls inside `Babysitter::DryRunEnv.with_env`, this read now returns exit 0 with empty output instead of the real value. The added dry-run test only covers `--get key` / `--get-all key` / `--list` in first position, so the regression is untested. Fix: skip leading `--bool|--int|--path|--null|-z|--type=*` modifiers before the `case`, or detect "contains a read verb AND no value-assignment positional."

## Medium
- [ ] Removing the height-aware viewport (`pane_height_for`/`rendered_line_count` in `bubble_model.rb`, `fit_rows`/`viewport_start`/`pad_rows` in `projects_pane.rb`, `visible_task_lines`/`fit_lines`/`viewport_start` in `tasks_pane.rb`) reintroduces vertical overflow: on a terminal shorter than the task/project list, panes render every row, pushing the footer — which now carries the `+N hidden — open Archive pane` notice — and the selected cursor row off-screen. This was previously fixed and guarded by `test_charm_tui_vertical_resize_keeps_panes_footer_and_selected_row_visible` and `test_grid_mode_clamps_rendered_height_to_terminal_rows`, both deleted here. Confirm the chrome-over-viewport tradeoff is intended and that footer/selected-row visibility on short terminals is acceptable.
- [ ] `ArchiveFilter.hide?` now hides **all** aged `9-done` rows regardless of marker (the `RESOLVED_MARKER_NAMES` allowlist was deleted), so an `error`-marked archived task silently drops out of daily `hive status`/TUI after 3 days, surfaced only as a `+N hidden` count. The tests/wiki were flipped to match (`test_hides_archived_rows_with_unresolved_markers`), so this is deliberate — but it reverses the prior "unresolved markers stay visible" guarantee and can bury a genuine error. Confirm this is the locked product decision.

## Nit
- [ ] `ArchiveFilter.hide?` still accepts `marker_name:` but no longer references it anywhere in the body (`lib/hive/archive_filter.rb:16`); both call sites still pass it. Dead parameter — drop it, or keep it only if a future marker-aware rule is planned.
- [ ] `write_idea_md` stores the full rendered `idea_text(finding)` in **both** the `original_text` frontmatter field and the markdown body (`lib/hive/patrol/review_handoff.rb:75-89`), duplicating the entire body inside frontmatter. Confirm this is intentional provenance rather than redundant bloat; `idea_text` is also computed twice in the same method.
- [ ] `idea_text` calls `finding.evidence.empty?` (`lib/hive/patrol/review_handoff.rb:172`). `Finding` is a keyword-init `Struct` whose `evidence` is `nil` when constructed directly; it is only guaranteed to be an array when built via `Finding.from_h` (which wraps `Array(...)`). Safe on the current pipeline path, but a direct `Finding.new` without `evidence:` would raise `NoMethodError`. Consider `Array(finding.evidence).empty?` for robustness.

ivankuznetsov:

# Codex CE Code Review - Pass 2

No plan.md was found in the task folder; review is grounded only in `git diff origin/main..HEAD`.

## High

## Medium
- [ ] Branch still lacks the current `origin/main` TUI vertical-resize viewport fix: the mandated diff removes height-aware pane clipping and its PTY/unit regressions from `lib/hive/tui/*` and tests, so merging this tree without rebasing risks losing footer/selected-row visibility on short terminals.

## Nit

ivankuznetsov:

# PR Review Toolkit — patrol-command-bin-hive-babysitter-stub-git (Pass 2)

**Plan grounding:** No `plan.md` was found in the task folder — this review proceeds without plan grounding.

**Diff reviewed:** `git diff origin/main..HEAD` (branch `hive-patrol/command-bin-hive-babysitter-stub-git-c8c1714e`).

**Stale-base caveat:** merge-base is `23cf47b9`, which predates `origin/main`'s viewport fix #326 (`b02c5fae`). The TUI "removals" in `lib/hive/tui/bubble_model.rb`, `projects_pane.rb`, and the deletion of `test/integration/tui_smoke_charm_test.rb` are diff artifacts of the stale base — the branch never touched those helpers. They are NOT reported as code bugs; only the merge hazard (below) is.

**Verification:** changed-area unit suites pass locally (per code-reviewer persona: 62 runs, 0 failures); rubocop clean on the new code.

**Personas run:** code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer.

## High
- [ ] [silent-failure] Bare `git config <name>` reads are silently swallowed by the dry-run stub: `read_only_config?` (`bin/hive-babysitter-stub-git:49-65`) only recognizes `--get`/`--get-all`/`--list`, so the no-flag read form hits `else => false`, logs a skip, and `exit 0` with empty stdout. Under `DryRunEnv.with_env`, `lib/hive/worktree.rb:91-93` (`git config remote.origin.url`, then checks `.success?`) wrongly concludes an origin exists with an empty URL, and `lib/hive/git_ops.rb:260` (`git config init.defaultBranch`) falls through to the `"master"` default. Pre-existing (the old `rest.include?` form also returned false here), but this PR rewrites the exact predicate and is the right place to close it by adding a bare-read branch (`args.length == 1 && !args.first.start_with?("-")`).
- [ ] [tests] Config write verbs are never asserted as skipped: nothing in `test/unit/babysitter/dry_run_env_test.rb` pins that `git config --add`/`--unset`/`--replace-all`/`--unset-all` are treated as writes. That is the primary regression surface of this security tightening; a future refactor moving one into the read allowlist would pass silently.

## Medium
- [ ] [code-reviewer/silent-failure/type-design] `evidence_text` (`lib/hive/patrol/review_handoff.rb:179-186`) assumes every evidence entry is a string-keyed Hash (`entry["file"]/["line"]/["snippet"]`). Evidence arrives from untrusted agent JSON (`reviewer.rb:127` → `Array(raw["evidence"])`); a scalar entry (e.g. `["app.rb:10"]`) raises `TypeError` aborting `enqueue`, and a symbol-keyed entry silently renders `- evidence` (data loss in `idea.md`). Sibling renderers `pr_opener.rb:171` and `fingerprint.rb` defensively handle both key types — this new code diverges. Suggest `next "- #{entry}" unless entry.is_a?(Hash)` plus `["file"] || [:file]` fallbacks.
- [ ] [type-design] `ArchiveFilter.hide?(stage:, mtime: nil, folder_mtime: nil, ...)` (`lib/hive/archive_filter.rb:16`) makes both timestamps optional, so omitting both silently returns `false` (never hides) instead of the prior hard `ArgumentError` from the required `folder_mtime:`. Fail-open is safe but the construction-time invariant is weakened; consider collapsing to a single required `archived_at:` and resolving `mtime || folder_mtime` at the two call sites.
- [ ] [type-design] The `mtime`-takes-precedence-over-`folder_mtime` rule lives only in the impl line `archived_at = mtime || folder_mtime` (`lib/hive/archive_filter.rb:18`); two same-typed undocumented `Time?` params give the reader no cue which wins. A single `archived_at:` param or a one-line precedence comment would carry the invariant.
- [ ] [silent-failure] `parse_time` (`lib/hive/tui/snapshot.rb:265-271`) rescues `ArgumentError` → `nil` with no `warn`/log, so a corrupted ISO-8601 mtime in a snapshot payload silently keeps a stale row visible and is undebuggable. Downstream effect is safe (nil → row stays visible) but the masking is total.
- [ ] [tests] Snapshot-layer mtime precedence is not actually exercised: `snapshot_test.rb:395` sets `mtime` and `folder_mtime` to the same value, so a bug swapping the args in `old_archived_row?` (`snapshot.rb:253-258`) would not be caught — only the pure-filter test proves precedence.
- [ ] [tests] 3-arg read `git config --get <key> <value-regex>` is now silently skipped (`args.length == 2` rejects it) with no test documenting the deliberate exclusion (`bin/hive-babysitter-stub-git:59`).
- [ ] [tests] `git config --get-regexp <pattern>` (a read) is unhandled — falls to `else => false` and is skipped — and untested; pin the current skip behavior so the gap is visible (`bin/hive-babysitter-stub-git:57-64`).
- [ ] [tests] `idea_text`/`evidence_text` conditional branches are uncovered: nil/empty description, nil/empty recommendation, empty/nil evidence, and the `location.empty? → "- evidence"` / `snippet.empty?` fallbacks — `review_handoff_test.rb:54` only exercises a fully-populated finding (the very `Array(finding.evidence).empty?` guard from e47f6aac is unasserted).
- [ ] [merge-hazard] Branch must be rebased onto current `origin/main` before merge: merge-base `23cf47b9` predates #326 (`b02c5fae`), so a non-rebased merge would revert #326's viewport fix in `lib/hive/tui/*`. The branch's only real TUI edit (`193c1e05`, `tasks_pane.rb` task-context/archive-counts) does not touch the viewport helpers, so a rebase should resolve cleanly.

## Nit
- [ ] [tests] Modifier-after-verb (`git config --get --bool commit.gpgsign`) is rejected (the `shift while` only strips leading modifiers) but untested, leaving the leading-vs-trailing asymmetry invisible to maintainers (`bin/hive-babysitter-stub-git:55`).
- [ ] [tests] `--list` with extra args (`git config --list --show-origin`, a common read) is skipped due to `args.length == 1`, with no negative/multi-arg `--list` test (`bin/hive-babysitter-stub-git:60`).
- [ ] [comment-analyzer] `bin/hive-babysitter-stub-git:51-53` comment ("These never assign a value, so they keep the command read-only") is imprecise: `--bool/--int/--path/--type=*` are type specifiers valid on writes too; read-only-ness comes from the subsequent `--get`/`--get-all`/`--list` verb check, not the modifiers. Reword to attribute it to the verb check.
- [ ] [type-design] Redundant `args = args.dup` at `bin/hive-babysitter-stub-git:54`: `rest[1..] || []` already returns a fresh array, so the dup guards nothing.
- [ ] [type-design] `idea_text` recomputes `Array(finding.evidence)` twice (guard + argument) at `lib/hive/patrol/review_handoff.rb:173-175`; a local `ev = Array(finding.evidence)` reads cleaner.
- [ ] [silent-failure] `ArchiveFilter.hide?` both-nil fail-open (`archive_filter.rb:16-22`) is correct but undocumented; a one-line "deliberate fail-open" comment would prevent a future reader from "fixing" it into hiding rows with unknown timestamps.
- [ ] [tests] `display_name` fallback to `finding.id` when title is nil (`review_handoff.rb:166`) is untested — only affects the idea.md heading.
- [ ] [tests] Mixed timestamp case (invalid `mtime`, valid `folder_mtime`) is untested at the snapshot layer (`snapshot_test.rb:417` sets both bad), leaving graceful fallback unverified.
- [ ] [code-reviewer] Informational: other read-only config forms not in the allowlist (`--get-regexp`, `--get-urlmatch`, bare `git config <name>`) are over-blocked (skipped) — safe direction, and the bare form was already skipped under the old logic, so not a regression. Confirmed no write form slips through: every truthy branch requires `args.first` to be exactly a read verb after modifier-stripping.

ivankuznetsov:

# Codex CE Code Review Pass 03

Scope: `git diff origin/main..HEAD`
Plan: no `plan.md` was found in the task folder; review proceeded without plan grounding.

## High
- [ ] Dry-run git stub still forwards exec/write-capable read commands (`bin/hive-babysitter-stub-git:87`): `git -c diff.external=... diff`, `git diff --output=...`, and `GIT_EXTERNAL_DIFF=... git diff` reach real git in dry-run, so restore main's exec/write screens while keeping the new `read_only_config?` handling.

## Medium

_No findings._

## Nit

_No findings._

PR #371, opened 2026-06-07
Hive patrol: gh dry-run stub permits browser execution

ivankuznetsov:
## Summary

Hardens the babysitter `gh` dry-run stub (`bin/hive-babysitter-stub-gh`) so it can no longer launch a browser, even when a read-only command is allowed through to the real `gh` binary.

Two gaps are closed:

- **`--web` / `-w` flags are now skipped.** Commands like `gh repo view --web` or `gh pr view 42 --web` were previously classified as read-only and passed through, opening a browser. A new `browser_option?` check skips any invocation carrying `--web`/`-w` (including `=value` forms).
- **`GH_BROWSER` / `BROWSER` env vars are scrubbed before `exec`.** Even without a browser flag, `gh` can open a browser via these environment variables. The stub now deletes both before handing off to the real `gh`, so an env-only browser launch can't slip through.

## Test plan

- `test/unit/babysitter/dry_run_env_test.rb` extends the existing allow/skip tables to assert that `--web` and `-w` invocations are skipped.
- A new test, `test_scrubs_browser_env_vars_before_exec_to_real_gh`, runs a recording stub as the "real" `gh` and asserts both `GH_BROWSER` and `BROWSER` are `<unset>` when a read-only command reaches it.
- `ruby -Itest test/unit/babysitter/dry_run_env_test.rb`

## Review summary

Two Codex CE code-review passes. Pass 1 raised one nit (missing regression coverage for `GH_BROWSER`/`BROWSER` env scrubbing), which was auto-fixed by adding `test_scrubs_browser_env_vars_before_exec_to_real_gh`. Pass 2 found nothing. No High or Medium findings in either pass; no user escalations.

## Linked task

Patrol finding `command-bin-hive-babysitter-stub-git-2` (fingerprint `4f947f63ca4fb7e89f27aa2bff4fad76de6f5f418dffc4d69f5d620b381086b7`).


ivankuznetsov:

# Codex CE Code Review Pass 1

No plan.md found in the task folder. Reviewed `git diff origin/main..HEAD`.

## High

No findings.

## Medium

No findings.

## Nit

- [ ] `GH_BROWSER` / `BROWSER` scrubbing lacks regression coverage: `bin/hive-babysitter-stub-gh` now deletes these env vars before `exec`, but the added test only covers `--web` / `-w` flag skips, so an env-only browser-launch regression would not fail.

PR #383, opened 2026-06-07
Hive patrol: Manifest generation can mask failures

ivankuznetsov:
## Summary

E2E artifact capture isolates each artifact writer behind `guard(...)`, but `write_manifest` ran *outside* that guard. Because manifest generation walks the whole scenario directory and calls `File.size`/`Digest::SHA256.file` on every file, a live TUI log being deleted, rotated, or made unreadable while the TUI/reaper was still active could raise mid-manifest — and that exception would replace the original scenario failure the capture step exists to preserve.

This change makes manifest generation fault-tolerant:

- `write_manifest` is now wrapped in `guard("manifest.json")`, matching the other artifact writers.
- Per-file manifest entries are built via a new `manifest_entry(path)` helper that rescues `StandardError`, records a `capture_errors` entry (`label: "manifest:<relative-path>"`), and skips the file (`filter_map`) instead of aborting the whole manifest.

Net effect: a disappearing or unreadable file degrades to a recorded capture error rather than masking the real failure under investigation.

## Test plan

- Added `test_manifest_records_capture_error_when_file_disappears_during_digest`, which stubs `Digest::SHA256.file` to raise `Errno::ENOENT` for one file and asserts the manifest omits that file, records a `manifest:racy.log` capture error, and does not raise.
- `bundle exec ruby -Itest test/e2e/lib/artifact_capture_test.rb` — 10 runs, 33 assertions, passing.

## Review summary

Codex CE code review (pass 01) ran against `git diff origin/main..HEAD` with `git diff --check` and the artifact-capture test suite. Zero findings (High / Medium / Nit all empty); no escalations required.

## Linked task

Patrol finding: `test-suite-test-babysitter-acceptance-dry-run-test-rb-2`
Fingerprint: `01ed3b4d64fcbe09bb79a0960a56c452b8ee8497f4e3158d1ef9907b71f2e7e0`

### DIFF
--- a/test/e2e/lib/artifact_capture.rb
+++ b/test/e2e/lib/artifact_capture.rb
@@ -53,7 +53,7 @@ module Hive
         guard("tui-subprocess") { copy_tui_subprocess_diagnostics }
         guard("tui-subprocess-live") { remove_live_tui_subprocess_diagnostics }
         guard("step-results.json") { write("step-results.json", JSON.pretty_generate(step_results)) }
-        write_manifest
+        guard("manifest.json") { write_manifest }
       end
 
       private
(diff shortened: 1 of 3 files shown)


ivankuznetsov:

Outcome: failure

Agent tail:

```

```

PR #393, opened 2026-06-07
fix(daemon): auto-retry finalize unpushed commits

ivankuznetsov:
## Summary
- auto-clear 8-finalize ERROR reason=unpushed_commits when no live task lock exists
- keep manual-only clean-exit failures red
- add bounded retry-budget coverage for ordinary interrupted-agent recovery

## Tests
- bundle exec ruby -Itest test/unit/daemon/stale_agent_healer_test.rb
- bundle exec ruby -Itest test/unit/task_action_test.rb
- bundle exec ruby -Itest test/integration/run_finalize_test.rb
- git diff --check


ivankuznetsov:

# Codex CE Code Review 01

No plan.md found in the task folder; review is based on `git diff origin/main..HEAD`.

## High
_No findings._

## Medium
- [ ] `lib/hive/daemon/stale_agent_healer.rb:156` clears finalize `ERROR` markers without the observed `marker_id`: matching only `reason=unpushed_commits` can clear a newer same-reason marker from a stale daemon snapshot, bypassing the race-safe guard used by existing ERROR recovery paths.

## Nit
_No findings._

ivankuznetsov:

# Code Review — fix(daemon): auto-retry finalize unpushed commits

- **No plan.md found** in the task folder — reviewed without plan grounding.
- **Scope note:** a rebase is mid-flight in the worktree (`origin/main..HEAD` is empty). The change under review is the **staged diff** (`git diff --cached`), equivalent to commit `d0608bac`.
- **Reviewed:** `lib/hive/daemon/stale_agent_healer.rb`, `test/unit/daemon/stale_agent_healer_test.rb`, plus wiki docs (`daemon.md`, `finalize.md`).
- **Lenses run (report-only):** correctness, reliability, adversarial, testing, maintainability. Cross-file claims verified against `dispatcher.rb`, `policy.rb`, `task_action.rb`, `finalize.rb`, `gh.rb`, `commands/run.rb`, `markers.rb`.
- **Local checks:** unit tests pass (36 runs, 0 failures); rubocop clean.

The budget accounting, the `marker_id`-guard vs recovery-key interaction (key omits `marker_id` so retries with fresh ids share one budget; `match_attrs` includes it to guard a TOCTOU clear of a newer marker), and the `clear_current` atomicity all trace correctly. The findings below are mostly about whether the *re-dispatch actually fires*, retry hygiene, and an inaccurate premise in the comment/docs.

## High
- [ ] Finalize re-dispatch path is not the clean ADVANCE path the review recovery relies on — it lands on `NEEDS_INPUT`/edit-resume, which can `:record_baseline` and skip: Clearing a **review** marker yields `ready_for_review` (an `ADVANCE_ACTIONS` member → immediate `:dispatch`). But clearing the **finalize** `unpushed_commits` error leaves the row markerless, so `TaskAction#finalize_action` returns `finalize_waiting` = `NEEDS_INPUT` (`task_action.rb:139-143,329-333`), which `Policy#decide` routes through `edit_resume?`→`decide_edit` (`policy.rb:73-75,125-129,169-180`). On first sight with `last_dispatched == nil` (e.g. the row errored under an external `hive run` the daemon never dispatched), `decide_edit` returns `:record_baseline` and **skips dispatch** — the heal already logged `marker_healed` and consumed a retry, but finalize never re-runs and the task strands as a markerless "Needs your input" row. No integration test proves finalize actually re-runs after the clear; the unit tests only assert the marker was removed.

## Medium
- [ ] No backoff and no grace window before clearing: a persistent push failure becomes a back-to-back retry burst: Unlike `heal_orphaned_review_row` (gated on `(now - mtime) > @grace_sec`, `stale_agent_healer.rb:363-365`), `heal_error_if_auto_recoverable` clears on the first tick it observes the marker. For a non-transient failure (expired `gh` auth, branch-protection rejection) all 3 retries fire across consecutive ticks, each re-dispatching a full finalize run, with zero spacing. The `hive run` auto-rebase pre-step (`commands/run.rb:68,80`) rebases the *default* branch, not the feature branch's own upstream, so it does not resolve auth or feature-branch-divergence failures.
- [ ] Budget exhaustion is silent — no event distinguishes "gave up after 3" from "never tried": `return if attempts >= @error_auto_recovery_limit` (`stale_agent_healer.rb:154`) exits with no log. After exhaustion the marker correctly stays red, but there is no observable `recovery_exhausted` signal to route the operator's attention; every later tick silently short-circuits. The pre-existing review path (`:218`) has the identical gap — worth fixing together.
- [ ] In-memory retry budget is reset by SIGHUP reload and process restart, defeating the "3×/process" cap: `reload_config!` rebuilds the healer (`dispatcher.rb:1493`) with fresh `Hash.new(0)` budgets **and** drops the `error_auto_recovery_limit:` / `review_error_auto_recovery_limit:` kwargs (silently reverting a tuned limit to 3). A routine config reload — or a daemon crash/OOM-restart loop, since the dispatch baseline persists but the budget does not — grants a fresh 3× budget against the same unpushable branch. Pre-existing for the review path; this PR extends the same latent issue to the finalize path.
- [ ] Doc comment and wiki misdescribe the trigger as interrupted-session "residue/commits," but `unpushed_commits` is strictly a push-gate failure: `finalize.rb:194-202` writes `reason=unpushed_commits` only when `git push -u` fails or the branch still isn't on the remote; residue is the separate `ensure_clean_on_exit_failed` reason (`finalize.rb:179-186`, deliberately *not* auto-recovered), and `gh.rb:17-20` confirms it means "a persistent push failure." The comment at `stale_agent_healer.rb:52-56` ("leaves relevant commits or safe residue in the feature worktree") could mislead a future maintainer into broadening `auto_recoverable_error?` to residue-class errors on a false premise.
- [ ] No test exercises cross-`marker_id` budget accumulation — the actual production case: Finalize mints a fresh `marker_id` on every push failure (`markers.rb:154-159`), so successive failures carry different ids. The recovery key intentionally omits `marker_id` (`stale_agent_healer.rb:200-207`) so the budget accumulates — but `test_auto_recoverable_finalize_unpushed_commits_stays_red_after_retry_budget_is_exhausted` reuses identical no-`marker_id` marker text on both passes, so it would still pass even if the key wrongly included `marker_id`. Add a test with distinct `marker_id`s per pass (limit=1) to pin this load-bearing invariant. Flagged independently by correctness, reliability, and testing.

## Nit
- [ ] `finalize_unpushed_commits` is added to `REASONS` but is never written onto a marker, breaking the constant's documented contract: `REASONS` is documented as "reason= attribute values the healer **writes onto** ERROR markers" (`stale_agent_healer.rb:70-73`); every other member is set via `Markers.set(reason:)`. This path *clears* the marker and the label appears only as a log field — and it also collides with finalize's commit label `{ commit: "finalize_unpushed_commits" }` (`finalize.rb:202`) and the marker's own `unpushed_commits` reason. Either move it out of `REASONS` or split the constant's doc into two categories; consider a distinct healer label (e.g. `finalize_retried_after_push_failure`).
- [ ] `heal_error_if_auto_recoverable` is a near-verbatim clone of `heal_review_error_if_auto_recoverable` with parallel helper families: Duplicated skeleton (live-lock return → recoverable? guard → key/attempts → `clear_current` → increment → `marker_healed`/rescue) plus parallel `auto_recoverable_error?`/`*_review_error?`, `error_auto_recovery_key`/`review_error_auto_recovery_key`, two ivars, and two limit constants both equal to 3. Consider one shared method parameterized by a small per-marker policy struct. Judgment call — the paths do differ (review computes phase/pass/signature; error computes marker_id), so weigh against premature over-generalization.
- [ ] `row.marker_attrs` Hash-guard idiom is copy-pasted ~7× (`stale_agent_healer.rb:184,193,250,255,278,299,416`); this PR adds two more. Extract a `marker_attrs_for(row)` accessor so the `respond_to?`/`is_a?(Hash)` guard lives in one place. Pre-existing, widened here.
- [ ] Recovery is structurally unable to fix non-transient push failures (auth, non-fast-forward): such failures retry identically up to 3× — spawning up to 3 finalize agent runs — then go red. Acceptable bounded degradation, but classifying transient (network/5xx) vs permanent push failures and only auto-recovering the transient class would avoid the wasted runs. Lower confidence given the rarity of feature-branch divergence and the auto-rebase pre-step.
- [ ] Uncovered defensive branches: no test for `marker_attrs` missing/non-Hash, the empty/absent `reason` case (`auto_recoverable_error?` should be false), or a positive assertion of the no-`marker_id` reason-only clear path (`stale_agent_healer.rb:191-198`). All are safe-by-construction but unpinned.
- [ ] Snapshot-vs-heal TOCTOU on `live_task_lock`: the snapshot's `row.live_task_lock` can be a stale `false` while an external `hive run finalize` acquires the `.lock` mid-tick; `controller.running_task?` only tracks in-process dispatches, so the healer can clear a marker the external run is already recovering, consuming a budget slot for no benefit. Mostly cosmetic; mirror `heal_wedged_review_row` by re-reading the live `.lock` at heal time if tightened.

ivankuznetsov:

# PR Review Toolkit — fix(daemon): auto-retry finalize unpushed commits

- **No plan.md found** in the task folder — reviewed without plan grounding.
- **Scope:** `git diff origin/main..HEAD` = commit `29ca91bc` (the rebase that was mid-flight earlier has since completed; the range now resolves to this single commit).
- **Working-tree caveat:** the on-disk files carry an **uncommitted** addition not in `29ca91bc` — a `observe_pre_clear_mtime(row)` method that seeds & persists the controller dispatch baseline with the pre-clear mtime (via `ConcurrencyController#observe_state_file_mtime`). It mitigates the edit-resume re-dispatch concern below; findings are scored against the committed diff and annotated where that addition already addresses them.
- **Personas run:** code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, code-simplifier. (type-design-analyzer skipped — no new types/structs introduced.)
- **Local checks:** unit tests pass (36 runs, 0 failures); rubocop clean on both files.

Line refs are approximate against the committed file; the uncommitted `observe_pre_clear_mtime` shifts later lines by ~8.

## High
- [ ] [comments] Class-header comment mischaracterizes the trigger as interrupted-session "residue/commits" (`stale_agent_healer.rb:52-56`): `reason=unpushed_commits` is written **only** when finalize's push gate fails (`finalize.rb:194-202`; `detail` is the push stderr; `gh.rb:17-20` calls it "a persistent push failure"). Dirty-worktree residue is the separate, deliberately-non-recovered `ensure_clean_on_exit_failed` reason — a maintainer could wrongly broaden `auto_recoverable_error?` on this false premise.
- [ ] [comments] `REASONS` doc comment ("reason= values the healer **writes onto** ERROR markers") is now inaccurate (`stale_agent_healer.rb:70-73,80`): `finalize_unpushed_commits` is the lone entry that is never written onto a marker — this path *clears* the marker, and the string appears only as a log-event field. It also differs from the real marker reason (`unpushed_commits`) by the `finalize_` prefix, so a consumer matching marker `reason=` against `REASONS` will never match it.
- [ ] [tests] The PR's central design invariant — budget key omits `marker_id` so retries accumulate across the fresh marker_ids finalize mints — is **untested**: `test_auto_recoverable_finalize_unpushed_commits_stays_red_after_retry_budget_is_exhausted` reuses identical no-`marker_id` marker text on both passes, so it would pass green even if `marker_id` leaked back into `error_auto_recovery_key`. Add a test driving distinct marker_ids (`err-a` then `err-b`, limit=1) asserting the second stays red.
- [ ] [tests] Budget-bucket isolation per `[project,slug,stage,reason]` is untested: every test uses the default `project:"p", slug:"s"`, so a regression dropping `slug`/`project` from the key (collapsing all tasks into one bucket, letting one noisy task starve every other task's recovery) would not be caught.
- [ ] [errors] Budget exhaustion is a silent failure (`stale_agent_healer.rb:154`): `return if attempts >= @error_auto_recovery_limit` emits no event, and the `Logger` enum has no `marker_heal_exhausted`. The "automation gave up, a human must act" transition is invisible — an operator sees three `marker_healed` events then silence. Emit a one-shot exhaustion event (guarded by a seen-set so it fires on the transition, not every tick); shared with the review path at `:218/:229`.

## Medium
- [ ] [code] Committed diff relies on an uncommitted mitigation for re-dispatch: clearing the marker leaves the finalize row markerless → `finalize_waiting` = `NEEDS_INPUT`, routed through `Policy#decide`'s edit-resume/debounce path, not a direct advance. The on-disk `observe_pre_clear_mtime` seeds the baseline so the post-clear mtime bump triggers dispatch — ensure that method ships **in the same commit/PR**, since `29ca91bc` alone does not re-fire the retry promptly (and first-sight-with-no-baseline would `:record_baseline` and skip).
- [ ] [errors] `clear_current` returning `false` is indistinguishable from a success-skip (`stale_agent_healer.rb:156-160`): a benign race (finalize already advanced) and a persistent `marker_id` mismatch both return silently with no event and no attempt consumed — so a stale-`marker_id` storm would no-op forever, never exhausting, never logging. Add a low-severity `skipped`-style trace or an inline comment explaining the silent return (unlike the self-evident returns at 148-149).
- [ ] [errors] `rescue StandardError` is too broad (`stale_agent_healer.rb:175-181`): it wraps the whole method (including `@logger.event(:marker_healed,...)`), so a `NoMethodError`/`KeyError`/logger `ArgumentError` gets mislabeled as `marker_heal_failed reason=finalize_unpushed_commits`, sending operators to chase a non-existent disk/push problem. Narrow to `SystemCallError, IOError` (the file already scopes `auto_recoverable_review_error?`/`review_error_signature` rescues to `SystemCallError`). Note `heal` has no per-row rescue, so confirm the propagation target before narrowing.
- [ ] [tests] In-memory budget reset on daemon restart / SIGHUP reload is untested: the exhaustion test reuses one `@healer` instance. A test that exhausts on one instance then heals the same row on a fresh instance would pin the reset as *intended* (so a future "persist the budget" change is conscious).
- [ ] [tests] No test for the no-`marker_id`-on-disk-vs-present interaction: `auto_recoverable_error_match_attrs` only adds `marker_id` to the clear-match when the row carries one, so a stale row lacking a `marker_id` could clear a fresh on-disk marker whose id it never validated — a real race given finalize always mints ids. Pin the intended behavior.
- [ ] [tests] No direct test that `clear_current → false` (raw marker changed, matching reason) consumes no budget and emits no event on the error path — the ordering of the budget increment (after the clear) is currently unprotected for that case.
- [ ] [comments] Comment omits the retry-budget bound (`stale_agent_healer.rb:55-56`): "without requiring a human to press retry" reads as unbounded; add "...up to `ERROR_AUTO_RECOVERY_LIMIT` attempts, after which the marker stays red for manual recovery."
- [ ] [simplify] Extract a `marker_attrs_for(row)` accessor: the guard `row.respond_to?(:marker_attrs) && row.marker_attrs.is_a?(Hash) ? row.marker_attrs : {}` is duplicated verbatim ~7× (lines ~184,193,250,255,278,299,416); this PR adds two more. One accessor gives a single place to change the guard.

## Nit
- [ ] [code] `prior_marker: row.marker` logs the bare string `"error"`, losing the reason (`stale_agent_healer.rb:166-167`); `marker_reason: reason` is emitted alongside so the diagnostic is complete, but the field is less informative than it looks in isolation.
- [ ] [code] In-memory `@error_auto_recoveries` reset on restart/SIGHUP is consistent-by-design with the review path (not a regression); worth a one-line doc noting the budget is per-process and SIGHUP reload (`dispatcher.rb:1493`) also drops a tuned limit kwarg.
- [ ] [tests] `observed_mtimes` assertion is coupled to the `make_row` mtime default (`NOW - 1000`), overstating what it guards — `observe_pre_clear_mtime` is an unconditional pass-through, so the assertion only confirms pass-through, not mtime logic.
- [ ] [tests] `refute_match(/ERROR/, ...)` is broader than intended (also passes on a partial/garbage write omitting the literal); assert `Hive::Markers.current(state_file)` is `:none` for a stronger, format-resilient contract.
- [ ] [tests] The "does not recover" negatives (live-lock / wrong-stage / wrong-reason) assert no `marker_healed` and that the marker survives, but don't `refute` a `marker_heal_failed` — a regression that throws on those paths would stay green while logging spurious failures.
- [ ] [simplify] After the `marker_attrs_for` extraction, `auto_recoverable_error_reason` and `auto_recoverable_review_error_reason` collapse to one `marker_reason(row)` helper (or inline at the few call sites).
- [ ] [simplify] Redundant ternary `reason == "review_agent_died" ? "review_agent_died" : ...` (`stale_agent_healer.rb:233`, pre-existing) — the true-branch re-yields the compared value; the new error path's fixed-`reason:` + separate `marker_reason:` is the cleaner convention to converge on.
- [ ] [simplify] Guidance (not a defect): do **not** unify `heal_error_if_auto_recoverable` with `heal_review_error_if_auto_recoverable` behind a shared policy struct — the two diverge in recovery key, match_attrs, side effects (`observe_pre_clear_mtime`), log payload, and recoverability predicate; this is load-bearing parallelism and unifying would be premature abstraction.

ivankuznetsov:

# Codex CE Code Review 02

No plan.md found in the task folder; review is based on `git diff origin/main..HEAD`.

Scope note: the endpoint diff also shows base-only changes from `origin/main` because this branch is not rebased on current `origin/main`; I did not flag those apparent deletions as branch defects.

## High
_No findings._

## Medium
- [ ] `lib/hive/daemon/stale_agent_healer.rb:214` samples `marker_id` outside the locked clear: a legacy no-`marker_id` row can still clear a newer same-reason `ERROR` if the marker changes between `Markers.current` and `clear_current`.

## Nit
_No findings._

ivankuznetsov:

# Code Review — review-pr-393-healer-auto-retry (pass 2)

- **No `plan.md` found** in the task folder; this review proceeds without plan grounding.
- **Scope note:** `git diff origin/main..HEAD` is noisy — `origin/main` has advanced past this branch's base (commits #394, #331), so the diff shows wiki/CLI/`wiki_log` *deletions* that are actually content present on `main` and absent here, **not** changes this branch makes. The real contribution is the 4 healer commits (`merge-base..HEAD`) touching `lib/hive/daemon/stale_agent_healer.rb`, `lib/hive/daemon/logger.rb`, and tests. Review is scoped to those. Rebasing onto current `origin/main` before merge will collapse the unrelated churn.
- **Working-tree state:** `lib/hive/daemon/stale_agent_healer.rb` has an **uncommitted modification** (`git status` = `M`). This matters — see the first High finding. Unit tests (43 runs, 0 failures) and `rubocop` pass against the working-tree state.

## High
- [ ] Committed HEAD carries a TOCTOU race in `auto_recoverable_error_match_attrs` that the uncommitted working-tree change fixes — but the fix is not committed: At HEAD (`stale_agent_healer.rb:210-223`), the `marker_id.empty?` branch reads `Hive::Markers.current(row.state_file)` *outside* the markers lock and only sets a sentinel when the on-disk marker already has an id. If the unlocked read sees *no* id, no `marker_id` constraint is added to `match_attrs`; a concurrent finalize that writes a fresh `ERROR ... marker_id=…` between that read and `clear_current`'s under-lock evaluation would then match on `reason` alone and **clear the brand-new marker** (and the heal miscounts as a recovery of the old failure). The working-tree version replaces this with `match_attrs["marker_id"] = nil`, which forces the under-lock match to require an empty on-disk `marker_id` — closing the window. Commit the working-tree change (or the PR ships the racy version). Narrow trigger (legacy no-`marker_id` ERROR markers only exist for tasks that errored before this feature), but the fix already exists and must be in a commit to be the reviewed/merged code.

## Medium
- [ ] Recovery budget (`@error_auto_recoveries`, keyed `[project, slug, stage, reason]`) is never reset on a *successful* finalize within the same process: In practice a successful push advances the task out of `8-finalize`, so re-entry is unlikely — but if a task ever re-reaches `ERROR reason=unpushed_commits` at `8-finalize` in the same daemon lifetime after an intervening success, it inherits the already-consumed count and gets a silently reduced (or zero) retry budget. Either reset the key on observed success/stage-advance, or document why same-process re-entry can't happen. The review path shares the same no-reset shape, so this is a pre-existing pattern, not a regression.
- [ ] `observe_pre_clear_mtime` silently no-ops when `row.state_file_mtime` is nil, but a retry has already been consumed: `observe_state_file_mtime` returns early on `mtime.nil?` (`concurrency_controller.rb:165`). The integration test (`daemon_stale_agent_healing_test.rb:167`) establishes that seeding the pre-clear baseline is *load-bearing* — without it the markerless `8-finalize` row strands on `:record_baseline` instead of re-dispatching. So if `state_file_mtime` is ever nil, the heal clears the marker and burns a budget slot while finalize fails to re-fire. An on-disk ERROR marker with a nil stat mtime is near-impossible, so severity is low, but the failure is asymmetric (clear succeeds, re-dispatch silently doesn't, budget still decremented). Worth a guard (skip the clear when mtime is nil) or at least a comment pinning the assumption.

## Nit
- [ ] No backoff/spacing is enforced between the 3 finalize retries; spacing is incidental (each retry requires a full finalize agent re-dispatch, which takes real time): This is fine today because a deterministic `unpushed_commits` cause (e.g. protected branch, genuine auth loss) is bounded to 3 extra agent runs before going red — but it's worth a one-line comment that the only thing preventing rapid-fire retries is finalize's own runtime, not any healer-side delay.
- [ ] `REASONS` constant gained `finalize_unpushed_commits`, but the doc comment (`stale_agent_healer.rb:72-75`) says it is "logged when the healer clears" — unlike the other entries it is never written *onto* a marker. The comment already distinguishes this; just confirm no consumer of `REASONS` (TaskAction synthetic diagnostic, bot rendering) assumes every member is a marker-writable reason.
- [ ] Integration helper `status_rows_via_consumer` (`daemon_stale_agent_healing_test.rb:82-94`) builds `Row` without `marker_attrs`/`live_task_lock`, diverging from the production `StatusConsumer` shape (`status_consumer.rb:90,97` populate both). The finalize test hand-builds its own row so this doesn't hide a bug today, but the helper is now slightly less faithful to production than its docstring claims ("whatever shape status produces"). Consider threading both fields through for fidelity.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — review-pr-393-healer-auto-retry (pass 2)

- **No `plan.md` found** in the task folder; this review proceeds without plan grounding.
- **Scope note:** `git diff origin/main..HEAD` is noisy — `origin/main` advanced past this branch's base, so the diff shows wiki/CLI/`wiki_log` *deletions* that are content on `main` absent here, **not** this branch's changes. Review is scoped to the feature commits (`4b33b92f..HEAD`) touching `lib/hive/daemon/stale_agent_healer.rb`, `lib/hive/daemon/logger.rb`, and the two test files.
- **State change since pass 1:** the working tree is now **clean**. The previously-uncommitted lock-safe fix to `auto_recoverable_error_match_attrs` (`match_attrs["marker_id"] = nil`) was committed as `ac01fb60 fix(daemon): make finalize marker clears lock-safe`. HEAD now contains the lock-safe version, so the pass-1 TOCTOU/uncommitted **High finding is resolved** — both `code-reviewer` and `comment-analyzer` independently verified the committed `nil` match is evaluated under the markers lock and correctly refuses to clear a marker that gained an id concurrently.
- **Validation:** rubocop clean; healer unit tests 44/44, integration 5/5, logger 8/8 pass (per personas).
- Personas run: `code-reviewer`, `silent-failure-hunter`, `pr-test-analyzer`, `comment-analyzer`, `type-design-analyzer`.

## High
*(none — the pass-1 TOCTOU concern was resolved by `ac01fb60`; no persona surfaced a Critical/Important issue at HEAD)*

## Medium
- [ ] [comment-analyzer] `REASONS` doc comment (`stale_agent_healer.rb:72-75`) inaccurately claims the "agent/review-orphaned values are also written onto ERROR markers": only `agent_died`/`agent_orphaned` are written onto markers (via `Markers.set` in `heal_row`); the review reasons and `finalize_unpushed_commits` are emitted only as `reason:`/`marker_reason:` log labels — those paths clear markers rather than stamping these strings on.
- [ ] [type-design] `REASONS` constant no longer expresses a coherent closed set: it mixes marker-writable reasons with log-only labels (`finalize_unpushed_commits`, `reviewer_tmux_session_terminated`) and omits the actual matched marker reason `"unpushed_commits"`; it has no production consumer (referenced only in its test), so the fix is cheap — split into `MARKER_REASONS` + `HEAL_LOG_LABELS` (or rename and correct the docstring). Same root as the comment-analyzer finding.
- [ ] [silent-failure] `marker_heal_exhausted` fires once per process and the per-process budget silently refills on daemon restart, while the event name implies a permanent terminal state: an operator reading the log may over-read "red forever" when a daemon bounce re-arms 3 more clears — document the per-process refill on the event and add an actionable remediation hint (the event carries `attempts`/`max_attempts` but nothing tells the operator to rerun finalize / push manually).
- [ ] [pr-test-analyzer] The legacy `marker_id=nil` reconciliation guard — the single subtlest correctness claim in the diff (a stale id-less row must not clear a marker that gained an id between status and heal) — is only tested as a static mismatch (`stale_agent_healer_test.rb:710`), not as a row-snapshot-vs-changed-on-disk race; an explicit test that rewrites the on-disk marker before `heal` would lock the TOCTOU intent against future refactors of `match_attrs`.

## Nit
- [ ] [type-design] Recovery-budget keys are anonymous stringly-typed positional tuples duplicated across two builders with different arity and field order (`reason` at position 3 in `review_error_auto_recovery_key:333` vs position 4 in `error_auto_recovery_key:239`); a transposition would silently mis-bucket budgets — a small `RecoveryKey` value object shared by both paths removes the drift risk (highest-leverage suggestion).
- [ ] [type-design] `ERROR_AUTO_RECOVERY_LIMIT` and `REVIEW_ERROR_AUTO_RECOVERY_LIMIT` are both literally `3` (`:84-85`) with nothing linking the intended symmetry; bumping one default silently leaves the other. Separate injectables are correct by design — optionally derive both from a shared default to express the relationship.
- [ ] [type-design] `marker_attrs_for` (`:498-503`) re-enforces the "is it a Hash" invariant the production `Row` producer already guarantees (`status_consumer.rb:90`); the nil/non-Hash branches are reachable only from test doubles. Pushing the guarantee into the `Row` Struct would let the call sites read `row.marker_attrs` directly and delete the duplicated guard.
- [ ] [pr-test-analyzer] No test feeds two rows for the *same* task in a single `heal` pass (the second `clear_current` returns false and must consume no budget / emit no event); the `clear_current=false` building block is tested in isolation (`:950`) but not the same-pass duplicate case.
- [ ] [pr-test-analyzer] The integration assertion `post_clear_mtime > baseline` (`daemon_stale_agent_healing_test.rb`) is safe only because of the 1000s pre-clear offset constant; a comment tying the assertion to that constant would prevent filesystem-mtime-resolution flakiness if the offset is later shrunk.
- [ ] [code-reviewer] Recovery-budget key omits `marker_id` (`:235-242`), so a genuinely new `unpushed_commits` failure with a fresh `marker_id` shares the same 3-strike budget rather than resetting — confirmed intended (per-process "3 total finalize-push attempts") and documented in `235b0e10`; flagged only so the no-reset-on-new-failure semantics are a conscious choice.

ivankuznetsov:

# Codex CE Code Review Pass 3

No plan found in the task folder; review proceeded without plan grounding.

Reviewed diff: `git diff origin/main..HEAD`.

## High
- [ ] Preserve merged wiki compile-log workflow: branch is 3 commits behind `origin/main` and the endpoint diff deletes #331's public `hive wiki compile-log` command, `Hive::WikiLog`, `wiki/log.d`, and their tests/docs.

## Medium

## Nit
- [ ] Preserve merged README Discord link: the endpoint diff also reverts #394's community link even though the healer commits do not touch README.

ivankuznetsov:

# Code Review — PR #393 healer auto-retry (unpushed_commits), pass 3

> **No plan.md found in the task folder** — review proceeded without plan grounding.

Scope reviewed: `git diff origin/main...HEAD` after refreshing `origin/main`
(the raw `origin/main..HEAD` initially showed unrelated wiki-command deletions;
those were a stale-ref artifact and are **not** part of this branch). Real
scope: `lib/hive/daemon/stale_agent_healer.rb`, `lib/hive/daemon/logger.rb`,
healer unit + integration tests, and four wiki pages.

Verification done: `test/unit/daemon/stale_agent_healer_test.rb` (46 runs, 0F)
and `test/integration/daemon_stale_agent_healing_test.rb` (5 runs, 0F) both
pass. Confirmed `hive status --json` populates `attrs` and `mtime` for error
rows (`lib/hive/commands/status.rb:458,461`), so the production status→healer
chain reaches this code. Finalize push is idempotent (`return nil if pushed?`,
`lib/hive/stages/finalize.rb:194`), so re-dispatch is safe to repeat.

## High
- [ ] (none) — no correctness/data-safety defect found; the marker_id TOCTOU guard, budget accounting, duplicate-row handling, and pre-clear mtime seeding are all tested and sound.

## Medium
- [ ] No transient-vs-permanent push-failure classification: `auto_recoverable_error?` (`stale_agent_healer.rb:217`) retries *any* `8-finalize ERROR reason=unpushed_commits`, so a permanent failure (protected-branch rejection, non-fast-forward, revoked auth) spends a full finalize agent run per retry; finalize records the push stderr in `detail` (`stages/finalize.rb:201`) but the healer ignores it — consider routing known-permanent stderr signatures to manual-only instead of retrying.
- [ ] Per-process budget reset + zero inter-retry backoff amplifies the above: the recovery budget is in-memory and re-armed on every daemon restart / SIGHUP healer rebuild (documented at `stale_agent_healer.rb:311-317`), and the only spacing is finalize's own runtime — so in a repo with frequent reloads a permanently-failing push can re-spawn finalize 3× *per reload* with no real ceiling. `marker_heal_exhausted` + the remediation hint soften it, but a permanent failure still burns agent runs indefinitely across reloads.
- [ ] `wiki/log.md` is hand-edited with 5 new `##` entries (and out-of-chronological order: `15:52` → `14:17` → `00:00`), against the project convention (CLAUDE.md) that compiled `wiki/log.md` is generated from `wiki/log.d/<ts>-<slug>.md` fragments in feature PRs — direct edits risk being clobbered or conflicting on the next `wiki compile-log`. Move these to `wiki/log.d/` fragments.

## Nit
- [ ] `MARKER_ERROR_REASONS` and `HEAL_LOG_LABELS` (`stale_agent_healer.rb:75,81`) are defined but referenced nowhere in `lib/` or `test/` — dead constants (the prior single `REASONS` was likewise unreferenced); either wire them into the code/log-label validation they document or drop them.
- [ ] Event key naming is inconsistent across the related heal events: `marker_healed` emits `attempt:` (singular, `:206`) while `marker_heal_exhausted` emits `attempts:` (plural, via `:174`) — log/dashboard consumers keying on attempt count will need to handle both.
- [ ] Finalize `error_auto_recovery_key` (`:246`) deliberately omits a failure signature, so two genuinely different push failures on the same task share one 3-retry budget — intentional per the comments, but it diverges from the sibling review path (`review_error_signature`, `:363`, hashes the errors file for a fresh budget); worth a one-line note in the class doc so the asymmetry isn't read as an oversight.
- [ ] No test drives a *real* `hive status` error row (carrying `attrs` + `mtime`) through the healer; the new integration test hand-builds the `Row`, leaving the status→error-row-attrs contract for this path unverified end-to-end. `wiki/gaps.md` #27 already acknowledges the "not live-smoked" gap, so this is informational.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — PR #393 healer auto-retry, pass 3

> **No plan.md found in the task folder** — review proceeded without plan grounding.

Scope: `git diff origin/main...HEAD` (origin/main fetched first; the raw
`origin/main..HEAD` showed stale unrelated wiki deletions — not part of this
branch). Changed files: `lib/hive/daemon/stale_agent_healer.rb`,
`lib/hive/daemon/logger.rb`, the healer unit + integration tests, and four
wiki pages. Personas run: **code-reviewer**, **test-analyzer**,
**comment-analyzer**, **silent-failure-hunter**. Unit (46 runs) and
integration (5 runs) suites pass; rubocop clean on both lib files.

Overall: no Critical/High defects. code-reviewer verified the three flagged
risk areas (marker_id TOCTOU guard, budget accounting, duplicate rows in one
heal pass) are all correct and tested. Findings below are correctness-adjacent
robustness, test-gap, and comment-accuracy items.

## High
- [ ] (none) — no persona found a correctness or data-safety defect; the marker_id TOCTOU guard, post-clear budget increment ordering, and duplicate-row handling were each traced and confirmed sound.

## Medium
- [ ] [comment-analyzer] Class-doc (`stale_agent_healer.rb:52-54`) lists "auth … agent loss" as a cause of the `ERROR reason=unpushed_commits` marker, but `finalize.rb:41` runs `ensure_authenticated!` (which *raises* on failure) before the push gate at `:194-202` — an auth failure never reaches the `unpushed_commits` marker write, so the stated cause is inaccurate.
- [ ] [test-analyzer] No end-to-end test drives the real `hive status --json` → error-row-attrs → healer chain for a finalize push failure; the integration test (`daemon_stale_agent_healing_test.rb:189-197`) and all finalize unit tests hand-build the `Row`, so a regression that stripped `marker_id` from status `attrs` would silently flip every production heal onto the legacy no-id path and defeat the TOCTOU guard with no failing test.
- [ ] [test-analyzer] The `log_recovery_exhausted_once` one-shot dedup guard (`stale_agent_healer.rb:318-321`) is never exercised: the exhaustion tests run exactly 2 heal passes and assert `exhausted.size == 1`, which only proves it fired once, not that the `seen` guard *suppresses a repeat* — a regression dropping the dedup (log spam every tick on a permanently-red task) would pass all tests.
- [ ] [silent-failure-hunter] `marker_heal_exhausted` reads as terminal, but both the budget counters and the exhausted `seen`-maps reset on every SIGHUP healer rebuild (`dispatcher.rb:1493`), not just on a full restart — so a routine config reload silently re-arms `max_attempts` fresh clears and re-fires the exhausted event; `budget_scope: "per_process"` is the only signal and undersells the reload-reset.

## Nit
- [ ] [silent-failure-hunter] `observe_pre_clear_mtime`'s `respond_to?` guard (`stale_agent_healer.rb:236-244`) silently no-ops if a controller lacks the method — production always satisfies it, but a future controller swap could reintroduce re-dispatch stranding with zero log signal; consider dropping the guard (let it raise into the rescue) or emitting a debug event when absent.
- [ ] [silent-failure-hunter] The `rescue` in `heal_review_error_if_auto_recoverable` hard-codes `reason: "reviewer_tmux_session_terminated"` (`stale_agent_healer.rb:307`), so a `marker_heal_failed` raised while healing a `review_agent_died` marker is mislabeled as the tmux-death channel; derive the label from the row or use a neutral `review_error`.
- [ ] [comment-analyzer] Both the code comment (`stale_agent_healer.rb:56-57`) and `wiki/modules/daemon.md` list the finalize re-run checks as "clean-exit, auth, … push", but `finalize.rb#run!` runs auth first (`:41`) then clean-exit (`:150-192`) then push — reorder the prose to match if it is meant to read as a sequence.
- [ ] [comment-analyzer] `stale_agent_healer.rb:57` says "up to ERROR_AUTO_RECOVERY_LIMIT times", but the effective ceiling is the constructor-injectable `@error_auto_recovery_limit` (the constant is only the default); soften to "the configured limit (default 3)" to match the injectable design and the wiki wording.
- [ ] [test-analyzer] The integration test hard-codes `action: "needs_input"` into the `Policy.decide` call (`daemon_stale_agent_healing_test.rb:218`) rather than reading it back from the real status pipeline on the cleared marker, so the load-bearing "would strand on :record_baseline" premise is asserted by comment, not derived — combining with the end-to-end gap above would close both.
- [ ] [test-analyzer] The signal-sequence assertions in `test_terminate_lock_holder_*` (stubbed `Time.now` + exact `[["TERM"…],["KILL"…]]` order) are the most implementation-coupled tests in the file and would break under a benign refactor of the escalation loop — acceptable but worth noting.

PR #399, opened 2026-06-07
Hive patrol: hv delegates to unverified hive candidates

ivankuznetsov:
## Summary

The `hv` fallback shim previously exec'd the first executable `hive` candidate on the search path without confirming it was actually this Hive CLI. Any candidate location (XDG, Homebrew, `/usr/local`) holding an Apache Hive binary would be run instead — re-opening the Apache Hive name-collision class the shim is meant to guard against.

`bin/hv` now version-gates each candidate before exec: it runs the candidate's `--version`, takes the first output line, and requires the strict bare semver contract (`MAJOR.MINOR.PATCH`) used by the installer docs. Candidates that fail the probe or print anything else (e.g. Apache Hive's `Hive 4.0.0`) are skipped, and the shim falls through to the next candidate.

## Test plan

- `test/unit/hv_test.rb` updated so the existing override fixture answers `--version` with a valid bare semver.
- New regression test `test_apache_style_xdg_candidate_is_skipped_for_next_valid_hive_cli`: an XDG candidate prints `Hive 4.0.0` (Apache style) and a Homebrew candidate prints `1.2.3`. Asserts the shim skips the Apache-style binary and execs the valid Hive CLI (`real:probe`).
- Run: `ruby -Itest test/unit/hv_test.rb`

## Review summary

Patrol-originated fix routed through the standard 6-review flow. Codex CE code review pass 1 reported zero findings (High/Medium/Nit all empty); triage recorded no escalations and no user questions.

## Linked task

- Patrol finding: `command-bin-hv-1`
- Fingerprint: `4bde6dfe0b5fdd478b1de1b524b8c02e622e85e989764c95596f7199cdae1617`
- Task folder: `.hive-state/stages/8-finalize/patrol-command-bin-hv-4bde6dfe`


ivankuznetsov:
Closing as **superseded / redundant**: `main` already enforces this behavior. Verified by reconstructing this PR's new test assertions onto current `main` and running them — they pass without this PR's changes, so the fix has already landed via earlier merges. No net behavior to add. (Triage performed during the patrol-PR finalization sweep.)

PR #413, opened 2026-06-08
feat(tui): scrollable, word-wrapping help overlay

ivankuznetsov:
## Summary

The TUI help overlay (`?` from the grid) rendered the entire keybinding cheatsheet into a single unbounded box. On short terminals the alt-screen clipped the **top** of the help and there was no way to scroll, so the upper sections were unreadable; long lines were silently truncated at the right edge.

Help now fits any terminal: content word-wraps to the available width, a height-bounded viewport shows one window at a time with a right-edge scrollbar, and you scroll with the keyboard or the mouse wheel. Below a 10×40 floor it falls back to a centered "terminal too small" message instead of clipping or crashing.

The change keeps the existing MVU shape (Model → KeyMap → Update → View, all pure) and mirrors the already-shipped scroll pattern from `RedStatusDetail` (offset-in-model + clamp helper + window slice).

| Area | Change |
|---|---|
| Viewport | `HelpOverlay.render` word-wraps each line to the inner width, slices a viewport-sized window starting at `help_scroll_offset`, and draws a right-edge scrollbar (`█` thumb / `│` track) only when content overflows |
| Scroll state | New `help_scroll_offset` field on `Model`, reset to 0 on each open and re-clamped on resize so a shrink keeps a valid position |
| Keys | `↑/↓`, `j`/`k`, `PgUp`/`PgDn`, `Home`/`End`, `g`/`G` scroll the viewport; `Esc`/`?`/`q` close. Navigation keys no longer dismiss the overlay |
| Mouse | SGR mouse reports are decoded in `InputDecoder`; the wheel scrolls help (3 lines per notch) and is inert in every other mode |
| Fallback | Below 10×40, a centered "Terminal too small for help (need 10×40)" message renders with no border and no crash; close keys still work |

Word-wrap was chosen over horizontal scroll: no horizontal state, nothing silently cut. The wrap helper was extracted from `RedStatusDetail` into `Views::Format.wrap` so both views share one cell-aware implementation rather than drifting.

**Trade-off:** enabling the wheel turns on mouse reporting program-wide (bubbletea has no per-mode toggle), which suppresses native terminal text-selection while the TUI runs. `mouse_cell_motion` (not `mouse_all_motion`) keeps the footprint minimal, and most terminals still allow Shift+drag to select.

## Test plan

Automated (run in the worktree):

- `bundle exec rake test` — full TUI unit suite green. The new behavior is covered by:
  - `help_overlay_test` — viewport windowing, scrollbar glyph + thumb position, narrow-width wrap, too-small fallback
  - `format_test` — cell-aware `Format.wrap` including wide-glyph wrapping
  - `input_decoder_test` — raw SGR wheel bytes → `HelpScroll` through the real decode path
  - `key_map_test` — scroll keys map to `HelpScroll`, close keys to `BACK`, unrelated keys no longer close
  - `update_test` — offset clamp at both ends, resize re-clamp (plus grid-mode no-op), unknown-direction parity log
  - `bubble_model_test` — wheel → `HelpScroll` gated to `:help` mode
  - `red_status_detail_test` — shared-wrap refactor leaves scroll math unchanged
- `bundle exec rubocop` — clean, no offenses.

Manual acceptance (alt-screen needs a real tty; cannot run headless):

1. 15×80 — open help (`?`): title at top, scrollbar on the right; `j`/`k`, arrows, and mouse wheel scroll to the bottom section.
2. Live resize with help open — shrink to 12 rows then grow back: pane re-fits, scroll position stays valid, no restart.
3. 60 cols — content wraps cleanly, no right-edge cutoff.
4. 8 rows (below floor) — centered "Terminal too small for help (need 10×40)", no crash; `Esc`/`?` still close.

## Review summary

Three compound-engineering review passes ran (`reviews/`) across claude-ce, pr-review-toolkit (multi-persona), and codex reviewers. All findings were auto-fixed or resolved from plan/brainstorm/project context — no open user escalations.

Notable items fixed during review:

- **Mouse wheel was non-functional end-to-end** (High): the custom `PasteAwareRunner` drained every byte through `InputDecoder`, which had no mouse branch, so wheel sequences decoded to `Esc`+garbage and *closed* help. Fixed by teaching `InputDecoder` to parse SGR mouse reports and emit a `Bubbletea::MouseMessage`, plus an integration test feeding raw wheel bytes through the decoder.
- **Cheatsheet alignment destroyed by wrap** (Medium): `Format.wrap` reflowed every line unconditionally, collapsing the `%-7s` key column and indent. Fixed to preserve already-fitting lines verbatim and indent wrapped continuations; alignment test added.
- **Lost header styling** (Medium): the rewrite dropped `Styles::HEADER`/`HINT`; restored by applying styles after width-bounding, matching sibling panes.
- **Too-small fallback dropped the `10×40` guidance** at degenerate sizes — fixed with a shorter fallback string that keeps the dimensions.
- **Footer hint truncation** — reordered so the close affordance survives at narrow widths.
- Added coverage for scrollbar thumb sizing/position, viewport advance, the resize re-clamp negative case, and the unknown-direction parity log.

The stale-base merge-hygiene finding (the branch was based on an older `main`, so the raw `origin/main..HEAD` range showed unrelated reverse-churn) was resolved by rebasing onto current `origin/main`; the feature touches only `lib/hive/tui/**`, tests, and one wiki page. One pass-1 reviewer (codex) hit a usage limit (`errors-01.md`); the same scope was covered in passes 2–3.

## Linked task

- hive task: `the-help-screen-in-tui-260519-bfeb` (id 47 — "Scrollable TUI Help Pane")
- Original report: *"the help screen in tui on smaller height terminals is cut on top and not scrollable so i can't see the whole help."*

---


ivankuznetsov:

# Code Review — Scrollable help screen in the hive TUI (pass 1)

Scope reviewed: the 4 task commits (`48607d0d..HEAD`) — TUI help overlay only.
Note: `origin/main..HEAD` is misleading because the branch is based on a stale
main (see Medium finding); the true task diff is the merge-base diff.

Verification performed:
- `bundle exec ruby -Itest` on help_overlay / key_map / update / bubble_model /
  messages / model / red_status_detail tests — all pass.
- `bundle exec rubocop` on all 9 edited source files — no offenses.
- Full `bundle exec rake test` — 4913 runs, 1 failure, 0 errors. The single
  failure (`StopHookInstallerTest#test_stop_hook_requires_stage_dir_env`) is an
  environment false-positive: the review runner exports `HIVE_TASK_STAGE_DIR`,
  which the hook subprocess inherits, so the "missing env var" path isn't taken.
  Passes cleanly with the var unset. Unrelated to this diff. Not a finding.
- Confirmed `Bubbletea::MouseMessage`, `#wheel?`, `BUTTON_WHEEL_UP/DOWN`, and
  `mouse_cell_motion` exist in bubbletea 0.1.4.

Overall: implementation faithfully matches the plan's Requirements Trace. All
in-scope requirements (height-bounded viewport, word-wrap, scrollbar, scroll
keys + mouse wheel, resize re-clamp, too-small fallback, Esc/?/q close) are
implemented and well tested. No correctness defects found.

## High

(none)

## Medium

- [ ] Branch is based on a stale main (`48607d0d`), but `origin/main` has advanced to `715a0be4` (retired Codex draft-assist, #410): rebase/merge onto current `origin/main` before merging so the PR diff and CI reflect only the help-overlay work — right now `origin/main..HEAD` shows large unrelated reverse-churn across `lib/hive/bot/*`, `bin/hv`, schemas, and bot tests.

## Nit

- [ ] Tall-terminal layout change: `render` always pads `visible` to `scrollable_rows`, so the overlay now fills the full terminal height (content at top, blank gap, footer pinned at bottom) instead of the previous content-sized box — cosmetic but a visible behavior change on large terminals (`lib/hive/tui/views/help_overlay.rb:38-42`).
- [ ] Footer hint (`"↑/↓ j/k PgUp/PgDn scroll · Esc/? close"`) is truncated with an ellipsis at minimum widths (≈40 cols, `footer_line` cuts to `box_width - 4`), so the scroll-key discoverability hint is partially lost on narrow terminals — acceptable degradation, but the close hint is what survives least (`help_overlay.rb:151-153`).
- [ ] `content_lines` (build_lines + per-line wrap) is recomputed multiple times per render (`render` → `content_lines` + `max_scroll_offset` → `content_lines` again) and again from `Update#clamp_help_scroll_offset`; harmless given help is rarely rendered and mirrors the `RedStatusDetail` pattern, but memoizing within a render would avoid the repeat (`help_overlay.rb`).

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — Scrollable help overlay (pass 1)

Personas run: code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer.

Scope reviewed: the help-overlay task commits only (`git diff 48607d0d..HEAD`). The
branch is based on a stale main, so `origin/main..HEAD` shows unrelated reverse-churn
(bot/codex/schemas) — that was excluded from review, not analyzed as task changes.

Cross-persona verdict: implementation is correct and well-tested; **no High-severity
findings**. Viewport/scrollbar geometry, clamp boundaries, resize re-clamp, mouse-wheel
gating, and the `red_status_detail` wrap refactor were independently verified as sound.
Findings below are improvements/edge-cases, plus the one merge-hygiene note (stale base).

## High

(none)

## Medium

- [ ] [tests/code] Footer dismiss hint is truncated at the minimum width: at `cols: 40`, `footer_line` (help_overlay.rb:139, `box_width - PADDING_COLS`) clips `"…Esc/? close"` so the word "close" is cut exactly when the user needs to know how to dismiss — UX bug, and no test asserts the close affordance survives at narrow/min widths.
- [ ] [silent-failure] `Format.wrap_line` silently hard-truncates (with ellipsis) any single word wider than the column instead of char-wrapping it (format.rb:69-87), dropping characters the now-scrollable overlay can no longer reveal — latent today (no current `Help::BINDINGS` token triggers it at inner width 33) but a real data-loss fallback.
- [ ] [type-design] `HelpScroll` doc comment (messages.rb:176-178) states "`amount` is a line count" but `HELP_SCROLL_KEYS` passes the `SCROLL_TO_EDGE = 1_000_000` sentinel for home/end/g/G (key_map.rb:357-361); the overload is documented only beside `RedStatusDetailScroll`, making the `HelpScroll` comment actively misleading.
- [ ] [tests] New public `Format.wrap`/`wrap_line` (format.rb:58-88) have no direct unit tests — the `max_width <= 8` truncate fallback, over-long-word path, multi-line `\n` input, and `[""]` empty guard are exercised only indirectly via overlay rendering.
- [ ] [tests] The `red_status_detail.rb:239` refactor to delegate `wrapped` to `Format.wrap` changes fit measurement from `String#length` to cell-aware `display_width` (and `[]` → `[""]` on blank lines); no test targets wide-char/empty-line wrapping to guard the semantic shift (existing rsd tests pass but don't cover it).
- [ ] [tests] Scrollbar thumb position/sizing is unverified: `test_overflow_renders_scrollbar_thumb` only asserts a `█` exists somewhere and would pass even if the thumb never moved with offset; add an assertion that the thumb occupies the last rows at max offset (help_overlay.rb:122-133).

## Nit

- [ ] [code] `mouse_cell_motion: true` (app.rb:104) enables SGR mouse tracking program-wide for the whole lifetime, disabling native click-drag text selection in every mode (Shift+drag workaround) — functionally fine (non-wheel/non-help events NOOP) but a conscious app-wide tradeoff; could be scoped to help mode if selection loss is unacceptable.
- [ ] [code] `clamp_help_scroll_offset` (update.rb:801) couples the Update layer to the View by calling `Views::HelpOverlay.max_scroll_offset`; the analogous red_status_detail path routes capacity math through a non-view module (`RedStatusDetailLayout`) — mild architectural inconsistency, helper is pure so not a defect.
- [ ] [code/type-design] Offset is clamped twice (in `clamp_help_scroll_offset` and again in `HelpOverlay.render`), so the invariant has no single owner — harmless belt-and-suspenders, consistent with view-layer defensive clamping elsewhere.
- [ ] [code] `content_lines` re-runs `build_lines` + full re-wrap on every call and `render` invokes it twice (directly and via `max_scroll_offset`), plus again from Update's clamp (help_overlay.rb:84) — negligible given help renders rarely; memoize within a render if desired.
- [ ] [silent-failure/comment] `TOO_SMALL_MESSAGE` reads "need 10×40" (rows×cols) while conventional terminal notation is cols×rows — minor user-confusion risk in the sole error message; threshold logic itself is correct.
- [ ] [comments] `Format` header "Naming note" (format.rb:12-15) lists `truncate`/`display_width` as bare-named cell-aware helpers but is not updated to mention the newly added public bare-named `wrap` — comment is now incomplete (not wrong).
- [ ] [type-design] `HelpScroll#direction` accepts any value; an invalid direction (typo `:upp`) is absorbed as a silent no-op in `apply_help_scroll`'s `else` (update.rb:794) rather than failing loudly — acceptable given the two trusted internal producers.
- [ ] [tests] Boundary tests are loose: `apply_help_scroll` unknown-direction `else` is uncovered (update.rb:794); the `too_small?` threshold (40×10 renders vs 39×10 falls back) is not pinned (existing test uses 30×8, deep in the region); `show_help` reset-on-open is tested only from the default mode, not re-entry while scrolled.

ivankuznetsov:

## High
- [ ] `lib/hive/daemon/status_consumer.rb:70` / `lib/hive/bot/status_watcher.rb:130` hard-fail on any `hive-status` schema_version mismatch again: this reverts the forward-skew tolerance from `origin/main`, so a long-running daemon or bot rejects valid newer `hive status --json` output after an upgrade.

## Medium
- [ ] `lib/hive/config.rb:478` no longer resolves `patrol.mode` before merging defaults: `patrol.mode: high` is retained but leaves `enabled=false`, `trigger=continuous`, and `poll_interval_sec=600`, so the mode scheduling dial from `origin/main` silently stops working.
- [ ] `lib/hive/patrol/reviewer.rb:149` / `lib/hive/patrol/fixer.rb:139` drop patrol usage recording: patrol review/fix agent usage is no longer written to `UsageDb`, and the token stats patrol row is removed, so patrol spend disappears from the TUI.
- [ ] `lib/hive/tui/state_source.rb:141` no longer watches each task `.lock` file in the mtime fingerprint: lock acquisition can be hidden until the 3s liveness fallback, leaving the TUI with a stale dispatchable row during that window.

## Nit

ivankuznetsov:

# Code Review — Scrollable help screen in the hive TUI (pass 2)

Scope: the 6 task commits (`b09b9c92..HEAD`) — TUI help overlay only.
`origin/main..HEAD` is misleading: the branch is based on a stale main, so that
range also shows large unrelated reverse-churn (see High finding). The true
feature diff is the merge-base diff (19 files, all `lib/hive/tui/**`, tests, and
one wiki page).

Verification performed:
- Relevant TUI unit tests (help_overlay / format / red_status_detail / key_map /
  update / bubble_model / messages / model) via a require-loop:
  **704 runs, 2449 assertions, 0 failures, 0 errors**.
- `bundle exec rubocop` on all 9 edited source files — **no offenses**.
- Confirmed `Bubbletea::MouseMessage`, `#wheel?`, `BUTTON_WHEEL_UP/DOWN`,
  `BUTTON_LEFT`, `ACTION_PRESS`, and the `mouse_cell_motion` runner option exist
  in bubbletea 0.1.4; the runner also disables mouse on suspend and re-enables on
  resume (`runner.rb:84,352,363`), so editor-suspend handling is correct.
- Verified at 40 cols the bordered box fits exactly (max rendered width = 40, no
  right-edge overflow) and the rendered help now contains **zero** ANSI escapes.

Overall: the feature faithfully implements the plan's Requirements Trace —
height-bounded viewport, word-wrap, right-edge scrollbar, scroll keys + mouse
wheel, resize re-clamp, too-small fallback, and Esc/?/q close are all present and
well tested. The pass-1 footer-truncation nit is resolved (`DISMISS_HINT` now
leads with `Esc/? close`, covered by a new min-width test). No correctness
defects in the feature code. Findings below are one merge hazard and two quality
items.

## High
- [ ] Stale base re-reverts shipped `origin/main` work: the branch is based on `b09b9c92` while `origin/main` has advanced, so the PR diff CI computes (`origin/main..HEAD`) reverts real merged changes — schema-skew tolerance (`lib/hive/daemon/status_consumer.rb`, `lib/hive/bot/status_watcher.rb`), `patrol.mode` resolution (`lib/hive/config.rb`), patrol usage recording (`lib/hive/patrol/reviewer.rb`, `lib/hive/patrol/fixer.rb`), and the `.lock` mtime fingerprint (`lib/hive/tui/state_source.rb`). Rebase/merge onto current `origin/main` before merging (help-overlay files don't overlap the churn, so it's a clean rebase); flagged in pass-1 and reconfirmed by codex-02 but still unresolved.

## Medium
- [ ] Help title and section headers lost their bold styling: the rewrite routes `build_lines` through `Format.wrap` as plain text and drops `Styles::HEADER`/`Styles::HINT` (`help_overlay.rb:46-55`), so the overlay is now a flat monochrome block — sibling views (`archive_pane`, `projects_pane`, `tasks_pane`) keep bold headers by applying `Styles::HEADER.render` *after* width-bounding the text, an approach the help view could follow to restore hierarchy on a reference screen.

## Nit
- [ ] `render_too_small`'s truncate fallback drops the actionable `10×40` dimensions at pathologically short heights: when wrapped lines exceed `height` (e.g. rows ≲ 3) it collapses to a single `Format.truncate(TOO_SMALL_MESSAGE, width)` line that can read `Terminal too…`, losing the size guidance exactly when the terminal is most cramped (`help_overlay.rb:106-109`); harmless at realistic sizes.

ivankuznetsov:

# PR Review Toolkit — Scrollable help screen in the hive TUI (pass 2)

Multi-persona review (code-reviewer, pr-test-analyzer, comment-analyzer,
silent-failure-hunter, type-design-analyzer). Personas were scoped to the actual
feature diff (`git diff b09b9c92..HEAD -- lib/hive/tui test/unit/tui
wiki/commands/tui.md`); the `origin/main..HEAD` reverse-churn is a stale-base
rebase artifact, captured separately under High.

Verification across personas: all TUI unit tests pass (1156 runs), RuboCop clean
on all 9 changed lib files, and the Bubbletea mouse API + `mouse_cell_motion`
runner option confirmed present in the installed gem. The implementation matches
the plan's Requirements Trace; the global-mouse-vs-text-selection tradeoff
flagged by code-reviewer is omitted here because the plan explicitly accepts it.

## High
- [ ] [review-scope] Branch is based on a stale main, so the CI-computed `origin/main..HEAD` diff reverts shipped main work (schema-skew tolerance, `patrol.mode`, patrol usage recording, `.lock` fingerprint): rebase onto current `origin/main` before merge — feature files don't overlap the churn, so it's clean.

## Medium
- [ ] [tests] `format_test.rb:51-56` CJK test calls `Format.wrap("一二三四五六", 8)`, which hits the `max_width <= 8` truncate early-return (`format.rb:69`) and never exercises the cell-aware word-wrap loop (`format.rb:75`): a regression reverting that comparison to codepoint `.length` would pass — add a wide-glyph case at width ≥ 10 asserting multiple rows each `display_width <= width`.
- [ ] [tests] `help_overlay.rb` `scrollbar_lines` "fits" branch (`total <= rows` → all spaces) has no direct assertion; `test_non_overflow_omits_scrollbar_thumb` only checks absence of `█`, not that the gutter is blank vs a full `│` track, so a regression rendering a full-height track on non-overflow would pass — add `assert_equal Array.new(rows, " "), scrollbar_lines(total: 5, rows: 10, offset: 0)`.
- [ ] [errors] `help_overlay.rb:115-126` `render_too_small`'s truncate fallback (`lines.length > height`) discards the actionable "10×40" guidance at degenerate sizes (rows == 1, or narrow-and-short), leaving `"Terminal too small for help (need 10×…"` — the one piece of info the user needs vanishes exactly when the terminal is smallest; a shorter fallback string would preserve it.
- [ ] [type-design] `update.rb:795` `apply_help_scroll`'s unknown-direction `else` silently no-ops with no debug log, breaking parity with the sibling `apply_red_status_detail_scroll` (`update.rb:778`) which logs `"scroll skipped: unknown direction"`: a `:sideways`/typo'd direction is dropped without a breadcrumb — one-line `Debug.log` restores parity.

## Nit
- [ ] [tests] `update_test.rb` covers the resize re-clamp in `:help` mode but not the false branch — resizing in `:grid` must leave `help_scroll_offset` untouched and must not call `Views::HelpOverlay`; the guard could be inverted without a failing test.
- [ ] [tests] `bubble_model.rb` `translate_mouse`'s wheel-but-not-up/down `else → NOOP` branch (e.g. wheel-left/right) is untested; defensive-only.
- [ ] [tests] Scrollbar `thumb_size` math (`(rows²/total).ceil` clamped to `[1, rows]`) is not directly asserted — only top/bottom tracking is; adding size assertions is optional to avoid over-coupling.
- [ ] [comments] `messages.rb:180` "Update clamps both against `max_scroll_offset`" is loosely worded — `max_scroll_offset` is only the upper bound (lower bound is 0); tighten to "clamps to [0, max_scroll_offset]".
- [ ] [comments] `messages.rb:177-178` documents `amount` as "a line count, except Home/End/g/G" but omits the second producer — the mouse wheel path also builds `HelpScroll` with `amount: HELP_WHEEL_SCROLL_LINES`; a one-word nod would help maintainers.
- [ ] [errors] `format.rb:72` `wrap_line`'s `max_width <= 8` path hard-truncates (tail content unrecoverable by scroll) — not reachable for help (`inner_content_width` ≥ 33 whenever help renders), latent only for other `Format.wrap` callers.
- [ ] [errors] `format.rb:81,85` an over-long single token is ellipsis-truncated rather than hard-split, dropping the tail — not triggered by help content (widest token 26 cells < 33 min), latent only.
- [ ] [type-design] `HelpScroll#amount` overloads the `SCROLL_TO_EDGE` (1_000_000) sentinel for "jump to edge" — accepted project pattern shared via the named constant and made safe by clamping; not worth a discriminated type for a 4-key feature.
- [ ] [type-design] `HelpScroll#direction` is an unconstrained Symbol and `amount`/`help_scroll_offset` permit negatives — representable but unreachable through the actual producers; a frozen `DIRECTIONS = %i[up down]` shared by producer/consumer would express the invariant if fixing both scroll messages together.

ivankuznetsov:

## High
- [ ] Mouse wheel input is not decoded in the custom runner path (`lib/hive/tui/app.rb:103`, `lib/hive/tui/paste_aware_runner.rb:75`): enabling mouse reporting sends SGR mouse bytes through `InputDecoder`, which turns wheel sequences into Esc/text instead of `Bubbletea::MouseMessage`, so real wheel input closes help rather than scrolling.

## Medium

## Nit

ivankuznetsov:

## High

- [ ] Mouse wheel never scrolls help and actively *closes* it through the real runner (`lib/hive/tui/app.rb:104`, `lib/hive/tui/paste_aware_runner.rb:75`, `lib/hive/tui/input_decoder.rb:224-241`): `PasteAwareRunner#process_input` drains *every* raw chunk through `InputDecoder`, which only emits `KeyMessage`/`RawTextInput`/`Flash` and has no mouse branch — so a wheel SGR sequence (`\e[<64;…M`) hits `drain_escape`, matches no `SEQUENCES` entry, is not an escape prefix, and falls to `consume(1)` → emits `KEY_ESC` (which `help_message` maps to `Messages::BACK`, closing help) leaving `[<64;…M` to spill out as literal text. `bubble_model.translate_mouse` can never fire because no `Bubbletea::MouseMessage` is ever produced — IU3's plan-required "mouse wheel" scroll is non-functional end-to-end and its unit test passes only because it hand-constructs a `MouseMessage` that the live input path never generates.
- [ ] Enabling `mouse_cell_motion: true` program-wide injects ESC + garbage into all modes (`lib/hive/tui/app.rb:104` + `input_decoder.rb` mouse gap): with mouse reporting on, every click/wheel anywhere now reaches `InputDecoder`, which decodes each report as `KEY_ESC` followed by literal `[<…M`/`m` text — in `:new_idea`/`:filter` composer modes a stray click corrupts the input buffer, a regression that did not exist when mouse reporting was off. Same root cause as above (decoder lacks mouse handling); the option should not be enabled until the decoder emits `MouseMessage`.

## Medium

- [ ] Help keybinding column alignment is destroyed at every terminal width (`lib/hive/tui/views/format.rb:64-86` via `help_overlay.rb:content_rows`): `Format.wrap_line` splits on `/\s+/` and rejoins with single spaces, so `"  b        run hive brainstorm…"` renders as `"b run hive brainstorm…"` — the `%-7s` key column and 2-space indent collapse, turning the aligned cheatsheet into ragged single-spaced text. This fires even on wide terminals where no wrapping is needed (every line is reflowed unconditionally), degrading the exact feature being built; no test asserts alignment so it is uncaught.
- [ ] Wrapped help descriptions have no hanging indent, so a continuation line is visually indistinguishable from a new keybinding entry (`help_overlay.rb:content_rows` / `format.rb:wrap_line`): the second line of a long `r`/etc. description starts at column 0 like a fresh key row, muddying the section structure. Compounds the alignment finding above; consider keeping the first line's `  key      desc` layout and indenting wrapped continuations.

## Nit

- [ ] Help overlay box now always fills the full terminal height even when the cheatsheet is short (`help_overlay.rb:render` pads `visible` to `scrollable_rows`): on a tall terminal the previous compact box becomes a near-empty full-screen box with a blank scrollbar track. Functional but a larger footprint than before; padding could be skipped when there is no overflow.
- [ ] Shared `Format.wrap` changes `RedStatusDetail` behavior for whitespace-only lines (`format.rb:81` `rows.empty? ? [""] : rows` vs. old `wrap_line` returning `[]`): a blank/whitespace log line that previously vanished is now preserved as an empty rendered row, subtly shifting line counts and scroll capacity in the red-status detail view. Likely benign/an improvement, but verify it does not perturb existing red-status scroll math.

ivankuznetsov:

## High

- [ ] [code] Mouse wheel scrolling is non-functional and actively breaks the overlay through the live runner (`lib/hive/tui/app.rb:104` + `lib/hive/tui/input_decoder.rb:233`): `mouse_cell_motion: true` is enabled, but `PasteAwareRunner#process_input` drains all bytes through `InputDecoder`, which has no mouse parsing — a wheel SGR sequence `\e[<64;…M` matches no `SEQUENCES` entry, isn't an escape prefix, so `drain_escape` emits `KEY_ESC` (→ `Messages::BACK`, closing help) then leaks `[<64;…M` as `RawTextInput`. Wheel closes help and injects garbage instead of scrolling.
- [ ] [code] `bubble_model.translate_mouse` is dead code in production (`lib/hive/tui/bubble_model.rb:270,323-334`): no `Bubbletea::MouseMessage` is ever produced by the live decoder, so the `when Bubbletea::MouseMessage` branch can never fire — IU3's plan-required mouse-wheel scroll is not actually wired end-to-end.
- [ ] [tests] The four `translate_mouse` tests give false confidence (`test/unit/tui/bubble_model_test.rb:446-518`): they hand-construct a synthetic `MouseMessage` and call `translate` directly, passing green while the real `PasteAwareRunner`→`InputDecoder` path does the opposite (decodes wheel to ESC+text) — a test that asserts shipped behavior the live pipeline never exhibits.
- [ ] [tests] No test that `mouse_cell_motion: true` produces any observable scroll through the real input pipeline (`lib/hive/tui/app.rb:104`): the only production change enabling mouse is untested at the integration boundary; nothing feeds raw wheel bytes through `InputDecoder#drain` and asserts a `HelpScroll` results.
- [ ] [comments] `HelpScroll` doc comment claims wheel is a live, equal `amount` source (`lib/hive/tui/messages.rb:177-181`): "produced by the keyboard line/page keys and by the mouse wheel" presents wheel scroll as functional, but the live runner never delivers a `MouseMessage`, so the comment misrepresents shipped behavior.
- [ ] [comments] No caveat that `translate_mouse` is exercised by tests only (`lib/hive/tui/bubble_model.rb:270,323-333`): a reviewer reading `messages.rb`/`bubble_model.rb` will believe wheel scroll is wired end-to-end; an inline note (or wiring mouse bytes through `InputDecoder`) is needed.

## Medium

- [ ] [code] `Format.wrap` refactor destroys the cheatsheet's key-column alignment and indent (`lib/hive/tui/views/format.rb:71`): `wrap_line` splits on `/\s+/`, collapsing the `format("  %-7s  %s", key, desc)` rows' 2-space indent + 7-wide key column to single spaces (`"  ctrl+r   rerun…"` → `"ctrl+r rerun…"`) — a visual regression vs. `origin/main` which rendered these lines raw, and it fires at every width (lines are reflowed unconditionally).
- [ ] [errors] `Format.wrap_line` ellipsis-truncates an over-long single word instead of hard-splitting (`lib/hive/tui/views/format.rb:72,81,85`): both `truncate(current, max_width)` branches and the `max_width <= 8` early return permanently drop characters even though the now-scrollable overlay has unlimited vertical room — latent today (widest `Help::BINDINGS` token is 26 cells < 33-cell floor) but silent data loss the instant a longer token (URL/path/flag) is added.
- [ ] [code] Per-keystroke full re-wrap on scroll (`lib/hive/tui/update.rb:798-808` → `help_overlay.rb` `max_scroll_offset`/`content_rows`): every scroll key re-wraps the entire cheatsheet via `content_lines`; the render path was deliberately written to avoid this re-wrap but the Update path was not — low impact (small modal) but worth memoizing/threading wrapped content.
- [ ] [tests] No test that line/page keyboard scroll moves the rendered viewport by the expected amount (`test/unit/tui/views/help_overlay_test.rb`): offset integer (update_test) and render-at-offset are checked separately, but nothing asserts e.g. PgDn advances the visible first line by ~10 rows — a swap of line/page amounts would pass every test.
- [ ] [tests] Scrollbar thumb position tested only at extremes (`test/unit/tui/views/help_overlay_test.rb:105`): no mid-range offset → mid-gutter thumb assertion, so an off-by-one or inverted interpolation in `((offset/max_offset)*(rows-thumb_size)).round` would survive.
- [ ] [tests] The `render_too_small` degenerate-height fallback branch is never exercised (`test/unit/tui/views/help_overlay_test.rb:178`): the `cols:30,rows:8` case shows the full `TOO_SMALL_MESSAGE`; the `lines.length > height` collapse to bare `"need 10×40"` (e.g. rows:1) — the exact branch the comment justifies — has no coverage.
- [ ] [tests] `apply_help_scroll` unknown-direction logging is unverified (`test/unit/tui/update_test.rb:1707`): the test only asserts the offset is unchanged, so the parity log call (commit `537cbb14`) could be deleted with no failure.
- [ ] [tests] Brittle implementation-coupled assertion (`test/unit/tui/update_test.rb:95`): `test_window_sized_in_grid_mode_leaves_help_scroll_offset_untouched` monkey-patches `HelpOverlay.max_scroll_offset` to assert it is NOT called — couples the test to the short-circuit detail; the behavioral assertions (mode/offset unchanged) already suffice.
- [ ] [types] `HelpScroll#direction` is an unconstrained Symbol (`lib/hive/tui/messages.rb:183`): nothing restricts it to `:up`/`:down`, so `update.rb:795` carries a live "unknown direction" `else` branch and tests exercise `:sideways`; a constructor-validated direction would make the branch dead code rather than a real path. (Mirrors pre-existing `RedStatusDetailScroll`.)
- [ ] [types] `HelpScroll#amount` overloads an Integer line count with the `SCROLL_TO_EDGE = 1_000_000` sentinel (`lib/hive/tui/messages.rb:183`, `key_map.rb:333`): two semantically distinct commands ("scroll N" vs "jump to edge") flattened into one field, working only because the consumer clamps a magic large number; an `:edge` variant/`amount: :edge` would express intent. (Convention-wide with `RedStatusDetailScroll`.)
- [ ] [types] `help_scroll_offset` invariant is enforced outside the type by two independent clamp sites (`lib/hive/tui/model.rb:49`): a bare Integer lets any caller set negative/out-of-range via `model.with`; Update clamps and the View re-clamps at render — upheld by duplication (drift risk) rather than by the type.
- [ ] [comments] Stale "if mouse ever gets enabled" comment (`lib/hive/tui/update.rb:138`): mouse IS now enabled (`app.rb:104`), so the conditional phrasing is outdated and reinforces the false impression that mouse is off.

## Nit

- [ ] [code] Undocumented behavior shift folded into the `red_status_detail` refactor (`lib/hive/tui/views/red_status_detail.rb:239-240`): the old `wrap_line` used `String#length` for the fit test; shared `Format.wrap_line` uses cell-aware `display_width` — an improvement for wide glyphs, but a silent behavior change to `RedStatusDetail` wrapping.
- [ ] [errors] `Format.wrap_line`'s `max_width <= 8` guard silently single-line-truncates at narrow widths (`lib/hive/tui/views/format.rb:72`): unreachable from `help_overlay` (too_small guard forces width ≥33) but `wrap` is advertised as general-purpose, so a future caller passing width 5–8 gets silent truncation with no signal.
- [ ] [comments] `wrap` doesn't fit the "many external callers" naming rationale (`lib/hive/tui/views/format.rb:14-15`): unlike `truncate` (23 sites) / `display_width` (7), `wrap` has exactly 2 callers, both added in this PR — the rationale should say "for consistency," not an existing caller base.
- [ ] [types] `content_rows` returns an unnamed positional `[String, Boolean]` tuple (`lib/hive/tui/views/help_overlay.rb`): readers must destructure `|line, header|` / `.first` and the padding literal `[ "", false ]` hard-codes the shape; a `Segment = Data.define(:text, :header?)` would be self-describing. Also a boolean-blindness case with no room to grow line styles.
- [ ] [types] Max-offset bound is expressed in two places (`lib/hive/tui/views/help_overlay.rb` `max_scroll_offset` vs inline `render`): they agree today but a change to one risks scroll/clamp drift against `clamp_help_scroll_offset` (`update.rb:803`).
- [ ] [types] `help_scroll_offset` lives unconditionally on the flat `Model` unlike sibling mode-state (`lib/hive/tui/model.rb:49`): `red_status_detail_state`/`token_stats_state` bundle mode-specific state into nested Data ("nil — :X mode only"); a `HelpState` sub-Data would localize the invariant and match convention (heavier than one Integer warrants, hence Nit).
- [ ] [tests] `test_help_scroll_down_clamps_to_bottom` drives the clamp with the raw `SCROLL_TO_EDGE` constant rather than the End/G key route (`test/unit/tui/update_test.rb:1665`): no single test proves End/G → clamped-to-bottom end to end.
- [ ] [tests] Footer scroll-hint truncation is asserted nowhere under truncation (`test/unit/tui/views/help_overlay_test.rb:19`): at default 80 cols the footer isn't truncated, so the `box_width - PADDING_COLS` drop of the scroll-key portion is only spot-checked by the min-width close-affordance test.

PR #428, opened 2026-06-09
Hive patrol: Bundled git grep -O pager option bypasses dry-run stub

ivankuznetsov:
## Summary

Closes an RCE seam in the babysitter dry-run git stub (`bin/hive-babysitter-stub-git`). The stub allowlists read-only `git grep` invocations, but `--open-files-in-pager` / `-O` runs an arbitrary pager command. The original guard only blocked the full long spelling and the leading `-O` short form, so several equivalent invocations slipped through and executed inside a dry-run worktree:

- **Clustered short options** — git parses `-nOtouch …` as `-n` + `-O<cmd>`, but the guard only checked `arg.start_with?("-O")`, so any cluster that buried `O` behind another short flag bypassed it.
- **Abbreviated long options** — git resolves any unambiguous long-option prefix, so `--open=cat`, `--open-files=cat`, down to the shortest unique `--op`, all reach `--open-files-in-pager`. The guard matched only the full name.

The fix scopes the pager block to `git grep` (where `-O` is the exec seam) and:

- Blocks any long-option prefix of `--open-files-in-pager` of length ≥ 4 (`--op` is the shortest spelling not shared with `--only-matching` / `--or`), with or without a glued `=<cmd>`.
- Walks short-option clusters character by character, returning unsafe on an `O` but stopping at the first value-taking option (`-e`, `-f`, `-m`, `-A`, `-B`, `-C`) so the rest of the cluster is treated as that option's operand.

That cluster-walk also fixes a false positive: read-only patterns like `git grep -eTODO` were being skipped as "unsafe" merely because the pattern value contained an uppercase `O`. They now reach real git.

`diff` / `log` / `show` are unaffected — there `-O<orderfile>` is the read-only `--output-ordering`, so the pager block stays scoped to grep.

## Test plan

- `bundle exec rake test` — passing. New regression cases in `test/unit/babysitter/dry_run_env_test.rb` cover:
  - rejected: `-nOtouch …`, `--open=…`, `--open-files=…`, `--op=…` (in addition to the existing `-Otouch …` and separate `-O <cmd>`).
  - still allowed (must reach real git, not be skipped): `git grep -eTODO -- .`, `git grep -fNEEDLEFILE.txt`, alongside the existing `-o` / `--only-matching` read-only cases.
- `bundle exec rubocop` — clean.

## Review summary

Two automated Codex CE review passes. Pass 01 raised two findings, both auto-fixed in this branch:

- **High** — abbreviated `--open` / `--open-files` long-option prefixes still bypassed the guard. Fixed by blocking any prefix (length ≥ 4) of `--open-files-in-pager`.
- **Nit** — the token-level `include?("O")` cluster check flagged read-only attached patterns like `-eTODO` as unsafe. Fixed by parsing short-option clusters so value-taking options consume their operand.

Pass 02 found no issues. No findings were escalated to the user.

## Linked task

Hive patrol finding `command-bin-hive-babysitter-stub-git-2` (fingerprint `e828a74030c28a572067587a52cd5c92f53b81d5024239debf7662fdea0fbb81`).


ivankuznetsov:

# Codex CE Code Review

No plan was found in the task folder; this review is grounded in `git diff origin/main..HEAD`.

## High
- [ ] Abbreviated `git grep` pager long options still bypass the dry-run stub: Git accepts forms like `--open=cat` and `--open-files=cat` as `--open-files-in-pager`, but the guard only blocks the full long spelling and short `-O` forms before execing real git.

## Medium

_No findings._

## Nit
- [ ] `git grep -e<pattern>` values containing uppercase `O` are now skipped as unsafe clusters: the new token-level `include?("O")` check treats read-only attached pattern operands such as `-eTODO` as pager flags, so dry-run searches can silently return skipped success instead of reaching real git.

ivankuznetsov:

Outcome: failure

Agent tail:

```

```

ivankuznetsov:

Outcome: failure

Agent tail:

```

```

ivankuznetsov:

Outcome: failure

Agent tail:

```

```

Issue #35, opened 2026-05-06
Migrate the 8 pre-existing --json emit sites onto Hive::Schemas::EnvelopeEmitter


ivankuznetsov:
**Source:** ce-code-review on PR #28 — maintainability finding #16 (P2). The minimal version landed in 29cb282 (Forget + Prune migrated). The full sweep across the 8 pre-existing sites is deferred here.

## Problem

Until 29cb282, the JSON-error-envelope rescue + emit scaffold was duplicated across 5 commands (Approve, Markers, Metrics, Forget, Prune). 29cb282 introduced \`Hive::Schemas::EnvelopeEmitter\` and migrated Forget + Prune onto it, leaving 8 pre-existing emit sites still rolling their own:

- \`Hive::Commands::Approve\`
- \`Hive::Commands::Markers\`
- \`Hive::Commands::Metrics\`
- \`Hive::Commands::FindingToggle\`
- \`Hive::Commands::Status\`
- \`Hive::Commands::Run\`
- \`Hive::Commands::Findings\`
- \`Hive::Commands::StageAction\`

## Why migrate

Each existing command duplicates: \`@stdout_written = false\`; \`rescue Hive::Error\` / \`rescue StandardError\`; \`emit_error_envelope(error)\` that calls \`Hive::Schemas::ErrorEnvelope.build\`; \`Errno::EPIPE / JSON::GeneratorError\` swallow on the JSON write. ~25 LOC per command times 8 commands = ~200 lines of essentially-identical scaffolding that drifts in lockstep when something changes.

## Proposed change

For each of the 8 commands:

1. \`include Hive::Schemas::EnvelopeEmitter\`
2. Replace the per-command \`call\` rescue scaffold with \`call_with_envelope { do_call }\`
3. Define \`envelope_schema\` returning the schema name string
4. Define \`envelope_error_kind(error)\` returning the closed-enum error_kind for the exception (this is intrinsically command-specific so it stays in the consumer)
5. Delete the now-duplicated \`emit_error_envelope\` body
6. Run the full test suite — the contract is identical, so no test changes should be needed beyond fixing any test that asserted on private method existence

Risk: each command has subtle differences in *when* it sets \`@stdout_written = true\` (some emit success on the same JSON write that needs the guard; some don't). A migration must preserve those exact moments. Suggested order: simplest first (Markers, Metrics) → reference (Approve) → most stateful last (Run, StageAction).

## Out of scope

[...]

PR #453, opened 2026-06-11
fix(status): stop transient duplicates and ENOENT during stage moves

ivankuznetsov:
## Summary

The TUI flashed a duplicate red row whenever a task moved between stages, and `hive status` could crash with exit 70 if a task folder vanished mid-scan. Both artifacts came from the same race: a stage move is a single atomic `File.rename`, but `Status#collect_rows` scans the nine stage directories sequentially. A rename landing mid-scan was either captured twice (the stale old-stage row then classified red because its files were gone) or disappeared between the glob and the per-entry reads, raising `Errno::ENOENT`.

The fix is read-side only, inside `collect_rows`. Because `hive status` text, `--json`, and the TUI poller all consume the same row collection, one change point covers all three. Write-side move mechanics (`hive approve` / `hive archive` locks and markers) are untouched.

**How it works**

- **Vanished-folder skip** — row construction rescues `Errno::ENOENT` and skips the entry only when the task folder itself is gone. An ENOENT while the folder still exists re-raises, so missing files inside an existing folder keep their Error classification — corrupted tasks stay visible.
- **Collision-scoped dedup** — after the scan, only slugs collected from more than one stage are re-stat'd, and rows whose folder no longer exists are dropped. When both folders still exist, both rows survive with the existing `stage_collision` behavior (`--from <stage>` command hints), so genuine duplicates still surface. Dedup runs before `annotate_actions`, so collision hints now fire only for real collisions.

Two residual windows are consciously accepted: a microseconds-wide red frame is still possible between the dedup re-stat and `TaskAction`'s lazy file reads, and a backward move (later-scanned dir → earlier-scanned dir) can hide the task for one 1 Hz poll. Neither produces a duplicate, and both self-heal on the next poll.

## Test plan

The race is pinned deterministically — no sleeps, no threads. A `stage_task_entries` glob seam lets tests fire the rename at exact scan points. New cases in `test/unit/commands/status_test.rb` cover:

- mid-read vanish (folder renamed away after the old-stage glob → entry skipped, no exception, no row);
- post-read duplicate (rename lands after the old-stage row is built → dedup drops the stale row);
- the same scenario asserted at the `json_payload` boundary;
- a vanish during a non-finalize stage and a 3-member duplicate group (count- and stage-agnostic);
- the re-raise contract (an ENOENT with a surviving folder still raises);
- genuine two-stage collision preservation (both rows survive with disambiguated commands);
- a corrupted finalize row (missing `pr.md`) still classifying Error.

Full suite green: 5091 runs, 0 failures, 0 errors. RuboCop clean on both changed Ruby files.

## Review summary

Reviewed across three Compound Engineering passes (claude-ce, pr-review-toolkit multi-persona, and codex), each running the status unit suite, the integration suite, and RuboCop. Consensus: **ready to merge** — no High- or Medium-severity defects in the fix itself. The dedup runs before `annotate_actions`, the `rescue Errno::ENOENT; raise if File.directory?(entry)` discriminator is a narrow skip rather than a swallow, and the JSON schema is unchanged.

Findings that landed as fixes were documentation-precision and test-hardening: narrowing the `InvalidTaskPath` rescue back to `Task.new`, restating the re-raise invariant comment, adding a re-raise regression test and a double-race test, non-finalize / 3-member dup-group coverage, a `renamed?` drift guard on the test double, and a symmetric collision-command assertion. The codex "High" flagging reverted upstream `v0.2.4` files was confirmed a reverse-diff artifact (origin/main advanced past the merge-base after branching; the three-dot diff touches only `status.rb`, its test, and four wiki/doc files) — nothing is reverted, since a 3-way merge preserves upstream.

## Linked task

Hive task `815` — *Suppress Duplicate Archive Rows* (`tui-briefly-shows-a-duplicate-260611-e9f8`).

Originally caught on camera in the Product Hunt demo recording: archiving a finished task rendered the stale `8-finalize` row as "Error 2m" beneath the fresh `9-done` Archived row for ~1.5 s before the next refresh cleared it. The duplicate now never renders.

---


ivankuznetsov:

## High

## Medium
- [ ] `lib/hive/commands/status.rb:468` ENOENT re-raise branch lacks coverage: `collect_rows` intentionally re-raises `Errno::ENOENT` when the task folder still exists, but the tests only cover vanished-folder skip and stable corrupted-task classification, so the folder-level-only boundary can regress silently.

## Nit
- [ ] `wiki/commands/status.md:77` Clarify still-existing-folder ENOENT wording: the docs currently conflate re-raised `ENOENT` for a still-existing folder with stable missing-file Error classification, while the code only classifies stable missing files and re-raises the race-shaped ENOENT.

ivankuznetsov:

# ce-code-review — pass 1

Scope: `git diff 0212469b..HEAD` (4 commits: U1 vanished-folder skip, U2 transient dedup, U3 race tests + seam, U4 wiki).
Reviewers: correctness, adversarial, testing, maintainability, project-standards, agent-native, learnings, reliability (8 agents).
Verification: `ruby -Itest test/unit/commands/status_test.rb` — 32 runs, 157 assertions, 0 failures; `rubocop` clean on both changed Ruby files.
Plan requirements trace: all in-scope units (1–4) implemented; no plan-required-but-missing gaps found. No code-defect findings in the production logic itself — correctness and adversarial reviewers both verified the broadened rescue swallows nothing new, the dedup is correct for 2-row and N-row groups, lock/marker readers already tolerate missing files, and the TUI mtime-fingerprint gate covers the new path.

## High

None.

## Medium

- [ ] `wiki/commands/status.md:77` misstates the ENOENT re-raise outcome: the sentence "an ENOENT for a still-existing entry is re-raised as a real read failure, so missing files inside an existing task folder still flow into the normal Hive::TaskAction error classification" implies the re-raise feeds Error classification, but a re-raised ENOENT propagates out of `collect_rows` uncaught and aborts the whole project scan (CLI `Hive::InternalError` exit 70 via `status.rb:62-65`; TUI held frame via `state_source.rb:113`) — missing files classify Error only because the guarded readers (`Markers.current` → `:none`) never raise; split the sentence into the two separate mechanisms. (maintainability, conf 100; verified against `lib/hive/commands/status.rb:467-470`, `:125`, `:345`)
- [ ] The new re-raise arm `raise if File.directory?(entry)` (`lib/hive/commands/status.rb:468`) has zero test coverage: if it regressed to a plain `next`, ENOENT raised while the task folder still exists would be silently swallowed and no existing test fails — add a regression that injects ENOENT through the `stage_task_entries` seam (e.g. an entry whose state-file read raises while the folder is kept alive) and asserts propagation. (testing, conf 100; independently listed as a gap by correctness, adversarial, maintainability, and learnings)
- [ ] `VanishAfterDirectoryCheckPath` (`test/unit/commands/status_test.rb:812`) pins the ENOENT-rescue window only through incidental Ruby path-coercion ordering: the rename fires on the 2nd `to_path` call (currently `File.basename` at `status.rb:425`) and ENOENT surfaces at `File.mtime` (`status.rb:429`); any future coercion of `entry` added before `basename` makes the rename fire at/before the top `File.directory?` guard, so the test stays green while no longer exercising the rescue at all — make the mechanism observable (record/assert the call index at which the rename fired, or assert `to_path` was called ≥4 times so `File.mtime` was reached) or at minimum document the call-count invariant on the helper. (adversarial conf 100 — empirically probed call order — corroborated by testing and maintainability; 3 independent reviewers)

## Nit

- [ ] Tests 2 and 3 (`test/unit/commands/status_test.rb:225` and `:251`) duplicate the identical `rename_before_stage: "9-done"` fixture and assertions, differing only in the surface exercised (annotate_actions vs json_payload) — extract a shared arrange/assert helper so a future fixture change cannot silently diverge between them. (maintainability, conf 75)
- [ ] Pre-existing, now entrenched behind the new seam: `stage_task_entries` uses `Dir[File.join(stage_dir, "*")]`, so a project path containing glob metacharacters (`[`, `{`, `?`, `*`) silently yields zero tasks across status/JSON/TUI while `detect_legacy_stage_dirs` and `Migrate` (both `Dir.children`-based) still see the folders — switching to `Dir.children(stage_dir).map { |c| File.join(stage_dir, c) }` keeps the seam signature and removes the inconsistency. (adversarial, conf 75, pre-existing)
- [ ] Pre-existing micro-window adjacent to the new rescue, behavior unchanged by this diff: `Hive::Markers.current` does an unguarded `File.exist?` → `File.read`; a state file removed between the two calls (only reachable via manual `rm` — atomic tempfile+rename writers cannot trigger it) raises ENOENT with the folder still alive → re-raise → whole-scan abort, exactly as pre-fix — consider rescuing ENOENT inside `Markers.current` to return `:none`, matching its existing missing-file semantics. (reliability, conf 75, pre-existing)
- [ ] Advisory, consumer-side of the plan-accepted residual window (no change required in this diff): the bot's zero-debounce first-sight send (`lib/hive/bot/notification_dispatcher.rb:181`) turns the accepted one-frame phantom Error row into a durable Telegram alert plus a later spurious "recovered" ping after the 60 s absence grace; if that proves noisy in practice, either re-stat all rows whose folder vanished by annotate time (not only duplicated slugs) or add a two-poll confirmation debounce for recovery-class fingerprints. (adversarial, conf 75, advisory)

---

## Coverage notes

- Suppressed: 1 finding below the confidence gate (`drop_transient_stage_moves` naming breadth, maintainability conf 50, single reviewer).
- Learnings check (`docs/solutions/`): no violated patterns. Relevant Known Patterns confirmed honored: legacy-stage-signal separation (dedup operates only on canonical-scan rows; `detect_legacy_stage_dirs` path untouched), required-and-nullable `diagnostic` contract (dedup runs before `annotate_actions`, only inspects `:slug`/`:folder`), StateSource background-thread polling (untouched per plan A7).
- Agent-native parity: clean — all three surfaces (text/JSON/TUI) share `collect_rows`; rows are only removed in race frames, never reshaped; genuine-collision `--from`/`--stage` disambiguation preserved and now fires only for real collisions.
- Project standards: wiki protocol followed (log.d fragment added, compiled log.md untouched, no new pages so no index.md change needed, backlinks valid); rubocop clean.
- Untested-but-accepted semantics noted by multiple reviewers (informational, plan-consistent): 3+ same-slug copies through the dedup, the double-race all-copies-gone drop, and the dedup fast path with zero duplicates have no direct assertions.

Verdict: **Ready with fixes** — no blocking code defects; the three Medium items are a one-sentence docs correction and two test-hardening changes that protect the regression guarantee this task exists to provide.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — pass 1

comment-analyzer, type-design-analyzer.

Scope reviewed: the branch's own changes vs merge-base `0212469b`
(`git diff origin/main..HEAD` is polluted by origin/main's divergence shown in
reverse; the real work is the 4 commits touching `lib/hive/commands/status.rb`,
`test/unit/commands/status_test.rb`, and four wiki/doc files). All personas ran
the suite (32 runs, 157 assertions, 0 failures) and rubocop (clean).

Consensus: no High-severity defects. code-reviewer found nothing at/above its
confidence threshold; comment-analyzer confirmed every wiki claim and all cited
commit SHAs (`bd0b965a`, `4099bbc4`, `520660e9`) are accurate; pr-test-analyzer
empirically confirmed tests 1–3 genuinely fail against pre-fix code (real
regressions, not tautologies). The deliberately-accepted residual annotate-time
TOCTOU micro-window (plan Risks) is not reported — it is an explicit plan scope
boundary.

## High
- [ ] None.

## Medium
- [ ] [silent-failure] `rescue Errno::ENOENT` at `status.rb:467-470` discriminates on folder-presence, not error-origin: when `entry`'s folder is gone, *any* ENOENT from the widened body (`Task.new`, `Markers.current`, lock reads, `worktree_path`, etc.) is swallowed and the row silently dropped — a non-race ENOENT from a genuinely wrong/buggy path is indistinguishable from the intended transient skip. Low probability (folder must also be absent), but the discriminator can't tell "folder moved out from under us" from "we computed a wrong path and the folder is also gone."

## Nit
- [ ] [silent-failure] Widened `rescue Hive::InvalidTaskPath` (`status.rb:465-466`): pre-fix only `Hive::Task.new(entry)` was inside this rescue; now the whole row body is, so an `InvalidTaskPath` (a `Hive::Error < StandardError`) raised by any body call would also be silently `next`-ed. Currently inert (only `Task.new` raises it on this path; `worktree_path` self-rescues), but a latent silent-skip surface and a deviation from the plan's "keep the existing guard as-is."
- [ ] [silent-failure] The ENOENT/InvalidTaskPath skip branches (`status.rb:466,470`) drop a task row with no `warn`/log, deviating from this file's own swallow-but-warn convention (`task_lock_holder`/`live_task_lock_holder` emit `warn "hive: status: …"`). For the intended race the silence is correct, but combined with the Medium above, a genuinely-missing task would disappear from `hive status` with nothing on stderr to explain it; a single `warn` on the ENOENT-skip branch would make a real drop debuggable without changing behavior.
- [ ] [test-analyzer][type-design] `VanishAfterDirectoryCheckPath#to_path` (`status_test.rb:804-831`) fires the rename on `@calls > 1`, a contract coupled to the exact production coercion order in `collect_rows` (call 1 = `File.directory?` guard, call 2 = `File.basename`, call 3 = `Task.new`); reordering or adding a `File.*` coercion at `status.rb:423-427` would silently make the test exercise a different race than its name claims, without failing loudly. Deterministic today; add a one-line comment naming the assumed call sequence so the implicit count survives refactors.
- [ ] [type-design] `VanishAfterDirectoryCheckPath#to_s` (`status_test.rb:818-820`) is inert on the tested path (only `to_path` triggers the vanish; omitting `to_str` is correct). Harmless, but the asymmetry could mislead a future reader into thinking both methods participate in the rename dance — a one-line note would clarify.
- [ ] [type-design] Duplicated `rename_once` guard (`return if @renamed; mkdir_p; File.rename; @renamed = true`) across `StatusRaceCommand` (`status_test.rb:795-801`) and `VanishAfterDirectoryCheckPath` (`status_test.rb:824-830`); minor, arguably fine to leave for two throwaway test classes — flagged for completeness only.
- [ ] [test-analyzer] The `refute_equal …ERROR` assertions in tests 1–3 (`status_test.rb:221,247,271`) are near-vacuous secondary checks — the load-bearing assertions are `size == 1` and `stage == "9-done"` (both proven to fail pre-fix); a healthy renamed 9-done folder would not classify ERROR regardless. Harmless mild redundancy.
- [ ] [test-analyzer] Optional coverage gap: the plan-acknowledged "double race drops the slug for one cycle" behavior (Risks) has no test; not one of the five required Unit 3 scenarios, so not a defect, but a test asserting "no Error row, slug may be absent" would harden the accepted semantics against regression.
- [ ] [comment-analyzer] `wiki/commands/status.md:77` accurately describes the ENOENT re-raise branch ("an ENOENT for a still-existing entry is re-raised … missing files inside an existing folder still flow into normal error classification"), but only the *skip* side (folder gone) is test-pinned; the re-raise side has no direct regression. Doc is code-accurate, but the implied coverage symmetry is slightly stronger than what the suite pins.
- [ ] [test-analyzer] Pre-existing (not introduced by this PR) dead ternary `stage == "9-done" ? "task.md" : "task.md"` in `create_status_task` (`status_test.rb:749`) — both arms identical; cosmetic, noted only because it sits beside the new helpers.

ivankuznetsov:

## High
- [ ] Branch diff reverts unrelated upstream work outside the status race scope: `origin/main` is not an ancestor of this branch, and `git diff origin/main..HEAD` includes removals in files such as `lib/hive/config.rb`, `schemas/hive-init.v1.json`, `Gemfile.lock`, and multiple wiki pages, violating the plan's read-side-only status change boundary.

## Medium

## Nit

ivankuznetsov:

# Code Review — tui-briefly-shows-a-duplicate (pass 2)

Scope reviewed: the branch's own changes via the three-dot diff
`git diff origin/main...HEAD` — `lib/hive/commands/status.rb`,
`test/unit/commands/status_test.rb`, and four wiki/doc files. The apparent
reverts of `lib/hive/config.rb`, `schemas/hive-init.v1.json`, `Gemfile.lock`,
and several wiki pages that show up in the **two-dot** `origin/main..HEAD`
diff are a reverse-diff artifact: `origin/main` advanced to v0.2.4 *after* this
branch's merge-base (`0212469b`), so the two-dot diff renders those upstream
commits in reverse. This branch does not touch those files (confirmed by the
three-dot diff and `git log origin/main..HEAD`). The codex-02 "High" on this
point is therefore a false positive, consistent with pass 1's resolution.

Verification run during review:
- `ruby -Itest test/unit/commands/status_test.rb` → 32 runs, 0 failures.
- `ruby -Itest test/integration/status_test.rb` → 8 runs, 0 failures.
- `rubocop lib/hive/commands/status.rb test/unit/commands/status_test.rb` → clean.

Four independent persona reviewers (correctness, reliability, testing,
adversarial) each traced the fix. Consensus: the change faithfully implements
the plan (U1 ENOENT vanished-folder skip, U2 collision-scoped
`drop_transient_stage_moves` before `annotate_actions`, U3 deterministic seam +
regressions, U4 docs); all A1–A7 requirements are met and no plan-required item
is missing. The adversarial pass confirmed the four "break it" scenarios
(surviving stale-red row, backward move, chained moves, genuine collision) all
fall *inside* the plan's consciously-accepted boundaries for ordinary atomic
stage moves. The findings below are residual edge cases, not defects in the
intended behavior.

## High
- [ ] None.

## Medium
- [ ] Re-raise guard has two narrow TOCTOU false-fire windows that can crash `hive status` (exit-70 `InternalError`) (`status.rb:469-470`): `rescue Errno::ENOENT; raise if File.directory?(entry)` re-raises when an ENOENT surfaced in the row body but the folder path is a directory at re-check. Two ways that assertion false-fires: (a) the vanished old-stage `<stage>/<slug>` path is recreated (move-back / new task at the same slug) in the one-statement window between the `File.mtime(entry)` ENOENT at `:433` and the `:470` re-check; (b) `task.state_file` is deleted inside a *surviving* folder so `File.mtime(task.state_file)` at `:434` raises while `File.directory?(entry)` stays true. Both are a strict *improvement* over pre-fix (which crashed on every mid-scan vanish) and are unreachable under today's atomic-rename writers (`Markers.write_atomic`), but neither is test-pinned, so a future non-atomic state-file writer would silently reintroduce a process-killing crash on a surface polled ~20×/min. Recommend documenting the invariant inline ("re-raise assumes any ENOENT with a surviving folder is genuine in-folder corruption; relies on all state-file mutations being atomic renames") and adding one regression test pinning the contract; a code change (always `next` on ENOENT, or scope the rescue to the `File.mtime` sites) is optional and would need to preserve the corrupted-but-present → Error path.

## Nit
- [ ] Rescue catches `Errno::ENOENT` only; sibling vanish errnos re-propagate as exit-70 (`status.rb:469`): a path component flipping to a non-directory (`Errno::ENOTDIR`) or an NFS stale handle (`Errno::ESTALE`) on the same race would escape the rescue. This is correct and in-scope for local same-device `.hive-state` renames (which always yield ENOENT), but a one-line comment documenting the deliberate narrowing would stop a future reader from assuming full vanish coverage.
- [ ] Genuine-collision test pins only one of the two surviving rows' disambiguated command (`status_test.rb:292`): `test_collect_rows_preserves_genuine_stage_collision_…` asserts `--stage 4-execute` on the execute row but never checks the `6-review` row's `suggested_command`. A regression that broke disambiguation for the higher-stage row would pass unnoticed; add the symmetric assertion (`hive artifacts #{slug} --from 6-review`).
- [ ] `drop_transient_stage_moves` dedups on folder existence, not task identity (`status.rb:483-490`): the predicate keys on `row[:slug]` + `File.directory?(row[:folder])` with no inode/`meta.yml` id check. A `<stage>/<slug>` path torn down and recreated by a *different* task within one scan could momentarily keep a row whose `id`/`display_name` no longer match the folder. Astronomically narrow and self-heals next poll; worth a one-line note in `wiki/gaps.md` or a code comment so the path-identity assumption is explicit.
- [ ] No test pins the dedup early-exit no-op for the non-duplicated path (`status.rb:485`): an explicit assertion that `drop_transient_stage_moves` returns rows untouched when no slug is duplicated would lock the `return rows if duplicated_slugs.empty?` branch against a future refactor that accidentally drops unique rows. Low value (incidentally exercised by existing single-slug tests), included for completeness.

ivankuznetsov:

# PR Review Toolkit — tui-briefly-shows-a-duplicate (pass 2)

Union of 5 persona agents (code-reviewer, pr-test-analyzer, silent-failure-hunter,
comment-analyzer, type-design-analyzer) over the branch's own merge-base diff
`git diff origin/main...HEAD` (the two-dot `origin/main..HEAD` is polluted —
origin/main advanced to v0.2.4 after the merge-base, rendering unrelated upstream
commits in reverse; this branch touches only `status.rb`,
`test/unit/commands/status_test.rb`, and 4 wiki/doc files).

Consensus across all five personas: the production fix is **correct and the error
handling is sound** — no Critical/Important issues on `collect_rows` logic. The
`rescue Errno::ENOENT; raise if File.directory?(entry)` discriminator is a
legitimate narrow skip (not a swallow), the inner `InvalidTaskPath` rescue is
correctly scoped to `Task.new`, `drop_transient_stage_moves` only drops
confirmed-gone duplicate-slug rows, the seam extraction is behavior-neutral, and
the wiki/gaps/tui docs accurately describe the behavior. `rubocop` clean; the
focused unit (32) + integration (8) suites pass. Findings below are
comment-accuracy, test-isolation, and test-double-robustness refinements.

## High
- [ ] None.

## Medium
- [ ] [comments] `VanishAfterDirectoryCheckPath` comment mis-attributes coercion call 2 to "the row build" (`test/unit/commands/status_test.rb:808`): empirically call 2 is `File.basename(entry)` (slug extraction, `status.rb:425`), and the `rows << { … }` hash literal (`status.rb:449`) never coerces `entry` at all — the ENOENT actually fires at `File.mtime(entry)` (`status.rb:433`). A maintainer debugging a future threshold change would look at the wrong line; correct the wording to name `File.basename` as call 2 and `File.mtime` as the raising read. Safe doc-only edit.
- [ ] [tests] The re-raise branch `raise if File.directory?(entry)` (`status.rb:470`) is untested (`test/unit/commands/status_test.rb`): no test forces an `Errno::ENOENT` while the folder still exists, so a refactor that turns the `raise if` into an unconditional `next` would silently start swallowing real filesystem errors with zero test failure. Add a test that raises ENOENT on the `File.mtime` coercion while leaving the directory present and asserts `collect_rows` re-raises. (Corroborates the re-raise concern raised in claude-ce-code-review-02.)

## Nit
- [ ] [tests/code/errors/type-design] `VanishAfterDirectoryCheckPath` couples to the production `to_path` coercion count via `@calls > 1` (`status_test.rb:813-840`): if a `collect_rows` refactor changes how many times the path is coerced before `File.mtime`, the double silently stops firing and `test_collect_rows_skips_old_stage_entry_that_vanishes_mid_read` passes vacuously. Four personas flagged this; cheapest fix is to expose a `renamed?` reader and assert `vanish_double.renamed?` after `collect_rows` so drift fails loudly instead of green-but-testing-nothing. (The fragility is already documented in the comment block — this makes it enforced rather than documentation-only.)
- [ ] [tests] The post-read duplicate tests (#2/#3, `status_test.rb:225,251`) inject the rename through the `stage_task_entries` seam the fix itself added: they remain valid *forward* regression guards (reverting `drop_transient_stage_moves` to a no-op makes them fail at `assert_equal 1`), but against pre-fix code they fail at the `"9-done"` stage assertion rather than the duplicate-count assertion, so they couple "fix present" with "seam present." Optional: add one test injecting the rename via a `Dir.[]`/`Dir.glob` seam present in both code versions, asserting exactly one row, to isolate `drop_transient_stage_moves` cleanly.
- [ ] [tests] Double-race (both old and renamed folders gone before `drop_transient_stage_moves`) is uncovered (`status.rb:483-490`): `drop_transient_stage_moves` rejects both rows → zero rows for the slug. One cheap test would lock the "vanish-to-zero, no error" behavior the plan's Risks section accepts.
- [ ] [errors] `drop_transient_stage_moves` removes rows with no log/trace (`status.rb:483-490`): silent removal of a confirmed-absent duplicate is acceptable, but a one-line `warn` on actual drop (matching the existing `task_lock_holder`/`live_task_lock_holder` warn pattern at `status.rb:616,643`) would make the rare event traceable if it ever fires for a non-race reason. Optional observability nit.
- [ ] [type-design] `StatusRaceCommand` constructor does not enforce mutual exclusivity of `rename_before_stage:` / `vanish_during_read_stage:` (`status_test.rb:772-802`): the two modes are intended to be exclusive (each test passes one), but passing both would half-activate the double with no error. A one-line `ArgumentError` guard ("pass exactly one of …") turns a silent misconfiguration into a loud failure. Optional (throwaway test infra).
- [ ] [comments] Optional clarity (`status_test.rb:811`): the accurate "`to_s` is inert / `to_str` omitted" note could name the mechanism — `File.*` coerces via the `to_path` protocol, not `to_str` — to make the omission rationale self-evident rather than implicit.

ivankuznetsov:

## High
- [ ] Release metadata is rolled back to v0.2.3: `lib/hive.rb:2`, `Gemfile.lock:4`, `README.md:43`, and `install.md:62` would make this branch report/install an older release than `origin/main` v0.2.4.
- [ ] Claude model/effort pins are silently ignored: `lib/hive/stages/base.rb:426`, `lib/hive/claude_launcher.rb:437`, and `lib/hive/agent.rb:269` remove `claude_cli_flags` propagation, breaking the released 0.2.4 cost-control contract for existing configs.
- [ ] `hive-init` v1 removes released answer fields without a schema bump: `schemas/hive-init.v1.json:39` and `test/integration/init_test.rb:80` bless removal of `answers.claude_model` / `answers.claude_effort` while `schema_version` stays 1.

## Medium
- [ ] Existing wiki log fragments are deleted: `wiki/log.d/20260611-103000-claude-model-effort.md:1`, `wiki/log.d/20260611T210441Z-patrol-native-reviewer-wiki-refresh.md:1`, and `wiki/log.d/20260612-101117-claude-model-effort-audit.md:1` remove append-only project history required by `AGENTS.md`.

## Nit

ivankuznetsov:

# Code Review — tui-briefly-shows-a-duplicate-260611-e9f8 (pass 3)

**Scope:** branch's own changes only (`git diff origin/main...HEAD`, merge-base `0212469b`):
`lib/hive/commands/status.rb`, `test/unit/commands/status_test.rb`, `wiki/commands/status.md`,
`wiki/commands/tui.md`, `wiki/gaps.md`, `wiki/log.d/20260611T201638Z-status-stage-move-race.md`.
(The init/config/agent/schema files in the two-dot diff are origin/main drift, not this task — excluded.)

**Intent:** Read-side fix in `Status#collect_rows` so a mid-scan stage move (atomic `File.rename`) no
longer (a) emits a transient duplicate row or (b) crashes with exit-70 ENOENT. Adds a vanished-folder
`rescue Errno::ENOENT` (re-raise if the folder survives), post-scan `drop_transient_stage_moves` dedup,
a `stage_task_entries` glob seam, regression tests, and wiki docs.

**Verdict: Ready to merge.** Implementation is correct and matches the plan's Requirements Trace
(A1–A7) and Units 1–4 in full. `drop_transient_stage_moves` runs before `annotate_actions`, so
`slug_counts`/`stage_collision` see deduped rows (genuine collisions preserved). Status unit suite
(35 runs, 166 assertions, 0 failures), integration suite (8 runs), and `rubocop` on both Ruby files
are all green. No P0/P1 issues. Findings below are documentation-precision and minor maintainability
items; none block merge. Reviewers: correctness, adversarial, reliability, testing, maintainability,
performance, cli-readiness, api-contract, project-standards, agent-native, learnings — correctness and
adversarial both returned empty after attempting to break the race logic.

## High

- [ ] _(none)_

## Medium

- [ ] `status.rb:479` re-raise comment overstates its premise — "This contract assumes EVERY state-file mutation is an atomic rename (see Markers.write_atomic)" is literally false: plain `File.write(task.state_file, …)` callers exist (`execute.rb:272`, `open_pr.rb:67`/`288`/`338`/`349`, `babysitter/pr_fixer.rb:238`). The re-raise stays safe in practice because `File.write` opens `O_CREAT|O_TRUNC`, so the state file is never transiently *absent* (only possibly truncated, on which `File.mtime` still succeeds) — but the stated justification is wrong and is the documented basis for a deliberate exit-70 crash on a poll-heavy surface; a future maintainer adding a delete-then-rewrite path would read "EVERY mutation is atomic rename" and miss that they'd be breaking the contract's real premise. Restate the invariant as "the state file is never transiently absent inside a surviving folder" (atomic-rename writers and `File.write` writers both uphold it).

## Nit

- [ ] `stage_task_entries` (`status.rb:500`) is public but exists solely as a test seam — mark it `private` / `# :nodoc:` so readers know there are no production callers; the test subclass can still override a private method and call `super`.
- [ ] `collect_rows` row-build body is a ~45-line `begin/rescue` nested inside the `each` block (`status.rb:426-494`); extracting it to a private `build_task_row(entry, stage, slug)` returning the hash-or-`nil` would place `rescue Errno::ENOENT` beside the code it guards and remove one level of nesting (rubocop is currently clean, so this is readability only).
- [ ] `VanishAfterDirectoryCheckPath` comment (`status_test.rb`) miscounts the `to_path` coercion sequence — it omits `Task.new`'s `File.expand_path` coercion between `File.basename` (call 2, where the rename fires) and `File.mtime` (the ENOENT site). Real order is `directory?`→`basename`→`expand_path`→`mtime` (confirmed empirically), not the 3 steps the comment narrates. Rename still fires at call 2 so behavior is correct; comment only.
- [ ] The vanish test's `@calls > 1` timing couples the double to `collect_rows`' internal call order. This is intentional — a synchronous rename inside the `stage_task_entries` override would fire before the `status.rb:423 File.directory?` guard and exercise the guard-skip path instead of the rescue path — and the `renamed?` assertion guards against silent drift. Keep both; just note that any added `File.*` call on `entry` before `File.basename` shifts the trigger and needs the threshold revisited.
- [ ] `test_collect_rows_reraises_enoent_when_folder_survives` monkeypatches `File.mtime` process-globally (restored in `ensure`). Safe under today's sequential Minitest, but would cause cross-test flakiness if parallel test execution is ever enabled; a `collect_rows`-override subclass seam (like `StatusRaceCommand`) would avoid mutating a core method globally.
- [ ] Coverage: race tests only exercise forward `8-finalize → 9-done` moves; a vanish during a non-finalize stage and a 3+ stage duplicate group through `drop_transient_stage_moves` are unexercised (the mechanism is stage- and count-agnostic, so low risk). Backward moves are explicitly out of plan scope and correctly not covered.

## Notes (non-findings, for the record)

- **Plan scope-boundary items dropped, not raised:** cli-readiness suggested adding a `CORRUPTED_TASK_PATH` error_kind and a `scan_notes`/dropped-slug array to the JSON envelope so agents can distinguish a transient drop from a deletion. Both propose new JSON fields, which the plan's Scope Boundaries explicitly exclude ("no new fields, no version bump"; one-cycle absence is the accepted behavior). Reasonable future enhancements, but out of this task's scope.
- **Residual risks (documented and consciously accepted by the plan/comments):** (1) spurious exit-70 only if a `<stage>/<slug>` folder is re-materialized in the sub-ms window between the failing `File.mtime` and the rescue's `File.directory?` re-check (e.g. `approve.rb` rollback `mv`) — astronomically narrow, self-heals next poll; (2) double-race → zero rows for one cycle (pinned by test); (3) slug-path identity assumption in dedup; (4) cross-device (`EXDEV`) stage moves would defeat the same-device ENOENT assumption, but all stage dirs live under one `.hive-state` root so this is not reachable in normal deployment.
- **Known-pattern alignment (learnings):** the `rescue ENOENT` + folder-existence discriminant matches the repo's documented "rescue the transient case, re-raise the real error" convention (`wiki/log.d/20260608-142201-status-schema-skew-fix-forward.md`) and complements `docs/solutions/architecture-patterns/silent-stage-rename-state-drift.md` (the twin `collect_rows` invariant). No contradiction.
- **Standards:** `wiki/index.md` correctly untouched (both edited pages pre-existed), `wiki/log.md` not hand-edited (fragment added under `wiki/log.d/`), `gaps.md` #35 sequential, no `sleep`/timeouts in tests. Clean.

ivankuznetsov:

# PR Review (pr-review-toolkit) — tui-briefly-shows-a-duplicate-260611-e9f8 (pass 3)

**Scope:** branch's own changes only (`git diff origin/main...HEAD`, merge-base `0212469b`):
`lib/hive/commands/status.rb`, `test/unit/commands/status_test.rb`, and wiki docs.
(init/config/agent/schema files in the two-dot diff are origin/main drift — excluded.)
Personas run: code-reviewer, silent-failure-hunter, comment-analyzer, pr-test-analyzer,
type-design-analyzer, code-simplifier. Read-only; no files edited.

**Verdict: Ready to merge.** No Critical/Important code defects. code-reviewer found the change
correct and standards-compliant (dedup runs before `annotate_actions`; rescue contract sound;
JSON schema unchanged; no `sleep` in tests). Suite green (35 runs / 166 assertions / 0 failures),
rubocop clean. All findings are documentation-precision, test-robustness, and polish items.

## High

- [ ] _(none)_

## Medium

- [ ] [comment-analyzer] `status.rb:479-480` comment "This contract assumes EVERY state-file mutation is an atomic rename (see Markers.write_atomic)" is factually false — confirmed plain `File.write(task.state_file,…)` callers exist (`execute.rb:272`, `open_pr.rb:67`/`288`/`338`/`349`, `babysitter/pr_fixer.rb:238`). The re-raise's safety conclusion still holds (those writes are in-place `O_CREAT|O_TRUNC`, so the file is transiently zero-*length*, never zero-*existence*, and `File.mtime` succeeds), but the stated premise is wrong; restate as "the only thing that makes `entry` (the folder) vanish is an atomic stage-move rename; in-place file writers never make the folder disappear" so a future maintainer adding a folder-level unlink/rename understands what actually breaks the contract. (silent-failure-hunter independently corroborates the catch-surface mismatch below.)

## Nit

- [ ] [silent-failure-hunter] `status.rb:469-490` rescue comment names only `File.mtime(entry)` at `:433` as the ENOENT source, but the `rescue` wraps the whole `begin` so it also covers `Markers.current`'s `File.exist?`→`File.read` TOCTOU (`markers.rb:59-61`) and the lock reads; behavior is uniform and correct, but amend the comment to "any in-folder read (mtime at :433, Markers.current read, lock read)" so the catch surface isn't undersold.
- [ ] [comment-analyzer][type-design-analyzer][pr-test-analyzer] `VanishAfterDirectoryCheckPath` comment (`status_test.rb`) under-counts the `to_path` coercion ledger — it omits `Task.new`→`File.expand_path` (an intervening pure-string coercion) between `File.basename` (call 2, rename fires) and `File.mtime`, and `File.mtime` itself coerces twice. Harmless (rename only needs *a* coercion after the directory guard, and `renamed?` guards drift), but the 3-call narration misleads anyone reasoning about the exact `@calls` threshold; clarify "fires on the first coercion after the directory guard; downstream count isn't load-bearing."
- [ ] [comment-analyzer] `status.rb:472` "the task resurfaces under its new stage in this same scan" is only guaranteed for forward moves (stages scanned 1→9); a backward `mv` resurfaces on the next poll — soften to note the backward case (no correctness impact; the row simply drops and reappears next poll).
- [ ] [pr-test-analyzer] The vanish test's `assert vanish_double.renamed?` guard catches coercion-count *drift* but not *replacement* of `File.basename` with a non-`File.*` string op (e.g. `entry.split("/").last`) that keeps some later `File.*(entry)` call — the test could then pass for the wrong reason. Add a focused unit test on `VanishAfterDirectoryCheckPath` asserting the rename fires exactly on `to_path` call 2, decoupling the timing contract from `collect_rows`' internal call order.
- [ ] [pr-test-analyzer] No coverage for a vanish during a NON-finalize stage; both vanish tests use `8-finalize → 9-done`. The rescue is stage-agnostic (keys on ENOENT + folder existence), so low risk, but a `vanish_during_read_stage: "4-execute"` → `6-review` case would close the "only tested at the terminal stage" blind spot.
- [ ] [pr-test-analyzer] `drop_transient_stage_moves` handles N≥3 members uniformly via `reject`, but no test feeds a 3+ stage duplicate group (two vanished + one surviving); add one to pin "exactly the surviving row remains" against a future "keep exactly one" rewrite.
- [ ] [pr-test-analyzer] The line-434 state-file TOCTOU (`File.exist?(state_file)` then `File.mtime(state_file)`) — the case the comment explicitly calls the intended re-raise path — is exercised only via the folder read at `:433`, not the state-file read; a test keyed on the state-file path would document intent more precisely (same rescue, identical behavior today).
- [ ] [silent-failure-hunter] Neither silent path (`next` skip-on-vanish at `:493`, row drop in `drop_transient_stage_moves`) emits any breadcrumb, so a mis-assumption (a slug perpetually flapping, or a recurring drop from a real producer bug rather than a one-off rename) would be invisible. Acceptable given the self-healing design and that `Status` has no process-wide logger, but a single debug-log line (e.g. via `tui/debug.rb` when reachable) would make a wrong assumption diagnosable.
- [ ] [code-simplifier] `status.rb:426-494` — flatten the nested `begin` (inner `rescue Hive::InvalidTaskPath`, outer `rescue Errno::ENOENT`) into one `begin` with both `rescue` clauses; the two exception classes are disjoint siblings and `Task.new` is already inside the ENOENT-covered region, so the covered region is identical, both `next` targets stay correct, and one indentation level is removed (behavior-preserving).
- [ ] [code-simplifier] `status.rb:512-518` — `drop_transient_stage_moves` builds a `group_by` Hash of per-slug row arrays that are materialized then discarded (only `.key?` is used); `rows.tally { _1[:slug] }` + `counts[slug] > 1` expresses "appears more than once" directly and keeps the `assert_same` early-exit (polish only; current form is correct and rubocop-clean).

## Notes (non-findings / context for triage)

- **Exit-70 re-raise is contained, not a crash loop (silent-failure-hunter).** The only in-process consumer — the TUI poller — wraps `json_payload`/`collect_rows` in `rescue StandardError => @last_error` (`state_source.rb:106-114`), so a re-raised ENOENT surfaces as a stalled frame and self-heals next poll; `hive status`/`--json` (daemon, bot) run as per-poll subprocesses (no in-process `Status.new`/`json_payload` callers in `lib/hive/daemon` or `lib/hive/bot`), so a re-raise is one failed poll. This addresses the "noisy crash on a poll-heavy surface" concern: the trade is silent-corruption-masking → contained, observable, self-healing failure.
- **code-simplifier explicitly recommends AGAINST extracting `build_task_row`** — it cannot cleanly carry the two `each`-targeting `next`s nor the `raise if File.directory?(entry)` re-check, so extraction would fracture the deliberately co-located race contract for ~no net clarity. (Counters a possible "extract the method" suggestion; the flatten in the Nit above is the better readability win.)
- **Pre-existing, out of scope (silent-failure-hunter):** `worktree_path = begin … rescue StandardError; nil end` (`status.rb:443-448`) swallows a malformed `worktree.yml` / config error into an indistinguishable `nil`. Unchanged by this diff (only re-indented) — flagged only for the record; narrow it if touched later.
- **type-design-analyzer** rated both new test doubles 8–9/10 (XOR-mode constructor guard, one-shot `@renamed` latch, `renamed?` observability hook, sound `to_path`-only/no-`to_str` design verified against real call sites); the `stage_task_entries` seam is a clean template-method, not test-concern leakage. No defects.

PR #463, opened 2026-06-12
wiki: gap — macOS colima smoke flake in release.yml

ivankuznetsov:
Post-publish verification job only; the v0.3.0 image itself is verified (multi-arch manifest + local container smoke incl. /health?deep=1).


ivankuznetsov:

Outcome: timeout

Agent tail:

```
The last job is still active in its minitest step. I’m checking its live log now to make sure it’s just a long root suite and not a silent failure state.
```

PR #477, opened 2026-06-14
feat(digest): daily shipped digest Telegram message

ivankuznetsov:
## Summary

Hive merged work silently: the single user had no daily readout of what actually shipped. This adds a daemon-scheduled cadence job that, once per day at local midnight, sends a Telegram message summarizing everything Hive merged in the local calendar day that just ended. The message is an LLM-written changelog, grouped by category (New features / Fixes / Patrol tasks) and split per registered project. Empty days send `Nothing shipped today 🌙`; a model failure sends a short "digest failed" notice. The digest is never silent.

The whole feature ships together because the six units form one end-to-end pipeline: nothing is useful until a scheduled tick can collect shipped tasks, categorize them, render Telegram Markdown, and deliver to the user.

### Architecture

The daemon cannot send Telegram messages: `hive daemon` only spawns subprocesses, and the separate `hive bot` process owns the Telegram client. So the scheduler lives in the daemon, but the actual digest work and send happen in a dispatched `hive digest` subprocess that constructs its own Telegram client and sends directly. This mirrors how `PatrolScheduler` dispatches `hive patrol <project>`, keeps the feature self-contained, and makes the whole thing testable by invoking one command. A `sendMessage` from the subprocess does not conflict with the bot's long-poll `getUpdates`.

```mermaid
graph TB
  D[hive daemon] --> S[DigestScheduler tick]
  S -->|hive digest --date D| P[digest subprocess]
  P --> C[Collector]
  C -->|9-done task files| FS[(on-disk pr.md + meta.yml)]
  C -->|ship time| G[(hive/state git log)]
  C --> CAT[Categorizer · Hive::Agent]
  CAT --> R[Renderer · MarkdownV2]
  R --> SND[Sender]
  SND --> TG[(Telegram)]
```

The work splits along a deterministic / generative seam, which satisfies the no-stub-AI rule while keeping most logic unit-testable:

- **Deterministic Ruby** decides the local-day time window (host TZ, DST-safe via `getlocal`), collects shipped tasks across all registered projects, groups them per project, renders the final Telegram Markdown, resolves the recipient, and drives scheduling plus catch-up.
- **The model** (the same `Hive::Agent` + `agent_profiles` path Patrol uses) does only two things per shipped task: assign a category and write a one-line friendly summary.

### Design decisions

| Decision | Rationale |
|---|---|
| Ship time comes from `hive/state` git history (`pr_finalized` commit, fallback `archived`), not from disk | Feature tasks in `9-done/` carry no events log and no on-disk ship timestamp; the `hive/state` commit is the authoritative "merged to main" signal. The resolver is injectable so tests need no real git. |
| `hive digest` exits 0 whenever *something* was sent (digest, empty, or failed-notice) | The scheduler advances its cursor on exit 0 and never spam-retries a handled failure. Non-zero is reserved for an unexpected crash, which leaves the day owed for the next tick. |
| Catch-up is a deterministic owed-days range with a cap | Owed = every local day in `(last_digested_date, today-1]`, oldest first, one dispatch per day. Downtime across midnights replays missed days; a first-run guard avoids blasting history; `max_catchup_days` (default 7) caps floods and logs what it skips. |
| Single global `digest.agent` over all projects' items | One combined message, one model run. Falls back to the patrol agent, then `claude`. The digest cursor is global state (`digest_state.json`), not per-project. |
| MarkdownV2 with a tested `escape_mdv2` over all dynamic text | Model prose can contain reserved characters; the escaper wraps everything except URL targets, and the model is instructed to emit plain text. |
| Recipient is the single primary chat | `bot.digest_chat_id`, falling back to `chat_id_allowlist.first`. No broadcast. |

### Surface added

New `hive digest [--date YYYY-MM-DD] [--dry-run] [--json]` command, the `Hive::Digest` service objects (collector, categorizer, renderer, sender, orchestrator, window, ship-times), `Hive::Daemon::DigestScheduler` with dispatcher wiring, a `templates/digest_prompt.md.erb` prompt, a `schemas/hive-digest.v1.json` `--json` contract, and a `digest:` config block (plus `bot.digest_chat_id`, `budget_usd.digest`, `timeout_sec.digest`) with defaults, validation, and documented `hive_config.yml.erb` comments. The feature is opt-in: `digest.enabled` defaults to `false`.

## Test plan

- **Deterministic digest unit suite** (`test/unit/digest/`, re-run green this finalize): window/DST boundary, ship-time resolution and fallback chain, per-project collection with graceful degradation, JSON parsing/mapping and unknown-category default, MarkdownV2 rendering + `escape_mdv2`, orchestrator routing (empty / failed-notice / success), recipient resolution, and the `Data` boundary invariants.
- **Scheduler + daemon wiring** (`test/unit/daemon/digest_scheduler_test.rb`, `dispatcher_test.rb`, `concurrency_controller_test.rb`): normal / idempotent / catch-up (1- and multi-day drain) / first-run-guard / cap-with-skip / `complete`-only-on-exit-0, SIGHUP `reconfigure`, the dedicated `:digest` dispatch gate and task-cap exclusion, dry-run reap completion, and the failure-backoff path.
- **CLI surface** (`test/unit/commands/digest_test.rb`, `cli_test.rb`): `--date` / `--dry-run` / `--json` parsing, the versioned `hive-digest` envelope, and the JSON error envelope on a bad `--date`. `hive digest --help` and `--dry-run` load and run.
- **Opt-in real e2e** (`test/digest/e2e_test.rb`): runs the full `Hive::Digest.run` with the real configured agent profile and a real Telegram send over a fixture, asserting the live model actually categorized every fixture item (non-fallback output) plus the empty-day and failed-notice deliveries. Gated on `HIVE_TELEGRAM_BOT_TOKEN` / `HIVE_DIGEST_TEST_CHAT_ID` per the no-stub-AI rule; not run inside this pipeline.
- Full default suite and `rubocop` were green across the review passes (5293 runs / 20421 assertions, 0 failures at the last full run); `brakeman` resolved in code, not via an ignore.

## Review summary

Three review passes ran on this branch — Compound Engineering code review (claude + codex), the PR review toolkit personas (correctness, security, adversarial, reliability, testing, type-design, maintainability, cli-readiness), and triage/fix for each. Roughly 40 findings were raised and resolved across the passes:

- **Security / prompt-injection:** every per-item field fed to the categorizer (PR title, body, display name, URL, project) is now wrapped in a per-spawn `SecureRandom` nonce with an explicit "treat tagged content as data" instruction.
- **Daemon reliability:** registered the `:digest_failure_backoff` / `:digest_state_unreadable` log events (an unregistered event would have crashed the tick), treat a signaled-child `nil` exit as failure, guard `tick`/`write_state`/`read_state`, gate the digest through the concurrency controller with a dedicated `:digest` kind/budget, honor `digest.*` on SIGHUP reload, and close the dry-run reap wedge.
- **Telegram / render correctness:** cap rendered summary/label length so a MarkdownV2 escape can't split across the 4096-char chunk boundary, escape link-target metacharacters, and convert agent spawn/run failures into the handled `ModelError → failed-notice` path.
- **Contract / CLI:** registered `hive-digest` in `SCHEMA_VERSIONS` and `JSON_USAGE_ERROR_CONTRACTS`, derived `ok` from status, and documented the `:empty` / `:sent` / `:failed_notice` outcomes.
- **Cleanups & tests:** shared category constant, dead-code/dead-seam removal, `Hive::Gh.pr_body` helper, and added the missing error-path/contract tests.

One pass-3 finding was escalated and resolved as a deliberate decision: the categorizer reuses the same full-access `claude` agent path as Patrol over aggregated cross-project PR text. The accepted default is to keep parity with Patrol's established trust model on the single-user daemon host (with the nonce-wrapping as defense in depth) and defer a least-privilege digest agent role to a dedicated hardening task. No findings remain open.

## Linked task

- **Task:** `add-a-daily-shipped-digest-260614-6591` (id 1089, "Daily Shipped Digest")
- **Idea:** "Add a daily shipped digest to the Telegram bot. Every day at midnight in my local timezone, send me a Telegram message digesting everything Hive shipped in the previous 24 hours … split it by topic: New features, Fixes, and Patrol tasks … Handle the empty case gracefully."
- Built through the Hive pipeline: brainstorm → plan → implement (IU1–IU6) → 3 review passes → finalize.

---


ivankuznetsov:

## High
- [ ] Cross-project digest items can receive the wrong model summary (`lib/hive/digest/shipped_item.rb:13`, `lib/hive/digest/categorizer.rb:84`): IDs are only PR numbers or slugs, so common PR numbers across registered projects overwrite each other in the model-output map.
- [ ] In-flight digest state is lost on daemon restart (`lib/hive/daemon/digest_scheduler.rb:46`): `@pending` is process-local and the cursor advances only on completion, so a crash/restart can dispatch the same date while the old subprocess may still send.
- [ ] Babysitter dry-run skip logs lost symlink and control-character hardening (`bin/hive-babysitter-stub-git:6`, `bin/hive-babysitter-stub-gh:6`): skipped commands now append through symlinks and write raw argv, allowing audit-log clobbering or forged log lines.

## Medium
- [ ] `hive new` task text is parsed as wrapper flags again (`bin/hive:68`, `bin/hive:80`): idea text containing `--help` opens command help and `--json=yes` exits usage instead of being captured.
- [ ] Daemon reload ignores digest config changes (`lib/hive/daemon/dispatcher.rb:1587`): SIGHUP reloads only daemon/update config, so `digest.enabled` and `digest.max_catchup_days` changes require a full restart.
- [ ] Partial Telegram chunk failures can duplicate delivered digest chunks (`lib/hive/digest/sender.rb:21`): a later chunk exception exits nonzero with the cursor unchanged, so retry resends earlier chunks that Telegram already accepted.
- [ ] Existing wiki changelog fragments are deleted (`wiki/log.d/20260615T065529Z-babysitter-skip-log-hardening.md:1`): removing this and sibling fragments loses per-change source history that AGENTS.md treats as append-only review context.

## Nit
- [ ] Normal PR bodies are reported to the model with title `Summary` (`lib/hive/digest/collector.rb:75`): the first `## Summary` section heading is used as `pr_title`, weakening the plan-required PR-title input.
- [ ] MarkdownV2 link targets are interpolated raw (`lib/hive/digest/renderer.rb:57`): a PR URL containing `)` or backslash can break Telegram entity parsing for the digest message.

ivankuznetsov:

# Code Review — Daily "Shipped Digest" (pass 1)

Scope: digest feature on `add-a-daily-shipped-digest-260614-6591` vs `origin/main`.
Deterministic digest suite (155 runs / 456 assertions) is green and `rubocop` is clean
on all new files; `hive digest --help`/`--dry-run` load and parse. All plan IUs (IU1–IU6)
and Requirements Trace rows appear implemented — no plan-required-but-missing gaps, so no
High findings. The non-digest reverse-diffs (`bin/hive`, babysitter stubs, golden-path,
`cli_version_test`) are branch-staleness artifacts (origin/main advanced past the merge-base),
not feature changes — see the Nit on rebasing.

## High

## Medium
- [ ] Digest dispatch bypasses the concurrency controller and is recorded as a `:task` kind (`lib/hive/daemon/dispatcher.rb:202-204,1030-1042`): unlike patrol (own `can_dispatch_patrol_scan?` budget + `:patrol_scan` kind), the digest is dispatched with no `can_dispatch?` gate and `dispatch_command`'s default `kind: :task`, so it can transiently exceed `max_concurrent_runs` and hold a real task slot for up to the 30-min digest timeout, delaying pipeline work once per day — gate it or give it a dedicated kind/budget.
- [ ] `digest.enabled` / `max_catchup_days` are read only at daemon start, not on SIGHUP reload (`lib/hive/commands/daemon.rb:188-194`, `lib/hive/daemon/dispatcher.rb` `reload_config!`): `reload_config!` never reconfigures `@digest_scheduler` and rebuilds `@config` without the `digest` key, so enabling the digest via config + reload has no effect until a full restart — inconsistent with the rest of the daemon's "takes effect within one tick" reload contract (timeouts, grace, per-project enable).
- [ ] `Collector#build_item` silently swallows `Hive::GitError` → `nil` with no logging (`lib/hive/digest/collector.rb:58-60`; collector has no logger): a failing `git log` on a project's `hive/state` makes every shipped item for that project vanish from the digest, so a corrupt/missing-branch repo can under-report shipped work or yield a false "Nothing shipped today 🌙" with zero operator signal — log the dropped slug/error instead of dropping silently.

## Nit
- [ ] `ShipTimes#log_commits` uses `--grep=#{slug}` without `--fixed-strings`/`-F` (`lib/hive/digest/ship_times.rb:20-23`), deviating from the plan's explicit "fixed-string grep"; git treats the pattern as a regex, so a slug containing a metachar (e.g. an unbalanced `[`) would make `git log` exit non-zero → `GitError` → the item is silently dropped (see Medium above). Add `-F`.
- [ ] `Collector#pr_title` returns the first markdown heading (`lib/hive/digest/collector.rb:75-80`), which for hive-generated `pr.md` is always `## Summary`; `pr_title` therefore resolves to the constant literal "Summary" for nearly every item, adding no signal over `display_name` and mildly misleading the model (acknowledged best-effort per Open Question 2, but worth revisiting).
- [ ] Confusingly similar config-loader names `load_global_digest` (returns just the `digest` block) vs `load_global_digest_config` (returns the full merged config incl. `bot`) (`lib/hive/config.rb:902-933`); the near-identical names make it easy to wire the wrong one into the scheduler vs the sender — rename for clarity.
- [ ] `digest.max_catchup_days` validation requires an integer `>= 1` (`lib/hive/config.rb` `validate_digest!`), but `DigestScheduler#apply_catchup_cap` treats `0` as "no cap" (`lib/hive/daemon/digest_scheduler.rb:20,75`); the unlimited branch is therefore unreachable from config — either allow `0` (= unlimited) in the validator or drop the dead branch.
- [ ] `Renderer#render_line` interpolates `pr_url` into the MarkdownV2 link target unescaped (`lib/hive/digest/renderer.rb:56-57`); per the MarkdownV2 spec `)` and `\` inside a link destination must be escaped, and one malformed URL would fail the entire Telegram `send_message` for that day. GitHub URLs are safe in practice (and the plan says "URL untouched"), but escaping the two link-breaking chars is cheap defense.
- [ ] Categorizer run dirs `<state_home>/digest/runs/<date>-<hex>` are created per run and never cleaned up (`lib/hive/digest/categorizer.rb:137-143`), so `items.json`/`state.yml`/`logs` accumulate indefinitely — add cleanup or document the retention expectation.
- [ ] The `origin/main..HEAD` diff includes reverse-diffs of unrelated files (`bin/hive`, `bin/hive-babysitter-stub-*`, `web/test/e2e/golden_path_e2e.rb`, `test/integration/cli_version_test.rb`) because origin/main advanced past this branch's merge-base; a 3-way merge won't actually revert main's work, but rebasing on current main before merge keeps the PR diff focused on the digest feature.

ivankuznetsov:

# PR Review Toolkit — Daily "Shipped Digest" (pass 1)

Multi-persona review (code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design,
comment-analyzer, code-simplifier) over the digest feature in `git diff origin/main..HEAD`.
Non-digest reverse-diffs (bin/hive, babysitter stubs, golden-path, cli_version_test) are
branch-staleness artifacts and were excluded from scope.

Reconciliation notes (cross-persona): two silent-failure-hunter "High" items were **dropped**
as non-defects — (a) `ModelError → failed-notice → exit 0 → cursor advances → no retry` is the
plan's *explicit* design ("never spam-retries a handled failure"); (b) the "rescue can re-raise"
claim is unreachable because `local_date`/`sender` are always assigned before any `ModelError`
can fire (confirmed by code-reviewer + code-simplifier).

## High
- [ ] [code-reviewer] Cross-project PR-number collision in `categorizer_id` (`lib/hive/digest/shipped_item.rb:14`, `lib/hive/digest/categorizer.rb:84-92`): `categorizer_id = (pr_number || slug).to_s`, but PR numbers are repo-local — two registered projects that ship the same PR number on the same day collapse to one id in `by_id`, so both items render with the same summary/category (and the agent can't disambiguate from the prompt). Make the id project-scoped (e.g. `"#{project_name}/#{pr_number || slug}"`) in both the prompt and the mapping.

## Medium
- [ ] [silent-failure] `Collector#build_item` swallows `Hive::GitError → nil` with no log or counter (`lib/hive/digest/collector.rb:58-60`): a project whose `hive/state` `git log` fails (corrupt repo, missing branch, permissions) has *all* its shipped items dropped via `filter_map`, producing a silently under-reported digest or a false "Nothing shipped today 🌙" — log the dropped slug/project/error.
- [ ] [silent-failure] No retry backoff or cap on a failing digest date (`lib/hive/daemon/digest_scheduler.rb:50-60`): non-`ModelError` failures exit non-zero, `complete` leaves the cursor unmoved, and the next tick re-dispatches the same date every poll interval — re-spawning the paid categorizer agent each time, with no escalating signal (contrast `dispatch_patrol_with_gates`' failure backoff).
- [ ] [code-reviewer] Recipient/token presence is checked only at send time, after the categorizer agent has already run (`lib/hive/digest.rb:30-33`, `lib/hive/digest/sender.rb:26-36`; validators check type only, never presence): a missing `bot.digest_chat_id`/allowlist or unset `HIVE_TELEGRAM_BOT_TOKEN` wastes a full LLM run and (with the no-backoff finding) hot-loops — resolve the recipient/token before categorizing.
- [ ] [silent-failure] Digest dispatch bypasses the concurrency controller entirely (`lib/hive/daemon/dispatcher.rb:1030-1042`): no `can_dispatch?`/`running_task?` gate (unlike patrol/archive), so it can push the global run count past `max_concurrent_runs`, and a crash-restart that loses in-memory `@pending` while a prior `hive digest` child is still alive can spawn a concurrent same-date run with no task `.lock` backstop.
- [ ] [code-reviewer] Inline-link URL is not MarkdownV2-escaped (`lib/hive/digest/renderer.rb:57`): a `)` or `\` in `pr_url` produces malformed entities and Telegram 400s the *entire* message, dropping the whole digest; GitHub URLs are safe in practice, but escape the two link-breaking chars in the link target as cheap defense.
- [ ] [type-design] The canonical category set is duplicated with no shared source of truth — `VALID_CATEGORIES` (`lib/hive/digest/categorizer.rb:20`) vs `CATEGORY_ORDER` (`lib/hive/digest/renderer.rb:8`); if the two lists ever diverge, the renderer silently drops items the categorizer accepted. Share one ordered constant.
- [ ] [tests] Error/fallback branches are untested: ship_times `GitError` raise (`lib/hive/digest/ship_times.rb:24-27`), the collector's `GitError → nil` drop (`lib/hive/digest/collector.rb:58-59`), and `matching_action`'s non-`9-done` suffix fallback (`ship_times.rb:46`) — the git-failure contract the empty/failed flows depend on is unverified (the fakes never raise).
- [ ] [tests] Catch-up cap full multi-day drain is not tested end-to-end (`test/unit/daemon/digest_scheduler_test.rb:78-92`): only the first capped dispatch + skip event are asserted, so a bug in `owed.last(@max_catchup_days)` slicing or the post-cap cursor that stranded or double-dispatched the remaining capped days would still pass.
- [ ] [comments] CLI `long_desc` omits the empty and failed-notice outcomes (`lib/hive/cli.rb:494-502`): it implies the agent always runs and a normal digest always sends, but `Hive::Digest.run` has three outcomes (`:empty` spawns no agent, `:sent`, `:failed_notice`) — the help is the main user surface and under-describes behavior.

## Nit
- [ ] [simplify] Dead defensive code in the `Digest.run` rescue (`lib/hive/digest.rb:36-38`): `local_date || date`, `local_date || Window.parse_date(date)`, and `sender || Sender.new(cfg:)` are all unreachable because `local_date` and `sender` are assigned before any `ModelError` can fire — drop them (this is also why the "rescue can re-raise" concern does not actually occur).
- [ ] [simplify] Confusingly near-identical config-loader names `load_global_digest` (returns the `digest` block only) vs `load_global_digest_config` (returns the full merged config incl. `bot`) (`lib/hive/config.rb:902,920`); rename to convey scope to avoid mis-wiring scheduler vs sender.
- [ ] [type-design] `Result.status`, `CategorizedItem.category/summary`, and `SendResult`'s dry-run shape are invariants enforced only by single private factories, not at the `Data` boundary (`lib/hive/digest.rb:14`, `lib/hive/digest/categorizer.rb:17`, `lib/hive/digest/sender.rb:9`): any other `.new` caller can construct an illegal instance (unknown category, blank summary, dry-run with a chat_id). Cheap to guard in `initialize`.
- [ ] [silent-failure] Failure diagnostics give no actionable pointer: `ModelError` omits the agent log dir/run dir (`lib/hive/digest/categorizer.rb:63,98-102`) and the Telegram failure notice (`lib/hive/digest/renderer.rb:32-34`) names neither the run dir nor a re-run command — debugging a failed digest means hunting for the run folder manually.
- [ ] [silent-failure] `--json` payload hardcodes `"ok" => true` even for `:failed_notice` (`lib/hive/commands/digest.rb:48-56`): machine consumers scraping `ok` can't distinguish a failed digest from a sent one (the truth is only in `status`).
- [ ] [silent-failure] `Collector#pr_body` swallows `SystemCallError`/`IOError → ""` with no log (`lib/hive/digest/collector.rb:71-72`): a correctable `pr.md` read problem silently degrades the item to a default summary with no operator signal.
- [ ] [comments] `digest.max_catchup_days` "0 = no cap" sentinel (`lib/hive/daemon/digest_scheduler.rb:20,74-75`) is unreachable via config (validator requires `>= 1`) and undocumented — either allow `0` (= unbounded) in the validator or drop the dead branch.
- [ ] [comments] Wiki wording drift (`wiki/modules/digest.md:24,118-119`): "first ship commit" understates the preference-ordered (not chronological-first) selection, and "one per child completion" omits that a non-zero exit re-dispatches the same day every tick.
- [ ] [tests] Brittle/weak assertions: `run_test.rb:54` couples the orchestrator test to the renderer's exact link markdown; `dispatcher_test.rb:265` ("without_project_gate") never asserts the gate was actually bypassed; and the `pr_title` heading-extraction / no-heading-fallback (`collector.rb:75-80`) has no dedicated test.
- [ ] [tests] The `agent_name` (`digest.agent → patrol.agent → "claude"`) and `budget_usd.digest`/`timeout_sec.digest` fallback chain (`lib/hive/digest/categorizer.rb:175-185`) is unverified — a wrong default silently routes the categorizer to the wrong model/budget.
- [ ] [simplify] `matching_action`'s two-pass `find` (`lib/hive/digest/ship_times.rb:44-47`) and the triple `VALID_CATEGORIES.include?` evaluation (`lib/hive/digest/categorizer.rb:109,114`) can each collapse to a single pass / single check with no behavior change.
- [ ] [silent-failure] (by design — not a defect, noting for triage) `ModelError → failed-notice` exits 0 so the scheduler advances and won't retry (`lib/hive/digest.rb:35-38`), matching the plan's "never spam-retry a handled failure"; the consequence is that a *transient* model blip is recoverable only via a manual `hive digest --date <D>`.

ivankuznetsov:

## High
- [ ] `lib/hive/daemon/digest_scheduler.rb:105` emits an unregistered daemon log event: `:digest_failure_backoff` is absent from `Hive::Daemon::Logger::EVENTS`, so any non-zero digest child can raise during completion and break retry/catch-up handling.
- [ ] `lib/hive/daemon/digest_scheduler.rb:70` treats signaled digest children as successful: `ChildSupervisor` reports `nil` exit status for signal exits, and `nil.to_i.zero?` advances the cursor, silently skipping that digest date.
- [ ] `templates/digest_prompt.md.erb:29` wraps PR bodies without an explicit untrusted-data instruction: shipped PR text can prompt-inject a normal Hive agent run before the digest JSON is produced and sent.

## Medium
- [ ] `lib/hive/digest/collector.rb:49` reads `pr.md` frontmatter outside the tolerant body fallback: an unreadable or directory-shaped `pr.md` can abort the whole multi-project digest instead of degrading that one task.
- [ ] `lib/hive/digest/sender.rb:74` calls `logger.error` on `Hive::Bot::Logger`: Telegram send failures are masked by `NoMethodError`, losing the original delivery error and partial-send context.

## Nit
- [ ] `lib/hive/daemon/digest_scheduler.rb:109` logs the wrong retry timestamp: `next_eligible_at` is recorded as the failure time instead of `now + interval`, misleading operators about backoff timing.

ivankuznetsov:

# Code Review — Daily "Shipped Digest" (pass 2)

Scope: `git diff origin/main..HEAD` on `add-a-daily-shipped-digest-260614-6591`.
Mode: report-only (no code edited). Reviewers: correctness, reliability,
adversarial, testing, maintainability, cli-readiness.

Intent: a daemon-scheduled global daily job dispatches a `hive digest` subprocess
that collects tasks shipped on a local calendar day across all registered
projects (on-disk `9-done/` + `hive/state` git-log ship-times), runs an LLM
categorizer, renders one Telegram MarkdownV2 message, and sends it; empty days
send a "nothing shipped" notice and LLM failures send a "digest failed" notice.

Requirements completeness: the plan's Requirements Trace and IU1–IU6 scenarios all
appear implemented and tested (window/DST, catch-up cap, first-run guard,
complete-on-exit-0, enabled:false, sectioning/line-format/empty, escape_mdv2,
parse fixture, malformed-JSON ModelError, unknown-category default, recipient
resolution, dispatcher routing, real e2e). No plan-required item is missing.

## High
- [ ] Prompt injection via the **unwrapped PR title** in the categorizer prompt (`templates/digest_prompt.md.erb:28`, fed by `lib/hive/digest/collector.rb:689` `pr_title`): the PR *body* is wrapped in a per-render `SecureRandom` nonce tag but the *title* (the first non-"Summary" heading of the attacker-influenceable `pr.md`) is interpolated raw, outside the wrapper. Because all items are classified in one shared prompt that emits one combined JSON, a crafted heading on item B can steer the model's category/summary for another item — and the categorizer is a real tool-capable `Hive::Agent` (claude) run with the configured permission mode. The deliberate injection control applied to the body should also cover the title (wrap it in the same nonce / treat every per-item field as data).

## Medium
- [ ] `DigestScheduler#tick` disk writes are unguarded and can crash the daemon tick (`lib/hive/daemon/dispatcher.rb:202`, `lib/hive/daemon/digest_scheduler.rb:125,197,230`): `tick` calls `write_state` (first-run seed + catch-up cap), which lets `File.write`/`rename` `SystemCallError` (ENOSPC/EACCES/EROFS) propagate. The sibling self-heal ops (`stale_agent_healer`, `display_name_backfiller`) are deliberately wrapped in `rescue StandardError` precisely because an unhandled raise crashes the tick and trips the unit's restart-loop cap; the digest tick (which does disk I/O) is not wrapped. Wrap the `@digest_scheduler&.tick` call (or rescue in `write_state`).
- [ ] Daemon restart between digest **spawn and reap** re-dispatches the same date → duplicate paid categorizer run + duplicate Telegram digest (`lib/hive/daemon/digest_scheduler.rb:114-138`): the only cross-restart de-dup is `last_digested_date`, which advances only in `complete(exit_code: 0)`; `@pending` and the controller in-flight state are in-memory. If the daemon dies after the child is spawned but before reap, the next startup tick re-dispatches the same day. There is no persisted "in-flight date", so `digest_dispatch_gate` cannot catch this. Consider persisting an in-flight/attempted marker before dispatch.
- [ ] `--json` payload omits `schema_version` and `hive-digest` is unregistered in `SCHEMA_VERSIONS` (`lib/hive/commands/digest.rb:52`, `lib/hive.rb:15`): every other `--json` command emits a versioned envelope (`schema` + `schema_version` via `SCHEMA_VERSIONS.fetch`); the digest command hardcodes a bare `"schema" => "hive-digest"` with no version and no registered schema, breaking uniform machine parsing/migration. Register `hive-digest` and use `fetch`.
- [ ] Partial Telegram delivery re-sends already-accepted chunks on scheduler retry (`lib/hive/digest/sender.rb:861-876`): `send_in_chunks` re-raises on a mid-stream chunk failure (logging "not auto-resending accepted chunks") → subprocess exits non-zero → scheduler retries the same date → `deliver` runs from chunk 0, duplicating chunks Telegram already accepted. The retry path defeats the code's own stated no-resend intent for multi-chunk messages.
- [ ] MarkdownV2 escape sequence can be split across the 4096-char Telegram chunk boundary (`lib/hive/digest/renderer.rb:759`, `lib/hive/bot/telegram.rb:~261` `split_message`): a single rendered line longer than 4096 chars with no newline is hard-cut, and the cut can land between a `\` and the char it escapes → Telegram 400 on that chunk → `send_in_chunks` re-raises → whole digest send fails (then permanently retry-fails, losing the day). Model summaries are not length-bounded; cap summary/line length or split on escape-safe boundaries.
- [ ] Unbounded PR body inlined into the categorizer prompt (`lib/hive/digest/collector.rb:670-684` `pr_body`, `templates/digest_prompt.md.erb:30`): `pr_body` returns the entire `pr.md` with no size cap and the template inlines all of it for every item. One oversized `pr.md` (or many items in a day) can exceed the model context / default $50 budget and raise `ModelError`, degrading the **entire global digest** for all projects to the generic failure notice. Truncate per-item body to a sane budget.
- [ ] `lib/hive/digest/renderer.rb:3` `require "hive/digest/categorizer"` is unused — the renderer references no symbol from it (its real dependency, `Categories::ORDERED`, is already required on line 2). The require transitively pulls in `hive/agent`/`hive/agent_profiles`/`hive/stages/base` and makes the dependency graph look circular (renderer→categorizer while both share `categories`). Drop it; the renderer consumes `CategorizedItem` purely by duck typing.

## Nit
- [ ] `Hive::Digest::Collector#@clock` is injected (`collector.rb:617,621`) but never read anywhere in the class — a dead seam implying time-dependence the collector does not have; remove it.
- [ ] `Categorizer::RunnerTask#project_root` is populated (`categorizer.rb:396,553`) with the same value as `folder` but never read by `Hive::Agent`; dead field copied from the real `Task` shape — drop or document as deliberate interface mirroring.
- [ ] pr.md frontmatter-stripping is duplicated: `Collector#pr_body` (`collector.rb:670-684`) re-reads `pr.md` (already read one line earlier by `Hive::Gh.pr_frontmatter`) and re-implements the `\A---…---` delimiter regex `Hive::Gh` already owns; the format is now encoded in two modules and the file is read twice. A `Hive::Gh.pr_body` helper would centralize it.
- [ ] `ShipTimes#matching_action` builds `preferred = "hive: 9-done/<slug> <action>"` (`ship_times.rb:947`) but `pr_finalized` commits are written by the **8-finalize** stage, so the exact-match `preferred` branch is dead for the primary action (the `end_with?` fallback still resolves it correctly). Either match the real `8-finalize` subject or drop the dead tie-break.
- [ ] Categorizer duplicate model ids silently last-write-wins (`categorizer.rb` `map_document`, `memo[id] = row unless id.empty?`): if the model emits two rows for the same `id`, the second overwrites the first with no warning (only missing/invalid ids warn). Combined with the title-injection finding this is attacker-steerable; log on duplicate id.
- [ ] Controller failure maps leak per digest date: a digest slug is a unique ISO date, so any non-success/non-CONFIG exit writes `@transient_failures`/`@cooldown_until` keyed by `["digest", "<date>"]` (`concurrency_controller.rb`), which the digest gate never consults and `prune_dispatch_baselines` never prunes → slow unbounded (failure-only) growth over the daemon's lifetime. Patrol avoids this via a fixed `"patrol"` slug.
- [ ] A digest subprocess `ConfigError` exit (78) tags a phantom `"digest"` project as dropped (`dispatcher.rb:~560` reap path → `@dropped_projects`), emitting a misleading `:project_dropped project=digest` event and a persistent phantom entry, though no real dispatch is blocked.
- [ ] Categorizer `ModelError` exits 0 and advances the cursor (`lib/hive/digest.rb:303-309`), so a *transient* LLM/agent outage permanently skips that day's real digest (only a failure notice is sent). This is the plan's deliberate "exit 0 on handled failure" design; consider a bounded retry that distinguishes transient agent failures from deterministic config failures.
- [ ] Test coverage gaps (all narrow error-path/contract omissions; happy paths are well covered): `can_dispatch_digest?` and the `kind: :digest` task-cap exclusion (`concurrency_controller.rb`) are untested; `DigestScheduler#complete` cursor-regression guard (`last >= local_date`), `reconfigure`, and `cancel` have no unit tests; the `Data` boundary-invariant raises (`Result`/`CategorizedItem`/`SendResult`) are untested; the `dispatch_digest` spawn-error rescue (asymmetric with the covered patrol path) is untested.
- [ ] CLI error ergonomics under `--json` (`lib/hive/commands/digest.rb`, `lib/hive/digest/sender.rb`): a bad `--date` raises `ConfigError` → unstructured stderr (digest is absent from `JSON_USAGE_ERROR_CONTRACTS`), and a Telegram send failure re-raises a non-`Hive::Error` → raw stack trace, even when `--json` was requested. Correct for the scheduler (non-zero → retry) but an interactive agent parsing stdout gets no JSON error envelope.

ivankuznetsov:

# PR Review Toolkit — Daily "Shipped Digest" (pass 2)

Scope: `git diff origin/main..HEAD` on `add-a-daily-shipped-digest-260614-6591`.
Read-only; no code edited. Personas run: code-reviewer, pr-test-analyzer,
silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier.
Findings are the union across personas; each is prefixed with the surfacing persona.

## High
- [ ] [code-reviewer] `:digest_failure_backoff` is **not** in the daemon Logger `EVENTS` allowlist (`lib/hive/daemon/digest_scheduler.rb:105` emits it; `lib/hive/daemon/logger.rb` lists only `digest_catchup_skipped`): `Logger#event` raises `ArgumentError` on any unknown name, and `record_failure` runs from `complete`/`reap_completed` (and `dispatch_digest`'s rescue) which have **no** surrounding rescue — so the *first* non-zero digest exit (ConfigError/exit 78, Telegram failure, spawn error) crashes the whole daemon tick, and repeated crashes trip systemd `StartLimitBurst`. Verified and reproduced; tests miss it because the scheduler tests use a permissive StubLogger. Fix: add `digest_failure_backoff` to `EVENTS`.

## Medium
- [ ] [silent-failure] `Collector#build_item` rescues only `Hive::GitError`, but the `Hive::Gh.pr_frontmatter` call on `lib/hive/digest/collector.rb:649` can raise `SystemCallError`/`IOError` (TOCTOU on `pr.md`: `File.exist?` then `File.read`, with concurrent archive/drop or a permission flip). That escapes `build_item`, `Digest.run` (catches only `ModelError`), and `bin/hive` (`rescue Hive::Error`) → one unreadable `pr.md` in any project kills the whole day's digest with a non-zero exit (which currently also triggers the High above). Inconsistent with the very next line's `pr_body`, which *does* rescue these; broaden the rescue.
- [ ] [silent-failure] `DigestScheduler#write_state` is unrescued (`digest_scheduler.rb:230-237`): a `SystemCallError` (disk full, RO mount, bad perms) from the first-run seed, catch-up-cap, or `complete`-cursor-advance propagates out of `reap_completed`/`tick` (no rescue) and crashes the entire daemon poll loop — not just the digest. Other schedulers' faults are isolated; this one is not.
- [ ] [silent-failure] `DigestScheduler#read_state` rescues `JSON::ParserError`/`SystemCallError` to `{}` with **no log** (`digest_scheduler.rb:221-228`): a corrupt/truncated `digest_state.json` makes `tick` treat it as first-run and silently reset the cursor forward to yesterday, permanently skipping every owed un-digested day with zero operator signal. Emit a distinct `:digest_state_unreadable` event before returning `{}`.
- [ ] [simplifier] `lib/hive/digest/renderer.rb:3` `require "hive/digest/categorizer"` is unused — the renderer references only `Categories::ORDERED` (already required line 2) and drags `hive/agent`/`hive/agent_profiles`/`hive/stages/base` onto the render path; delete it.
- [ ] [comment-analyzer] `config/brakeman.ignore` note (fingerprint `a11d1f57…`) misstates the argv shape — "slug is interpolated only into the `git --grep=<slug>` argv token" — whereas `ship_times.rb:921-924` passes `"-F", "--grep", slug` as **separate** argv elements, directly contradicting the sibling `ship_times.rb` comment. Fix the security-rationale note to match the code (standalone `--grep` argv element).
- [ ] [comment-analyzer] Wiki references a non-existent `Config.load_global_digest` (the real scheduler-facing loader is `load_global_digest_block`) in 4 places: `wiki/commands/daemon.md:64`, `wiki/modules/config.md:145` and `:164`, `wiki/log.d/20260614T222807Z-digest-command-surface-refresh.md:7`.
- [ ] [comment-analyzer] Wiki mis-describes the categorizer id as just `pr_number`/slug, omitting the mandatory `project_name/` prefix that `ShippedItem#categorizer_id` adds for cross-project collision-safety (`wiki/modules/digest.md:68`, `wiki/commands/digest.md:49`).
- [ ] [pr-test-analyzer] No test asserts the daemon actually calls `DigestScheduler#reconfigure(enabled:, max_catchup_days:)` on SIGHUP/config reload — `FakeDigestScheduler#reconfigure` (`test/unit/daemon/dispatcher_test.rb`) is a no-op, so a wiring regression leaving the digest stuck enabled/disabled after an operator edit passes all tests.
- [ ] [pr-test-analyzer] The plan requires "unknown category defaults **and warns**", but `test/unit/digest/categorizer_test.rb:24,48` pass `logger: nil`, so the warn side-effect — the only operator signal the model misbehaved — is never asserted; a regression dropping the `log_warning` calls would not be caught.

## Nit
- [ ] [type-design] `ShippedItem` has no construction guard for its de-facto primary key `categorizer_id` — a blank `project_name`/`pr_number`/`slug` yields `"/"` and collides all blank items into one map key in `map_document`, the exact collision the prefix comment claims to prevent; latent only because `collector.rb` fills every field (`shipped_item.rb:3,18`).
- [ ] [type-design] `CategorizedItem`'s constructor guard is unreachable from production — the sole caller `categorized_item` pre-coerces both `category` and `summary` immediately before `new` (`categorizer.rb:502-504`), duplicating the validity logic in the type and the caller where they can drift.
- [ ] [type-design] `Result#delivery` (and transitively `SendResult#responses`/`#text`) is populated but never read in production — every `Commands::Digest` consumer reads only `status`/`date`/`message`; dead/write-only state (`digest.rb:266`, `sender.rb:804`). Also flagged by simplifier.
- [ ] [type-design/simplifier] `Categorizer::RunnerTask#project_root` is set to `folder` but never read by `Hive::Agent` — dead field copied from the real task shape (`categorizer.rb:396,553`).
- [ ] [simplifier] `Collector#@clock` is injected (`collector.rb:617,621`) but never read anywhere — a dead seam implying a time-dependency the collector does not have.
- [ ] [type-design] `CategorizedItem#category` is a guarded **string** enum while sibling `Result#status` is a guarded **symbol** enum — two new sibling enums with inconsistent representation.
- [ ] [simplifier] pr.md frontmatter-strip regex is duplicated in `Collector#pr_body` (`collector.rb:670-684`) and `Hive::Gh.pr_frontmatter`, and `pr.md` is read twice; a `Hive::Gh.pr_body` helper would define the boundary once.
- [ ] [simplifier] The "local day that just ended" default is computed in both `Commands::Digest#parse_date` (`commands/digest.rb:32`) and `Digest.run` (`digest.rb:283`); the run-side default and its `clock:` param are test-only since the command always passes a concrete Date.
- [ ] [simplifier] `Commands::Daemon#load_digest_config` (`commands/daemon.rb:247-249`) is a single-use wrapper over `Config.load_global_digest_block`; the SIGHUP path calls the config method directly, so inline for symmetry.
- [ ] [simplifier] Categorizer's class-method/instance split threads `logger:` through 5+ class methods that each re-default `Logger.new($stderr)`, where the instance already holds `@logger`; weigh the duplication against the direct-call test seam.
- [ ] [silent-failure] `categorized_item` uses `logger&.warn`, so a `nil` logger silently swallows every "model omitted/miscategorized item — using default" signal, letting a half-bad model produce a plausible digest with no trace; fall back to `$stderr` so the signal can't be configured away (`categorizer.rb:491-517`).
- [ ] [silent-failure] `dispatch_digest`'s rescue can call `complete` a second time if `dispatch_command` raises *after* the child is spawned+recorded; `complete` is otherwise idempotent but `record_failure` would double-increment the backoff count for one logical failure (`dispatcher.rb`).
- [ ] [pr-test-analyzer] The backoff 900s ceiling / clamp-at-last-tier path (`[count-1, size-1].min`, `digest_scheduler.rb:102`) is untested — only the `[60, 300]` tiers are exercised; a third consecutive failure's cap is unverified.
- [ ] [pr-test-analyzer] No live e2e covers the `ModelError → failed_notice` real Telegram send or the `enabled: false` short-circuit; the failed-notice MarkdownV2 path (the exact class of bug the escaping exists to prevent) is only unit-tested with a fake sender.
- [ ] [pr-test-analyzer] `test/unit/digest/sender_test.rb:86` pushes both the factory-capture hash and the send hash onto one `sent` array and asserts via `first`/`last` — brittle to ordering, conflates two concerns.
- [ ] [comment-analyzer] "one/single Telegram message" (wiki/modules/digest.md, wiki/commands/digest.md, `cli.rb` digest `long_desc`, `templates/hive_config.yml.erb`) understates that `send_in_chunks` issues one `send_message` per chunk for long digests; add a "(chunked above Telegram's length limit)" parenthetical.
- [ ] [comment-analyzer] Wiki "`--json` output is success-only" (`wiki/commands/digest.md`) is imprecise — a `failed_notice` still prints a JSON document with `ok:false`; it is "success-only" only in lacking a separate JSON *error envelope* (hard failures use stderr + exit code).

ivankuznetsov:

## High
- [ ] `lib/hive/digest/collector.rb:78` truncates PR bodies before the prompt: the plan requires the model to receive the full PR body, but long bodies are silently capped at 8,000 chars.
- [ ] `lib/hive/config.rb:1979` does not validate `budget_usd.digest` or `timeout_sec.digest`: malformed values pass config load and can crash the categorizer instead of producing a handled digest failure.
- [ ] `lib/hive/daemon/dispatcher.rb:1101` only uses in-memory state to suppress duplicate digest dispatches: after a daemon crash/restart an orphaned digest child is invisible and the same date can be sent again.
- [ ] `lib/hive/daemon/digest_scheduler.rb:69` clears `@pending` before the cursor is durably written: a write failure after a successful send makes the same date immediately eligible for duplicate delivery.

## Medium
- [ ] `test/digest/e2e_test.rb:40` does not prove the live model returned rows for every item: an empty `items` response still passes via categorizer defaults while the plan requires e2e coverage of every fixture item.
- [ ] `lib/hive/daemon/dispatcher.rb:696` never completes digest dry-run pseudo-children: the scheduler's pending date remains set forever after the first daemon dry-run dispatch.
- [ ] `lib/hive/digest/renderer.rb:70` bounds only summaries, not labels or PR URLs: an overlong display name/URL can still produce a >4096-char MarkdownV2 link line that Telegram chunking splits mid-syntax.
- [ ] `templates/digest_prompt.md.erb:30` leaves display names outside the untrusted-data tags: a crafted `meta.yml` name can inject instructions into the categorizer prompt.
- [ ] `templates/digest_prompt.md.erb:38` sends PR bodies through the Claude argv/log path: private PR text can be exposed in process listings and digest run logs instead of being passed via stdin/redacted logging.
- [ ] `lib/hive/daemon/digest_scheduler.rb:52` treats unreadable or wrong-shaped state as first run: it overwrites the cursor with yesterday and can permanently skip owed catch-up days instead of failing closed.

## Nit

ivankuznetsov:

# Code Review — Daily "Shipped Digest" (pass 3)

Scope: `git diff origin/main..HEAD` on `add-a-daily-shipped-digest-260614-6591`.
Mode: report-only (no code edited). Reviewers: correctness, security, adversarial,
reliability, testing, maintainability, cli-readiness.

Intent: a daemon-scheduled global daily job dispatches a `hive digest` subprocess
that collects tasks shipped on a local calendar day across all registered projects
(on-disk `9-done/` + `hive/state` git-log ship-times), runs an LLM categorizer,
renders one Telegram MarkdownV2 message, and sends it; empty days send a "nothing
shipped" notice and LLM failures send a "digest failed" notice.

Requirements completeness: the plan's Requirements Trace and IU1–IU6 scenarios all
appear implemented. The only gap against the literal plan wording — "full PR body"
vs. the 8,000-char cap (`collector.rb:20`) — is a **deliberate pass-2 hardening**
(documented; preserves the full body for realistic PRs) and is not re-raised here.

Passes 1–2 already fixed ~32 findings (title+body nonce-wrapping, summary length
cap, schema_version envelope, nil-exit-as-failure, tick/complete rescue guards,
digest pseudo-project map exclusion, run-dir pruning, etc.); those are not re-listed.
The duplicate-on-restart / write-failure-retry windows that the peer codex-03 review
re-raises as High are the plan's **accepted at-least-once delivery model** (resolved
NO-FIX in pass 2: DigestScheduler deliberately mirrors PatrolScheduler's in-memory
`@pending`); they are not re-escalated here. The findings below are issues that
survive into the **current** diff.

## High
- [ ] Categorizer agent runs claude with `bypassPermissions` and no tool allowlist over aggregated, attacker-influenceable cross-project PR bodies (`lib/hive/digest/categorizer.rb:239`, default `claude.permission_mode` = `bypassPermissions` at `lib/hive/config.rb:17`): the per-spawn nonce wrapper added in pass 2 is an in-band heuristic, not a trust boundary — a successful injection in any one project's `pr.md` (the digest is the first surface to aggregate untrusted PR text from **every** registered project into one prompt) reaches a full-access agent (Bash/Read/Write/WebFetch via `--dangerously-skip-permissions`), so it can read secrets / other projects' private state or exfiltrate. The categorizer's only legitimate action is writing one `items.json`; give the digest role a least-privilege tool allowlist (a permission-mode change won't do — `plan` blocks the required Write, and `default` hangs a non-interactive child) so a successful injection has nothing to act on. This is the residual the prior passes' wrapping was meant to mitigate but cannot close, and it needs a permission-policy decision.

## Medium
- [ ] Hung Telegram send wedges the single global digest slot and silently disables ALL future digests (`lib/hive/digest/sender.rb:70`→`lib/hive/bot/telegram.rb:88`, `lib/hive/daemon/concurrency_controller.rb` `can_dispatch_digest?`): the Telegram client is `::Telegram::Bot::Client.new(token)` with no read/open timeout, and `send_message` is **not** inside the `Faraday::TimeoutError` rescue (only `poll_updates` is), so a black-holed socket blocks the subprocess forever. With `daemon.child_timeout_sec` defaulting to `0` (no cap, `config.rb:300`) and no digest verb-timeout, the child is never reaped, `DigestScheduler#complete` never runs, `@pending[date]` never clears, and `can_dispatch_digest?` returns `:digest_in_flight` permanently → no digest ever dispatches again until daemon restart. This is a permanent *wedge*, distinct from the accepted at-least-once *duplicate* tradeoff. (The pre-agent `git log` calls in `ship_times.rb:24` are the other unbounded leg.) Add an explicit Telegram read/open timeout on the digest send path, or seed a non-zero default `digest` verb timeout so a wedged child is reaped and the date retries via backoff.
- [ ] The one real-model e2e does not prove the live model categorized any item (`test/digest/e2e_test.rb:40-41`): both assertions check `display_name` substrings, which render as the link label `[display_label](url)` regardless of model output — and an empty `{"items":[]}` response still yields `status: :sent` with default (fallback) summaries. So the test passes even if the live agent returns nothing, defeating IU6's stated purpose ("categorized output covers every fixture item") and the project's "does this test actually test anything" rule. Assert that at least one rendered summary differs from the `pr_title`/`display_name` fallback (i.e. the model actually contributed), and ideally that both fixture ids are present in the model's `items.json`.
- [ ] `reap_dry_run` omits the digest completion hook that `reap_completed` has (`lib/hive/daemon/dispatcher.rb:696` vs. the digest-stage branch at ~`:583`): a dry-run daemon dispatches the digest pseudo-child, frees the controller slot on exit, but never calls `@digest_scheduler&.complete(...)`, so the scheduler's `@pending` marker never clears and `tick` returns `[]` forever (`@pending.any?`). The dry-run daemon dispatches the digest once and is then wedged (no further digests, no catch-up progress). Add the same digest-stage `complete` hook (wrapped in the sibling `rescue StandardError`) to `reap_dry_run`.
- [ ] PR URL / Display name / Project are interpolated into the categorizer prompt OUTSIDE the per-spawn nonce wrapper (`templates/digest_prompt.md.erb:28-31`): only `pr_title` and `pr_body` are fenced. `pr_url` comes from on-disk `pr.md` frontmatter (attacker-influenceable, same provenance as the body), so it is an un-fenced injection slot — inconsistent with the pass-2 fix's own stated principle ("treat every per-item field as data"). Wrap these fields in the same nonce tag, or drop `pr_url` from the prompt entirely (the categorizer never uses it; the renderer re-reads it for the link).
- [ ] `hive digest --json` emits no JSON error envelope on a bad `--date` (`lib/hive/commands/digest.rb:25-39`): the `--date` format check raises `Hive::ConfigError`, caught by `bin/hive`'s `rescue Hive::Error` (`bin/hive:174`) which prints bare stderr for every command; the `JSON_USAGE_ERROR_CONTRACTS["digest"]` registration only fires on the `rescue Thor::Error` path (unknown flags / malformed `--json`), which a string-typed `--date` value never triggers. ~10 sibling commands (run, approve, markers, findings, stage_action, …) emit an in-command JSON error envelope before re-raising; digest does not, so an agent parsing `--json` stdout gets nothing on the most common usage error. Add the in-command `rescue Hive::Error → emit envelope if @json → raise` pattern. (Low operational impact: the scheduler always passes a valid ISO date; exit code + stderr stay correct.)
- [ ] Aggregated, possibly-private PR bodies from all registered projects are passed to claude as an argv element, not stdin (`lib/hive/agent.rb` claude `prompt_via_stdin?` is false), exposing them in `/proc/<pid>/cmdline` while the categorizer runs. The prompt-passing mechanism is **pre-existing** (affects every claude agent) and the fix lives in `Hive::Agent`, not digest code — but the digest is the first surface to inline up-to-8,000-char private bodies from *every* project into one argv, a real data-sensitivity increase. Low risk on a single-user daemon host; route the digest/claude prompt via stdin (as codex already does) if hardened.

## Nit
- [ ] `Hive::Digest::Sender.blank?` is a public class method with no external callers (`lib/hive/digest/sender.rb:55`); it is an implementation detail of `resolve_chat_id`. Make it `private_class_method` or inline it.
- [ ] `ConcurrencyController#can_dispatch_digest?(now:)` accepts a `now:` keyword it never reads (`lib/hive/daemon/concurrency_controller.rb`, ~`:131`) — a dead seam implying time-dependence the method does not have. Drop it (note: `can_dispatch_patrol_scan?` carries the same pre-existing unused `now:`).
- [ ] `hive digest` `long_desc` (`lib/hive/cli.rb:494`) omits an exit-code table and `--json`/`--dry-run` examples, and the `--json` envelope drops the already-resolved `chat_id` from `SendResult`. Adding an exit-code line, the schema name, and `chat_id` (or a `message_length`/`message_sha256`) would improve agent discoverability and post-send auditability.

ivankuznetsov:

# PR Review Toolkit — Daily "Shipped Digest" (pass 3)

Scope: `git diff origin/main..HEAD` on `add-a-daily-shipped-digest-260614-6591`.
Mode: report-only (no code edited). Personas: code-reviewer, pr-test-analyzer,
silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier.

Union of all persona findings below. This is pass 3 (two prior fix passes); each
finding was verified against the current code. Personas confirmed the bulk of the
feature is clean and unusually well-hardened — DST/window tests pin the timezone,
the unit suite avoids no-op-passing assertions, the new `Data` types enforce their
invariants at the constructor boundary, and nearly every dense comment checks out.
The duplicate-on-restart delivery window (plan's accepted at-least-once model) and
the deliberate 8,000-char body cap are not re-raised.

## High
- [ ] [pr-test-analyzer] The one real-model e2e cannot prove the live model categorized/summarized any item (`test/digest/e2e_test.rb:38-42`): its only content assertions check `display_name` substrings, which `Renderer#render_line` emits via `item.display_label` (`renderer.rb:68`) independent of model output — and the categorizer's documented fallback (`categorizer.rb:151-160`, proven by `categorizer_test.rb:60-71` on an empty `{"items":[]}`) renders defaulted lines. So an empty/garbage live `items` response still passes all three assertions, leaving IU6 ("categorized output covers every fixture item") unverified and violating the project's "does this test actually test anything" rule. Assert against the model-produced summary text (or that each fixture id appears with a non-fallback category/summary).

## Medium
- [ ] [code-reviewer] `hive digest --json` emits no JSON error envelope on a bad `--date` or missing chat config (`lib/hive/commands/digest.rb:13` `call`): `#call` has no `rescue`+`emit_error`, so `Hive::ConfigError` propagates to `bin/hive`'s `rescue Hive::Error` (only `Thor::Error` routes through `emit_json_usage_error`) — verified live, `hive digest --date not-a-date --json` → empty stdout, exit 78. The cited model command `patrol` *does* emit these via an internal `rescue Hive::Error/StandardError → emit_error` (`commands/patrol.rb:36-43,177-188`), and `schemas/hive-digest.v1.json`'s `ErrorPayload` declares `error_kind ∈ {usage,config,internal}` (exit 78/70) where only `usage` is reachable. Mirror patrol's in-command rescue, or narrow the schema to drop the unreachable `config`/`internal` kinds.
- [ ] [pr-test-analyzer] `reap_dry_run` is missing the digest-stage `@digest_scheduler.complete(...)` branch that `reap_completed` has (`lib/hive/daemon/dispatcher.rb:696-706` vs. `:583-594`), and the path is untested (the only dry-run reap test, `dispatcher_test.rb:1761`, uses stage `"3-plan"`). Under `daemon --dry-run`, a digest pseudo-child is reaped without ever calling `complete`, so `@pending` never clears (`digest_scheduler.rb:43`) and the dry-run digest dispatches once then wedges permanently. Add the digest-stage hook to `reap_dry_run` and a `stage: "digest"` dry-run reap test.
- [ ] [silent-failure-hunter] `Digest.run` rescues only `ModelError` (`lib/hive/digest.rb:51`), so non-`ModelError` failures bypass the "digest failed" notice and degrade to a silent (no-Telegram) backoff-retry loop with only stderr/daemon-event signal: a missing/misconfigured `digest.agent` binary raises `Errno::ENOENT` from `Process.spawn` (verified — `Agent#run!`/`spawn_and_wait` do not rescue it; only `Errno::ESRCH` is caught), a scratch-dir disk fault raises from `run_dir` `mkdir_p` (`categorizer.rb:189-197`), and a registry/`Dir` glob error raises from `collector.for_date` (outside the per-task `build_item` rescue). A persistent agent-config typo therefore silently never delivers and never notifies the user. Confirm the model-failure-only notice scope is intended for a persistently broken agent config, or widen to send a failed-notice on agent-spawn failure.

## Nit
- [ ] [code-reviewer] `@daily_counts[["digest", <date>]]` is incremented unconditionally in `record_dispatch` (`concurrency_controller.rb:202`) but never refunded — `record_completion` early-returns for `kind == :digest` (`:258`) before `dec_daily` (`:278`), and nothing else prunes it → one never-read entry leaks per calendar day (the pass-2 early-return fix introduced this by skipping the refund). Skip the increment for `:digest` in `record_dispatch` for symmetry.
- [ ] [code-simplifier] Dead constant alias `VALID_CATEGORIES = Categories::VALID` (`lib/hive/digest/categorizer.rb:33`) has zero references in lib or test (`Categories.valid?`/`DEFAULT_CATEGORY` are used directly); delete it.
- [ ] [code-simplifier] Unreachable guard `return [] if owed.empty?` after `apply_catchup_cap` (`lib/hive/daemon/digest_scheduler.rb:60`): `owed` is non-empty from `:57`, and `apply_catchup_cap` returns either `owed` unchanged or a non-empty `owed.last(max>0)`. Drop it, or keep as documented belt-and-suspenders.
- [ ] [type-design-analyzer] `ShippedItem#display_label` (`shipped_item.rb:40`) uses `!display_name.to_s.empty?` (no `strip`) while the rest of the pipeline uses strip-based blank checks, so a whitespace-only `display_name` renders as a visible blank label instead of falling back to `slug`. Use `display_name.to_s.strip.empty?` for parity.
- [ ] [type-design-analyzer] `SendResult` dry-run invariant is one-directional (`sender.rb:12` guards `dry_run && !chat_id.nil?` but not the paired `!dry_run && chat_id.nil?`); add the symmetric guard so "a real send always has a recipient" is structural, not incidental.
- [ ] [type-design-analyzer] `Result#delivery` (`digest.rb:14`) and `SendResult#responses`/`#text` (`sender.rb:9`) are read only by tests (e2e message-id assertion / one unit assert), not production — justified for the plan's "assert the send returns a message id," but add a one-line "exists for test/observability" comment so their absence from `json_payload` doesn't read as a bug.
- [ ] [comment-analyzer] `sender.rb:62-69` "at-least-once across restarts" understates the window: after a mid-stream failure the *same* daemon re-owes and re-dispatches the date once the backoff expires, so duplication happens on any retry, not only across restarts. Reword to "across retries (in-process or after a restart)".
- [ ] [comment-analyzer] `renderer.rb:9-13` "the chunker cuts on that boundary" reads as an unconditional hard cut, but `Telegram#split_message` prefers the last newline in the 4096 window and hard-cuts only as a fallback (the 600-char defense is still sound); reword.
- [ ] [comment-analyzer] `categorizer.rb` `RUN_DIR_RETENTION` "keep at most this many" is transiently off-by-one — `prune_old_runs` runs before the new dir is created, so there are `RETENTION + 1` dirs until the next run. Say "roughly the most recent".
- [ ] [silent-failure-hunter] Model-omitted/invalid items render as plausible default lines (`categorizer.rb:144-161`) indistinguishable from real ones in the delivered Telegram digest, with operator signal only via a stderr `warn`. This is plan-sanctioned (IU2: "default category + verbatim title summary (logged)"), but a half-bad model run ships a partly-fabricated digest with no in-message marker — consider a visible marker or a structured daemon event when any item is defaulted.
- [ ] [code-simplifier] `Hive::Digest.empty_grouped?`/`render_digest` (`digest.rb:59,63`) and `Categorizer#flat_items` (`categorizer.rb:185`) are single-use public/helper hops; marking the module functions `private_class_method` (or inlining `flat_items`) would shrink the public surface to just `run`. Altitude nit — they read fine as-is.

PR #479, opened 2026-06-15
Hive patrol: Dry-run skip log follows symlinks

ivankuznetsov:
## Summary

Hardens how the babysitter dry-run `git` / `gh` stubs (`bin/hive-babysitter-stub-git`, `bin/hive-babysitter-stub-gh`) write their skipped-command audit log.

Previously `log_skip` opened the configured `HIVE_BABYSITTER_DRY_RUN_LOG` (or the default `.babysitter-dry-run-skipped.log`) with a plain append, which follows symlinks — so a pre-planted symlink could redirect the write to an attacker-chosen file under the babysitter's uid. Both stubs now:

- Open the log with `File::NOFOLLOW` and create new logs as mode `0600`.
- Require the opened target to be a regular file owned by the current uid, and `warn` (without un-skipping the command) when the log cannot be written — so dry-run stays default-deny even when the audit sink is unsafe or unavailable.
- Escape ASCII control characters in argv as `\xHH` before writing to the log or stderr, so a skipped argument carrying a newline cannot forge extra audit lines.

A follow-up fix makes `escape_control_chars` binary-encode argv (`arg.to_s.b`) before the byte-class `gsub`, so non-UTF-8 argv (git paths, agent-supplied messages) no longer raises `ArgumentError` inside `log_skip` and silently drops the audit record.

## Test plan

`test/unit/babysitter/dry_run_env_test.rb` adds two regressions:

- `test_stubs_refuse_symlinked_skip_log` — a symlinked log target is refused, the real file is left untouched, and both stubs emit the `failed to write skip log` warning while still exiting successfully.
- `test_stubs_escape_control_characters_in_skip_log` — a newline in argv is rendered as `\x0A` in both the skip log and stderr, and the raw newline never reaches either sink.

Verified locally:

```
bundle exec ruby -Itest test/unit/babysitter/dry_run_env_test.rb
# 10 runs, 788 assertions, 0 failures, 0 errors, 0 skips
```

## Review summary

Two `codex-native-review` passes ran. The only substantive finding was a P2 in pass 01: the new `escape_control_chars` `gsub(/[\x00-\x1f\x7f]/)` raised `ArgumentError: invalid byte sequence in UTF-8` on non-UTF-8 argv, inside `log_skip` ahead of its `rescue`, so a skipped command exited 1 with no audit-log entry. Auto-fixed by binary-encoding argv before the `gsub` (commit `2fa53c74`); re-running the dry-run env suite confirmed the fix. Pass 02 returned no actionable findings. The remaining reviewer lines were unfilled template scaffold and were resolved as no-fix.

## Linked task

`/home/asterio/Dev/hive/.hive-state/stages/6-review/patrol-command-bin-hive-babysitter-stub-git-16817f87`


ivankuznetsov:

Outcome: timeout

Agent tail:

```
The final GitHub `rake test` job is still pending; local `bundle exec rake test` already passed, but I’ll keep polling while the hosted check is active.
```

Issue #12, opened 2026-04-27
Code-review residual: 17 P3 polish items from ce-code-review run 20260427-110800-0e135bb6


ivankuznetsov:
## Summary

The ce-code-review run \`20260427-110800-0e135bb6\` on \`feat/hive-tui\` surfaced 32 findings. The top 15 were validated and applied / deferred individually (see #(F1) and #(F9)). The remaining 17 P3 items below didn't make the validation budget but are real and worth a polish pass.

### Correctness (P3)
- **#18 \`LogTail::Tail#open!\` discards leading whole line on backbuffer-newline alignment.** When seek-to-tail boundary lands exactly on a \`\\n\`, the discard-leading-partial-line logic drops the first whole line of the visible window. \`lib/hive/tui/log_tail.rb:65-78\`.
- **#19 \`LogTail::Tail#ingest\` never flushes a final partial line.** Logs without a trailing \`\\n\` render as empty. \`lib/hive/tui/log_tail.rb:144-180\`.

### Maintainability (P3)
- **#20 \`TriageState#develop_command\` is dead code** (only its own test calls it). Either delete or have KeyMap delegate. \`lib/hive/tui/triage_state.rb:57-59\`.
- **#21 \`Help::BINDINGS\` Array is frozen but inner Hashes are not.** Use \`.each(&:freeze).freeze\` or \`Data.define\`. \`lib/hive/tui/help.rb\`.
- **#22 \`translate_key\` lives on \`Hive::Tui\` but belongs on \`Hive::Tui::KeyMap\`** — KeyMap owns the symbol contract; it should own the curses-int → symbol translation.

### API contract (P3)
- **#23 \`KeyMap\` public constants** (\`VERB_KEYS\`, \`ENTER_KEYS\`, \`ESCAPE_KEYS\`, \`DOWN_KEYS\`, \`UP_KEYS\`, \`ENTER_FLASH_MESSAGES\`) lack stability annotation. Document as \`@api private\` or move into a nested private namespace.
- **#24 \`Snapshot::Row.action_key\` reads from JSON's \`"action"\` key** — silent rename at the boundary. Add a \`schema_files_test\` assertion pinning the 1:1 correspondence between \`Status#task_payload\` keys and \`Snapshot::Row\` field set.

### Reliability (P3)
- **#25 SIGHUP trap never restored after \`Hive::Tui.run\` exits normally.** Restore in the \`ensure\` of \`run_loop\`.
[...]

PR #491, opened 2026-06-15
feat: show PR numbers across all task-list surfaces

ivankuznetsov:
## Summary

A task's pull-request number was invisible in every human-facing task list. To find which PR a task had produced you had to open `pr.md` or GitHub. Now the TUI tasks pane, `hive status` text output, and the Telegram `/status` and `/queue` lists all show the PR number the moment a PR exists, rendered as a clickable link where the surface supports it and as a dash `—` when there is no PR yet. The existing "Ready for review" Telegram push also carries the link.

The number renders after the `id` in each surface, so a `6-review` task reads `#561` and a `2-brainstorm` task reads `—`.

### How the number flows

One data source feeds four surfaces through a new `pr_url` field on the `hive status --json` envelope:

| Surface | Rendering | Clickable |
|---|---|---|
| TUI tasks pane | fixed-width `PR` column after `id` | OSC 8 hyperlink (tty only) |
| `hive status` text | `#NNN` field left of the name | OSC 8 hyperlink (tty only) |
| Telegram `/status` & `/queue` | `title — #561 — Review` | HTML `<a>` link |
| Telegram review push | `PR: #561` line appended | HTML `<a>` link |

`Hive::Pr.number(url)` is the single authority for the displayed `#NNN` token, so every surface agrees.

### Design decisions

**Source from `pr.md` frontmatter, not `row.attrs`.** The `pr_url` marker attribute only exists on the current state file at stages `5-open-pr` and `8-finalize`. At `6-review`, `7-artifacts`, and `9-done` the marker lives in a different state file with no `pr_url`, so reading `row.attrs["pr_url"]` would show a dash for the very stages the feature targets. Reading `pr.md` frontmatter through the existing error-tolerant `Hive::Gh.pr_frontmatter` helper makes the value stable across stages 5 through 9, and a mid-write or malformed `pr.md` degrades to a dash rather than raising. The per-row read is guarded by `stage_index >= 5`, so stages 1 through 4 never touch the file.

**A dedicated `pr_url` field, not another `attrs` key.** The Telegram notification de-dup fingerprint hashes `attrs`. Adding `pr_url` as a sibling field keeps the fingerprint byte-identical, so the enriched review push still fires exactly once and is never re-sent. The field is added to `schemas/hive-status.v3.json` as a required property (the task object is `additionalProperties: false`) with no `SCHEMA_VERSIONS` bump, matching the additive precedent of `live_task_lock`.

**HTML over MarkdownV2 for Telegram links.** Switching the `/status` reply and the review push to a parse mode means every dynamic token must be escaped or Telegram rejects the whole message. HTML has a 3-character escape surface (`& < >`) versus MarkdownV2's 18, so it is the lower-risk default; titles and stage labels pass through `Hive::Bot::Format.html_escape`, and the link href is escaped via `CGI.escapeHTML`. The injection gates (`Hive::Pr.valid_http_url?` for both OSC 8 and HTML links, plus `Tui::Text.sanitize`) reject non-`http(s)` schemes so a hand-edited `pr.md` cannot inject escapes.

**The TUI PR column never drops.** It is a fixed-width column subtracted in every `compute_layout` branch, including the narrowest, so the elastic `name` column absorbs the change and the PR number survives on narrow terminals. Column width math runs on the plain `#NNN` text; the OSC 8 wrapper (zero display width) is applied only to the visible token after padding.

## Test plan

- `bundle exec rake test` — full suite green (rubocop clean on all changed `.rb` files).
- New unit coverage for `Hive::Pr.number` (URL variants, trailing slash, query/fragment suffix, blank and non-PR input → nil) and the new `Hive::Pr.valid_http_url?` gate.
- `Hive::Tui::Views::Hyperlink` (`osc8`/`splice`): wraps on enabled + valid URL, passes through when disabled/blank/non-http, strips control chars, and a non-tty render emits zero OSC 8 bytes.
- Telegram `Hive::Bot::Format` HTML escaper and `html_pr_link` builder, including attribute-breakout and special-char title escaping; legacy queue line escaping.
- Per-surface rendering and `compute_layout` width math: PR column present after `id`, `#NNN` vs `—`, and the column never zeroes across every narrow-width branch.
- TTY/OSC 8 path for `hive status` text pinned so the identity column stays aligned when escape bytes are spliced in (the alignment bug a TTY would otherwise hit).
- Extended the status JSON schema-conformance test (populated task with `pr_url` string and null) and the notification-exactly-once test (`fingerprint` byte-identical with/without `pr_url`).

Manual dogfood still outstanding (tracked in `wiki/gaps.md`): a live-TTY frame confirming lipgloss passes the OSC 8 sequence through intact, and a live Telegram `/status` + review-push round-trip.

## Review summary

Three review passes (compound-engineering code-review, pr-review-toolkit multi-persona, and Codex) ran on the branch's actual contribution (`git diff origin/main...HEAD`). All findings were triaged with zero escalations to the user — every item was auto-fixed or resolved from the plan/prior decisions.

- **No High-severity defects remain.** All plan requirements (A1–A10 + acceptance cases) are implemented; injection gates and the A10 once-only push invariant hold under adversarial probing.
- **Consensus High fixed:** `hive status` text columns misaligned in a TTY for PR rows because `ljust` counted the spliced OSC 8 escape bytes — fixed by padding the plain token first, then wrapping (mirroring the TUI's `rjust_cells` discipline).
- **Cleanup applied:** removed dead code and dead `respond_to?(:pr_url)` guards, hoisted the duplicated `valid_http_url?` to `Hive::Pr`, unified the PR-column width constant and the splice helper, delegated `ReviewHandoff#pr_number` to the canonical parser, derived the stage threshold from `Hive::Stages::DIRS`, enforced the `PARSE_MODES` closed set, and narrowed an over-broad rescue.
- **Resolved as out-of-scope / by-design:** a runtime plain-text fallback for a Telegram parse-mode 400 (escaping is verified complete; the plan accepted the parse-mode blast radius via a shared escaper + tests) and the >4096-char mid-entity split (low-probability dogfood note — real display names are short slugs).
- **Phantom reverts:** the two-dot `origin/main..HEAD` diff shows `agent_limit.rb`, `finalize.rb`, `tmux_runner.rb`, `web/agents_auth.rb`, and `stages/review/triage.rb` as deletions only because the branch is behind `origin/main` (#490/#492/#496/#497). The branch never touched those files (confirmed via the three-dot diff and `git merge-tree`); a 3-way merge preserves main's versions. Rebasing onto `origin/main` before merge is advised so CI runs against latest main and the diff reads clean.

## Linked task

- Hive task: `show-pr-when-pr-created-260606-75a8`
- Branch: `show-pr-when-pr-created-260606-75a8`
- Stage: `8-finalize`

---


ivankuznetsov:

## High
- [ ] `lib/hive/agent_limit.rb:54` reintroduces the broad `limit reached/reset/exceeded` detector: ordinary agent output such as `window limit reached` is again classified as provider exhaustion and can kill healthy runs.
- [ ] `lib/hive/stages/finalize.rb:41` removed the already-merged PR short-circuit: out-of-band merged PRs now fall through to auth/push/finalize work and can strand tasks in `8-finalize` again.

## Medium
- [ ] `lib/hive/commands/status.rb:373` pads the CLI identity after injecting OSC8 bytes: TTY rows with linked PR numbers lose the fixed identity-column spacing, with the same archive-row issue at line 399.

## Nit

ivankuznetsov:

# Code Review — show-pr-when-pr-created (pass 1)

Reviewed `git diff origin/main..HEAD`. Scope: U1–U5 (surface `pr_url`, `Hive::Pr.number`,
TUI/CLI PR columns, Telegram `/status` list link, `ready_for_review` push enrichment).
Build is green: `rubocop` clean on all 10 changed Ruby files; the changed unit + schema +
integration tests pass (487 runs, 0 failures). Plan grounding (read `pr.md` frontmatter, not
`row.attrs`) is correctly implemented; `stage_index < 5` guard correctly includes `5-open-pr`;
fingerprint stays `pr_url`-independent (A10 verified by test). Findings below.

## High

(none — no plan-required Goal/Requirements-Trace item is missing or undeferred; no crash,
data-integrity, or security defect found. URL handling is correctly hardened: both
`Hyperlink.osc8` and `Format.html_pr_link` validate http(s) scheme, strip control chars, and
HTML-escape, so a hand-edited `pr.md` cannot inject terminal escapes or break out of the `href`.)

## Medium

- [ ] `hive status` text columns misalign in a real terminal (tty) for every PR-bearing row — `lib/hive/commands/status.rb:372-373,396-397,417-428`: `display_identity_with_pr(r).ljust(TEXT_IDENTITY_WIDTH)` uses plain `String#ljust`, which counts the OSC 8 escape bytes that `pr_cell` embeds when `$stdout.tty?` is true. Verified empirically: with the link present the identity field pads to 72 *characters* but only 20 *visible* cells (target 49), so the state/command/age columns shift ~29 cells left and rows go ragged (dash rows align, PR rows don't). It contradicts the plan's own U3 technique ("pad the plain token first, then wrap"): here the PR token is wrapped *before* the outer field is padded. Invisible to the test suite because the only coverage runs non-tty (`refute_match(/\e\]8;;/…)`), and `wiki/gaps.md` item 47 confirms no live tty frame was captured. Fix: pad the plain identity to width first, then OSC8-wrap the token (or measure padding on visible width).

## Nit

- [ ] Branch is stale (behind `origin/main`); the diff's apparent reverts of `lib/hive/agent_limit.rb` and `lib/hive/stages/finalize.rb` (+ their tests) are merge-base artifacts, NOT changes this branch makes — `git log fce1a3a3..HEAD` shows the branch never touches those files. Resolve by **rebasing onto `origin/main` before merge** (a standard merge/rebase/squash already keeps main's versions); do **not** "fix" this by editing code. Rebasing also gets CI running against current main.
- [ ] TUI OSC 8 hyperlink is enabled but the OSC8-through-lipgloss path is unverified — `lib/hive/tui/views/tasks_pane.rb:189-196`: the plan required a tty dogfood frame before enabling OSC 8 (lipgloss may strip/mangle the sequence or miscount width), and `wiki/gaps.md` item 47 confirms no live tty frame exists. Tests only assert non-tty structure (lipgloss strips ANSI there). Capture a dogfood frame to confirm the link renders + stays clickable, or fall back to plain `#NNN` (A1 permits it).
- [ ] `ready_for_review` push can permanently lose its PR link on a race — `lib/hive/bot/notification_builders.rb:99-107`: if `pr_url` is momentarily nil when the row first appears (e.g. `pr.md` mid-write), the push fires plain and the fingerprint never changes, so the enriched link is never re-sent (A10 once-only). Low likelihood (pr.md is written at open-pr COMPLETE before the `ready_for_review` transition), but unhandled and silent.
- [ ] Queue line for a PR-less row reads `Title — — — Stage` (separator, em-dash, separator) — `lib/hive/bot/supervisor.rb:1409,1418-1420`: three adjacent em-dashes are visually noisy. Cosmetic; consider distinct spacing or glyph for the empty-PR slot.

ivankuznetsov:

# PR Review Toolkit — show-pr-when-pr-created (pass 1)

Multi-persona review of `git diff fce1a3a3..HEAD` (the real branch changes; the
`origin/main..HEAD` diff also shows phantom reverts of `agent_limit.rb` /
`finalize.rb` that are stale-base artifacts — the branch never touches those
files, so all personas were scoped to the merge-base diff). Personas run:
`[code]` code-reviewer, `[tests]` pr-test-analyzer, `[errors]`
silent-failure-hunter, `[type-design]` type-design-analyzer, `[comments]`
comment-analyzer. Union of findings below; per-persona severities preserved.

## High
- [ ] [code] `hive status` text columns misalign on a real TTY when a PR hyperlink is present (`lib/hive/commands/status.rb:373`, archive path `:397`): `display_identity_with_pr(r)` embeds OSC 8 escape bytes when `$stdout.tty?`, then `String#ljust(TEXT_IDENTITY_WIDTH)` counts those ~30+ invisible chars, so a linked row's raw length already exceeds 49 → zero padding added, visible identity ≈20 cells, every later column shifts left only for PR rows. Fix: pad the plain identity to width first, then inject the link.
- [ ] [errors] Permanent Telegram 400 (malformed HTML entity) is swallowed and misclassified as a transient outage on the approval-gate push (`lib/hive/bot/notification_dispatcher.rb:236-242`, `lib/hive/bot/supervisor.rb:399-403`): the new `parse_mode: :html` path, on a 4xx `ResponseError`, logs `:send_failure` and feeds the same exponential-backoff retry used for network outages, so a permanent reject is retried forever, never delivered, never surfaced — the `ready_for_review` prompt silently never arrives. Distinguish 4xx (permanent → surface/operator-visible, or retry once as plain text) from transient transport errors.

## Medium
- [ ] [tests] The entire tty/clickable OSC 8 + HTML-link rendering path is unexercised across all four surfaces (`lib/hive/commands/status.rb:419`, `lib/hive/tui/views/tasks_pane.rb:189`): every test runs under captured (non-tty) stdout so `osc8(..., enabled: true)` inside a rendered row never runs — the alignment regression above ships undetected. Add one `$stdout.tty? => true`-stubbed render test per surface asserting the escape is present AND visible (escape-stripped) width is unchanged.
- [ ] [tests] No test pins display-width preservation when the OSC 8 escape is actually embedded (`lib/hive/tui/views/tasks_pane.rb:189`): the "embedded escape adds zero display cells, columns stay aligned" property is only true-by-vacuity today (escape never present in tests); the existing alignment guards need a tty-mode sibling for the PR column.
- [ ] [tests/code] `cell.sub(/.../, linked)` OSC 8 injection is correct only by luck and untested (`lib/hive/tui/views/tasks_pane.rb:195`, `lib/hive/commands/status.rb:428`): `String#sub`'s replacement string interprets `\`-sequences, and `linked` contains raw `\e\\` ST backslashes — safe only because each `\` is followed by `#`/end-of-string. Use the block form `cell.sub(pattern) { linked }` (immune to backreference interpretation) and pin it with a tty-path byte-roundtrip test.
- [ ] [tests] Late-arriving `pr_url` is never re-pushed and the consequence is untested (`lib/hive/bot/notification_builders.rb:80,99-107`): if the `ready_for_review` push fires while `pr.md`/`pr_url` is momentarily absent, the operator gets the plain-text push and never receives the clickable HTML version (fingerprint excludes `pr_url`, so no re-send). Narrow window, but pin it as intended-behavior or document it.
- [ ] [errors] `pr_url_for` broad `rescue StandardError => nil` masks an unreadable-but-present `pr.md` silently (`lib/hive/commands/status.rb:539-547`): `Gh.pr_frontmatter` already handles `Psych::Exception` and deliberately lets `Errno::EACCES`/`EISDIR`/etc. propagate; the blanket rescue swallows exactly those to a silent `—`, unlike the sibling lock-readers in the same file that `warn`. Add a `warn` and/or narrow the rescue to `SystemCallError, IOError` so a `NoMethodError` still surfaces.

## Nit
- [ ] [code] Dead method `display_identity` (`lib/hive/commands/status.rb:412`): both call sites now use `display_identity_with_pr`; unreferenced in `lib/` and `test/`. Safe to delete.
- [ ] [code] TUI hyperlink silently drops for 7+ digit PR numbers (`lib/hive/tui/views/tasks_pane.rb:191-195`): `rjust_cells(token, 6)` truncates `#1234567` → `#1234…`, so the following `.sub(/token\z/, ...)` no longer matches and the row shows a truncated number with no link (text path uses `String#rjust`, unaffected). Low impact.
- [ ] [tests] PR-number column overflow (6-digit `#NNNNNN`) is never tested (`lib/hive/commands/status.rb:419`, `lib/hive/tui/views/tasks_pane.rb:189`): `PR_WIDTH = 6` fits only 5 digits; pad-only rjust shifts every downstream column on a 6-digit PR. Add a `#1234567` fixture to pin the chosen overflow behavior.
- [ ] [tests] `valid_http_url?` is only tested via `javascript:` rejection (`lib/hive/bot/format.rb:24`, `lib/hive/tui/views/hyperlink.rb:18`): plain `http://` acceptance and `ftp://` rejection are unpinned, so a future tighten-to-HTTPS would go unflagged. Low value.
- [ ] [tests] The `:html` legacy-only queue path escaping is unverified for `&`/`<` (`lib/hive/bot/supervisor.rb:1401`): the test uses a clean `/tmp/hive` path so `html_escape` is a no-op; a project path containing `&` would actually prove the escape.
- [ ] [type-design] `Notification#parse_mode` is a free-form symbol, not a closed `:html | nil` type (`lib/hive/bot/notification_builders.rb:407-411`): contract lives only in convention; blast radius is tiny (one producer, one consumer). A one-line comment documenting the allowed set suffices — full enum would be over-engineering.
- [ ] [type-design] `valid_http_url?` is copy-pasted across `Hive::Bot::Format` and `Hive::Tui::Views::Hyperlink` (`lib/hive/bot/format.rb:25-30`, `lib/hive/tui/views/hyperlink.rb:20-26`): identical scheme/host guard; consider a shared predicate, though independence may be the project convention.
- [ ] [type-design/code] `Hive::Pr.number` returns nil for the `/pull/NNN/files` (and `/commits`) subpath form (`lib/hive/pr.rb:6`): unreachable today (input is canonical `pr.md` frontmatter), but the invariant is narrower than the general-purpose name implies; document or broaden.
- [ ] [type-design] `osc8` sanitizes `url` but passes the `visible` label through raw (`lib/hive/tui/views/hyperlink.rb:10-18`): safe today (callers pass a trusted `Hive::Pr.number` token), but the "caller guarantees `visible` is display-safe" contract is undocumented. Add a one-line comment.
- [ ] [type-design] A third independent PR-number extractor remains at `lib/hive/patrol/review_handoff.rb:198-199` (`pr_url.to_s[%r{/pull/(\d+)\z}, 1]`): pre-existing and out of this branch's scope, but `Hive::Pr.number` is now the natural single home — note for follow-up, not a blocker, do not expand scope here.
- [ ] [comments] `wiki/commands/tui.md:25-27` ASCII mock grid is schematic and its column spacing is inconsistent across rows: cosmetic (no behavioral claim), but someone counting spaces to infer widths would be misled; add a "diagram is schematic" note or leave as-is.

---
**Personas with no actionable findings:** `[comments]` found no High/Medium comment/doc inaccuracies (column-math, width constants, stage gates, regex, fingerprint, and schema comments all verified accurate; `wiki/gaps.md` #47 honestly records the clickable-link path as un-smoked, so no overclaiming). `[errors]` confirmed the four URL/sanitization rescues (`valid_http_url?`, `strip_control_chars`/`sanitize`, the `stage_approval` plain-text fallback) are correctly-scoped intentional degradations, not masking. `[type-design]` confirmed the `Data.define` custom initializer keeps positional + keyword construction working and the `pr_url` nullable field matches sibling-field treatment end-to-end.

ivankuznetsov:

## High
- [ ] `lib/hive/agent_limit.rb:54` reintroduces broad provider-limit matching: `limit_reached?` now treats healthy output like `scroll limit reached` or `page limit reset` as quota exhaustion, parking otherwise healthy runs as `limits_reached`.
- [ ] `lib/hive/stages/finalize.rb:41` removed the already-merged PR short-circuit: an out-of-band merged PR now continues into auth, push verification, the finalize agent, and `gh pr ready`, reintroducing stranded `8-finalize` tasks.

## Medium
- [ ] `lib/hive/pr.rb:24` accepts invalid rendered hrefs as safe links: URLs with spaces or other illegal URI characters still pass `valid_http_url?` and reach Telegram HTML / OSC 8 output instead of degrading to plain text.
- [ ] `lib/hive/stages/review/triage.rb:81` falls back to stale triage limits: partial configs now use `$15` / `300s` instead of `Config::DEFAULTS` `$75` / `1800s`, increasing false triage timeout or budget failures.

## Nit

ivankuznetsov:

# Code Review — show-pr-when-pr-created (pass 2)

Reviewed `git diff origin/main..HEAD`. The feature (PR number surfaced in TUI tasks
pane, `hive status` text/archive, Telegram `/status`+`/queue`, and the
`ready_for_review` push) is implemented across all five plan units, is security-gated
against `javascript:`/attribute-injection, preserves the A10 fingerprint contract
(verified: `fingerprint` hashes `[project, slug, stage, marker, attrs]`; `pr_url` is a
sibling field, pinned by `test_fingerprint_ignores_pr_url`), and is comprehensively
tested. Rubocop is clean and the full changed-area test suite passes. No
correctness/security defects found — only the nits below.

## High

None.

## Medium

None.

## Nit
- [ ] `wiki/modules/pr.md` claims `Hive::Pr` is used "in `Hive::Commands::Status#pr_cell`": that method does not exist — in `status.rb` `pr_cell` is only a local variable; the real method is `Hive::Commands::Status#display_identity_with_pr` (`lib/hive/commands/status.rb:432`). The `Hive::Tui::Views::TasksPane#pr_cell` reference on the next line is correct. Inaccurate symbol name violates the project's "ground everything in code" wiki standard.
- [ ] TUI OSC 8 clickability through lipgloss is unverified by automated tests: `pr_cell` is pinned in isolation, but the full `render_row` path wraps the row in `Styles.for_action_key(...).render` / `CURSOR_HIGHLIGHT.render` after the link is built (`lib/hive/tui/views/tasks_pane.rb:193-194`). Plan U2 explicitly called for a manual tty dogfood frame to confirm lipgloss passes the sequence intact before relying on it; worth that one manual check (plain `#NNN` fallback is acceptable per A1, so this is low-risk).
- [ ] `Hive::Pr.valid_http_url?(nil)` raises `TypeError` rather than returning false — only `URI::InvalidURIError` is rescued (`lib/hive/pr.rb:25-28`). Not reachable from the two current callers (both pass an already-stripped string), but as a public shared security predicate it would be safer to make total (rescue non-string/nil) like its sibling `Hive::Pr.number`.
- [ ] Process note (not a code defect): the branch is 2 commits behind `origin/main` (#490 AgentLimit/finalize, #492 triage fallbacks), so `git diff origin/main..HEAD` shows phantom "reversions" of `lib/hive/agent_limit.rb`, `lib/hive/stages/finalize.rb`, and `lib/hive/stages/review/triage.rb`. The feature commits never touched those files, and a three-way merge keeps main's versions — these are NOT real changes by this branch. Recommend rebasing/merging `origin/main` so CI validates the integrated tree and the diff stops showing the phantom hunks.

ivankuznetsov:

# PR review (pr-review-toolkit, multi-persona) — show-pr-when-pr-created (pass 2)

Union of findings from six persona agents (code-reviewer, silent-failure-hunter,
type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier) run on
the branch's actual contribution (`git diff origin/main...HEAD`). Tests pass
(465 runs, 0 failures) and rubocop is clean. The feature is well-built and the
security gating (`Hive::Pr.valid_http_url?`, `html_attr_escape`, OSC 8
sanitization) is sound; no persona found a confirmed correctness crash. Findings
below are reliability, type-enforcement, coverage, comment-accuracy, and cleanup.

> Scope note: the five files the two-dot `origin/main..HEAD` diff shows as
> deletions (`agent_limit.rb`, `finalize.rb`, `tmux_runner.rb`,
> `web/agents_auth.rb`, `stages/review/triage.rb`) are phantom reverts — the
> branch is 4 commits behind origin/main (#490/#492/#496/#497) and never touched
> them. Not regressions from this branch; rebase before merge.

## High
- [ ] [silent-failure] No plain-text fallback when an `:html` message is 400-rejected — the `/status` reply routes through `safe_send_message` (supervisor.rb:399-404,975-978), which rescues `StandardError`, logs `send_failure`, and returns nil: on a Telegram 400 the operator gets a blank non-response, indistinguishable from a hung bot.
- [ ] [silent-failure] `ready_for_review` `:html` push treats a permanent 400 (parse error) as a transient outage (notification_dispatcher.rb:182-242) — `record_send_failure` just backs off and re-sends the identical malformed payload forever, so the push is never delivered and the operator is never told.
- [ ] [tests] `telegram.rb` `split_message` (255-271) splits purely on length/newline with zero HTML-tag/entity awareness yet applies `:html` to every chunk, and no test feeds an `:html` body >4096 chars — an uncapped `display_name` in `render_queue`/`ready_for_review` can sever an `<a>` tag or `&amp;` entity → Telegram 400 (the plan's "queue capped ≪ 4096" premise is wrong: the cap is on row count, not line length).

## Medium
- [ ] [type-design][comment] `PARSE_MODES = [nil, :html]` (notification_builders.rb:413) is a documented closed set that is never referenced or enforced — `Notification.new` accepts any symbol for `parse_mode`, so an unsupported value (e.g. `:markdown`) reaches the Telegram API as a runtime 400 instead of failing at construction; the comment claiming it prevents shipping raw `<a>` as plain text is therefore inaccurate. Enforce via `PARSE_MODES.include?` in the existing init wrapper, or soften the comment.
- [ ] [silent-failure] `pr_url_for`'s `warn` (status.rb:566-575) lands on stderr that the dominant consumer `StatusWatcher` captures via `capture3` and discards (status_watcher.rb:109-117) — so on the bot path a persistently broken `pr.md` (EACCES/ESTALE) re-warns into a void every poll while silently showing a dash; route it through the structured bot logger instead.
- [ ] [silent-failure] `pr_url_for` rescues `Errno::ENOENT` and `SystemCallError` but not `IOError` (status.rb:557-575) — `IOError` is not a `SystemCallError` subclass, so a bare environmental read fault escapes and exit-70 crashes the whole status poll (empty reply on the bot path), violating the never-crash-on-bad-pr.md contract. Low-probability (`File.read` usually raises `Errno::*`).
- [ ] [simplify] `Hive::Pr.number` and the pre-existing `Hive::Patrol::ReviewHandoff#pr_number` (review_handoff.rb:198) are two `/pull/N` parsers that DISAGREE — trailing-slash/query URLs give `#12` vs `nil`; review_handoff writes `pr_number` into `pr.md` frontmatter, so a non-canonical URL stores nil there while every display shows `#12`. Have review_handoff delegate to `Hive::Pr.number(url)&.delete_prefix("#")`.
- [ ] [tests] `pr_url_for`'s non-ENOENT `SystemCallError` (EACCES) warn+nil branch is untested — only ENOENT/malformed/missing are covered, leaving the deliberate `StandardError`→`SystemCallError` narrowing unpinned; mirror the existing `.lock` EACCES test (status_test.rb:1018-1026).
- [ ] [tests] The TUI OSC 8 link is pinned only on `pr_cell` in isolation; no test runs the full `TasksPane.render` through lipgloss under a real tty (the plan's gating condition), so a lipgloss miscount of the zero-width OSC 8 bytes → tasks-pane misalignment/strip would go uncaught. `wiki/gaps.md` item 47 records the live-TTY frame was never smoked.
- [ ] [type-design] `Hive::Pr.valid_http_url?` (pr.rb:24-29) rescues only `URI::InvalidURIError`, but `URI.parse(nil)`/non-String raises an unrescued `TypeError` — asymmetric with `number`'s `.to_s` tolerance; safe today (callers pre-stringify) but a latent footgun. Use `URI.parse(url.to_s)` and/or rescue `TypeError`.

## Nit
- [ ] [tests] `render_queue` overflow-cap path (supervisor_test.rb:1434-1447) is never tested combined with PR links + HTML-special `display_name`s — the exact fixture most likely to exceed 4096 chars and interleave entities with anchors.
- [ ] [tests] `pr_cell` truncation assertions (tasks_pane_test.rb:165-180) hard-code the exact OSC 8 escape framing, so a legitimate framing change (e.g. adding an `id=` param) breaks them; assert behavior (link wraps token, padding outside) instead.
- [ ] [tests] `test_wide_icon_does_not_shift_the_id_column` (tasks_pane_test.rb:519-526) uses `split(/\d/, 2)`, which now works only because the fixtures have no `pr_url`; adding one would make the split land on the PR number, not the id.
- [ ] [simplify] `display_identity_with_pr` (status.rb:432) and `pr_cell` (tasks_pane.rb:197) are two near-identical "rjust token then splice OSC 8 link after the padding" routines whose offset math already differs slightly; extract one shared helper so the subtle splice invariant is fixed once.
- [ ] [simplify] Twin width constants `TEXT_PR_WIDTH` (status.rb:28) and `PR_WIDTH` (tasks_pane.rb:72), both 6, "kept in sync by convention"; tasks_pane already requires status, so reference one shared constant to remove the manual-sync drift hazard.
- [ ] [simplify] Dead defensive guards `row.respond_to?(:pr_url) ? row.pr_url : nil` (notification_builders.rb:100, supervisor.rb:1419) — `StatusWatcher::Row` now always defines `:pr_url` (and the test Structs were updated too); simplify to `row.pr_url`.
- [ ] [simplify] Magic literal `task.stage_index < 5` in `pr_url_for` (status.rb:554) hard-codes the `5-open-pr` ordinal; derive from `Hive::Stages::DIRS` or name a constant (the file already names `BRAINSTORM_STAGE_DIR` for this).
- [ ] [simplify] `Format.html_attr_escape` (format.rb:14) re-implements part of `CGI.escapeHTML` (already used in service_installer.rb); the attribute path could use it, though the 3-char `html_escape` text-node path is a deliberate Telegram-specific subset and should stay.
- [ ] [comment] status.rb:425 "~45 invisible escape bytes" undercounts — OSC 8 framing (~14 bytes) plus the full URL is ~50+ and URL-length-dependent; the ljust-counts-them argument is correct regardless.
- [ ] [comment] status.rb:427-428 "Mirrors the visible-width discipline Views::TasksPane uses via Format.rjust_cells" is misleading — this method uses plain `String#rjust`/`ljust` (char-counting), not cell-aware math; a CJK/wide `display_name` would misalign the text column (latent limitation worth noting in the comment).
- [ ] [comment] status.rb:431 "it cannot match the token inside the name" — wording implies a regex match; the splice is a fixed-offset slice, so reword to "the offset always targets the PR cell."
- [ ] [comment] pr.rb:12-13 `number` docstring says "GitHub pull-request URL" but the regex is host-agnostic and returns nil for valid sub-views like `/pull/123/files`; document the exact `/pull/<digits>` (+ trailing slash/query/fragment) contract.
- [ ] [code-reviewer] `header(row)` is computed twice on the HTML path (notification_builders.rb:97 then :102) — harmless micro-redundancy; compute once.
- [ ] [code-reviewer] `Hive::Pr.number`'s end-anchored regex extracts a number from a crafted suffix like `?x=/pull/123` (the `=/` supplies the leading `/`); not exploitable today since `pr_url` is Hive-authored and `valid_http_url?` gates the scheme, but worth tightening if pr.md ever becomes hand-edited/untrusted.

ivankuznetsov:

# Code Review — show-pr-when-pr-created (pass 3)

Scope: `git diff origin/main...HEAD` (merge-base; the branch is 4 commits behind
`origin/main`, so the two-dot diff's "deletions" of `agent_limit.rb`,
`tmux_runner.rb`, `web/agents_auth.rb`, `stages/review/triage.rb`,
`stages/finalize.rb` are origin-only commits #490/#492/#496/#497 the branch
hasn't merged — not this task's changes and excluded from review).

Reviewers: correctness, adversarial, security, testing, maintainability,
reliability, api-contract, performance, cli-readiness, project-standards,
agent-native. Validators: rubocop clean on all 28 changed `.rb` files; 481
focused unit/integration tests pass (0 failures).

Verdict: **Ready with minor fixes.** All plan requirements (A1–A10 + acceptance
cases) are implemented. No correctness, security, schema-contract, or
agent-parity defects found. The injection gates (`Hive::Pr.valid_http_url?` +
`Tui::Text.sanitize` for OSC8, `CGI.escapeHTML` for HTML href) hold under
adversarial probing; the de-dup fingerprint excludes `pr_url`/text so the
open-pr push stays once-only (A10). Findings below are latent/edge issues.

## High

_(none)_

## Medium

- [ ] Telegram `split_message` is HTML-entity/tag-unaware, but `/status` and the open-pr push now send `parse_mode: :html` (`supervisor.rb:977`, `notification_dispatcher.rb:238`): a single task line over 4096 chars (an operator-set ~4000-char display name) is hard-cut at `telegram.rb:264`, bisecting an `&amp;` entity or `<a>` tag → Telegram 400 → the whole message is silently dropped (and multi-chat sends may re-deliver chunk 0 to already-served chats on the next retry tick). Multi-row messages split safely on `\n`; only a single over-long line triggers it. Plan flagged this "for dogfood" but did not bound it — consider capping `display_title` length in `render_queue`/`stage_approval` or making `split_message` entity-aware when `parse_mode` is set. (adversarial primary; security + reliability corroborated as residual)
- [ ] `TEXT_IDENTITY_WIDTH = 49` (`status.rb:37`) is a hand-derived literal whose comment claims it "has to grow if `TEXT_PR_WIDTH` grows," but it can't: every other coupled width (`TEXT_PR_WIDTH`, `tasks_pane PR_WIDTH`) derives from `Hive::Pr::NUMBER_WIDTH`, while this one is magic. A future `NUMBER_WIDTH` bump widens the PR cell everywhere except this budget, silently misaligning `hive status` text output, and no test pins the relationship — derive it from `NUMBER_WIDTH` or add an assertion that fails on drift.

## Nit

- [ ] `compute_layout` separator counts `5` and `4` (`tasks_pane.rb:165-166`) are magic literals while the adjacent column widths are symbolic; adding a column (which bumps `SEPARATORS`) leaves the dropped-stage/minimal branches stale with no test failure — use `(SEPARATORS - 1)` / `(SEPARATORS - 2)`.
- [ ] `render_queue` legacy-line HTML escaping is untested (`supervisor.rb:1402`): `/status` now ships `parse_mode: :html`, but no test feeds an HTML-special char (`& < >`) through a legacy stage-dir line, so the `html_escape(legacy.text)` call would still pass if removed — add a special-char legacy-line assertion mirroring the actionable-row escaping test.
- [ ] `pr_url_for` warns to stderr on every status reparse for a permanently-unreadable `pr.md` (`status.rb:582`, `SystemCallError` branch): on the bot/daemon/TUI poll loops a single mis-permissioned `pr.md` emits one stderr line per affected row indefinitely (degrades safely otherwise) — consider de-duping the warn per `(path, error_class)` like `Telegram#build_update` does.
- [ ] Latent row-shape divergence: `NotificationBuilders.stage_approval` calls `row.pr_url` unconditionally on the `ready_for_review` branch (`notification_builders.rb:101`). `StatusWatcher::Row` carries `pr_url`, but the daemon's `StatusConsumer::Row` Struct does not; no crash today (that row never routes here), but any future path routing a non-`StatusWatcher` `ready_for_review` row into `stage_approval` raises `NoMethodError` — guard with `respond_to?` or unify the row contract.
- [ ] `hive status --help` long_desc (`cli.rb:544`, unchanged file) doesn't mention the new nullable `pr_url` row field, so an agent reading help (not the schema) won't discover it; the same line still says "schema v2" while the envelope is v3 (pre-existing drift) — optionally add a one-line `pr_url` note and interpolate `SCHEMA_VERSIONS.fetch("hive-status")` so the version can't drift again.

ivankuznetsov:

# PR Review Toolkit — show-pr-when-pr-created (pass 3)

Scope: `git diff origin/main...HEAD` (merge-base; the branch is 4 commits behind
`origin/main`, so two-dot "deletions" of `agent_limit.rb`, `tmux_runner.rb`,
`web/agents_auth.rb`, `stages/review/triage.rb`, `stages/finalize.rb` are
origin-only commits #490/#492/#496/#497 the branch hasn't merged — excluded).

Personas run (report-only): code-reviewer, pr-test-analyzer,
silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier.
Union of findings below; severities calibrated by the synthesizer (persona's own
rating noted where it differs). Validators: rubocop clean (28 changed `.rb`);
481 focused tests pass. **code-reviewer found no Critical/Important issues.**
All plan requirements (A1–A10 + acceptance cases) are implemented — no
plan-required-but-missing gap, so no High finding.

## High

_(none)_

## Medium

- [ ] [silent-failure] Telegram `:html` sends have no plain-text fallback on a 400: `send_notification` (`notification_dispatcher.rb:232-244`) and `safe_send_message` (`supervisor.rb:400-404`) rescue-and-swallow, and `split_message` (`telegram.rb:264`) hard-cuts a >4096-char line mid-`&amp;`/`<a>` tag — so a pathological ~4000-char task title makes the `/status` reply return total silence (no fallback reply) and the open-pr push fail and be backoff-retried (bounded by the task's ready_for_review lifetime) without ever delivering. Fix: on a parse-mode 400, retry once with `parse_mode: nil` (degraded-but-delivered) and/or make `split_message` entity-aware. (persona rated F2/F3 Medium; "infinite loop" framing is bounded by backoff + TTL, and the 400 is only reachable via the pathological-title cut.)
- [ ] [type-design][comment] `TEXT_IDENTITY_WIDTH = 49` (`status.rb:37`) is a hand-derived literal whose comment claims it "has to grow if `TEXT_PR_WIDTH` grows," but unlike the sibling `TEXT_PR_WIDTH`/`PR_WIDTH` (both derived from `Hive::Pr::NUMBER_WIDTH`) it does NOT auto-track — a `NUMBER_WIDTH` bump silently misaligns `hive status` text output, and no test pins the relationship. Fix: derive it from `NUMBER_WIDTH` (or assert the minimum in a test) and soften the comment to say "bump by hand." (type-design + comment-analyzer)

## Nit

- [ ] [silent-failure] `pr_url_for`'s I/O-fault `warn` (EACCES/ESTALE on `pr.md`, `status.rb:582`) goes to stderr, but `StatusWatcher#fetch` (`status_watcher.rb:110-115`) discards stderr on a successful exit — so in the daemon (the primary operating mode) a genuinely unreadable `pr.md` shows a blank PR with zero log signal. Persona rated HIGH; calibrated to Nit (impact is a missing clickable link, rare permission-drift trigger, and silent degrade-to-dash is the feature's intended contract) — worth a structured `@logger.event` for observability.
- [ ] [type-design] `stage_approval` calls `row.pr_url` unconditionally on the `ready_for_review` branch (`notification_builders.rb:101`); safe today only via a three-layer guard argument (build short-circuit + READY_ACTIONS + ready_for_review), but `Daemon::StatusConsumer::Row` (`status_consumer.rb:19`) lacks `pr_url` — a future non-`StatusWatcher` row routed here raises `NoMethodError`. Guard with `respond_to?(:pr_url)` (idiomatic in this module) or document the row contract.
- [ ] [type-design] Token-vs-link consistency is an emergent property of three call sites, not one invariant: a non-http `…/pull/12` renders `#12` plain in the TUI/text surfaces but is hidden entirely by the bot (`html_pr_link` requires both gates). Both safe; an optional `Hive::Pr.linkable?(url)` predicate would make "token shown ⟺ linkable" a single invariant.
- [ ] [comment] `pr.rb:17-19` — the centralized `NUMBER_WIDTH` comment says PR numbers ≥100000 "overflow this cell on purpose" for BOTH surfaces, but the TUI truncates (`#1000…` via `rjust_cells`) while only text mode (`String#rjust`) overflows; the sibling `tasks_pane.rb:70` comment correctly says "overflow/truncate." Align the centralized comment.
- [ ] [tests] No test asserts `Hyperlink.osc8`/`splice` reject a PR-shaped non-http URL (e.g. `ftp://h/o/r/pull/12`, which yields token `#12` but fails `valid_http_url?`); the link-builder side of the paired gate is only pinned for `javascript:`. Add an `osc8` + `splice` rejection assertion.
- [ ] [tests] No symmetric negative test that the TUI `pr_cell`/pane emits zero OSC8 bytes in a non-tty render — the status-text path has `refute_match(/\e\]8;;/, …)` (`status_test.rb:782`) but the TasksPane path only asserts the tty branch. Add the matching `refute_match` for a non-tty pane render with a populated `pr_url`.
- [ ] [tests] `Hive::Tui::Views::Hyperlink.splice` has no direct unit test — its trickiest contract (padding stays outside the link, token read back from `padded`, no-op when the URL is rejected) is exercised only indirectly via `pr_cell`/`display_identity_with_pr`, so a slice-arithmetic refactor could break trailing-text preservation undetected. Add direct `splice` tests (padding-outside-with-trailing-text, disabled→unchanged, invalid-URL→unchanged).
- [ ] [simplify] `compute_layout` separator counts `5` and `4` (`tasks_pane.rb:165-166`) are magic literals while `fixed_full` uses the named `SEPARATORS`; replace with `(SEPARATORS - 1)`/`(SEPARATORS - 2)` so the dropped-column branches can't drift on a future column add — byte-identical today.
- [ ] [simplify] `stage_approval` (`notification_builders.rb:97-117`) builds a plain `text` then discards it on the html path inside two-level nesting; extracting `approval_body(row, verb) → [text, parse_mode]` flattens the method and removes the throwaway assignment that reads as a latent bug. Optional clarity refactor (behavior-preserving; covered by existing tests).
- [ ] [tests] Minor brittleness (awareness only): the integration `s3_queue_test` couples to the exact `— — —` separator punctuation, and `status_test.rb:818` reimplements an OSC8-strip regex distinct from the production stripper — both fine today but fragile against formatting tweaks.

ivankuznetsov:

Outcome: timeout

Agent tail:

```
The poll has not returned yet; I’ll keep waiting until this session exits.
```

ivankuznetsov:

Outcome: timeout

Agent tail:

```
The final Ruby test check is still active. It behaved similarly on the previous head and then passed, so I’m continuing to wait.
```

ivankuznetsov:

Outcome: timeout

Agent tail:

```
GitHub now reports the PR as mergeable, with checks newly in progress; `mergeStateStatus` is `BLOCKED` only because required checks have not finished yet. I’m going to poll CI briefly so I can catch an immediate red job instead of leaving right after the push.
```

PR #515, opened 2026-06-18
feat(artifacts): capture visual demos and surface them in hivebox

ivankuznetsov:
## Summary

A finished UI or CLI task used to leave no visual trace: the `7-artifacts`
stage only wrote a text `artifact.md`, and hivebox only rendered text. This
adds best-effort visual proof to the artifacts handoff. When a task has an
observable surface, the artifacts agent boots and drives it, writes labelled
PNG stills plus a short GIF, and records a `media/manifest.json`. The stills
get pushed to screenote for hosted PR links, hivebox renders the committed
media inline, and the finalize prompt links the screenote copies in the PR
body.

Capture is fail-soft by design. A project that cannot boot, a missing capture
tool, a screenote outage, or a corrupt manifest all degrade to a recorded
state and the stage still completes. Visual capture can never turn the
pipeline red.

The three pieces ship together because they share one seam: the
`media/manifest.json` schema is the contract between the agent that captures,
the Ruby stage that enriches, and hivebox that displays.

```mermaid
graph TB
  agent[Artifacts agent] -->|writes stills + GIF + manifest| manifest[(media/manifest.json)]
  stage[Artifacts.run! Ruby] -->|reads captured manifest| manifest
  stage -->|uploads PNG/JPEG stills| screenote[screenote service]
  screenote -->|annotate_url| stage
  stage -->|writes screenote_url back| manifest
  hivebox[hivebox task page] -->|reads manifest| manifest
  hivebox -->|streams committed media| route[GET .../media/:filename]
  finalize[Finalize prompt] -->|reads screenote_url| manifest
  finalize -->|Demo links| pr[PR body]
```

## Manifest states

| `status` | When the agent writes it | hivebox renders |
|---|---|---|
| `captured` | observable surface, boot and drive succeeded | inline gallery of committed stills/GIF plus screenote links |
| `skipped` | no observable surface (backend, refactor, docs, config) | nothing; text artifacts only |
| `failed` | boot, driving, or capture tooling failed | a warning banner with the recorded reason, no broken images |

## Design decisions

**The screenote token never reaches the agent.** Upload runs in Ruby after the
agent returns `COMPLETE`: `Artifacts.run!` reads the manifest and pushes
PNG/JPEG stills via `Hive::ScreenoteUploader`, sourcing `base_url` and
`api_token` from global config or `HIVE_SCREENOTE_*` env. The prompt only flags
which files to push (`push_to_screenote: true`); it never sees the secret, and
the `HIVE_SCREENOTE_*` env vars are scrubbed from the artifacts-agent child
environment (mirroring the existing `ANTHROPIC_API_KEY` blanking precedent). A
blank token disables upload while committed media still renders.

**Storage split.** Stills and a short GIF are committed under the task folder
(already on `hive/state`) and served directly by hivebox, so they exist on any
clone with no external dependency. screenote receives PNG/JPEG stills only,
matching its current content-type allowlist, to give the PR a hosted link.
mp4/video and GIF-to-screenote are deferred until the screenote video endpoint
lands.

**The media route is path-safe.** `GET /tasks/:project/:slug/media/:filename`
accepts a single filename component matching `png/jpe?g/gif`, resolves the real
path under `<task>/media/` with `File.realpath`, and refuses anything that
escapes that directory via traversal or symlink. It streams read-only with an
inline content type under the same auth as the rest of the controller.

**No new gem.** The uploader is a stdlib `Net::HTTP` multipart POST with an
injectable transport, so it unit-tests without a network and adds no dependency.

## Test plan

- Uploader: 201 returns the annotate URL; 401/500/timeout/exception return `nil`
  and never raise; missing credentials skip without touching the transport;
  non-ASCII captions/filenames, multipart wire framing, and JPEG content types
  are covered.
- Stage wiring: only PNG/JPEG stills upload (GIF skipped), URLs write back
  idempotently, `skipped`/`failed` and corrupt manifests are swallowed, the
  upload budget and `ConfigError` paths are exercised, and the full `run!` path
  pushes only after the `COMPLETE` marker.
- Web: the media route streams committed PNG/JPEG/GIF and refuses traversal,
  symlinks, disallowed extensions, missing files, and unknown tasks; integration
  and Capybara-Playwright system tests cover the captured gallery, screenote
  links, the failed banner, and the hidden-section skipped/absent cases.

`bundle exec rake test` (`5440 runs, 21267 assertions`), `cd web && bin/rails
test` (`69 runs, 344 assertions`, including the system suite), and both rubocop
runs pass. The Docker image build was not run locally; `asciinema`/`ffmpeg` were
added for terminal capture and browser tooling stays environment-dependent
(absence records a `failed` manifest).

## Review summary

Three Compound-Engineering review passes ran on this branch — Claude CE, Codex
CE, and the pr-review-toolkit persona panel (correctness, silent-failure,
type-design, comment, test, and simplicity lenses) — converging by pass 3.

Auto-fixed across the review-fix commits:

- **Resilient manifest reads** — the hivebox reader and stage guard
  `is_a?(Hash)` so a valid-but-non-object `manifest.json` can no longer 500 the
  task page.
- **Bounded screenote upload** — an aggregate deadline caps total upload time so
  a slow/hostile endpoint can't stall the `hive run artifacts` child.
- **Containment** — symlink/traversal-safe path resolution via `File.realpath`
  on both the gem upload side and the web serving route, plus anchored filename
  constraints.
- **Token isolation** — `HIVE_SCREENOTE_*` is scrubbed from the artifacts-agent
  child environment (following the `ANTHROPIC_API_KEY` precedent), so an
  env-sourced token never reaches the LLM.
- **Uploader robustness** — `.b` binary coercion for non-ASCII captions and
  filenames, `https?://` scheme validation, Hash-guarded JSON parsing, and fixed
  `Content-Disposition` escaping.
- **Forward-compat** — the `media/manifest.json` schema version is hoisted to a
  single gem constant and gated by both readers.

One open item is flagged for reviewer attention before merge:

- **Agent-supplied `screenote_url` provenance** (escalation Q1, unresolved).
  `upload_item_to_screenote?` skips any item that already carries a
  `screenote_url` for idempotency, so a contract-violating or prompt-injected
  manifest value is preserved verbatim and could surface as a false "screenote"
  link in hivebox and in this PR body via the finalize prompt. Bounded (the
  agent already authors `artifact.md`), but it reaches a public surface.
  Suggested default: make the stage the sole writer of `screenote_url`
  (null/host-match agent-supplied values) plus web render-time host validation.

## Linked task

- Hive task **Step Workflow Update** (id 1361), slug
  `we-now-have-step-where-260617-1929`.
- Implements the plan *"visual artifacts (screenshots / GIF) in 7-artifacts +
  hivebox"* (units U1–U6), produced by the hive pipeline through stage
  `8-finalize`; the plan and all three review passes are recorded in the task
  folder.

---


ivankuznetsov:

## High
- [ ] `lib/hive/stages/artifacts.rb:87` screenote upload follows symlinked media files: `media_item_path` only does a lexical basename/prefix check before `File.file?` and `File.binread`, so an agent-produced `media/01-home.png` symlink can exfiltrate any readable host file to screenote; resolve the media dir and candidate with `File.realpath` and require the real target to remain under the real media directory.

## Medium

## Nit

ivankuznetsov:

## High
- [ ] `web/app/controllers/tasks_controller.rb:196` `media_manifest` only rescues `JSON::ParserError`/`SystemCallError`, so a syntactically-valid manifest whose top-level value is a JSON array, number, or `null` raises an unrescued `TypeError`/`NoMethodError` and 500s the whole task show page: `JSON.parse("[]")["status"]` → TypeError, `JSON.parse("null")["status"]` → NoMethodError; the sibling `open_questions` deliberately rescues `StandardError` for this exact "half-written agent file must not 500 the page" reason, and the plan explicitly requires reading the manifest "resiliently … like `open_questions`" — guard `manifest.is_a?(Hash)` or rescue `StandardError`.

## Medium
- [ ] `lib/hive/stages/artifacts.rb:84` / `lib/hive/screenote_uploader.rb:36` the screenote upload loop runs synchronously in `run!` after COMPLETE with only per-request timeouts (open 10s / read 60s) and no aggregate cap, so a captured manifest with several stills against a slow/hostile endpoint serializes to N×60s of blocking inside the daemon dispatch cycle (bounded per request, unbounded in aggregate) — stalls processing for minutes even though it never fails the stage.

## Nit
- [ ] `lib/hive/screenote_uploader.rb:79` `quote`'s backslash-doubling `gsub("\\", "\\\\")` is a no-op (a gsub replacement of `"\\\\"` collapses to a single literal backslash), so a backslash in a still filename is not escaped in the `Content-Disposition` header; only reachable via an exotic on-disk filename containing `\` (field values aren't quoted, so only the filename path uses `quote`), but the method's stated escaping intent isn't met.
- [ ] `lib/hive/screenote_uploader.rb:90` `parse_success` only rescues `JSON::ParserError`; a 201 whose body is valid-but-non-object JSON (`[...]`/`42`) raises `TypeError` on `payload["annotate_url"]`, which the broad `rescue StandardError` in `upload` catches (fail-soft holds) but mislabels as a generic "upload failed", losing the invalid-JSON diagnostic.
- [ ] `lib/hive/config.rb:988` an invalid `HIVE_SCREENOTE_BASE_URL` env value (e.g. `ftp://…` or a scheme-less typo) makes `load_global_screenote` raise `ConfigError` (env override is applied before `validate_screenote!`), which is swallowed by the call-site rescue in `push_manifest_media_to_screenote` → uploads silently disabled with only a stage-stderr warning instead of a loud startup failure.
- [ ] `lib/hive/screenote_uploader.rb:91` `parse_success` accepts any non-empty `annotate_url` without scheme validation; a relative/garbage URL (`/s/123`) is written into the manifest, then dropped by the web layer (which requires `\Ahttps?://`), while `upload_item_to_screenote?` now sees a non-empty `screenote_url` and skips the item on every re-run, so it can never re-upload to obtain a valid URL.
- [ ] `web/config/routes.rb:27` the media route constraint `/[\w.-]+\.(?:png|jpe?g|gif)/i` is unanchored and combines with Rails' implicit `(.:format)`, so `GET …/media/evil.png.rb` is accepted as `filename: "evil.png", format: "rb"`; non-exploitable (the served basename always ends in an allowed image extension and the content-type is pinned from that extension, never disclosing the `.rb`), but it accepts more URL shapes than intended — anchor with `\A…\z` and/or `format: false`.

ivankuznetsov:

## High
- [ ] [silent-failure] `web/app/controllers/tasks_controller.rb:196` `media_manifest` only rescues `JSON::ParserError, SystemCallError`, so a syntactically-valid manifest whose top level is a JSON array/number/`null` raises an uncaught `TypeError`/`NoMethodError` on `manifest["status"]` and 500s the whole task show page; sibling `open_questions` rescues `StandardError` for exactly this case and the plan required reading "resiliently … like open_questions" — guard `manifest.is_a?(Hash)` or rescue `StandardError`.

## Medium
- [ ] [silent-failure] `lib/hive/stages/artifacts.rb:79,101` a malformed `HIVE_SCREENOTE_BASE_URL` makes `load_global_screenote` raise `ConfigError` (env override is applied before `validate_screenote!`), which the broad `rescue StandardError` flattens into one generic warn — an operator's fat-fingered env var is indistinguishable from a transient upload blip; let `ConfigError` escape or log it distinctly.
- [ ] [type-design] `lib/hive/screenote_uploader.rb:91` / `templates/finalize_prompt.md.erb:25` screenote_url scheme asymmetry: the uploader only guarantees `annotate_url` non-empty (no scheme check), and the finalize PR-body reader links it verbatim with NO `https?://` validation while the web reader (`tasks_controller.rb:225`) does — the most security-sensitive consumer (the public PR body) is the least strict, so a relative/`javascript:` URL could land in the PR.
- [ ] [type-design] manifest schema has no single source of truth: `"schema": 1` is written (`templates/artifacts_prompt.md.erb`) but never read by any consumer, and `status`/`surface`/`type`/extension rules are re-implemented and already diverge across the prompt, `artifacts.rb`, and `tasks_controller.rb`; a future schema bump will be silently mishandled differently per consumer — consider a shared `Hive::MediaManifest` normalizer + constants.
- [ ] [comment] `lib/hive/config.rb:340` the comment claims the token "may be supplied only via HIVE_SCREENOTE_API_TOKEN", but `load_global_screenote`/`validate_screenote!` fully honor `screenote.api_token` from the global config file — the comment contradicts the code (the "agent never receives it" half is accurate; "only via env" is wrong).
- [ ] [tests] `web/test/integration/tasks_test.rb` has no case for a valid-JSON-but-wrong-shape manifest (top-level array/number/null) reaching `show`, nor for a malformed `items` entry (non-Hash element, or junk/traversal `file`) — the `normalized_media_items` guards and the High bug above are unexercised.
- [ ] [tests] `test/unit/screenote_uploader_test.rb` has no case for a 201 whose body is non-object JSON, nor for a relative/garbage `annotate_url`; the response-shape contract (the whole point of `parse_success`) is unverified.

## Nit
- [ ] [code-review] `lib/hive/screenote_uploader.rb:271` `quote`'s `gsub("\\", "\\\\")` is a no-op (the gsub replacement `"\\\\"` collapses to a single backslash), so backslashes are not escaped in the `Content-Disposition`; unreachable today (only literal field names + PNG/JPEG basenames flow through `quote`) but the method's stated escaping intent isn't met.
- [ ] [code-review] `lib/hive/screenote_uploader.rb:202` success is hard-pinned to HTTP `201`; a screenote `200` would silently fail-soft (warn + nil, leaving `screenote_url` null). Accept `Net::HTTPSuccess` (2xx) for robustness.
- [ ] [silent-failure] `lib/hive/screenote_uploader.rb:90` `parse_success` only rescues `JSON::ParserError`; a non-object 201 body raises `TypeError`, caught by `upload`'s outer `rescue StandardError` but surfaced as a generic "upload failed" instead of the accurate "non-object response body" diagnostic.
- [ ] [silent-failure] `lib/hive/stages/artifacts.rb:82-103` items are mutated in memory and written once after the loop; a mid-loop raise would discard `screenote_url`s for stills already uploaded to screenote (unreachable today since `upload` swallows its own errors) — write-after-each-success would be safer.
- [ ] [type-design] `lib/hive/screenote_uploader.rb:12` `attr_reader :api_token` exposes the secret publicly though only `base_url` is read externally — drop it to shrink the secret's reachable surface.
- [ ] [type-design] `lib/hive/screenote_uploader.rb:96` `screenshot_id` is carried in the return shape but consumed nowhere (only tests reference it) — wire it (e.g. manifest idempotency) or drop it to keep the return contract honest.
- [ ] [type-design] extension drift: gif is renderable (`MEDIA_FILENAME_RE` + route constraint) but never uploadable (`upload_item_to_screenote?` + `content_type` know only png/jpeg) — intentional but spread across four unlinked sites with no shared constant, so a parity edit can silently produce an un-renderable upload or an `image/png`-mislabeled body.
- [ ] [comment] `wiki/stages/artifacts.md:30` overstates observability: it says traversal-shaped/non-still/missing-file items are "ignored with warnings", but those per-item skips are silent `next`s; only invalid-JSON and the outer rescue actually warn.
- [ ] [comment] `wiki/commands/web.md:103` attributes the strict filename guarantee to the route, but the route constraint `/[\w.-]+\.(?:png|jpe?g|gif)/i` is unanchored — the real enforcement (anchored regex + `File.basename` equality + `File.realpath` containment) lives in the controller's `resolved_media_path`.
- [ ] [comment] `templates/artifacts_prompt.md.erb` documents a TUI/CLI branch and the wiki says "UI/TUI/CLI", but the `surface` enum is only `ui | tui | none` — no `cli` value or mapping is defined (display-inert; no code branches on `surface`).
- [ ] [tests] `test/unit/stages/artifacts_test.rb:145` `test_complete_agent_run_pushes_manifest_after_marker` stubs the private `screenote_uploader` via `define_singleton_method`, coupling the test to an implementation detail rather than the `screenote_config:` injection seam the other tests use.
- [ ] [tests] no test covers a `captured` manifest with empty/all-filtered `items` (the view gates on `items.present?`), so a regression into an empty `<section class="demo">` with a bare heading would go uncaught.
- [ ] [tests] `lib/hive/stages/artifacts.rb:125` `media_item_path`'s basename/containment guard — which gates what is uploaded to an external service — is never directly exercised with a `../`-shaped `item["file"]`.

ivankuznetsov:

## High
- [ ] `HIVE_SCREENOTE_API_TOKEN` can reach artifacts agents (`lib/hive/config.rb:999`, `lib/hive/agent.rb:127`): the new env override keeps the secret in process ENV while agent spawns inherit that environment, violating the plan's token-never-reaches-LLM requirement.
- [ ] `media/` directory symlinks bypass containment (`lib/hive/stages/artifacts.rb:149`, `web/app/controllers/tasks_controller.rb:242`): both upload and serving code trust the symlink target as the media root, so a task can upload or stream readable png/jpg/gif files outside its task folder.

## Medium
- [ ] `screenote_url` manifest writes do not trigger hivebox live refresh (`lib/hive/stages/artifacts.rb:112`): status broadcasts track `artifact.md` and the task folder mtime, but rewriting nested `media/manifest.json` only changes the media directory, so open task pages can miss hosted links until reload.
- [ ] Screenote upload accepts filenames hivebox refuses (`lib/hive/stages/artifacts.rb:142`, `web/app/controllers/tasks_controller.rb:221`): upload eligibility only checks the extension while the route/display contract requires `[\w.-]+`, so captures with spaces or newlines can be uploaded and then hidden from the Demo gallery.

## Nit

_None._

ivankuznetsov:

## High
- [ ] `lib/hive/screenote_uploader.rb:64` (`append_field`) — a non-ASCII caption silently breaks the screenote upload. The multipart `body` starts as `Encoding::BINARY`; appending an ASCII-only title leaves it binary, but appending a title containing any non-ASCII byte (em-dash `—`, curly quotes, accented letters, Cyrillic, emoji) flips `body`'s encoding to UTF-8, after which `append_file`'s `body << File.binread(path)` (line 74) raises `Encoding::CompatibilityError: incompatible character encodings: UTF-8 and BINARY`. That is swallowed by `upload`'s `rescue StandardError` → returns `nil` → the still is never uploaded and `screenote_url` stays `null`, so the U2 screenote push and the U5 PR-body Demo links silently fail for that item (confirmed empirically). Title is `item["caption"]`, an agent-written human string — em-dashes and (for this Russian-context project) Cyrillic captions are likely, so this triggers in normal use. Fix: force the title bytes to binary, e.g. `body << value.to_s.b`. Existing uploader tests only use ASCII titles, so add a non-ASCII-caption case.

## Medium
- [ ] `lib/hive/stages/artifacts.rb:30,86` — the screenote upload loop runs synchronously inside `run!` *after* the agent has already stamped `<!-- COMPLETE -->`, adding up to `SCREENOTE_UPLOAD_BUDGET_SEC` (120s, effectively ~180s since the budget is checked only before each request) of wall-clock to the `hive run artifacts` child past the terminal marker. If an operator sets `daemon.child_verb_timeouts.artifacts`, the daemon can SIGTERM/SIGKILL the child mid-upload after `:complete` is on disk. That contradicts `stale_agent_healer`'s documented invariant (`lib/hive/daemon/stale_agent_healer.rb:290-293`: "the timeout means `:complete` was never stamped, so artifacts does NOT hit its own `:complete` short-circuit"), so a heal would re-dispatch and force a wasteful full agent re-capture even though the stage genuinely completed. Default-disabled (verb timeouts default to 0) and self-healing, but worth bounding the post-`:complete` work tighter or making the healer tolerant of a complete-but-killed artifacts child.

## Nit
- [ ] `lib/hive/stages/artifacts.rb:13-16` — the `SCREENOTE_UPLOAD_BUDGET_SEC` comment says the loop "runs inside the daemon dispatch cycle"; it actually runs inside the spawned `hive run artifacts` child process (the daemon dispatches stages as supervised children, not in its own tick loop), so a slow endpoint stalls that one child, not the daemon's main loop. Minor comment accuracy.
- [ ] `web/app/controllers/tasks_controller.rb:198` (`media_manifest`) — the manifest's `schema` field is read by neither the stage nor hivebox; a future `schema: 2` with an incompatible `items[]` shape would be rendered as-is rather than ignored. Harmless at `schema: 1`; consider gating on the known schema version for forward-compat.

ivankuznetsov:

## High
- [ ] [comments] `lib/hive/stages/artifacts.rb:14-15` — the comment "the loop runs inside the daemon dispatch cycle" is architecturally false: stages run as separate `hive run` child processes (`ChildSupervisor#spawn` → `commands/run.rb`), so the budget bounds that child's wall-clock, not the daemon tick. Misleads about where blocking occurs.
- [ ] [comments] `lib/hive/stages/artifacts.rb:17-18` — "unfinished stills are retried on a later run" is false: `run!` early-returns on a `:complete` marker, so budget-skipped stills keep a blank `screenote_url` permanently unless an operator manually clears the marker and re-runs. The documented auto-retry does not exist.
- [ ] [tests] `lib/hive/screenote_uploader.rb:52-76` — multipart body framing is only substring-sniffed; nothing asserts the per-part CRLF framing (`\r\n\r\n`), the file-part `Content-Type`, verbatim `File.binread` bytes, or the closing `--boundary--\r\n`. A framing regression passes every current assertion but breaks every real upload (the `@http` seam is stubbed in all tests, so no server ever parses the body).
- [ ] [tests] `lib/hive/stages/artifacts.rb:92-98` — the `SCREENOTE_UPLOAD_BUDGET_SEC` deadline/`break`/warning loop is never exercised; its entire "don't serialize to N×60s of blocking" contract is unpinned (deterministically testable by stubbing `Process.clock_gettime` or a negative budget).

## Medium
- [ ] [errors] `lib/hive/screenote_uploader.rb:93-114` — a 201 with a non-JSON/non-object body warns generically and returns `nil`, but the upload actually succeeded server-side; the URL is lost and a manual stage re-run re-uploads the same still (duplicate screenote record). A distinct condition (succeeded-but-unparseable) is masked as a transient blip.
- [ ] [errors] `lib/hive/stages/artifacts.rb:145-166` — `media_item_path` rescues only `SystemCallError`; an `ArgumentError` (e.g. a null byte in an agent-written `item["file"]`) escapes to the caller's broad `StandardError` rescue at :121-122 and aborts uploads for ALL remaining valid stills instead of skipping the one poisoned item. Widen to `rescue SystemCallError, ArgumentError`.
- [ ] [errors] `lib/hive/stages/artifacts.rb:137-143` — `upload_item_to_screenote?` skips any item already carrying a `screenote_url` with no mtime/content check, so a re-captured still (same filename, new bytes) keeps pointing at the stale screenote annotation of the previous image; no warn on skip.
- [ ] [tests] `lib/hive/stages/artifacts.rb:104` + `lib/hive/screenote_uploader.rb:60-66` — non-ASCII/multibyte captions (realistic for this Russian-context project) are never sent through the upload path; appending a UTF-8 caption into the `BINARY`-encoded multipart buffer is exactly the kind of input that breaks in production and is untested. (Confirmed elsewhere to raise `Encoding::CompatibilityError`.)
- [ ] [tests] `lib/hive/screenote_uploader.rb:86-91` — `content_type` jpeg-vs-png selection (including the `.downcase`) is untested; no `.jpg`/`.jpeg` upload asserts `Content-Type: image/jpeg`, and the `else→image/png` default arm is dead from the suite's view.
- [ ] [tests] `lib/hive/screenote_uploader.rb:78-84` — `quote()` Content-Disposition escaping has no test despite a comment noting it regressed to a no-op before; a `"`/`\` filename should assert correct backslash-escaping (security-adjacent, defense-in-depth guard left unprotected).
- [ ] [tests] `web/app/controllers/tasks_controller.rb:58` — the media route's `expires_in 60.seconds, public: false` is unasserted; a regression to `public: true` would let a shared proxy cache an authenticated user's task screenshots.
- [ ] [tests] `web/app/controllers/tasks_controller.rb:60` + route — the JPEG path through the web media route is untested (png and gif covered), even though the route, `MEDIA_FILENAME_RE`, and the manifest filter all accept jpeg.
- [ ] [tests] `lib/hive/stages/artifacts.rb:142` — no test proves a `push_to_screenote: true` GIF is rejected by the extension gate; the existing gif fixture sets `push_to_screenote: false`, so the extension filter is never the thing doing the rejecting.
- [ ] [tests] `test/unit/stages/artifacts_test.rb:175` — `test_complete_agent_run` stubs `ScreenoteUploader.new` with the exact kwarg list `{ |base_url:, api_token:| }`, so any added constructor kwarg breaks the test with an unrelated `ArgumentError`; stub the `screenote_uploader` factory or use `|*, **|`.
- [ ] [types] `lib/hive/stages/artifacts.rb` vs `web/app/controllers/tasks_controller.rb:216-233` — the manifest's two readers hand-roll different item-acceptance rules (stage: `%w[.png .jpg .jpeg]` + basename; controller: `MEDIA_FILENAME_RE` including `gif`). Future media-kind changes must touch both sites or silently diverge (renderable-but-never-processed items, or vice versa); no shared "valid item" definition.
- [ ] [types] `templates/artifacts_prompt.md.erb:52` vs readers in `artifacts.rb`/`tasks_controller.rb` — `schema: 1` is written but read/enforced by neither reader; a future `schema: 2` is parsed as v1 and silently yields empty/garbage items. Inconsistent with `update_check_state`/status-feed which gate on their version field. Either drop the field or branch on it.
- [ ] [types] manifest is a three-way stringly-typed `Hash` with the `status` enum duplicated and inconsistent (controller accepts `captured|skipped|failed`; stage acts only on `captured`); a key typo (e.g. `sceenote_url`) degrades silently to "no link" via `to_s` on a missing key. Extract key names + status enum into one shared constant referenced by both gem and web.
- [ ] [errors] `web/app/controllers/tasks_controller.rb:189-233` + `web/app/views/tasks/_media.html.erb:13` — a `captured` manifest whose items all fail to resolve reduces `items` to `[]`, so the partial renders nothing (no gallery, no banner); a claimed-success capture with missing files is indistinguishable from success on the page. Consider a "captured, but media files missing" banner.

## Nit
- [ ] [comments] `lib/hive/screenote_uploader.rb:81-82` — the `quote` comment overstates the prior bug as a total "no-op"; only the backslash branch was broken (the quote branch `gsub('"','\"')` already worked). Tighten the description (the fix itself is correct).
- [ ] [errors] `lib/hive/screenote_uploader.rb:36-43` — `upload` calls `response.code` with no shape guard; a buggy injected `@http` returning a non-response is swallowed by the broad `rescue StandardError` and looks like a flaky endpoint forever (programming error masked as transient).
- [ ] [errors] `web/app/controllers/tasks_controller.rb:53-62` — the `media` action returns a bare `head :not_found` with no log when a manifest lists a file that won't resolve; a `Rails.logger.warn` would make a half-cleaned media dir diagnosable (contrast the deliberate warns at :270/:288/:212).
- [ ] [tests] `lib/hive/screenote_uploader.rb:100-101` — the empty-string `annotate_url` branch (returns nil with NO warning, unlike the relative-URL/non-object branches) and the `screenshot_id` passthrough are untested.
- [ ] [tests] `lib/hive/stages/artifacts.rb:102` — the stage's own "manifest item names a png missing on disk → skip without crashing" guard (`next unless file_path && File.file?`) has no dedicated test (the web side covers its analogue).
- [ ] [tests] `test/unit/stages/artifacts_prompt_test.rb` / `finalize_prompt_test.rb` — assert literal ERB copy (`"Visual demo capture"`, `"## Demo"`, `"Preserve or re-add"`); fine as anti-deletion tripwires but will churn on wording edits and are not behavioral.
- [ ] [tests] `test/unit/stages/artifacts_test.rb:241` — `StubUploader#upload` returns `screenshot_id => title`, so no stage test would catch the stage wrongly starting to depend on `screenshot_id` (harmless today; stage ignores it).
- [ ] [types] `lib/hive/screenote_uploader.rb:107-110` — `upload` returns a bare `Hash` whose `screenshot_id` key is never read by any caller, and the caller re-checks `annotate_url` emptiness (`result["annotate_url"].to_s`), evidence the http-only/non-empty invariant isn't self-evident from outside; a small `Result` struct or returning the URL string would localize it.
- [ ] [types] `lib/hive/screenote_uploader.rb:14` — the injected `http:` transport's response duck-type (`#code`→String, `#body`→String) is implicit; a one-line param doc would make the seam self-documenting for a future transport author.
- [ ] [types] `lib/hive/screenote_uploader.rb:22-27` vs `lib/hive/stages/artifacts.rb:125-131` — `configured?` duplicates the stage factory's credentials check; same drift smell at small scale if "configured" ever grows (e.g. a required project id).
- [ ] [code] `lib/hive/stages/artifacts.rb:67` — `render_prompt` passes `user_supplied_tag:` into `TemplateBindings` but `artifacts_prompt.md.erb` never references it; harmless dead binding (other stages use the tag, so likely intentional consistency).
- [ ] [code] `lib/hive/stages/artifacts.rb:112` — the manifest write-back is a non-atomic `File.write`; fine for once-per-dispatch, but if concurrent artifacts runs against one task were ever possible an atomic write-then-rename would be safer.
- [ ] [simplify] `web/app/controllers/tasks_controller.rb:220-223` — `normalized_media_items` line 221 (`file.match?(MEDIA_FILENAME_RE)`) is strictly weaker than the line-222 basename check + `resolved_media_path`'s own regex; droppable, though defensible as layered validation (confirm before applying).
- [ ] [simplify] `web/app/controllers/tasks_controller.rb:235-255` vs `lib/hive/stages/artifacts.rb:145-166` — near-identical symlink/traversal path-resolution duplicated across the gem and the Rails app; both the simplify and type-design passes judge the duplication justified (separate load paths, ~12 lines) — recorded, not recommended to change.
- [ ] [simplify] `lib/hive/screenote_uploader.rb:37` — `response.code.to_s == SUCCESS_CODE` double-converts an already-String Net::HTTP code; defensible given the injectable test seam, so leave unless tightening.

ivankuznetsov:

## High
- [ ] `HIVE_SCREENOTE_API_TOKEN` can still reach artifacts agents (`lib/hive/agent.rb:127`): env-configured screenote credentials are inherited by `Process.spawn`, violating the plan's token-never-reaches-LLM requirement.
- [ ] Nested hivebox paths were removed from the review fix allowlist (`lib/hive/config.rb:150`): legitimate `web/app/**` and `web/test/**` fixes now fail the auto-commit scope check, blocking review fixes for this feature's Rails changes.
- [ ] LLM wiki post-commit refresh now dirties the committing worktree (`lib/hive/llm_wiki_bootstrap/scripts.rb:173`): the generated hook runs Codex in the feature checkout and no longer stages/commits wiki refresh output on the main checkout.

## Medium
- [ ] Agent-supplied `screenote_url` values are trusted (`web/app/controllers/tasks_controller.rb:244`): prefilled manifest URLs skip Ruby upload and are rendered/linked, so an agent can inject non-screenote Demo and PR links.
- [ ] Babysitter dry-run artifacts are no longer ignored or cleaned (`lib/hive/babysitter/dry_run_env.rb:26`): `.hive-babysitter-dry-run-bin/` and skip logs can remain as untracked residue and trip clean-exit scope checks.
- [ ] TUI/CLI GIF tooling is incomplete in the Docker image (`packaging/docker/Dockerfile:14`): installing `asciinema` and `ffmpeg` without `agg`/`vhs` leaves the prompt-required terminal GIF path unable to run in hivebox.

## Nit
- [ ] Screenote config is documented in the wrong template (`templates/project_config.yml.erb:55`): the snippet appears in per-project config even though uploads read only global config or environment overrides.

ivankuznetsov:

# claude-ce-code-review — pass 03

Scope note: this branch is diverged from `origin/main` (merge-base `0d0cac16`), so
`git diff origin/main..HEAD` mixes the feature's real changes with origin/main's later,
unrelated commits shown inverted (llm-wiki scripts, babysitter dry-run, daemon self-heal,
digest). Findings below are scoped to the feature's actual commits
(`0d0cac16..HEAD`: U1–U6 + the three fix commits). The codex-03 High findings on
`lib/hive/config.rb:150` (allowlist), `lib/hive/llm_wiki_bootstrap/scripts.rb:173`, and the
Medium on `lib/hive/babysitter/dry_run_env.rb:26` are **false positives** from this divergence —
the feature commits never touch those files/lines — and are not reproduced here.

## High
- [ ] `HIVE_SCREENOTE_API_TOKEN` env path leaks the token to the artifacts agent, contradicting the plan's "token must not reach the LLM" requirement and the code's own promise: `config.rb`'s `apply_screenote_env_overrides!` and `docs/visual-artifacts.md` document `export HIVE_SCREENOTE_API_TOKEN=...` as a supported source, but `config.rb:340` states "the agent never receives them" and `docs/visual-artifacts.md` "The agent never receives the token" — when the token is sourced from the env var, the artifacts agent child (`lib/hive/agent.rb` `Process.spawn`, no env scrubbing) inherits it via the OS environment and can read it with its `Bash` tool. The config-file path is safe (agents never read it), so this is env-path-only, but it's a documented config path that violates the explicit promise. Fix: scrub `HIVE_SCREENOTE_API_TOKEN`/`HIVE_SCREENOTE_BASE_URL` from the agent child env at spawn, or soften the absolute "never receives them" claim and steer operators to config-file sourcing. (Unanswered escalation Q1 from pass 02.)

## Medium
- [ ] Agent-supplied `screenote_url` is trusted end-to-end with false provenance: the plan's contract is "`screenote_url` is `null` from the agent and filled in by the stage," but the stage's `upload_item_to_screenote?` (`lib/hive/stages/artifacts.rb:163`) skips any item that already has a non-empty `screenote_url`, preserving an agent-prefilled value, and the web reader (`web/app/controllers/tasks_controller.rb:239-244`) renders any `https?://` URL as a "View / annotate on screenote ↗" link while U5 (`templates/finalize_prompt.md.erb`) flows it into the public PR body. A prompt-injected manifest could thus plant an arbitrary link presented as a hosted screenote artifact. Bounded (the agent already controls `artifact.md`), but the false "screenote" provenance + PR-body reach warrant nulling any stage-unwritten `screenote_url`.
- [ ] In-image TUI→GIF capture cannot actually produce a GIF: `packaging/docker/Dockerfile:14` installs `asciinema` + `ffmpeg` but no `agg` or `vhs`, and `ffmpeg` cannot transcode an asciinema `.cast` to a GIF — so the prompt-required "record with asciinema/vhs and convert to a GIF" path always degrades to a `failed` capture in hivebox, and `packaging/docker/README.md` + `docs/visual-artifacts.md`'s claim that "asciinema and ffmpeg … can produce short GIFs" is factually wrong. U6 is plan-deferred/optional and the failure degrades safely, so this is not a scope gap — but add `agg`/`vhs` or correct the docs to name the missing GIF encoder.
- [ ] Branch is stale against `origin/main` and unmergeable-as-diffed: the 10 feature commits sit on merge-base `0d0cac16`, so the `origin/main..HEAD` diff is misleading and — more concretely — this branch's `lib/hive/config.rb` lacks origin/main's `web/**` auto-commit allowlist, which already failed pass-02's auto-commit scope check for the feature's own `web/app/controllers/tasks_controller.rb` and `web/test/integration/tasks_test.rb` (see `reviews/auto-commit-scope-02.md`). Rebase onto `origin/main` before merge so review and the fix-commit flow operate on the true integration diff.

## Nit
- [ ] `lib/hive/stages/artifacts.rb:107-113` upload budget is checked only before each request, so a single in-flight upload can overrun the 120s `SCREENOTE_UPLOAD_BUDGET_SEC` by up to one read-timeout (~60s) past the deadline. Already accepted NO-FIX in pass 02 (bounded, self-correcting, never fails the stage); restated only for completeness.

ivankuznetsov:

# pr-review-toolkit — pass 03

Multi-persona review (code-reviewer, pr-test-analyzer, silent-failure-hunter,
comment-analyzer, type-design-analyzer, code-simplifier). Union of findings below.

Scope note: this branch is diverged from `origin/main` (merge-base `0d0cac16`), so
`git diff origin/main..HEAD` folds in origin/main's later, unrelated commits shown
inverted (llm-wiki, babysitter, daemon, digest, wiki/log.md). All personas were scoped
to the feature's real changes (`0d0cac16..HEAD`); findings on those inverted files are
not this PR's defects and are excluded. The general code-reviewer pass found no
correctness/security/guideline defect at or above its reporting bar (path containment,
token-in-prompt secrecy, schema gating, multipart wire format, and resilient manifest
reads are all solid and test-backed); the items below are the other personas' findings.

## High
- [ ] [comments] `config.rb:340-343` and `docs/visual-artifacts.md:44` claim "the agent never receives the token", but it is false for the `HIVE_SCREENOTE_API_TOKEN` env path the same docs recommend (`export HIVE_SCREENOTE_API_TOKEN=...`): the artifacts agent is spawned without scrubbing it — headless `agent.rb` `Process.spawn` passes no `env:` (full inheritance), and the tmux launcher (`claude_launcher.rb:363-367` → `tmux_runner.rb:55`) deliberately blanks `ANTHROPIC_API_KEY`/`CLAUDE_API_KEY` but NOT the screenote token — so an exported token is readable by the agent's `Bash`. Violates the plan's token-isolation requirement; the per-secret scrubbing precedent already exists in the tmux launcher, so the fix (blank it alongside the Anthropic keys, or reword to the accurate "never written into the prompt" as `wiki/modules/config.md:162` already says) is clear. (Also raised in claude-ce-code-review-03; unanswered escalation Q1 from pass 02.)

## Medium
- [ ] [comments] `docs/visual-artifacts.md:71` + `packaging/docker/README.md:112-113` claim the image's `asciinema`+`ffmpeg` "can produce short GIFs", but `packaging/docker/Dockerfile:14` ships no `agg`/`vhs` and `ffmpeg` cannot read an asciinema `.cast`, so the prompt-required TUI/CLI record→GIF path dead-ends to a `failed` capture in-box. U6 is plan-deferred/optional (degrades safely), so not a scope gap — but the doc claim is factually wrong; add `agg`/`vhs` or reword. (Persona rated High; corroborates claude-ce-code-review-03.)
- [ ] [silent-failure] `normalized_media_items` (`tasks_controller.rb:230-247`) silently drops any manifest item failing `resolved_media_path` with no log line, so a `status: captured` manifest whose stills were cleaned/renamed renders an empty Demo section that is diagnosable only by hand-reading the manifest — unlike the deliberate `Rails.logger.warn` breadcrumb at the sibling `media` action (line 61). Add a warn on discard.
- [ ] [silent-failure] `parse_success` (`screenote_uploader.rb:104-116`) returns `nil` for a `201` whose body has a blank/missing `annotate_url` with NO `warn` — the lone non-logging branch (every other anomaly warns), so a real screenote contract break is invisible and looks identical to "screenote disabled".
- [ ] [silent-failure] `media_item_path` (`artifacts.rb:172-198`) rescues `ArgumentError` over the whole method body though it documents only the null-byte case, with no `warn` in the rescue — a future non-null-byte `ArgumentError` is masked as "skip item". Gated upstream by `SCREENOTE_FILENAME_RE` today, so latent; guard the null-byte case explicitly or warn in the rescue.
- [ ] [type-design][simplify] The media-manifest contract is stringly-typed and duplicated across the gem stage, the controller, and the prompt: `MEDIA_MANIFEST_SCHEMA = 1` is redeclared in `artifacts.rb:26` and `tasks_controller.rb:198`, and the filename-shape regexes in `artifacts.rb:32` / `tasks_controller.rb:194`, even though the web tier already loads the `hive` gem (it references `Hive::Config` etc.). A schema bump in one file silently leaves the other reading v2 as v1. Hoist the schema int + regex + field/enum vocabulary into one gem module both reference.
- [ ] [simplify] The symlink-safe media-path resolution core is duplicated between `media_item_path` (`artifacts.rb:172-198`) and `resolved_media_path` (`tasks_controller.rb:249-274`) — an identical 8-line trusted-root/containment core in security-sensitive code kept byte-for-byte in sync by hand; the rescue-set asymmetry (controller omits the `ArgumentError` rescue the stage has) is itself a symptom. Extract one shared gem helper, preserving each caller's pre-checks.
- [ ] [type-design] `ScreenoteUploader#upload` (`screenote_uploader.rb:31-49`) collapses not-configured, file-missing, HTTP-failure, and any exception into a single `nil`; the sole caller (`artifacts.rb:119-121`) cannot distinguish "skip" from "failed, retry later", relying on `warn` side effects for observability. A small Result value object would express the outcome axis. Medium-leaning-nit.
- [ ] [tests] No test covers the media route for a valid-shape-but-unknown task slug → `InvalidTaskPath`→404 (plan U3/U5 scenario): existing coverage only hits `show` (unknown project / constraint-blocked traversal). Behavior is correct via the shared `task_row!`+`rescue_from` (covered through `show`), so risk is bounded — but the media endpoint's 404 contract is unasserted. (Persona rated High; moderated since the shared codepath is tested via `show`.)
- [ ] [tests] Partial budget-exhaustion is untested: `artifacts_test.rb:334-374` trips the deadline on the FIRST iteration (zero uploads). The realistic path — item 1 uploads + writes its URL, budget trips before item 2, manifest partially rewritten + folder touched + remaining still left blank — exercises the `changed`/`break`/`File.write`/`FileUtils.touch` interaction the production comment documents, and is unasserted. (Persona rated High; moderated — components are individually exercised.)
- [ ] [tests] The dedicated `rescue Hive::ConfigError` branch in `push_manifest_media_to_screenote` (`artifacts.rb:138-143`) is untested; a regression that let `ConfigError` escape would crash the `:complete` path (`run!` calls it at `artifacts.rb:43`). Only the corrupt-JSON and no-config branches are covered.

## Nit
- [ ] [code-review][comments] `ScreenoteUploader#append_file` (`screenote_uploader.rb:78-86`) interpolates the filename into the BINARY buffer without `.b`, unlike the title fix at line 74; a non-ASCII on-disk filename would raise `Encoding::CompatibilityError` before the image bytes append. Unreachable from production (`SCREENOTE_FILENAME_RE` gates `[\w.-]`), so a latent edge in the standalone class only.
- [ ] [comments] `SUCCESS_CODE = "201"` (`screenote_uploader.rb:10`) hardcodes only-201-is-success (a `200 OK` would be treated as a failure) with no comment noting the external-API assumption.
- [ ] [comments] The `quote()` comment (`screenote_uploader.rb:88-95`) references a prior-commit regression ("so only that branch had regressed") with no in-repo anchor — orphaned transitional context once the PR lands; the "why block form" rationale is the durable part.
- [ ] [comments] `artifacts.rb:30-31`/`:168` states "Screenote only hosts PNG/JPEG stills, never GIFs" as a third-party-service fact unverifiable from the repo; it is actually a Hive-side policy (the prompt + `push_to_screenote` gating decide it). Reword to what this code controls.
- [ ] [tests][type-design] `ScreenoteUploader` returns `screenshot_id` (`screenote_uploader.rb:120`) but no production code consumes it (stage reads only `annotate_url`); the test asserts a pass-through purpose ("for the manifest") that nothing fulfills. Drop the field until a consumer needs it, or mark it forward-looking.
- [ ] [tests] `surface` is normalized and carried into `@media` (`tasks_controller.rb:222`) but never rendered or asserted — effectively dead data in the output.
- [ ] [tests] A captured manifest with a literal empty `items: []` is not separately asserted on the web side (it rides the all-items-filtered-out test); `_media.html.erb:15` guards with `items.present?`, so behavior is correct but unpinned.
- [ ] [simplify] The media filename shape appears three times (`SCREENOTE_FILENAME_RE`, `MEDIA_FILENAME_RE`, the route constraint in `routes.rb:29`); the route copy is Rails-forced and the screenote-vs-display GIF divergence is intentional — leave, but a new extension must be updated in lockstep.
- [ ] [simplify] `normalized_media_items` re-validates the filename (`basename` + regex, `tasks_controller.rb:235-237`) that `resolved_media_path` validates again — cheap, intentional double-check; collapsible only if the resolver is extracted (see Medium).
- [ ] [type-design] The "present token" invariant is encoded in three places (`validate_screenote!`, the stage's `screenote_uploader` factory at `artifacts.rb:148-153`, and `ScreenoteUploader#configured?`); defensible defense-in-depth, noted only because the factory check duplicates the constructor's.
- [ ] [silent-failure] The plan-mandated aggregate `StandardError` rescue in `push_manifest_media_to_screenote` (`artifacts.rb:144-146`) also wraps `File.write`/`JSON.pretty_generate`/`FileUtils.touch`, so a manifest-write failure (disk full, permissions) is reported only as a generic "manifest processing failed" warn rather than pointing at the persistence step. Acceptable per the fail-soft contract; noted.

ivankuznetsov:

Outcome: timeout

Agent tail:

```
Still waiting on the final coverage report; the full suite is long but remains active.
```

ivankuznetsov:

Outcome: timeout

Agent tail:

```
I’m midway through the direct Actions poll for the Ruby job. Local coverage already matches the CI command, so this is now just waiting for GitHub’s final result.
```

PR #521, opened 2026-06-18
feat(tasks): add depends_on dependency gate and stacked PRs

ivankuznetsov:
## Summary

Tasks can now declare a single `depends_on: <id-or-slug>` prerequisite. While the prerequisite has not reached the configured gate stage (default `8-finalize`, i.e. PR open), the daemon freezes the dependent in place — every auto-advance and dispatch is suppressed, and the freeze is logged. Once the prerequisite clears the gate, the dependent flows normally: at `4-execute` its feature worktree branches off the prerequisite's branch, and at `5-open-pr` its PR opens with `--base <prerequisite-branch>` (a live stacked PR). Humans can always override the gate manually (`hive run` / `mv` / `hive approve`); only the daemon honors it.

Before this, tasks had no way to express ordering: a dependent would advance independently and always branch off the default branch, even when it logically built on another in-flight task.

### Architecture

Dependency state is computed **once**, inside `hive status`, by a pure I/O-free resolver, and every consumer reads back that single source of truth. The daemon does not re-derive the threshold comparison — it trusts the `blocked` flag exactly as it already trusts `stage`/`action`/`marker` from the same status payload.

```mermaid
graph TB
  meta["meta.yml: depends_on"] --> resolver["Hive::Dependencies (pure resolver)"]
  resolver --> status["hive status --json: blocked / blocked_by / dependency_stage"]
  status --> gate["daemon Policy: blocked → :blocked_on_dependency (freeze)"]
  status --> indicator["text/TUI: ⏸ blocked by slug (stage)"]
  resolver --> base["base_branch_for → execute worktree base + open_pr --base"]
```

### Design decisions

| Decision | Rationale |
| --- | --- |
| `blocked` computed in `hive status`, honored by the daemon | One place owns the threshold comparison; status, TUI, and the gate can't drift |
| `depends_on` stored in `meta.yml` | The only per-task file durable across every stage transition; already read for every task |
| Unresolved dependency reuses the four status fields (`blocked: true`, `blocked_by: null`, `dependency_stage: null`) + a distinct `(unresolved)` text suffix | Typo / deleted / rejected prereq stays blocked without a fifth JSON field |
| Daemon-only gate | Freeze-in-place intercepts only the dispatch/advance branch; manual paths never pass through the dispatcher, so humans always override |
| Merged-prereq fallback to `default_branch` | Once the stack's base merges and GitHub deletes the branch, both the worktree base and PR `--base` fall back to the default branch and rely on GitHub auto-retarget — no auto-restack |
| `dependency_gate_stage` global config knob | Single knob, default `8-finalize`, alternative `9-done`, validated at config load |

The status schema bumps `hive-status` v3 → v4 (`schemas/hive-status.v4.json`) because the daemon now depends on the `blocked` field. This correctly forces the daemon and the `hive` binary to be reinstalled together.

### Scope

In: single `depends_on` value, same-project resolution, the pure resolver, the config knob, the daemon freeze, stacked worktree base + PR `--base`, status JSON fields + inline indicator, unit tests for the eligibility and base-branch decisions.

Out (deferred): multiple dependencies, full cycle detection (only a degenerate self-reference guard ships — 2-cycles freeze both tasks by design and surface as blocked), auto-restack at merge time, a real-git end-to-end stacked-branch test, and a per-task threshold override.

## Test plan

Unit coverage lands the two decision points the design hinges on:

- `test/unit/dependencies_test.rb` — eligibility against a synthetic snapshot, slug/id resolution, unresolved + self-reference (slug and numeric), threshold honoring, base-branch selection.
- `test/unit/dependency_snapshot_test.rb` — the disk→resolver glue (`tasks`, `current_task`, `depends_on`, non-directory skipping).
- `test/unit/worktree_test.rb` — override base fetch with a deterministic colon-refspec tracking ref (so narrow/single-branch clones still stack), and the merged-prereq fallback.
- Wiring specs: `policy_test`, `dispatcher_test` (incl. the `unresolved:` discriminator and the single-source-of-truth pin), `status_consumer_test`, `status_test`, `config_test`, `new_test`, `migrate_test` (`depends_on` survives id-backfill), `open_pr_test`, `execute_test`, `tasks_pane_test`, and `schema_files_test` (v3 pin + populated-v4 validation).

Full suite green: **5468 runs, 21360 assertions, 0 failures, 0 errors, 2 skips** (`bundle exec rake test`). Rubocop clean on touched lib files. Real-git end-to-end stacked-branch test deferred per plan A6.

## Review summary

Three review passes ran (compound-engineering code review, pr-review-toolkit multi-persona, and codex), each with adversarial triage. **0 escalations across all passes** — every finding was either auto-fixed or resolved from plan / prior decisions.

Notable fixes applied:

- **Fleet-freeze guard**: per-project isolation in `hive status` so one project's invalid `dependency_gate_stage` can no longer abort the whole status fetch and freeze daemon dispatch globally.
- **Narrow-refspec stack-collapse** (reproduced live): `git fetch origin <branch>` only wrote `FETCH_HEAD` on single-branch/shallow clones, silently collapsing the stack to the default base. Fixed with an explicit colon-refspec write of `refs/remotes/origin/<branch>` + a regression test.
- **Observability**: surfaced the status subprocess's stderr breadcrumbs to `daemon.log` on a successful fetch, added the `:status_warning` event, and named the gate-loosening direction in the degrade warning so a fail-open is never silent.
- **Reliability**: HTTPS fetch low-speed timeout, narrowed `TaskMeta.read` rescue so a corrupt `meta.yml` warns instead of silently dropping `depends_on`.
- **Type/maintainability**: derived `Result#unresolved?` predicate, shared `blocked_label`/`stacked_base` helpers, dead-code removals, CLI wording + examples, comment/YARD corrections.

The stranded `wiki/log.md` recompile commit (`9324cbbc`) flagged in passes 01–02 was dropped via rebase; `git diff 5035bcaa..HEAD` no longer touches `wiki/log.md` (only adds the feature's own `wiki/log.d/` fragments). Deliberate scope boundaries (no fifth `unresolved` JSON field, same-project-only resolution, no auto-restack, deferred real-git e2e, 2-cycles freeze forever) were adjudicated and left as designed.

## Linked task

- Hive task: **#1452** — Task Dependencies With Stacked PRs (`task-dependencies-depends-on-gate-260618-4418`)
- Pipeline: planned via brainstorm decisions A1–A6, implemented across units U1–U7, finalized at stage `8-finalize`.

---


ivankuznetsov:

## High
- [ ] Babysitter dry-run artifacts are no longer cleaned or ignored: `DryRunEnv.with_env` now leaves `.hive-babysitter-dry-run-bin/` behind and the branch removed the `.gitignore` patterns for the overlay/skip files, so a dry-run babysitter pass can leave untracked out-of-scope residue that CleanExit will stage and fail with `ensure_clean_on_exit_failed`.

## Medium

## Nit

ivankuznetsov:

# Code Review — task-dependencies (`depends_on` gate + stacked PRs)

Scope: branch contribution vs merge-base `5035bcaa` (`git diff origin/main...HEAD`). The
two-dot `origin/main..HEAD` is polluted by two unrelated PRs (#519 babysitter, #520 digest)
that landed on main after this branch forked — their spurious reverse-diffs were excluded.
Reviewers: correctness, adversarial, reliability, api-contract, cli-readiness, testing,
maintainability, project-standards, agent-native. All affected unit tests pass locally.

Findings that flag deliberate plan scope boundaries were dropped (no fifth `unresolved` JSON
field per Design Decision #3; deferred `hive depends`/edit command; same-project-only
resolution per A2; deferred auto-restack per A3; deferred real-git e2e per A6).

## High
- [ ] Unrescued per-project `Config.load` in `annotate_dependencies` can freeze daemon dispatch for ALL projects (lib/hive/commands/status.rb:537): `Hive::Config.load(project.fetch("path")).fetch("dependency_gate_stage")` runs for every project with no rescue, and `json_payload`'s `projects.map` (status.rb:106) has no per-project isolation — one project with an invalid `config.yml` (e.g. `dependency_gate_stage: 5-open-pr`) raises `ConfigError`, aborts the entire `hive status --json`, the daemon's `StatusConsumer` then gets `ok:false` and skips the whole tick, freezing auto-advance globally; this is a regression because the pre-existing per-project `Config.load` (via `worktree_path`) is wrapped in `rescue StandardError → nil` (status.rb:465) while this new path is not. Wrap it to fall back to `Hive::Config::DEFAULTS["dependency_gate_stage"]` or isolate `project_payload` per project.
- [ ] Feature branch directly edits compiled `wiki/log.md` and bundles unrelated stranded commits (commit 9324cbbc, +3619 lines to wiki/log.md plus three log.d fragments for web/codex/finalize work unrelated to this feature): violates the explicit CLAUDE.md rule "do not edit compiled wiki/log.md directly in feature PRs" and pollutes the PR / risks log.md merge conflicts — drop commit 9324cbbc from this branch (the feature's own log.d fragment `20260618T194328Z-task-dependencies.md` + module page + index entry are correct and should stay).

## Medium
- [ ] HTTPS `git fetch` has no transfer/low-speed timeout (lib/hive/worktree.rb:7-10, `NONINTERACTIVE_FETCH_ENV`): the env only disables prompts and bounds SSH connect (`ConnectTimeout=10`); on an HTTPS origin a stalled `git fetch` can hang indefinitely. This branch extends fetches to the non-daemonized `4-execute` (`freshest_override_base`) and `5-open-pr` (`origin_branch_exists?`) stage paths, so a stalled remote wedges the stage. Add `GIT_HTTP_LOW_SPEED_LIMIT`/`GIT_HTTP_LOW_SPEED_TIME` (pattern is partly pre-existing in `freshest_base`).
- [ ] `origin_branch_exists?` issues a network `git fetch` on every open-pr run for a dependent (lib/hive/stages/open_pr.rb:202, lib/hive/worktree.rb:184): couples the PR-open path to remote availability/latency on every retry, with no caching and no reuse of the worktree's recorded base — derive the PR base from the existing `worktree.yml` base record or memoize the probe.
- [ ] Text-mode resolved indicator `⏸ blocked by <slug> (<stage>)` is untested via `render_project` (lib/hive/commands/status.rb:421-428 `dependency_indicator` else-branch): only the `(unresolved)` variant is covered in status_test; the resolved branch is exercised solely through the TUI's separate renderer, so swapping the two branches in `Commands::Status` would pass every test. Add a `render_project` test for a below-gate resolvable prereq.
- [ ] Duplicated blocked-message rendering with divergent unresolved-detection predicates (lib/hive/commands/status.rb:418-428 vs lib/hive/tui/views/tasks_pane.rb:184-196): status checks `row[:unresolved_dependency]` while the TUI checks `row.blocked_by.to_s.empty?`, and the row key `:unresolved_dependency` is an asymmetric rename of `Dependencies::Result#unresolved` (status.rb:557) — they agree today only because unresolved ⟺ `blocked_by==nil`; extract one shared `blocked_label` helper and align the row key to `Result#unresolved`.
- [ ] `:blocked` daemon log event cannot distinguish unresolved from resolved-blocked (lib/hive/daemon/dispatcher.rb:772-779): an unresolved dependency emits `reason: "dependency_unmet"` with `blocked_by: nil`/`dependency_stage: nil`, identical in shape to a real waiting-on-prereq block, so a mistyped `depends_on` (a config error) is indistinguishable in the log — add `unresolved:` to the event or use a distinct reason string.
- [ ] Resolution wiring duplicated between `dependency_base_override` (lib/hive/stages/execute.rb:101-112) and `dependency_pr_base_branch` (lib/hive/stages/open_pr.rb:190-204): both read `depends_on`, build the snapshot, call `Dependencies.base_branch_for`, and null-out when equal to default — a future change to the resolution call must be made in both places; extract the shared scaffolding (e.g. `Dependencies.resolved_base_branch(task, default_branch)`), leaving open-pr's extra origin-existence guard.

## Nit
- [ ] `--depends-on` option description says "until this task id or slug reaches PR-open" (lib/hive/cli.rb:290) but the default gate is `8-finalize`, not `5-open-pr` — the imprecision could lead an operator to set `dependency_gate_stage: 5-open-pr`, which the new validator rejects and (per the High finding) freezes status; align the wording with the actual default threshold.
- [ ] `validate_dependency!` error message exposes raw regex source `#{SLUG_RE.source}` (lib/hive/commands/new.rb:170-174) rather than a plain-English format — keep the regex in the structured `value:` field but make the human/agent-facing message describe the accepted shape (positive integer id, or lowercase slug starting with a letter).
- [ ] `hive new` has no `long_desc`/examples for `--depends-on` (lib/hive/cli.rb:289), unlike most other commands — an agent must trial-and-error the accepted formats; add examples for both id and slug forms.
- [ ] Dead tertiary `:dependency_stage` fallback in `stage_index_for` (lib/hive/dependencies.rb:78-79): no production caller produces a task shape that reaches it (the `dependency_stage` field is the *prerequisite's* stage, not the task's own), and it would silently misfire on a wrong shape rather than raising — remove it and let the existing `ArgumentError` stand.
- [ ] `DEPENDENCY_GATE_STAGES` hardcodes the last two `Hive::Stages::DIRS` entries (lib/hive/config.rb:515): a stage rename/addition won't propagate — derive it as `Hive::Stages::DIRS.last(2)` or add a cross-reference comment.
- [ ] `DependencySnapshot#current_task` and `#depends_on` carry unreachable duck-typing fallbacks (lib/hive/dependency_snapshot.rb:28-39): production callers always pass a `Hive::Task`, which already responds to `id`/`depends_on`, so the `respond_to?` guard and `TaskMeta.read` fallback are dead — simplify and document the expected type.
- [ ] Cosmetic warning inaccuracy (lib/hive/worktree.rb:114-118): `freshest_override_base` warns "branching from #{default_branch}" then calls `freshest_base`, which on success returns `origin/<default>`, not local `<default>` — the chosen base is correct; only the message names the wrong ref.
- [ ] Leftover local branch named `<slug>` makes `create!` take the attach path and silently ignore `base_override` (lib/hive/worktree.rb:40-44): only reachable on a re-run with a stale branch (pre-existing attach behavior; `base_override` is honored on normal first creation) — note in a comment or guard.
- [ ] `DependencySnapshot.tasks` does not dedupe transient stage-moves the way `collect_rows`→`drop_transient_stage_moves` does (lib/hive/dependency_snapshot.rb:8): during a folder-rename race a prerequisite could transiently vanish from the disk scan and the dependent would branch off `default_branch` with no warning — narrow timing window, but the stacking decision uses a different snapshot view than the `blocked` decision.
- [ ] No back-compat pin test for `schemas/hive-status.v3.json` (test/unit/schema_files_test.rb): the v3 file is preserved and the daemon rejects v3 payloads (fails closed via schema-skew), so correctness holds, but unlike `hive-approve` there is no test guarding v3 against accidental mutation — add one.
- [ ] Additional unit coverage worth adding: `base_branch_for` for a resolvable prereq *below* threshold (asserts it's threshold-blind by design), `dependency_pr_base_branch` self-reference → nil, `dependency_base_override` missing-prereq → nil, `freshest_override_base` fetch-succeeds-but-ref-absent fallback, and a multi-project `hive status --json` where one project's config is invalid (pins the High finding's fix).
- [ ] `hive new` emits no `--json` envelope (lib/hive/commands/new.rb:131, pre-existing): an agent chaining tasks cannot read back the new task's `id` to pass as a later `--depends-on` without a follow-up `hive status --json` — pre-existing gap made more relevant by this feature.

ivankuznetsov:

# PR Review (pr-review-toolkit) — task-dependencies (`depends_on` gate + stacked PRs)

Scope: branch contribution vs merge-base `5035bcaa` (`git diff origin/main...HEAD`). The
two-dot `origin/main..HEAD` is polluted by two unrelated PRs (#519, #520) merged to main after
this branch forked — their spurious reverse-diffs were excluded. Personas run: code-reviewer,
pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier.
All affected unit tests pass; `bundle exec rubocop` was run to verify lint findings.

Findings that flag deliberate plan scope boundaries were dropped (no fifth `unresolved` JSON
field per Design Decision #3; `DEPENDENCY_GATE_STAGES` kept as an explicit policy literal rather
than derived from `Stages::DIRS`; deferred multi-dep / cycle-detection / auto-restack / real-git
e2e).

## High
- [ ] [silent-failure-hunter] Unrescued `annotate_dependencies` can freeze daemon dispatch for ALL projects (lib/hive/commands/status.rb:537, invoked unguarded inside `projects.map` at :106): `Hive::Config.load(project.fetch("path"))` (raises `ConfigError` on any malformed project config, including an invalid `dependency_gate_stage`) and `Hive::Dependencies.resolve` (`stage_index_for`/`threshold_index` raise `ArgumentError`) run with no rescue and no per-project isolation, so one bad project aborts the entire `hive status --json` → `StatusConsumer` gets `ok:false` → the dispatcher skips the whole tick → auto-advance freezes fleet-wide, surfaced only in daemon.log; wrap `project_payload` per-project and degrade `annotate_dependencies` per-row (fall back to `Config::DEFAULTS` threshold / fail-open with a warn).

## Medium
- [ ] [code-reviewer] Rubocop offenses introduced by this branch in lib/hive/worktree.rb:60-62 (`Layout/ElseAlignment` + `Layout/EndAlignment` on the new `base = if/else/end` block): confirmed via `bundle exec rubocop` (2 correctable offenses), violating the CLAUDE.md rule "always run necessary validators and linters after file edit" — run `rubocop -a lib/hive/worktree.rb` or align the `else`/`end` to the `base = if` column.
- [ ] [silent-failure-hunter] `base_branch_for` fails open silently while the gate fails closed (lib/hive/dependencies.rb:23-31; callers lib/hive/stages/execute.rb:101-112 and lib/hive/stages/open_pr.rb:190-204): a set-but-unresolvable `depends_on` (typo, or prereq transiently missing from the snapshot) makes the worktree/PR base silently fall back to `default_branch` with no warning at execute/open-pr time, so the requested stack silently collapses — add a `warn` at the stage call sites when a non-empty `depends_on` does not resolve to a stacked base.
- [ ] [silent-failure-hunter] `TaskMeta.read`'s broad `rescue StandardError => empty` now silently drops `depends_on` on partial meta.yml corruption (lib/hive/task_meta.rb:15-27): a YAML/permission/encoding error returns an empty meta, the resolver reads `depends_on: nil` as "no dependency" → `blocked:false`, and the daemon dispatches the dependent ahead of its prerequisite — narrow the rescue to `Psych::Exception, SystemCallError, IOError` (+ explicit `Errno::ENOENT → empty`) and log, matching the existing `read_pointer` pattern (rescue is pre-existing; the fail-open gate consequence is new).
- [ ] [silent-failure-hunter] `freshest_override_base` no-origin branch drops the stacked base with NO warning (lib/hive/worktree.rb:116): when `origin` is unconfigured it silently `return freshest_base(default_branch)`, unlike the fetch-failure/ref-absent branches which warn — add a matching warn so a dropped stack request is observable.
- [ ] [comment-analyzer] `Policy` module doc block is stale and self-contradictory (lib/hive/daemon/policy.rb:3-13): the prose says "one of four outcomes" and omits `:blocked_on_dependency` (and `:record_baseline`), while the authoritative `@return` tag at :98-100 lists `:blocked_on_dependency` — update the bulleted list and drop the brittle hard count.
- [ ] [type-design-analyzer] `Dependencies::Result` stores `unresolved` as a 4th field redundant with `blocked_by.nil? && blocked`, permitting contradictory tuples (lib/hive/dependencies.rb:5): nothing prevents constructing `unresolved:true + blocked:false` or `blocked_by:"x" + unresolved:true`; drop the field and expose `unresolved?` as a derived predicate so the correlation holds by construction.
- [ ] [type-design-analyzer] Three divergent representations of the same "unresolved" state (lib/hive/dependencies.rb:5 `Result#unresolved` → lib/hive/commands/status.rb:557 renamed row key `row[:unresolved_dependency]` → lib/hive/tui/views/tasks_pane.rb:192 re-derived via `blocked_by.to_s.empty?`): they coincide only because `unresolved ⟺ blocked_by.nil?` today and are computed in three files, so a future change can make text-mode and the TUI disagree — converge on one shared predicate/helper.
- [ ] [pr-test-analyzer] `Hive::DependencySnapshot` has zero direct tests: the disk→resolver seam (`tasks`, `current_task`, `depends_on`, non-directory skipping, the `TaskMeta.read` fallback) is exercised only indirectly through execute/open_pr happy-path stubs, so a regression in the glue would pass every resolver test — add `test/unit/dependency_snapshot_test.rb`.
- [ ] [pr-test-analyzer] No status-level test wires `dependency_gate_stage` override to the resolver: nothing asserts that writing `dependency_gate_stage: 9-done` into a project config makes a prereq at `8-finalize` report `blocked:true` in `hive status --json` (config→`annotate_dependencies` threshold path is unverified end-to-end) — add a status_test case.

## Nit
- [ ] [comment-analyzer] `freshest_override_base` fallback warnings say "branching from #{default_branch}" (lib/hive/worktree.rb:121,126) but the code then calls `freshest_base`, which returns `origin/<default>` in the common case — the message names the wrong ref and can mislead an operator debugging a collapsed stack (corroborates the cosmetic warning noted in the ce-code-review pass).
- [ ] [comment-analyzer] Edit-resume gating comment under-describes the code (lib/hive/daemon/policy.rb:131-136): it mentions only the brainstorm answers-pending hold, but the code now checks `blocked` first (→ `:blocked_on_dependency`) and only then `answers_pending` — note both gates and their precedence.
- [ ] [pr-test-analyzer] No test for the `blocked` over `answers_pending` precedence on a debounced edit-resume (a row both blocked and answers-pending must resolve to `:blocked_on_dependency`, not `:wait_for_answers`) — a future reordering of those branches would pass CI.
- [ ] [pr-test-analyzer] Self-reference `depends_on` is asserted only at the resolver level, not through the status JSON path (the `same_task?` hash-input branch runs in production via `task: row` but is unasserted there) — add a status_test for a task depending on its own slug → `blocked:true, blocked_by:nil`.
- [ ] [pr-test-analyzer] No dispatcher test confirms a blocked first-sight edit-resume row still records its mtime baseline (the policy `:record_baseline` arm is covered, but not threaded through `tick` with `blocked:true`).
- [ ] [pr-test-analyzer] `base_branch_for` is not asserted for a prereq at/after the gate (e.g. `9-done`): the threshold-blind stacking contract (returns the prereq slug regardless of stage) is only implicitly covered at `stage_index:8`.
- [ ] [silent-failure-hunter] `Dependencies.task_id` swallows `ArgumentError, TypeError → nil` (lib/hive/dependencies.rb:63-71): a malformed prereq `id` in meta.yml makes a numeric `depends_on` silently mis-resolve (fails closed in the gate, open in `base_branch_for`) with no breadcrumb distinguishing "typo" from "corrupt prereq id" — add a debug warn.
- [ ] [silent-failure-hunter] `row.fetch(:task)` in the snapshot builder is an unguarded `KeyError` on the never-fail status surface (lib/hive/commands/status.rb:491): unreachable today (production rows always carry `:task`), but a future row source lacking it would land in the High-finding blast radius — guard with `filter_map`/`next`.
- [ ] [code-simplifier] Dead tertiary `:dependency_stage` lookup in `stage_index_for` (lib/hive/dependencies.rb:78-79): no production caller or test produces a task with `dependency_stage` but no `stage`/`stage_index`, so the branch is unreachable — remove it.
- [ ] [code-simplifier] Unreachable `respond_to?` guards in `DependencySnapshot.current_task`/`depends_on` (lib/hive/dependency_snapshot.rb:28-39): both production callers pass a `Hive::Task` that always responds to `id`/`depends_on`/`folder`, so the duck-typing and `TaskMeta.read` fallback are dead defensiveness — simplify to direct calls.
- [ ] [code-simplifier] Duplicated dependency-base resolution skeleton between `dependency_base_override` (lib/hive/stages/execute.rb:101-112) and `dependency_pr_base_branch` (lib/hive/stages/open_pr.rb:190-204) — extract a shared `Dependencies.stacked_base(task, default_branch:)`, leaving open_pr's extra `origin_branch_exists?` guard.
- [ ] [code-simplifier] Duplicated `⏸ blocked by …` string rendering between `Commands::Status#dependency_indicator` (lib/hive/commands/status.rb:419-426) and `Tui::Views::TasksPane#dependency_status` (lib/hive/tui/views/tasks_pane.rb:191-197), with different unresolved discriminators — extract one shared formatter (relates to the representation-divergence Medium above).
- [ ] [code-reviewer] `default_branch` source differs between the two stacking call sites (lib/hive/stages/open_pr.rb:194 uses `cfg["default_branch"] || GitOps#default_branch`; lib/hive/stages/execute.rb's `dependency_base_override` uses `ops.default_branch` without consulting config) — harmless today since a task slug never equals the default branch, but worth aligning if those values ever diverge.
- [ ] [type-design-analyzer] `field(task, key)` is an under-specified four-way input contract (Hash-or-object × symbol-or-string, lib/hive/dependencies.rb:101-107) that returns `nil` silently on any unrecognized shape; every production caller passes symbol keys, so narrow the contract or make it raise/log on a fully-unrecognized object instead of degrading to "field missing".
- [ ] [type-design-analyzer] The tuple `blocked:true, unresolved:false, blocked_by:nil` is representable and would render as "unresolved" (lib/hive/dependencies.rb:20): unreachable today because all snapshot builders guarantee a slug, but `resolve` does not assert "resolved ⟹ slug present" — add the guard so the misrender state is unrepresentable.

ivankuznetsov:

## High
- [ ] `hive status` ignores invalid `dependency_gate_stage` and uses the default: `lib/hive/commands/status.rb:588` rescues `Config.load` failures and substitutes `8-finalize`, so a typo such as an intended `9-done` gate can silently unblock dependents at PR-open time instead of failing fast as the plan requires.

## Medium

## Nit

ivankuznetsov:

# Code Review (pass 02) — task-dependencies (`depends_on` gate + stacked PRs)

Scope: branch contribution vs merge-base `5035bcaa` (`git diff origin/main...HEAD`); the
two-dot `origin/main..HEAD` is polluted by unrelated PRs #519 (babysitter) / #520 (digest)
that landed on main after this branch forked — confirmed those reverse-diffs are absent from
the three-dot feature diff and excluded. Reviewers: correctness, adversarial, reliability,
api-contract, project-standards, maintainability, testing, agent-native. All touched unit
tests pass locally (dependencies, dependency_snapshot, policy, dispatcher, status_consumer,
config, status, execute, open_pr, worktree, tui, schema_files, new) and rubocop is clean on
all 10 touched lib files.

Pass-01 was reviewed in its post-fix (`dbb998c8`) state. Most pass-01 findings were applied
and verified (HTTPS fetch timeout, `project_payload_or_degraded` per-project isolation,
shared `blocked_label`/`stacked_base` helpers, `Result#unresolved?`, dispatcher `unresolved:`
event field, dead-code removals, CLI wording/long_desc, schema-v3 pin test, the added unit
set). Findings below are limited to what remains unaddressed or newly observed. Plan-deferred
scope boundaries (no fifth `unresolved` JSON field, deferred `hive depends`/edit, same-project
resolution, no auto-restack, deferred real-git e2e) were dropped per the anti-finding rule.

## High
- [ ] Pass-01 AUTO-FIX "drop stranded commit `9324cbbc`" was NOT applied — it is still an ancestor of HEAD: commit `9324cbbc` directly edits the compiled `wiki/log.md` (+3612 lines), violating the explicit CLAUDE.md rule "do not edit compiled wiki/log.md directly in feature PRs," and bundles unrelated web/codex/finalize/llm-wiki-refresh wiki churn (`wiki/active-areas.md`, `wiki/commands/web.md`, `wiki/decisions.md`, `wiki/modules/{agent,reviewers}.md`, `wiki/stages/{brainstorm,finalize,review}.md`, `wiki/testing.md`, plus three log.d fragments — `20260615T222240Z-agent-finalize-review-refresh`, `20260616T223044Z-web-codex-review-tmux-refresh`, `20260617T222312Z-llm-wiki-refresh-audit`) that have nothing to do with task dependencies: `fix-success-01.md` records the pass complete but this history rewrite was skipped, so the PR still pollutes log.md and risks merge conflicts — `rebase --onto 5035bcaa` to drop it, keeping the feature's own `20260618T194328Z-task-dependencies.md` fragment + `wiki/modules/task_dependencies.md` + index/gaps entries.

## Medium
- [ ] Invalid `dependency_gate_stage` degrades silently to the default on the gate-computing surface (lib/hive/commands/status.rb:588-595 `dependency_gate_stage_for`): an operator-intended `9-done` typo (any value `Config.load` rejects) is swallowed and the gate falls back to `8-finalize` with only a stderr breadcrumb — the misconfiguration is surfaced nowhere in the `hive status --json` payload (no project-level `error`/`warning`/`diagnostic`), so the daemon (the primary consumer that gates on this) and any agent reading the JSON silently use the wrong threshold, unblocking dependents one stage earlier than configured; this diverges from the plan U3 fail-fast/surface intent. The fleet-freeze risk pass-01 fixed is already handled independently by `project_payload_or_degraded`, so the inner rescue mostly just hides the bad config rather than surfacing it — emit a payload-level diagnostic on the degraded project so the gate degradation is observable to JSON consumers, not just to whoever tails daemon.log. (Behavior is a deliberate, tested contract — `test_json_payload_isolates_a_project_with_invalid_config` — so this is a refinement of that tradeoff, not a regression.)

## Nit
- [ ] TUI hides a blocked row's underlying action state where text mode shows both (lib/hive/tui/views/tasks_pane.rb:184-190 `status_label` returns the dependency block label and short-circuits, vs lib/hive/commands/status.rb:423 which appends `dependency_indicator` to `state_label` via `compact.join(" ")`): a task that is `blocked` AND in an `error`/`recover_review` state (reachable when a human manually `hive run`s a frozen dependent) renders only `⏸ blocked by …` in the TUI, dropping the error/recover label that text mode would still surface — the two renderers already share `blocked_label` for the unresolved discriminator, so align the composition too (append rather than replace) for full text/TUI parity.

ivankuznetsov:

# PR Review (pr-review-toolkit, pass 02) — task-dependencies (`depends_on` gate + stacked PRs)

Multi-persona review (code-reviewer, silent-failure-hunter, comment-analyzer, pr-test-analyzer,
type-design-analyzer, code-simplifier) over the feature contribution `git diff 5035bcaa..HEAD`
(merge-base `5035bcaa`); the two-dot `origin/main..HEAD` is polluted by unrelated PRs #519
(babysitter) / #520 (digest) which were excluded. Findings are the union across personas,
deduped. Near-duplicate breadcrumb-visibility findings were consolidated into one root-cause
item; the `field`-comment overstatement (surfaced by 3 personas) is reported once with
corroboration noted. One code-reviewer sub-threshold item ("double `git fetch` in open_pr") was
dropped as a misread — `stacked_base`→`base_branch_for` is pure and issues no fetch, so open_pr
fetches origin only once via `origin_branch_exists?`. Plan-deferred scope (multi-dep, full cycle
detection, auto-restack, real-git e2e, fifth `unresolved` field, same-project-only) was not flagged.

All 16 touched test suites pass locally and rubocop is clean. code-reviewer reported **no findings
at or above its confidence-80 bar** (the gate, the unresolved invariant, per-project isolation, and
the schema/`required` contract were each checked and judged sound).

## High
- [ ] [silent-failure] The daemon discards the status subprocess's stderr on a successful (exit-0) fetch (lib/hive/daemon/status_consumer.rb:47-50 — `err` is read only in the `unless status.success?` branch), so every fail-open dependency-gate breadcrumb the feature added lands on stderr that is thrown away: `apply_dependency_result` "treating as unblocked" (status.rb:613), `dependency_gate_stage_for` "using default" (status.rb:592), `TaskMeta.read` "depends_on dropped" (task_meta.rb:36), `Dependencies#task_id`/`#field` warns (dependencies.rb:106,155), and `stacked_base` "stack collapsed to default" (dependency_snapshot.rb:65): when the gate fails open under a compound error (corrupt meta.yml / config-load failure / resolver crash / corrupt prereq id), the daemon can dispatch a dependent ahead of its prerequisite with **zero trace in daemon.log**, contradicting A2's "never silently advance" — fix by emitting the status subprocess's non-empty stderr as a daemon event on successful fetches, and/or by carrying the signal in the payload (see Medium below).

## Medium
- [ ] [silent-failure] `apply_dependency_result` fail-open stamps `blocked:false` (lib/hive/commands/status.rb:602-618) that is indistinguishable in the JSON payload from a legitimate unblock — there is no per-row field (e.g. `blocked_reason`/`dependency_error`) letting the daemon (which reads `blocked` verbatim and never re-checks) gate or log on "resolver crashed → assumed unblocked" vs "genuinely not blocked"; a wrong-but-plausible `blocked:false` is invisible to the only consumer that acts on it.
- [ ] [silent-failure] `dependency_gate_stage_for` silently substitutes the default `8-finalize` when `Config.load` raises (lib/hive/commands/status.rb:588-595): an operator who set `9-done` loses the stricter gate (dependents release one stage earlier) if config later fails to load, and per the High finding the explanatory warn never reaches daemon.log — surface the misconfiguration in the payload rather than only on discarded stderr.
- [ ] [comment] The `field` shape comment overstates production callers (lib/hive/dependencies.rb:138-145): it names "a `Hive::Task`" as a production caller shape, but no production path passes a Task or any reader-exposing object — `Status#apply_dependency_result` and `DependencySnapshot#stacked_base` both project to symbol-keyed Hashes first, so the reader arm is test/forward-compat only; the claim directly contradicts dependency_snapshot.rb's own header ("the snapshot does not duck-type other inputs"). (Corroborated by type-design and code-simplifier as test-only over-generality.)
- [ ] [pr-test] `depends_on` preservation on the migrate path is untested (lib/hive/commands/migrate.rb:195 re-passes `depends_on: meta[:depends_on]`, but every test in test/integration/migrate_test.rb asserts `depends_on: nil`, e.g. :316-321, :356): dropping that arg during id-backfill would silently strip dependencies and no test would catch it — seed a task with a populated `depends_on` and assert it survives a full `migrate`.
- [ ] [pr-test] The dispatcher `:blocked` event's `unresolved:` discriminator is never asserted (lib/hive/daemon/dispatcher.rb:785; dispatcher_test.rb:281 asserts depends_on/blocked_by/dependency_stage but not `unresolved`, and only drives a *resolved* prereq): a regression hardcoding `unresolved:false` or dropping the field would pass CI — add a case driving an unresolved blocked row (`blocked:true, blocked_by:nil`) and assert `unresolved:true`.
- [ ] [pr-test] No JSONSchemer validation of a *populated* v4 Task against `schemas/hive-status.v4.json` (test/unit/schema_files_test.rb validates SuccessPayloads only with `tasks:[]`, and the required-keys test compares only the key set): a too-narrow field type (e.g. `dependency_stage` not allowing null, or `blocked` not boolean) would be rejected at runtime by external consumers and no test would catch it — validate one Task carrying real `depends_on`/`blocked_by`/`dependency_stage`/`blocked` values.
- [ ] [type-design] `Result` does not enforce its documented invariant by construction (lib/hive/dependencies.rb:11): the comment promises `unresolved? ⟺ (blocked && blocked_by.nil?)` "by construction," but `Data.define`'s public `.new` is unguarded, so a future second producer can build a contradictory tuple (e.g. `blocked_by:"x"` with `dependency_stage:nil`, or a `dependency_stage` with no slug) — make `.new` private behind a validating factory (or `resolved`/`unresolved`/`unblocked` constructors) so the guarantee lives in the type, not in one method.
- [ ] [type-design] The resolver's input "task snapshot" crosses the module boundary as an unnamed symbol-keyed Hash `{slug:,id:,stage:,stage_index:}` built independently in two places (lib/hive/dependency_snapshot.rb:25 and lib/hive/commands/status.rb:567): the two builders can drift (one omitting `stage_index` silently falls through `field`'s nil arm) with no construction-time check that they agree — give the resolver input a named type so the contract is enforced once.

## Nit
- [ ] [comment] The resolved⟹slug guard comment says the `ArgumentError` is isolated "per project" (lib/hive/dependencies.rb:31-32), but it is caught **per row** by `apply_dependency_result`'s rescue (status.rb:612) — the prose understates the blast-radius isolation (one bad task, not a whole project).
- [ ] [comment] The `field` string-key arm is justified as covering "config-sourced hashes" (lib/hive/dependencies.rb:140-141), but no production caller passes a string-keyed hash today — only the unit test does; reword as test-ergonomics/forward-compat.
- [ ] [type-design] `Result#unresolved?` is defined but thrown away at the JSON boundary and re-derived three times by hand (dispatcher.rb:783 `row.blocked_by.nil?`, dependencies.rb:48, and the TUI via `blocked_label`) — carry the predicate across the boundary instead of reinventing it at each consumer.
- [ ] [type-design] `StatusConsumer::Row` is a mutable `keyword_init` Struct (lib/hive/daemon/status_consumer.rb:19) while the sibling `Tui::Snapshot::Row` is a frozen `Data.define` — the daemon's gate row is the one place you'd most want immutability for a value carried verbatim into a fail-closed dispatch decision.
- [ ] [type-design] `blocked` is re-coerced with `== true` at three boundaries (status.rb:206, tui/snapshot.rb:154, status_consumer.rb:178), a tell that the in-process row carries it as a loosely-typed (possibly-nil) field even though hive-status.v4.json mandates a strict boolean — guarantee it once at the single write site.
- [ ] [type-design] `Tui::Snapshot::Row`'s initializer now defaults six fields incl. `blocked: false` (lib/hive/tui/snapshot.rb:81-85) "for older test factories," which erodes `Data.define`'s fail-fast completeness — a production builder that forgets `blocked` silently gets the dispatchable value.
- [ ] [type-design] `blocked_label` takes three loose keyword strings rather than the `Result` it conceptually renders (lib/hive/dependencies.rb:47) — the one helper designed to keep two renderers in sync can be called with a `blocked_by` that contradicts its `dependency_stage`; passing the typed `Result` would close the gap.
- [ ] [pr-test] `Hive::Dependencies.blocked_label` has no direct unit test (lib/hive/dependencies.rb:47) — both branches are pinned only end-to-end through the two renderers; a direct module test would localize a regression to the helper.
- [ ] [pr-test] The `same_task?` id-only self-reference arm is untested (lib/hive/dependencies.rb:86-88) — the self-reference test triggers by slug match; a numeric self-reference (`depends_on` = own id with differing/absent slugs) never exercises the id-equality branch.
- [ ] [pr-test] The `task_id` corrupt-prerequisite-id warn path is untested (lib/hive/dependencies.rb:101-108) — no test feeds a non-integer prereq id to confirm the breadcrumb fires and the gate degrades to unresolved rather than mis-resolving.
- [ ] [simplify] `field`'s object-reader and string-key arms (lib/hive/dependencies.rb:146-159) are broader than any production caller needs (reached only by tests), but code-simplifier advises **no change** — collapsing them would force test rewrites for no production benefit and drop the defensive warn breadcrumb; flagged only so the over-generality stays a conscious choice. (code-simplifier found no other actionable simplification — the feature is already at the right altitude, with shared `blocked_label`/`stacked_base` from pass 1.)

ivankuznetsov:

## High
- [ ] Stacked dependency branches are falsely treated as missing in narrow-refspec clones (`lib/hive/worktree.rb:145`, `lib/hive/worktree.rb:153`, `lib/hive/worktree.rb:224`): `git fetch origin <branch>` can succeed with only `FETCH_HEAD`, so worktrees fall back to default and PR prompts omit `--base`.
- [ ] Invalid `dependency_gate_stage` is swallowed by `hive status` (`lib/hive/commands/status.rb:588`): a bad config falls back to `8-finalize` instead of the plan-required fail-fast validation, allowing earlier daemon unblocks.
- [ ] Babysitter dry-run `gh` passthrough no longer blocks host/repo overrides (`bin/hive-babysitter-stub-gh:38`, `bin/hive-babysitter-stub-gh:189`, `bin/hive-babysitter-stub-gh:201`): allowlisted reads can run against agent-chosen GitHub hosts/repos with inherited enterprise auth context.

## Medium
- [ ] Native Codex reviewer accepts prompt-template echoes and drops captured failure tails (`lib/hive/reviewers/codex_review.rb:92`, `lib/hive/reviewers/codex_review.rb:130`, `lib/hive/reviewers/codex_review.rb:150`): hollow `- [ ] <finding>` output can be recorded as a valid review and usage-limit failures lose the text needed for cooldown routing.
- [ ] Babysitter dry-run artifacts are no longer ignored or removed (`.gitignore:39`, `lib/hive/babysitter/dry_run_env.rb:26`): dry-run leaves `.babysitter-dry-run-skipped.log` / `.hive-babysitter-dry-run-bin` as untracked residue that can trip CleanExit scope checks.
- [ ] Compiled wiki changelog was truncated directly (`wiki/log.md:1`): the branch deletes 3,612 generated log lines despite the repo rule to add only `wiki/log.d/*` fragments in feature PRs.

## Nit

ivankuznetsov:

# Code Review (pass 03) — task-dependencies (`depends_on` gate + stacked PRs)

Scope: the branch's own contribution vs merge-base `5035bcaa`
(`git diff 5035bcaaaac93b6bf67e60205d40f40ba854c842..HEAD`). The two-dot
`origin/main..HEAD` is polluted — origin/main advanced past the merge-base with
unrelated work (babysitter, codex_review, digest, a 3.6k-line `wiki/log.md`
compaction), which appears as spurious reverse-diffs. Those files are NOT part of
this feature and were excluded; codex-03's babysitter / codex_review / `wiki/log.md`
findings are pollution artifacts of the stale two-dot diff and do not apply.

Reviewers: correctness (always, session model), adversarial (diff >50 LOC + git
mutation), reliability, cli-readiness, api-contract, testing, maintainability,
project-standards. Touched specs pass locally (worktree, dependencies,
dependency_snapshot — 37 runs / 0 failures) and rubocop is clean on the touched
lib files. The headline Medium below was empirically reproduced with live git.

Pass-01 and pass-02 findings were applied/adjudicated and are not re-raised.
Plan-deferred scope boundaries dropped per the anti-finding rule: no 5th
`unresolved` JSON field (derived from `blocked_by` presence), invalid
`dependency_gate_stage` silent-degrade on the status surface (deliberate, tested
per-project isolation contract — observability lands via the daemon stderr
breadcrumb added in `51969d6f`), same-project-only resolution, no auto-restack,
2-task cycles freeze forever (documented in `wiki/modules/task_dependencies.md:39`),
deferred real-git e2e, and the deferred `hive depends` edit command. U1–U7 are all
implemented; no plan-required-but-missing gaps.

## High
None.

## Medium
- [ ] Stacking silently collapses to the default branch on single-branch / shallow / narrow-refspec clones (`lib/hive/worktree.rb:145-159` `freshest_override_base`, `lib/hive/worktree.rb:219-246` `origin_branch_exists?` / `fetch_origin_branch` / `origin_branch_ref_exists?`): `git fetch origin <branch>` only updates `FETCH_HEAD` — it does NOT write `refs/remotes/origin/<branch>` unless a configured refspec matches it (the opportunistic-update only fires on the wildcard `+refs/heads/*:refs/remotes/origin/*`). Reproduced live: on a `clone --single-branch` repo, `git fetch origin prereq-branch` exits 0, prints `-> FETCH_HEAD`, and leaves `refs/remotes/origin/prereq-branch` absent, so `origin_branch_ref_exists?` returns false even though the branch genuinely exists on origin. Both stacking call sites then fall back to the default base — `4-execute` branches the dependent off `origin/<default>` and `5-open-pr` omits `--base <prereq>` — silently defeating U6/U7 (the dependent loses the prereq's unmerged commits from its diff base, the exact outcome the feature exists to prevent). It fails open (no crash) so it is invisible. The primary daemon checkout (`~/Dev/hive`, a full clone with the wildcard refspec) is unaffected, which is why this is Medium not High; shallow/single-branch CI or hosted setups are affected. Fix: make the tracking ref deterministic regardless of clone config — fetch with an explicit colon-refspec `git fetch origin <branch>:refs/remotes/origin/<branch>` (or resolve the base from `FETCH_HEAD`), and add a `worktree_test` that sets a narrow `remote.origin.fetch` (or `--single-branch`) and asserts the dependent still branches from the prereq tip and the PR still gets `--base <prereq>`. Every existing override test uses `git remote add origin` (wildcard refspec), so the suite cannot catch this.

## Nit
- [ ] The collapse warning is both misleading and hard to find (`lib/hive/worktree.rb:153-155`): when the narrow-refspec collapse above fires, the only breadcrumb is `warn "origin/<branch> not found after fetch; falling back to the default branch base"` — but the branch genuinely EXISTS on origin (only the local tracking ref was never written), so an operator hunts for a missing branch that is present. It also lands on the execute/open-pr child's combined stdout/stderr per-task log (captured by ChildSupervisor), not in daemon.log or the `hive status` stderr breadcrumbs, so it is invisible at the fleet level. If the colon-refspec fix lands this resolves; otherwise differentiate the message (e.g. "local refspec did not create refs/remotes/origin/<branch>; check your clone config") so "prereq never pushed" is distinguishable from "your clone config defeated the stack."
- [ ] No regression test pins the single-source-of-truth invariant that the daemon reads `blocked` verbatim from the status JSON and never re-derives the threshold (`lib/hive/daemon/status_consumer.rb:196` → dispatcher → `lib/hive/daemon/policy.rb:129-130`). This property is the linchpin of the status↔daemon fail-open/fail-closed consistency the plan leans on (decision #2); it currently holds only by construction. A cheap guard — feed a synthetic status doc with `blocked:true` and assert the dispatcher emits `:blocked_on_dependency` without recomputing — would prevent a future refactor from silently reintroducing a second threshold comparison.

ivankuznetsov:

# PR Review Toolkit (pass 03) — task-dependencies (`depends_on` gate + stacked PRs)

Scope: the branch's own contribution vs merge-base `5035bcaa`
(`git diff 5035bcaaaac93b6bf67e60205d40f40ba854c842..HEAD`). The two-dot
`origin/main..HEAD` is polluted — origin/main advanced past the merge-base with
unrelated work (babysitter, codex_review, digest, a `wiki/log.md` compaction)
that appears as spurious reverse-diffs; those files are NOT part of this feature
and were excluded.

Personas run (parallel): code-reviewer, silent-failure-hunter, pr-test-analyzer,
type-design-analyzer, comment-analyzer, code-simplifier. Union of NEW findings
below. Rubocop is clean on the touched lib files and the touched specs pass.

Suppressed per the anti-finding rule (already adjudicated in passes 01–02 or
plan-deferred): invalid `dependency_gate_stage` fail-fast at the status surface
(deliberate tested per-project isolation contract — `validate!` fail-fast lives
in `Config.validate!`); no 5th `unresolved` JSON field (derived from `blocked_by`
presence); same-project-only resolution; no auto-restack; 2-task cycles freeze
forever (documented `wiki/modules/task_dependencies.md:39`); deferred real-git
e2e; deferred `hive depends` edit command; the single-branch/narrow-refspec
`git fetch` → FETCH_HEAD stack-collapse (recorded separately this pass in
`claude-ce-code-review-03.md`; corroborated by code-reviewer + pr-test-analyzer
here but not duplicated). code-simplifier and type-design-analyzer found no new
actionable defect (type-design confirmed the `Result`/`unresolved?` invariant and
duck-typed `field` access are sound).

## High
None. No persona raised a P0/P1.

## Medium
- [ ] [silent-failure] `dependency_gate_stage_for` degrades to the **more-permissive** `8-finalize` default on ANY `Config.load` failure (`lib/hive/commands/status.rb:588-595`): a project deliberately configured `dependency_gate_stage: 9-done` silently loosens its gate — and dispatches a dependent one stage early — on a transient load failure OR an unrelated bad config key (e.g. a typo in `claude.mode`) that aborts the whole load. This is distinct from the adjudicated fail-fast/diagnostic asks: the fix is a one-line degrade-**direction** change (fall back to the strictest stage `9-done`, or degrade the whole project to empty tasks like `project_payload_or_degraded` already does), so the degrade fails *closed* like the resolver instead of *open*. At minimum make the `warn` state the security-relevant direction ("gate LOOSENED to 8-finalize; dependents may dispatch one stage early"). Verified the rescue substitutes the looser value; confidence ~78 (real behavior, low operational likelihood since it needs a non-default gate + load failure).
- [ ] [pr-test] The narrowed `TaskMeta.read` rescue's `warn` is load-bearing but untested (`lib/hive/task_meta.rb:29-38`): a dropped `depends_on` reads as "no dependency" (`blocked:false`) and lets the daemon dispatch a dependent ahead of its prerequisite — the `warn` is the only signal. A regression that re-broadens to a silent `rescue StandardError => e; empty` passes every existing test (`test_missing_or_malformed_file_returns_nil_fields` asserts only the return value, not the warn). Add `capture_io` asserting `/depends_on dropped/` on the malformed-YAML arm AND that the ENOENT (missing-file) arm stays silent (`assert_equal "", err`) to pin the deliberate two-arm split. Confidence 95.
- [ ] [pr-test] `Worktree.origin_branch_exists?` is stubbed in every test, so its real-git `fetch` + `rev-parse` composition has zero coverage (`lib/hive/worktree.rb:219-228`; stubbed at `test/unit/stages/open_pr_test.rb:477,492`). It is the guard `open_pr` relies on to avoid a `gh pr create --base` 422 on an unpushed prereq; an inverted boolean or wrong refspec would ship green. Add a real-git test (reuse the bare-origin harness already in `worktree_test.rb`) asserting `true` for a pushed branch and `false` for a missing one — a narrow/single-branch-refspec variant of this same test would also catch the recorded FETCH_HEAD stack-collapse.

## Nit
- [ ] [silent-failure] Dependency-degradation breadcrumbs are logged under the misleading event name `:status_schema_skew` (`lib/hive/daemon/dispatcher.rb:227`): the pass-02 fix repurposed `StatusConsumer`'s `warning` channel (`status_consumer.rb:134-140`) to also carry status-stderr breadcrumbs (fail-open gate, dropped `depends_on`), but the dispatcher still emits the combined warning under the schema-version event name, so an operator grepping daemon.log for dependency-gate problems never finds them — partly undermining the observability the breadcrumb plumbing was added for. Rename to a neutral `:status_warning`, or branch on `success_warning`'s `" | "` parts and emit a distinct `:status_degraded` for the stderr arm. Verified at dispatcher.rb:227; confidence 85.
- [ ] [comment] `status_consumer.rb:40-41` lists "collapsed stack" as a `hive status` stderr breadcrumb source, but `DependencySnapshot.stacked_base`'s collapsed-stack `warn` fires only from the `execute`/`open_pr` child processes (`stages/execute.rb:99`, `stages/open_pr.rb:192`) — never the status command — so it never reaches this channel. Drop "collapsed stack" from the parenthetical (the other two examples are accurate). Confidence 90.
- [ ] [silent-failure] `Dependencies.task_id`'s corrupt-id `warn` is effectively dead on the production status path (`lib/hive/dependencies.rb:103-110`): `TaskMeta.normalize_id` (`task_meta.rb:74-81`) already silently coerces a corrupt id to `nil` before the resolver sees it, so the breadcrumb never fires there (no behavior bug — the numeric-`depends_on` case still fails closed as not-found). The genuinely-silent drop is in `normalize_id` (no warn). Move the breadcrumb there, or soften the comment. Confidence 80.
- [ ] [comment] `dependencies.rb:104` rescue comment says a corrupt id makes a numeric `depends_on` "silently mis-resolve" — but `find_task`'s numeric arm compares `nil == <int>` → false, so the prereq *fails to resolve* (treated as missing), not resolves to the wrong task. Tighten to "fail to resolve (treated as a missing prerequisite)". Confidence 55.
- [ ] [comment] `Policy.decide`'s YARD `@param` block omits the new `blocked:` keyword (`lib/hive/daemon/policy.rb:86-114`): the keyword is covered in the module header, `@return`, and inline precedence comments, but the structured `@param` list is incomplete. Add a `@param blocked [Boolean] …gates a would-be :dispatch to :blocked_on_dependency, taking precedence over answers_pending.` line. Confidence 95.
- [ ] [pr-test] No absence assertion that an unblocked-but-dependent row (`blocked:false`, `blocked_by` present — a resolved prereq past the gate) does NOT render the `⏸ blocked by …` indicator in either renderer (`status.rb:439`, `tui/views/tasks_pane.rb:186`). A plausible regression keying the indicator off `depends_on`/`blocked_by` presence instead of `blocked` would label a dispatchable task as held; per CLAUDE.md test-rule 10 only an `assert_no_selector`-style absence check catches it. Add an at-gate dependent case + `refute_includes out, "blocked by"` to both `status_test.rb` and `tasks_pane_test.rb`. Confidence 88.
- [ ] [pr-test] `Status#apply_dependency_result`'s per-row fail-open rescue is untested (`lib/hive/commands/status.rb:602-616`): stub `Hive::Dependencies.resolve` to raise for one slug and assert that row serializes `blocked:false` with the breadcrumb while sibling rows keep their resolved state — otherwise removing this finer-grained net (which currently prevents one raising row from escalating to whole-project degradation) is undetectable. Confidence 80.
- [ ] [pr-test] `Worktree.freshest_override_base`'s no-origin warn arm is untested (`lib/hive/worktree.rb:136`): the existing `test_create_falls_back_to_local_default_when_no_origin_remote` passes no `base_override`, so it exercises `freshest_base`, not this arm. Add a no-origin repo + `base_override:` case asserting the fallback and the `/cannot stack/` warn. Confidence 85.
- [ ] [code-review] A prerequisite task whose slug literally equals the project's `default_branch` (e.g. a task slugged `main`) makes `DependencySnapshot.stacked_base` treat the resolved base as a collapse (`base == default_branch` → warn + nil), silently disabling stacking for that one pair. `SLUG_RE` permits such a slug, but it is pathological and self-correcting (falls back to the default base, which is where that branch already lives). No fix recommended; recorded for completeness. Confidence 55.
- [ ] [type-design] Optional, by-design: `blocked_label` interpolates `dependency_stage` unconditionally (`lib/hive/dependencies.rb:49-55`), so a degenerate tuple (`blocked_by` present, `dependency_stage` nil) would render `⏸ blocked by <slug> ()`. The producer (`resolve`) precludes this by construction (pairs slug+stage or raises), so this is unreachable in production; the reviewer explicitly recommends NOT adding a renderer-side guard now (it defends a producer-precluded state). Recorded only to make the union complete.

ivankuznetsov:

Outcome: timeout

Agent tail:

```
The staged diff is limited to the worktree hardening, regression tests, and wiki fragment. I’m creating a commit with that scope.
```

PR #529, opened 2026-06-19
refactor(workflow): derive stage constants from one descriptor

ivankuznetsov:
## Summary

The coding pipeline's 9-stage list was triplicated across three files that had to be hand-kept in lockstep: `Stages::DIRS`, `Task::STAGE_NAMES` / `STATE_FILES`, and `Workflows::VERBS`. Renaming or reordering a stage meant editing three literals and hoping none drifted. This PR makes the pipeline a single in-Ruby descriptor and derives all four constants from it at load time, so there is now one place to change a stage.

This is the foundation unit (U1 of 8) of a workflow-as-data inversion. It is deliberately behavior-preserving: every public constant keeps the same name, order, values, and frozen-ness it had before, proven by a golden test that pins each one to its original literal. No call site migrates yet.

```mermaid
graph TB
    D["Coding::DESCRIPTOR<br/>id: :coding · 9 stages"] --> R["Workflows::Registry.default"]
    R --> S["Stages::DIRS"]
    R --> N["Task::STAGE_NAMES"]
    R --> F["Task::STATE_FILES"]
    R --> V["Workflows::VERBS"]
```

### What changed

- `Hive::Workflow`, `Hive::Workflow::Stage`, and `Hive::Workflow::AdvanceVerb` (new `lib/hive/workflow.rb`): frozen `Data` value objects modeling a workflow as an ordered list of stages, each optionally carrying the verb that advances into it.
- `Hive::Workflows::Coding::DESCRIPTOR` (new `lib/hive/workflows/coding.rb`): today's 9-stage pipeline as one pure Ruby literal, `id: :coding`.
- `Hive::Workflows::Registry` (new `lib/hive/workflows/registry.rb`): `fetch(:coding)` / `default`, raising the typed `Hive::Workflows::UnknownWorkflow` on an unknown id.
- `Stages::DIRS`, `Task::STAGE_NAMES`, `Task::STATE_FILES`, and `Workflows::VERBS` now derive from `Registry.default` instead of inline literals. `NAMES`, `SHORT_TO_FULL`, `VERB_BY_SOURCE`, and `VERB_BY_TARGET` keep deriving from those rebuilt constants unchanged.

### Design decisions

[...]

PR #530, opened 2026-06-19
feat(stages): dispatch stage runners from workflow descriptor

ivankuznetsov:
# feat(stages): dispatch stage runners from workflow descriptor

## Summary

`pick_runner` previously hardcoded a 9-arm `case task.stage_name`, and the `kind:` field on the workflow descriptor (shipped in U1) was declared-inert — nothing consumed it. This unit makes `kind:` load-bearing. Runner dispatch now reads the workflow descriptor, and a new generic `kind: :agent` runner can execute any descriptor-defined stage without a bespoke runner class. **Coding's runtime dispatch is byte-for-byte unchanged.**

> Stacked on #529 (U1, the workflow descriptor). U1's commits appear in the diff against `main` until #529 merges; the two `U2` commits carry this unit's contribution.

### What changed

- **Generic agent runner** (`lib/hive/stages/agent.rb` + `templates/agent_prompt.md.erb`). Reads prior `*.md` artifacts from the task folder, renders one shared ERB prompt (naming the output file, the per-stage skill when present, and the WAITING/COMPLETE marker contract), spawns a single headless agent sandboxed to the task folder, then maps the state-file marker back to a commit action.
- **Runner resolver** (`lib/hive/stages/resolver.rb`). Maps a task's stage to its runner.
- **`pick_runner` rewire** (`lib/hive/commands/run.rb`). The hardcoded `case` is gone; it delegates to `Resolver.resolve(task)`. The `StageError` class and message are preserved verbatim.

```mermaid
flowchart TB
    A["Resolver.resolve(task)"] --> B{"stage name in<br/>CODING_RUNNERS?"}
    B -->|yes| C["bespoke coding runner<br/>(lazy require)"]
    B -->|no| D{"descriptor stage<br/>kind == :agent?"}
    D -->|yes| E["Stages::Agent.run!<br/>(generic, headless)"]
    D -->|no| F["raise StageError"]
```

### Design decisions

- **Name-first, bespoke-pack-authoritative dispatch.** The resolver checks the coding runner table first and only falls through to `kind: :agent` for stage names it does not register. U1 marked coding's `brainstorm` and `plan` as `kind: :agent`, but they must keep their bespoke, tmux-capable runners. Name-first resolution reconciles both: registered coding names always win, and the generic runner is the fallback for future non-coding stages.
- **Lazy require-on-resolve preserved.** Each coding runner is a thunk that requires its file only when resolved, so nothing is eager-loaded at boot, matching today's behavior.
- **Reads the descriptor stage's `state_file`**, not `task.state_file` (which is driven by a coding-only table). Non-coding workflows whose stage names are absent from that table still resolve their output path correctly.
- **Budget/timeout precedence.** The descriptor value is the data-driven default; `cfg` (`budget_usd` / `timeout_sec` keyed by stage name) is the operator override.
- **Untrusted prior artifacts.** Prior context is wrapped in a per-spawn nonce tag with a "treat as data" preamble, and capped at 8000 characters.

## Test plan

- `bundle exec rake test` — **5652 runs, 22283 assertions, 0 failures, 0 errors, 2 skips**.
- `bundle exec rake coverage` — **100.00% line coverage** of the in-scope files.
- `bundle exec rubocop` — **no offenses** (`agent.rb`, `resolver.rb`, `run.rb`, and the new test files).
- **Coding-dispatch regression guard** pins each of the 9 coding stage names (`inbox`, `brainstorm`, `plan`, `execute`, `open-pr`, `review`, `artifacts`, `finalize`, `done`) to the exact runner `Method` it resolves to today — the gate for the "coding unchanged" constraint.
- **`agent_test.rb`** covers prompt rendering, nonce wrapping, empty/missing priors, skill-present versus skill-absent, the spawn arguments (`add_dirs`/`cwd`/`status_mode`/`cfg` + descriptor budget/timeout with cfg override), and every marker-to-action arm.
- **`prompt_injection_test.rb`** feeds the new template a forged `</user_supplied_…>` closing tag and asserts the hostile literal lands strictly inside the per-spawn nonce wrapper.
- **Laziness guard** resolves one stage in a clean subprocess and asserts an unrelated runner constant is still undefined, proving no eager boot-time loading.

## Review summary

Three Compound Engineering review passes (claude, codex, and pr-review-toolkit personas) ran against the diff. **No High findings in any pass.**

- **Pass 1** — no High/Medium defects. Auto-fixed: the plan-mandated U3 seam comment, a nil-stage `StageError` guard before `stage.state_file`, a redundant `.to_s`, and several test/comment hardening items (resolver `StageError` branch, truncation-cap behavior, drift guard).
- **Pass 2** — one Medium (a missing prompt-injection regression test for the new template) plus doc/coverage polish, all auto-fixed. Forward-looking robustness notes (nil budget/timeout, prior-artifact input hardening) were scoped to the U3 activation of a real non-coding `:agent` workflow and triaged NO-FIX, since the generic runner is name-shadowed and unreachable under the only registered (coding) workflow today.
- **Pass 3** — one Medium (the laziness regression guard was vacuous because `execute.rb` was preloaded) fixed by asserting in a clean subprocess; remaining items were forward-looking U3 seam/hardening notes carried over from prior passes.

Coding dispatch was independently re-confirmed byte-for-byte identical across all three passes (`StageError` constant + message preserved verbatim; `Agent#action_for` a verbatim copy of `Brainstorm#action_for`). Prompt-injection containment of the new template was adversarially verified to hold. All findings were resolved from the plan, brainstorm, or prior decisions — **zero escalations to the user.**

## Linked task

- **Hive task:** `u2-pick-runner-registry-lookup-260618-66d7` (id 1454) — *Agent Runner Registry Lookup*.
- **Depends on:** `u1-hive-workflow-descriptor-and-260618-c732` (U1 — the workflow descriptor, PR #529). U2 is unit 2 of 8 in the "workflow-as-data inversion" effort.

---


ivankuznetsov:

## High

## Medium
- [ ] `lib/hive/stages/agent.rb:17` ignores `spawn_agent` failure results: if profile preflight/version checks return `{ status: :error }` before an agent writes a marker, `run!` still reads a missing or stale state file and can commit `none` or an old terminal action instead of surfacing the spawn failure.

## Nit

ivankuznetsov:

# Code Review — U2: `pick_runner` registry lookup + generic `kind: :agent` runner (pass 1)

**Scope:** `git diff origin/main..HEAD` — focused on U2's deliverables
(`lib/hive/stages/agent.rb`, `lib/hive/stages/resolver.rb`, `lib/hive/commands/run.rb`,
`templates/agent_prompt.md.erb`, and their tests). The U1 descriptor files
(`workflow.rb`, `workflows/coding.rb`, `workflows/registry.rb`, and the
`stages.rb`/`task.rb`/`workflows.rb` derivations) are explicit plan-level scope
boundaries (U1/U4–U7) and were not re-litigated.

**Grounding / verification:**
- `rubocop` clean on all three changed Ruby files (no offenses).
- `agent_test.rb` (7) + `resolver_test.rb` (5) + `workflow_golden_test.rb` (6) green.
- Full suite: 5652 runs, **1 failure** — `StopHookInstallerTest#test_stop_hook_requires_stage_dir_env`.
  This is an **environment-leak flake unrelated to this diff**: the test runs the
  stop-hook expecting it to fail when `HIVE_TASK_STAGE_DIR` is unset, but the reviewer
  process has that var set, so the hook succeeds and `refute status.success?` fails.
  Passes in a clean env; **not a U2 regression.**
- Coding dispatch confirmed **byte-identical** by the resolver regression test (all 9
  stage names → same `.method(:run!)` targets); `Agent#action_for` is a verbatim copy
  of `Stages::Brainstorm#action_for`; the `StageError` constant and message are preserved.

The implementation faithfully matches the plan's Requirements Trace. No High or Medium
defects found. Remaining items are forward-looking nits about the U3 hand-off seam and
input robustness in the new (currently coding-shadowed) generic runner.

## High

_None._

## Medium

_None._

## Nit

- [ ] Missing plan-mandated seam comment in `Agent.run!` (`lib/hive/stages/agent.rb:12`): the plan explicitly directed to "leave a short comment marking that seam" where the runner re-resolves the stage via `Registry.default` instead of the dispatched descriptor (the U3 hand-off); the comment is absent, so the U3-threading intent is undocumented at the call site.
- [ ] Opaque nil-stage failure at the U3 seam (`lib/hive/stages/agent.rb:12-13`): when a future descriptor-threaded `:agent` stage is absent from `Registry.default`, `stage` is `nil` and `stage.state_file` raises a bare `NoMethodError`; a one-line `stage or raise Hive::StageError, "no agent stage #{task.stage_name}"` guard would yield the same typed error the resolver already raises. Not reachable under coding (out of U2 scope) — forward-looking hardening for U3.
- [ ] `prior_artifacts` reads untrusted artifacts without input hardening (`lib/hive/stages/agent.rb:50-54`): `File.read` (line 53) has no NUL-scrub — a NUL byte in any prior `*.md` raises `ArgumentError` at `Process.spawn` (default claude profile passes the prompt as argv), leaving a stale `AGENT_WORKING` marker — and no per-file size bound (a multi-GB artifact is read fully into memory before the 8000-char cap applies). Harmless today (the generic runner is shadowed by `CODING_RUNNERS` under the only registered workflow); add `.scrub` / `delete("\x00")` + a size guard before U3 wires a real non-coding `:agent` workflow.
- [ ] Redundant `.to_s` (`lib/hive/stages/agent.rb:54`): `String#[](0, 8000)` always returns a String here (the `join` result is never `nil`), so the trailing `.to_s` is dead defensiveness.

ivankuznetsov:

# PR Review Toolkit — U2: `pick_runner` registry lookup + generic `kind: :agent` runner (pass 1)

**Scope:** `git diff origin/main..HEAD`, focused on U2 (`lib/hive/stages/agent.rb`,
`lib/hive/stages/resolver.rb`, `lib/hive/commands/run.rb`, `templates/agent_prompt.md.erb`,
and their tests). U1 descriptor files reviewed only for introduced regressions.

**Personas run:** code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter,
type-design-analyzer, code-simplifier.

**Cross-persona consensus:** coding dispatch is byte-for-byte unchanged (name-first
`CODING_RUNNERS` shadows the `kind: :agent` `brainstorm`/`plan` stages); `action_for` is a
verbatim copy of `Brainstorm#action_for`; `StageError` class/message preserved; rubocop clean;
all 12 new tests pass and achieve 100% line coverage of the in-scope files. **No Critical /
High defect found.** The findings below are the union of all persona output, de-duplicated.

Note: several personas independently flagged that `Agent.run!` re-resolves the stage against
`Registry.default` rather than the descriptor the resolver matched — the plan **explicitly
defers** descriptor-threading and non-coding `:agent` workflows to U3, so those are recorded as
forward-looking nits (not escalated) per the scope rules.

## High

_None._

## Medium

- [ ] [comments] Plan-mandated seam comment missing at `lib/hive/stages/agent.rb:12`: IU1 explicitly directed "leave a short comment marking that seam" where the stage is re-resolved via `Registry.default` instead of a threaded descriptor (the U3 hand-off); the only comment in the file is the line-49 guardrail, so the descriptor/registry split is undocumented at the call site that U3 will depend on.

## Nit

- [ ] [type-design/silent-failure] No nil guard before `stage.state_file` (`lib/hive/stages/agent.rb:12-13`): if the stage is absent from `Registry.default`, `run!` raises an opaque `NoMethodError` (wrapped as `InternalError`) instead of the resolver's `Hive::StageError`. Unreachable under coding/U2 (descriptor-threading deferred to U3); a one-line `raise Hive::StageError, "no agent stage #{task.stage_name}" unless stage` would harden the U3 seam and keep the failure mode consistent.
- [ ] [silent-failure] `prior_artifacts` bare `File.read` over a glob with no `File.exist?`/rescue and no `.scrub` (`lib/hive/stages/agent.rb:50-54`): a task-folder file removed between glob and read (`Errno::ENOENT`), an unreadable file (`Errno::EACCES`), or invalid-UTF-8 content can make it raise — contrary to the plan's "never raises" invariant. Matches existing `brainstorm`/`plan`/`execute` convention so it's no regression; a defensive per-file read (`File.read(path, encoding: "UTF-8").scrub`, skip-on-`Errno`) would honor the invariant for arbitrary task-folder files.
- [ ] [silent-failure] Marker-less error envelope yields `marker=none` (`lib/hive/stages/agent.rb:30-31`): if `spawn_agent` returns `{status: :error}` (e.g. preflight failure) without writing a marker, `Markers.current` returns `:none`, so the runner reports `commit="none"`/`marker=none` rather than an attributed `:error`. Loud upstream (stderr `warn`), not silent, but less specific than the bespoke tmux path; consider routing the error envelope into an `:error` marker write in U3.
- [ ] [type-design] `CODING_RUNNERS` ↔ descriptor drift (`lib/hive/stages/resolver.rb:7-44`): the table hardcodes the same 9 stage names as `Coding::DESCRIPTOR` with no cross-check, so a descriptor rename silently leaves a dead entry (and the renamed stage falls through to the agent/`StageError` path). A test asserting `CODING_RUNNERS.keys` ⊆ `Coding::DESCRIPTOR.stages.map(&:name)` converts latent drift into a failing test.
- [ ] [type-design] Duplicated runner-result contract (`lib/hive/stages/agent.rb:30-31,57-64`): `action_for` is a verbatim copy of `Brainstorm#action_for` and the "read marker → `{commit:, status:}`" epilogue is now repeated across ≥3 runners; hoisting the shared contract into `Stages::Base` (where `spawn_agent`/`render` live) would enforce it in one place. Deferrable.
- [ ] [tests] Uncovered resolver branch (`test/unit/stages/resolver_test.rb:64`): the `StageError` test uses an empty descriptor (stage `nil`), so the distinct branch "stage present but `kind != :agent`" is never exercised — a regression loosening the `kind == :agent` guard to bare truthiness would pass. A `kind: :marker` non-coding stage asserting `StageError` closes it.
- [ ] [tests] Truncation guardrail behaviorally unasserted (`lib/hive/stages/agent.rb:54`): no fixture exceeds a few bytes, so the `[0, 8000]` cap is line-covered but always a no-op; a >8000-char prior artifact asserting bounded `prior_context` would lock the cap against silent removal.
- [ ] [tests] Misleading test name (`test/unit/stages/agent_test.rb:165`): `test_marker_actions_match_brainstorm_mapping` actually drives the `plan` stage, not brainstorm; the assertion is correct and strong, only the name misleads.
- [ ] [comments] Guardrail comment imprecise (`lib/hive/stages/agent.rb:49`): "Guardrail against flooding the prompt with historical artifacts" omits the actual mechanism — a hard 8000-char cap on the joined string; naming the cap prevents a reader assuming a per-file or token-based limit.
- [ ] [comments/type-design] Name-first precedence undocumented (`lib/hive/stages/resolver.rb:48-56`): the load-bearing "bespoke name wins over descriptor `kind`" invariant (why `brainstorm`/`plan` stay on bespoke runners despite `kind: :agent`) is inferable only from the early `return`; a one-line comment would protect it from future "simplification."
- [ ] [simplify] Redundant `.to_s` (`lib/hive/stages/agent.rb:54`): `String#[](0, 8000)` always returns a String here (the `join` result is never `nil`), so the trailing `.to_s` is dead defensiveness.
- [ ] [simplify] Unused `_cfg` positional in `render_prompt` (`lib/hive/stages/agent.rb:34`, call site line 15): the parameter exists only to mirror `brainstorm.rb`'s signature (which actually uses `cfg`); droppable, or keep deliberately for sibling-stage parity — judgment call.

ivankuznetsov:

# Code Review — U2: `pick_runner` registry lookup + generic `kind: :agent` runner (pass 2)

**Scope:** `git diff origin/main..HEAD`, focused on U2's deliverables
(`lib/hive/stages/agent.rb`, `lib/hive/stages/resolver.rb`, `lib/hive/commands/run.rb`,
`templates/agent_prompt.md.erb`, and their tests). The U1 descriptor files
(`workflow.rb`, `workflows/coding.rb`, `workflows/registry.rb`, and the
`stages.rb`/`task.rb`/`workflows.rb` derivations) are an inherited dependency from a
separate task (U1/U4–U7) and were not re-litigated.

**Method:** ce-code-review (report-only). Parallel persona reviewers — correctness,
testing, maintainability, security, adversarial, reliability, project-standards — then
merge + independent validation of the one non-pre-existing functional claim.

**Gates verified (this pass):**
- `rubocop` clean on `agent.rb`, `resolver.rb`, `run.rb` (3 files, no offenses).
- `agent_test.rb` (9) + `resolver_test.rb` (9) green; the testing reviewer confirmed
  **100% line + branch coverage** of both new files via stdlib `Coverage`.
- Coding dispatch confirmed **byte-identical**: the regression guard compares each of
  the 9 coding stage names against the exact `Method` the old `case` returned (Ruby
  `Method#==` matches owner + name), `Agent#action_for` is a verbatim copy of
  `Brainstorm#action_for`, and the `StageError` constant + `"no runner for stage <name>"`
  message are preserved unchanged.
- Security: the nonce-wrapping in `agent_prompt.md.erb` mirrors the established ADR-019
  pattern (fresh per-spawn `SecureRandom.hex(8)` tag), `add_dirs`/`cwd` are not widened
  beyond the task folder. No exploitable issue found.

**Already triaged in pass 1 — not re-raised here:** the `spawn_agent` preflight result
being ignored / marker-less error envelope (resolved NO-FIX: mirrors brainstorm/plan/
execute per plan IU1; error-envelope→marker change scoped to U3), and `prior_artifacts`
input hardening (NUL-scrub / per-file size bound, resolved NO-FIX, scoped to "before U3").
Both remain valid forward-looking items but were deliberately deferred and are out of
re-litigation scope for this pass.

The implementation faithfully matches the plan's Requirements Trace and the hard
acceptance constraint (coding unchanged, generic runner reachable + unit-tested). No High
or Medium defects. The findings below are forward-looking nits in the new (currently
coding-shadowed) generic runner.

## High

_None._

## Medium

_None._

## Nit

- [ ] Generic runner passes `nil` `timeout_sec` straight through to spawn — latent crash + orphaned child (`lib/hive/stages/agent.rb:27`): `timeout_sec: cfg.dig("timeout_sec", task.stage_name) || stage.timeout_sec`; `Workflow::Stage#timeout_sec` defaults to `nil` and config `DEFAULTS["timeout_sec"]` is keyed only by built-in stage names, so a `kind: :agent` stage with no `timeout_sec` yields `nil`. `Hive::Agent#spawn_and_wait` computes `deadline = Time.now + @timeout_sec` (`lib/hive/agent.rb:177`) **after** `Process.spawn` (`agent.rb:127`), so `nil` raises `TypeError` with the child already running and untracked (leaked process). Not reachable in U2 (coding's `kind: :agent` stages route to bespoke runners via `CODING_RUNNERS` and set `timeout_sec` anyway), so the generic runner is never dispatched today — but it activates the moment U3 registers a non-coding `:agent` descriptor stage lacking `timeout_sec`. A one-line guard now (raise `StageError` when both cfg and descriptor `timeout_sec` are nil, or supply a sane fallback) prevents a hard-to-diagnose crash later. New (not raised in pass 1).
- [ ] U3 seam comment covers the stage re-resolution but not the second path derivation (`lib/hive/stages/agent.rb:17`): the runner reads the terminal marker from `File.join(task.folder, stage.state_file)` (re-derived from `Registry.default`), while `Run#commit_after`/`#report` read from `task.state_file`. Both derive `state_file` from `Registry.default` keyed by `task.stage_name` today (`Task::STATE_FILES` is built from `Registry.default`), so they always agree under U2. When U3 threads a non-default descriptor into the runner but `task.state_file` still uses `Registry.default`, the two marker reads can silently diverge. Extend the existing seam comment to note `output_path` must stay equal to `task.state_file` (or read from `task.state_file`).
- [ ] No test locks the ADR-019 nonce guarantee for the new runner (`test/unit/stages/agent_test.rb`): no case feeds a forged closing tag (literal `</user_supplied_…>`) through `prior_artifacts` and asserts the opening/closing nonce tags still bracket it. The brainstorm pattern is trusted, but an explicit payload test would pin the prompt-injection containment for the generic runner that future `kind: :agent` workflows depend on. (Robustness test only; the 100% coverage gate is already met.)

ivankuznetsov:

## High

## Medium

## Nit
- [ ] `lib/hive/stages/agent.rb:27` can pass a nil timeout to `Hive::Agent`: a future non-coding `kind: :agent` stage without descriptor/config `timeout_sec` reaches `Time.now + nil` after the child process is spawned, causing a crash with a leaked agent process.
- [ ] `lib/hive/stages/agent.rb:17` leaves the U3 descriptor seam split across two state-file derivations: the generic runner reads its marker from descriptor-derived `output_path`, while `Run#commit_after`/`#report` still read `task.state_file`, so future descriptor threading can make the two disagree.
- [ ] `test/unit/stages/agent_test.rb` does not pin the nonce escape case: the tests assert wrapping but never feed prior content containing a forged `</user_supplied_...>` closing tag, leaving the generic runner's ADR-019 prompt-injection containment unguarded.

ivankuznetsov:

## High
- [ ] Out-of-scope visual artifact pipeline removal (`templates/artifacts_prompt.md.erb:14`, `web/config/routes.rb:24`): the branch deletes existing demo capture, media serving, screenote upload, and their tests even though U2 only scopes runner lookup/generic agent work, so merging would regress shipped artifacts behavior.
- [ ] Symlinked prior artifacts bypass the task-folder sandbox (`lib/hive/stages/agent.rb:58`): `File.read` follows any task-folder `*.md` symlink before the sandboxed agent spawn, leaking outside-readable files into the prompt and also making dangling priors raise despite the plan's "missing -> empty" contract.

## Medium
- [ ] Generic runner ignores failed spawns (`lib/hive/stages/agent.rb:21`): `spawn_agent` can return `{ status: :error }` before writing a marker on preflight/version failure, but `run!` discards that result and may report a stale or missing marker as a non-error commit action.
- [ ] Resolver laziness regression test is ineffective (`test/unit/stages/resolver_test.rb:101`): the snapshot runs after earlier assertions have loaded all bespoke runners, so it would still pass if resolving `inbox` started eager-loading `execute`, missing a plan-required guard.

## Nit

ivankuznetsov:

# Code Review — U2: `pick_runner` registry lookup + generic `kind: :agent` runner (pass 3)

**Scope:** U2's deliverables — `lib/hive/stages/agent.rb`, `lib/hive/stages/resolver.rb`,
`lib/hive/commands/run.rb` (`pick_runner` rewire), `templates/agent_prompt.md.erb`, their
tests, and the wiki updates. The U1 descriptor files (`workflow.rb`, `workflows/*`, and the
`stages.rb`/`task.rb`/`workflows.rb` derivations) are prior-task scope and were sanity-checked
only.

**Grounding / verification (pass 3, this worktree):**
- `rubocop` clean on `agent.rb`, `resolver.rb`, `run.rb`, and all three new/edited test files
  (6 files, no offenses).
- Tests green: `agent_test.rb` (10), `resolver_test.rb` (7), `prompt_injection_test.rb` (14),
  `workflow_test.rb` (6), `workflows/coding_test.rb` (4), `workflows/registry_test.rb` (2).
- Coding dispatch re-confirmed **byte-identical**: the resolver's name-first early return routes
  all 9 coding stage names (including `kind: :agent` `brainstorm`/`plan`) to today's bespoke
  runners; `Agent#action_for` is a verbatim copy of `Brainstorm#action_for`; and the generic
  `Agent.run!` mirrors `Brainstorm#run_headless!` line-for-line (spawn → read `Markers.current`
  → `action_for`). In coding, the generic-runner branch and the `StageError` branch are never
  reached at runtime — only exercised via tests with synthetic descriptors.
- Pass-2 Medium **closed**: `test_agent_prompt_wraps_hostile_prior_artifacts_in_unique_tags`
  (`prompt_injection_test.rb:92-130`) now feeds a forged `</user_supplied>` and asserts the
  nonce open/close appear exactly once each with the hostile literal strictly between them.
  Containment holds; the safety-net test is present and effective.
- Pass-2 wiki nits **landed**: `wiki/index.md` now lists `[[stages/agent]]` before
  `[[stages/artifacts]]`; `wiki/modules/workflows.md` runner-consumed-metadata wording corrected
  (commit `bf2daa12`).

**Scope note — neutralizing a likely false positive in the parallel reviewer file
(`codex-ce-code-review-03.md`):** that file raises a **High** for "out-of-scope visual artifact
pipeline removal" (screenote/artifacts/media deletions). This is an artifact of the two-dot diff
range, **not** a regression. `git log origin/main..HEAD -- lib/hive/screenote_uploader.rb`
(and the artifacts/web files) is **empty** — this branch touches none of them. They appear as
deletions only because `origin/main` *added* them after this branch forked at `71e92a94`
(commit `a240f033`, "feat(artifacts): capture visual demos … (#515)"). The true branch diff
(`git diff origin/main...HEAD`, three-dot) shows **only** the U1+U2 additions and deletes
nothing. A normal merge/rebase onto current `main` preserves the #515 features (plan R2 already
requires rebasing onto U1's merge before integration). **Recommend triage drop the codex High.**

The implementation faithfully matches the plan's Requirements Trace and the byte-identical-coding
acceptance constraint. No High defects. One Medium (a hollow regression test). Two forward-looking
Nits that are unreachable in U2 and mirror prior NO-FIX deferrals — included to calibrate the
codex-03 escalations for triage.

## High

_None._

## Medium

- [ ] Resolver laziness regression test is ineffective (`test/unit/stages/resolver_test.rb:101-109`): `test_resolving_one_bespoke_stage_does_not_load_other_bespoke_files` cannot fail under the eager-load regression it purports to guard — under a hypothetical eager `CODING_RUNNERS`, `execute.rb` would be required at `require "hive/stages/resolver"` (file top, line 3), i.e. *before* the `before = $LOADED_FEATURES.dup` snapshot, so `newly_loaded` never contains it; equally, any earlier test that resolved `execute` puts it in the baseline. Either way the `refute` passes whether resolution is lazy or eager, so the guard is decorative. Production laziness (require-on-call lambdas) is correct and the prod path is fine; only the test is hollow. Fix: assert the requires triggered by a single `resolve` call (track `require`/run in a forked clean process) rather than a process-global snapshot. Plan IU2 sanctioned a "light-touch" check, so this is should-fix not must-fix; corroborated by codex-03.

## Nit

- [ ] Forward-looking, unreachable in U2: `prior_artifacts` (`lib/hive/stages/agent.rb:55-59`) reads `*.md` with `File.read`, which follows symlinks — a *dangling* task-folder `*.md` symlink raises `Errno::ENOENT` (violating the plan's "missing → empty, never raises"), and a symlink pointing outside the task folder leaks that file's content into the prompt. Only reachable once U3 wires a real non-coding `kind: :agent` workflow (coding shadows the generic runner today; the task folder is operator-controlled; nonce containment is unaffected). Continues the pass-1/pass-2 input-hardening deferral to before-U3. codex-03 rated this High — downgrading: not reachable under U2's only registered workflow.
- [ ] Forward-looking, unreachable in U2 (already RESOLVED/NO-FIX in pass 2): the generic runner discards `spawn_agent`'s `{ status: :error }` preflight return (`lib/hive/stages/agent.rb:21-35`) and maps a marker-less clean exit through `action_for`'s `else marker_name.to_s` (e.g. `:none → "none"`), so a preflight failure on a future non-coding stage would commit a non-error action with no terminal/error signal. This is byte-identical to `Brainstorm#run_headless!`/`Plan` — the verbatim contract plan IU1 prescribes — and is shadowed by `CODING_RUNNERS` in U2, so it is **not** a U2 regression (the bespoke runners that DO run today share the same behavior). codex-03 rated this Medium; recorded here only for continuity with pass-2's deferral (error-envelope→marker change scoped to U3).

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — U2 (pass 3)

**Personas run:** `[code]` code-reviewer, `[tests]` pr-test-analyzer, `[errors]` silent-failure-hunter,
`[comments]` comment-analyzer, `[types]` type-design-analyzer, `[simplify]` code-simplifier.

**Scope:** U2's four deliverables + tests — `lib/hive/stages/agent.rb`, `lib/hive/stages/resolver.rb`,
`templates/agent_prompt.md.erb`, the `pick_runner` change in `lib/hive/commands/run.rb`, and
`test/unit/stages/agent_test.rb`, `test/unit/stages/resolver_test.rb`,
`test/integration/prompt_injection_test.rb`. U1's descriptor files are prior scope. The
screenote/media/artifacts "deletions" in `origin/main..HEAD` are an artifact of the branch's base
(merge-base `71e92a94`) lagging `origin/main`'s unrelated #515 feature — **not** removals by this
task — and were ignored.

**Verified by reviewers:** rubocop clean (3 ruby files); 31 U2 tests green; coding dispatch
byte-for-byte identical (name-first early return keeps `kind: :agent` `brainstorm`/`plan` on their
bespoke runners; all 9 names → same `Method` objects; `StageError` constant + message verbatim;
`action_for` verbatim vs `brainstorm.rb:130-137`); prompt-injection nonce wrapper holds and is
regression-tested; the generic `Agent.run!` is **unreachable in U2 production** (`Task` validates
`stage_name` against `Registry.default`, and its only `:agent` stages are name-shadowed), so
robustness gaps in the generic path are latent/forward-looking, not active U2 bugs.

**No High (merge-blocking) findings.** Two personas rated an item "High": `[tests]` the vacuous
laziness guard (a test-effectiveness gap, no production defect) and `[errors]` the discarded
`spawn_agent` return (real, but **latent** — cannot fire while the generic runner is unreachable and
coding stays byte-identical). Both are placed in Medium with their original persona ratings noted, so
triage has the strongest framing without treating a latent/test issue as a U2 merge blocker.

## High

_None._

## Medium

- [ ] [errors] Discarded `spawn_agent` return → **false-green commit on preflight failure** (`lib/hive/stages/agent.rb:21-35`): `spawn_agent` returns `{status: :error, error_message: …}` *without writing a marker* when `check_version!`/`preflight!` fail (`base.rb:414-420`); `run!` ignores the return and reads the on-disk marker, so an absent marker → `action_for(:none)` → `{commit: "none"}` which `commit_after` commits with `ok:true`/exit 0 (a stale `:complete` masquerades as success). `execute.rb:203-206` (`agent_failed?` gating `%i[error timeout]`) is the proven-correct pattern; agent.rb copied brainstorm's weakest *headless* discard. _Persona rated **High**; latent (unreachable until U3, shape shared with brainstorm/plan headless) — fix before U3 activates the path._ Fix: capture the result and set an `:error` marker (`reason: "agent_preflight_failed"`) when `result[:status]` is `:error`/`:timeout`, before reading back.

- [ ] [tests] Laziness regression guard is **vacuous in every gated run** (`test/unit/stages/resolver_test.rb:101-109`): the `$LOADED_FEATURES` delta can only contain not-yet-loaded files, but `execute.rb` is always preloaded by the coverage gate's `load_all_sources!` and by `prompt_injection_test.rb:4` (and same-file sibling resolution), so an injected eager-load regression passes silently — the persona reproduced both outcomes. Defeats the plan's Q6/A6 "no eager boot loading" + the A7 laziness clause. _Persona rated **High** with reproduction; `codex-ce-code-review-03` and `claude-ce-code-review-03` rated it **Medium**._ Fix: assert in a clean subprocess that resolving `inbox` leaves `Hive::Stages.const_defined?(:Execute, false)` false.

- [ ] [errors] `prior_artifacts` does unbounded `File.read` over globbed `*.md` with **no per-file rescue** (`lib/hive/stages/agent.rb:52-60`): one unreadable/binary/racing (`ENOENT` between glob and read) sibling `.md` aborts the whole stage as an unattributed `InternalError` with **no marker written**, so `hive status` shows it un-run and the next run re-reads the same file and fails identically (a non-obvious loop). U2-introduced — `brainstorm`/`plan` read a single `File.exist?`-guarded file, not an unbounded glob. Extends the pass-1/2-deferred input-hardening (broader than the previously-noted encoding facet). Fix: per-file `rescue StandardError` that `warn`s and skips so advisory context degrades instead of killing the stage.

- [ ] [tests][code] Generic `Agent.run!` is **never driven end-to-end via a custom (non-`Registry.default`) descriptor** (`test/unit/stages/agent_test.rb` all tests; `lib/hive/stages/resolver.rb:56-59` vs `lib/hive/stages/agent.rb:15`): the resolver matches `kind: :agent` against the *passed* descriptor, but `Agent.run!` re-resolves the stage from `Registry.default` and ignores it — so a custom `research` stage *resolves* successfully yet `run!` raises `StageError "no agent stage research"`. Tests only assert resolution, hiding the resolve-path/run-path source mismatch U3 must close. _Persona rated **Medium** (also `[code]` Nit, `[comments]`)._ Fix: add a test that threads a synthetic descriptor through `Registry.default` and invokes `run!` end-to-end; when U3 lands, pass the matched descriptor into `Agent.run!`.

- [ ] [comments] `agent.rb:12-14` U3-seam comment asserts a "**descriptor the dispatcher matched**" that is not plumbed: `run.rb:138` calls `Resolver.resolve(task)` with the default and the `descriptor:` param is test-only, so no dispatcher threads a descriptor anywhere; the note also implies the coupling is unique to agent.rb though `resolver.rb:48` shares it. Could send a maintainer hunting for non-existent plumbing. _Persona rated **Medium**._ Fix: reword to state descriptor-threading as not-yet-existing and acknowledge the `resolver.rb:48` sibling coupling.

- [ ] [tests] **Nil budget/timeout fallback path unexercised** (`test/unit/stages/agent_test.rb:161-198` vs `lib/hive/stages/agent.rb:26-27`): the both-nil case (a custom `:agent` stage declaring no `budget_usd`/`timeout_sec`, legal per `Workflow::Stage` defaults) is never tested, and the `plan`-stage descriptor defaults (100/3600) are never asserted (only `brainstorm`'s 50/1800). Branch coverage is reported but not gated. _Persona rated **Medium**._ Fix: assert nil flows for a budget-less custom stage and pin `plan`'s defaults. (Note: the *robustness* of nil `timeout_sec` → `Time.now + nil` crash in `Hive::Agent` was triaged NO-FIX/deferred-to-U3 in pass 2; this finding is only the test gap.)

## Nit

- [ ] [tests] `prior_artifacts` sort + multi-file cap asserted loosely (`test/unit/stages/agent_test.rb:59-92`): order-independent `assert_includes` would not catch a dropped `.sort`, and the cap test uses one 20k-char file so it does not prove the cap applies to the *joined* multi-file string. Fix: assert relative ordering (`prompt.index("## idea.md") < prompt.index("## brainstorm.md")`) and add a two-file (5k+5k) fixture truncating to 8000.

- [ ] [errors] `action_for(:none)` → `"none"` clean-exit-no-marker gap (`lib/hive/stages/agent.rb:62-69` → `:35`): a clean `exit 0` with no WAITING/COMPLETE/ERROR marker yields a green `"none"` commit with no terminal signal. Inherited from `brainstorm`/`plan`; the deeper fix belongs in `Agent#handle_exit_state_file_marker` so all `:state_file_marker` stages benefit. **Already triaged NO-FIX in pass 2** — listed for union/corroboration; the new detail is the suggested central fix location.

- [ ] [types] `CODING_RUNNERS` subset guard is one-directional (`test/unit/stages/resolver_test.rb:93-99`): it asserts no orphaned runner keys but not that every coding descriptor stage resolves to a runner, so adding a bespoke-less coding stage would raise at dispatch rather than fail a test. Fix: add a round-trip test asserting every `Coding::DESCRIPTOR` stage name resolves via `Resolver.resolve` (test-only, zero production cost).

- [ ] [comments] Hard-coded line-number references are a rot vector: `agent_test.rb:201` cites "`agent.rb:11`", and `agent.rb:12`'s seam note is the highest-churn comment (must be deleted when U3 lands). Fix: reference the method/expression (e.g. "the `cfg ||= {}` guard in `run!`") rather than line numbers.

- [ ] [code] `action_for` is duplicated verbatim vs `brainstorm.rb:130-137` (`lib/hive/stages/agent.rb:62-69`): expected and plan-mandated for U2 (must stay verbatim; `brainstorm` keeps its own runner). No change now; a later unit where touching `brainstorm` is in scope could centralize to `Stages::Base.action_for`.

- [ ] [simplify] `render_prompt`'s unused `_cfg` param (`lib/hive/stages/agent.rb:38`) is signature-parity ballast with `Brainstorm#render_prompt`; simplifier recommends keeping it. Separately, the simplifier explicitly **declined** extracting the 9× `CODING_RUNNERS` thunk repetition — greppability and the non-mechanical `open-pr`→`OpenPr` mapping make the explicit table the clearer choice. No action.

- [ ] [types] Runner result `:status` key is dispatcher-dead parity ballast (`run.rb` reads only `result[:commit]`; state is re-derived from `Markers.current`); retained for `brainstorm` parity. No action — flagged only so it is not mistaken for a live signal.

ivankuznetsov:
Superseded by #562, which consolidates the full U1–U10 workflow-as-data inversion into one branch off current main (all unit fixes integrated, gates green) for single-feature evaluation. Closing u2's standalone PR.

PR #538, opened 2026-06-19
feat(task): resolve stages and state_file from workflow descriptor

ivankuznetsov:
# feat(task): resolve stages and state_file from workflow descriptor

## Summary

A `Hive::Task` now resolves its stage set and `state_file` from the workflow descriptor selected for that task, instead of the coding-global constants every task was hardwired to. This is unit 3 of the workflow-as-data inversion (built on U1's `Hive::Workflow`/`Registry` and U2's runner `Resolver`): the piece that lets a single task belong to a non-coding workflow and dispatch that workflow's runners. A task with no workflow selected behaves byte-identically to before.

Previously the `Task` constructor validated the stage name against `STAGE_NAMES`/`Hive::Stages.parse` and read `state_file` from `STATE_FILES` — both derived from the one coding workflow, so there was no seam for any other workflow to flow through. Now the constructor resolves a descriptor, validates the stage name **and its numeric prefix** against that descriptor's own stages, and `state_file` comes from `workflow.state_file_for(stage_name)`. `Run#pick_runner` passes `task.workflow` into the U2 resolver so the selected workflow's runners actually run.

This branch is stacked on U1+U2, so the diff against `main` also carries the already-reviewed `Hive::Workflow`/`Registry`/`Resolver`/generic-agent-runner foundations U3 depends on. The U3 surface is `task.rb`, `workflow.rb`, `task_meta.rb`, `config.rb`, and `commands/run.rb`.

### Workflow resolution

The descriptor is resolved through a 3-tier precedence chain, read once and memoized in the constructor:

| Tier | Source | Default |
|---|---|---|
| 1 | task `meta.yml` `workflow:` selector | — |
| 2 | per-project `default_workflow` (new `Config::DEFAULTS` key) | `"coding"` |
| 3 | built-in `coding` descriptor | always present |

`TaskMeta.read`/`empty` surface the `workflow:` key through the same graceful-degradation path `depends_on` already uses, so a missing or malformed `meta.yml` yields `workflow: nil` and falls down the chain rather than raising.

### Backward compatibility

For a field-less task with `default_workflow` unset, the resolver returns the coding descriptor (`== Registry.default`). Stage-name + index validation over the coding stages is equivalent to the old `STAGE_NAMES.include?` + `Stages.parse` check, and `state_file_for` returns exactly what `STATE_FILES.fetch` did. `Task::STAGE_NAMES` and `STATE_FILES` are left untouched for the ~20 downstream consumers (status, daemon, TUI) that later units will migrate.

### Two failure-mode decisions

Moving resolution into the constructor means it must not break the status-listability invariant, where `Status` only rescues `InvalidTaskPath`:

- **Typo'd registered workflow** → `Registry.fetch` raises `UnknownWorkflow`, which `Task` re-raises as `InvalidTaskPath` (exit 64). A misconfigured `workflow:` value skips that one task out of `hive status` exactly as an unknown stage dir does today, instead of crashing the whole listing. A typo'd project-level `default_workflow` additionally `warn`s, since its blast radius is every task in the project.
- **Broken project `config.yml`** → reading `default_workflow` defensively rescues `ConfigError` and the narrow IO/YAML group, `warn`s to stderr, and falls back to coding. A broken config can't brick every `Task.new` in the project.

## Test plan

- `bundle exec rake test` — passes
- `bundle exec rake coverage` — passes at 100% line **and** branch coverage on every U3 file
- `bundle exec rubocop` — clean on all changed Ruby files

New/extended coverage:

- `task_test.rb` — field-less byte-identical path across all 9 coding stages; index/prefix mismatch and unknown-stage raises (exit 64); per-project `default_workflow`; task-selector-wins-over-project-default precedence; both rescue-to-coding fallbacks (`ConfigError` + malformed YAML) and the unregistered-name re-raise via both selector and project default.
- `task_meta_test.rb` — `workflow:` read, blank→nil normalization, missing-file and malformed-YAML fallbacks (warn asserted).
- `config_test.rb` — `default_workflow` default and merge survival.
- `workflow_test.rb` / `workflow_golden_test.rb` — `stage_named`/`state_file_for`/`stage_names` helpers and the descriptor-derived-constants golden pin.
- `test/support/workflow_helpers.rb` — throwaway non-coding descriptor (distinct stage names + `:agent`-kind stage) driving the integration assertion and the `pick_runner` dispatch test in `run_test.rb` / `resolver_test.rb`.

## Review summary

Three `ce-code-review` passes (claude + codex + pr-review-toolkit personas) plus escalation triage. Final verdict across reviewers: **Ready** — no outstanding High/Medium findings on the U3 surface.

- **Pass 01** — auto-fixed: restored byte-identical stage-prefix parity (`stage.dir == stage_dir` per plan IU-4, fixing the leading-zero `04-execute` edge); named `workflow` alongside `depends_on` in the dropped-fields warn; documented the soft-lookup-vs-hard-resolve dual contract; simplified selector resolution; refreshed stale wiki source anchors; added per-branch coverage tests.
- **Pass 02** — auto-fixed: one-line `warn` when a project `default_workflow` resolves to an unregistered name (whole-project blast radius the per-task D1 rationale didn't cover); added malformed-YAML-config and non-coding-workflow validation tests; pinned the task-selector-over-project-default precedence.
- **Pass 03** — verdict Ready; auto-fixed remaining nits: re-anchored a tautological `stage_names` assertion to a literal, wrapped a malformed-meta `Task.new` in `capture_io`, softened the agent-runner seam comment, threaded the resolved workflow explicitly into the validator, froze `Workflow#stage_names`.

Resolved-as-deferred (documented, not defects): the generic `Agent.run!` re-resolving its stage from `Registry.default` is the plan-scoped U6 descriptor-threading seam (dormant in production — only `:coding` is registered, and coding's agent stages use bespoke `CODING_RUNNERS`); the `TaskMeta.write`/`update_display_name` selector round-trip is the deferred tagging-unit writer (U3 is read-only); the `STAGE_NAMES`/`STATE_FILES` consumer migration is U6. No reviewer questions were escalated to the user.

## Linked task

- Hive task: `u3-task-resolves-its-stages-260618-f532` (id 1455) — "Workflow Aware Task Stages"
- Depends on: `u2-pick-runner-registry-lookup-260618-66d7` (U2 registry-driven runner resolver)
- Unit 3 of 8 in the workflow-as-data inversion

---


ivankuznetsov:

## High
- [ ] Generic agent runner ignores the selected task workflow: `Run#pick_runner` can resolve a non-coding `2-gather` task to `Hive::Stages::Agent`, but `lib/hive/stages/agent.rb:19` re-reads `Registry.default` and raises `StageError` instead of using the task descriptor's stage metadata.

## Medium

## Nit

ivankuznetsov:

## High
- [ ] `Hive::Stages::Agent.run!` re-resolves the stage from `Registry.default` (coding), not `task.workflow` (`lib/hive/stages/agent.rb:19`): U3's `pick_runner` now correctly dispatches a non-coding `kind: :agent` stage (e.g. research `2-gather`) to `Agent.run!`, but the runner then looks the stage up in the coding descriptor and raises `Hive::StageError: no agent stage gather` (reproduced), so IU-5's "a non-coding task dispatches its own workflow's runners" is wired at dispatch yet non-functional end-to-end. Dormant in production (no real non-coding workflow is registered, and coding's agent stages use bespoke `CODING_RUNNERS` so they never reach `Agent.run!`); agent.rb's own comment frames this as a deferred descriptor-threading seam — confirm whether closing it belongs in U3 or a later unit.

## Medium
- [ ] Stage-prefix validation is narrowed vs the removed `Hive::Stages.parse` (`lib/hive/task.rb:127-138`): the new check `stage.index == @stage_index` collapses `"04".to_i == 4`, so a hand-built `04-execute` folder is now accepted (reproduced) where `Stages.parse("04-execute")` previously rejected it for not being exactly in `DIRS`; the on-disk basename stays `04-execute` while `stage.dir` is `4-execute`, breaking the byte-identical guarantee for non-canonical prefixes (pathological — hive never emits leading-zero dirs).

## Nit
- [ ] `TaskMeta.update_display_name` → `write` does not round-trip the new `workflow:` field (`lib/hive/task_meta.rb:42-69`): once a later unit persists `workflow:` in meta.yml, a display-name rename will silently drop it and revert the task to coding. Dormant in U3 (selector is never written yet) — flag so the future writer unit extends `write`/`update_display_name` too.
- [ ] Broken project `config.yml` now emits one "failed to read default_workflow … falling back to coding" warning per task (`lib/hive/task.rb:118-125`): `hive status` over a project with a malformed config logs N stderr warnings (one per field-less `Task.new`) where it previously listed silently; stdout/JSON unaffected, but consider memoizing the resolved default per `project_root` (noted as R-1 mitigation).
- [ ] Stale warn text: `TaskMeta.read`'s malformed-YAML branch still says "(depends_on dropped)" though it now also drops the new `workflow` field (`lib/hive/task_meta.rb:37-38`).
- [ ] Stale source reference: `wiki/state-model.md` cites `Hive::Task::STATE_FILES` at `lib/hive/task.rb:6`, but the constant now lives at `lib/hive/task.rb:11`.

ivankuznetsov:

## High
- _No High-severity findings. (The two `state-model.md` line-anchor items a persona tagged High are stale doc references — recorded as Nit below. End-to-end non-coding execution risk is captured as a Medium testing gap.)_

## Medium
- [ ] [code-reviewer, silent-failure-hunter] `lib/hive/task.rb:118-125` — `project_default_workflow` calls `Hive::Config.load(@project_root)` uncached in every field-less `Task.new` (pre-U3 the ctor did zero config I/O); `hive status`/daemon/TUI build one Task per folder, so this adds M redundant full YAML-parse+`validate!` per project per poll, and on a project with a broken `config.yml` it also emits one "failed to read default_workflow" warn per task per poll — perf regression + stderr flood that scale with task count. Suggest memoizing the resolved default per `project_root`.
- [ ] [code-reviewer] `lib/hive/task.rb:30,134` (`validate_workflow_stage!`) — a non-canonical leading-zero prefix like `04-execute` now passes (`"04".to_i == 4`) where the removed `Hive::Stages.parse("04-execute")` rejected it (verified); the on-disk basename stays `04-execute` while `stage.dir` is `4-execute`, breaking byte-identical parity for malformed paths. Compare `stage.dir == stage_dir` for strict parity.
- [ ] [silent-failure-hunter, comment-analyzer] `lib/hive/task_meta.rb:31-38` — the rescue comment + `warn` still say "(depends_on dropped)" but the same path now also drops the new `workflow:` selector, so a corrupt `meta.yml` silently reverts a tagged task to `coding` and the only breadcrumb misdirects the operator. Update the message to name `workflow` too.
- [ ] [pr-test-analyzer] `test/unit/stages/resolver_test.rb:56`, `test/unit/commands/run_test.rb:97` — tests only assert `pick_runner`/`resolve` RESOLVES a non-coding agent stage to `Hive::Stages::Agent.method(:run!)`; none drive `runner.call(task, cfg)`, so end-to-end execution of a generic agent stage is unproven — and it would currently raise (`Agent.run!` re-resolves the stage from `Registry.default`, `agent.rb:19`, not `task.workflow`).
- [ ] [type-design] `lib/hive/workflow.rb:15-24` — `stage_named` returns `nil` on an unknown stage while `state_file_for` raises `KeyError` on the same miss; the soft-lookup-vs-hard-resolve split is defensible but undocumented and reads as accidental — one comment makes the dual contract intentional.
- [ ] [pr-test-analyzer] `lib/hive/task.rb:120` — the `workflow.empty? ? "coding"` arm (an explicitly blank `default_workflow:` override) has no dedicated test; plan R-5 calls for a test per new branch. (The coverage gate is line-only, so CI still reports 100% — this only violates the plan's stated per-branch intent.)

## Nit
- [ ] [code-reviewer, type-design, code-simplifier] `lib/hive/task.rb:111-112` — `selector.nil? || selector.empty?`: the `.empty?` branch is unreachable because `TaskMeta.read` runs `normalize_string` (blank → nil); simplify to `selector ||= project_default_workflow`. Also a fragile cross-file assumption — it would `NoMethodError` if `meta[:workflow]` ever became a Symbol.
- [ ] [code-simplifier] `lib/hive/task.rb:118-120` — the local variable `workflow` shadows the public `#workflow` accessor; it holds a raw config String, so the shadow obscures the type — rename to `name`/`configured`.
- [ ] [code-simplifier] `lib/hive/task.rb:30,127` — `validate_workflow_stage!` is passed a hand-built `"#{m[:stage_idx]}-#{m[:stage_name]}"` for the error label though `Stage#dir` already formats `"<idx>-<name>"`; the dir-label convention now lives in two places.
- [ ] [code-simplifier] `lib/hive/task.rb:120,124` — the `"coding"` fallback literal is duplicated and re-hardcodes `Config::DEFAULTS["default_workflow"]` (config.rb:14); three spots can silently disagree if the project default ever changes.
- [ ] [type-design] `lib/hive/task.rb:13,15,41-43` — dual stage-name API: the coding-pinned `Task::STAGE_NAMES` constant now coexists with the descriptor-driven `Task#stage_names` instance method (they diverge for non-coding workflows); add a "coding-default only; prefer the instance accessor" comment (full consumer migration is U6).
- [ ] [type-design] `lib/hive/task_meta.rb:24,72` vs `:42-57` — read/write asymmetry: `read`/`empty` surface `workflow:` but `write`/`update_display_name` never persist it, so a non-coding selector can only arrive via hand-edited `meta.yml`. (Plan-deferred — writing the selector is a later unit; noted so the writer unit closes the round-trip.)
- [ ] [type-design] `lib/hive/workflow.rb:3-16` — `stage_named` relies on a stage-name-uniqueness assumption (`find`) that `Workflow#initialize` doesn't enforce; a duplicate-name descriptor is silently constructible (only one descriptor exists today, so optional).
- [ ] [silent-failure-hunter] `lib/hive/task.rb:110-116` — D1 is correct, but a typo'd `workflow:` selector now makes that task silently invisible in `hive status` (skipped by `rescue InvalidTaskPath; next`, status.rb:524) with no per-task breadcrumb; consider surfacing skipped-for-bad-selector tasks (the message is preserved on the direct `hive run <path>` path).
- [ ] [comment-analyzer] `wiki/state-model.md:58` — `STATE_FILES` anchor `lib/hive/task.rb:6` is stale (now line 11); U3 re-edited this exact sentence and left the anchor.
- [ ] [comment-analyzer] `wiki/state-model.md:38` — `PATH_RE` anchor `lib/hive/task.rb:14` is stale (now line 12; U3 added lines 10-11 which shifted it).
- [ ] [comment-analyzer] `wiki/modules/workflows.md:15` — stale `agent.rb:17`/`agent.rb:30` anchors (real lines 21/35; pre-existing drift but U3 rewrote the surrounding paragraph) and the paragraph is still dated "In U2".
- [ ] [pr-test-analyzer] `test/unit/task_test.rb:166` — `test_malformed_meta_selector_falls_back_to_coding` actually exercises `TaskMeta.read`'s YAML rescue (meta → empty → project default), not a malformed-selector arm inside `resolve_workflow`; the name is mildly misleading.
- [ ] [pr-test-analyzer] `test/unit/task_test.rb:153` — the `UnknownWorkflow → InvalidTaskPath` re-raise is tested only via a meta selector, not via a project-default `default_workflow: nope`; same rescue, low risk but undt covered.
- [ ] [pr-test-analyzer] `test/support/workflow_helpers.rb:9-11` — `research_workflow` omits `advance_verb`/`status_mode`/`budget_usd`/`timeout_sec`; fine as a resolution-only fixture, but reuse for dispatch-time (budget/timeout) behavior would silently get nils — worth a comment.

ivankuznetsov:

## High
- [ ] `lib/hive/stages/agent.rb:19` re-resolves agent stages from `Registry.default`: a non-coding `kind: :agent` stage selected via `Run#pick_runner` still raises `Hive::StageError` before spawn, so Q6's workflow-specific run path is not actually implemented.

## Medium

## Nit

ivankuznetsov:

## High
- [ ] `bundle exec rake test` is RED on HEAD — `TaskMetaTest#test_malformed_yaml_warns_that_depends_on_was_dropped` (`test/unit/task_meta_test.rb:88`) still asserts `assert_match(/depends_on dropped/, err)`, but the pass-01 fix (commit `cfc80b77`, `lib/hive/task_meta.rb:38`) changed the warn to `treating meta as empty (depends_on, workflow dropped)`; `/depends_on dropped/` no longer matches (the `, workflow` breaks the substring), so the suite fails (`2 failures` — the other is an env-leak false positive, see Notes). The plan's gates (`rake test`, and therefore `rake coverage` 100%) do not pass: fix is updating the regex to `/depends_on, workflow dropped/` (and ideally the test name/comment at `:74-82` which still say only `depends_on`).

## Medium
- [ ] Stale/incorrect comment introduced by U3 at `lib/hive/stages/agent.rb:12-18`: it states `pick_runner` "always calls `Resolver.resolve(task)` with the default, and that resolver's `descriptor:` param is test-only." After U3, `pick_runner` (`lib/hive/commands/run.rb:138`) passes `descriptor: task.workflow` and the param is production-used — the comment is now false on both counts and contradicts the refreshed wiki (`wiki/modules/workflows.md`, `wiki/stages/agent.md`), misleading the next maintainer about the exact seam this unit changed (the remaining gap is only that the resolver doesn't thread the chosen descriptor *into* the runner).

## Nit
- [ ] `wiki/modules/task.md` documents the pre-fix `PATH_RE` literal (no `(?<stage_dir>…)` group) while the code now has `(?<stage_dir>(?<stage_idx>\d+)-(?<stage_name>…))` at `lib/hive/task.rb:16`; the page was otherwise refreshed for U3, so the regex literal drifted. (The same section also still calls `PATH_RE` "the only validator for task paths," though stage-name/index validity now lives in `validate_workflow_stage!`.)

## Notes (not findings)
- The second `rake test` failure, `StopHookInstallerTest#test_stop_hook_requires_stage_dir_env` (`test/unit/stop_hook_installer_test.rb:143`, "Expected true to not be truthy"), is an **environment-leakage false positive** in this review sandbox: the reviewer runs inside the 6-review stage with `HIVE_TASK_STAGE_DIR` exported, which the hook subprocess inherits, so it succeeds where the test expects failure. The test is byte-identical on `origin/main` and passes in clean CI (no `HIVE_TASK_STAGE_DIR`). Not attributable to U3; do not chase it.
- Verified clean: `rubocop` (7 touched files, no offenses); coding descriptor maps byte-identically to the old `STAGE_NAMES`/`STATE_FILES`; pass-01 fixes confirmed present (`stage.dir == stage_dir` parity at `task.rb:140`, `selector ||=` simplification, `Config::DEFAULTS["default_workflow"]` reuse, `stage_dir` PATH_RE group); IU-4 rescue branches and IU-5 `pick_runner`/resolver wiring are covered by tests.
- The unanimous `Agent.run!` re-resolves-from-`Registry.default` finding (raised again in `codex-ce-code-review-02`) is **dropped** as a deliberate plan scope boundary: brainstorm Q6 / IU-5 scope U3 to the single `pick_runner` swap, `agent.rb` is outside the in-scope file list, acceptance requires `pick_runner` to *resolve* (not execute) the agent runner, and it is dormant in production (registry holds only `:coding`). Captured here only as the Medium comment-accuracy defect above.

ivankuznetsov:

# pr-review-toolkit multi-persona review — u3-task-resolves-its-stages (pass 02)

Scope: `git diff origin/main..HEAD`, U3 surface (`task.rb`, `workflow.rb`,
`task_meta.rb`, `config.rb`, `commands/run.rb` + tests). Stacked branch carries
landed U1/U2; findings concentrate on U3. Personas run: code, tests, errors
(silent-failure), comments, type-design, simplify.

Gate status (re-confirmed by personas): rubocop clean on all changed lib files;
U3 unit tests green; golden test pins `STAGE_NAMES`/`STATE_FILES` byte-identical
to the prior literals. The only `rake test` failure in this environment
(`StopHookInstallerTest#test_stop_hook_requires_stage_dir_env`) is environmental
(reviewer process exports `HIVE_TASK_STAGE_DIR`; that test doesn't sanitize the
child env) and is not part of this diff. The code-simplifier found **no**
actionable simplifications — the change is tight and deliberately contained.

## High
- [ ] [silent-failure] Typo'd project-level `default_workflow` (valid YAML, unregistered name e.g. `codng`) silently removes the **entire** project from `hive status` and `--json`: `project_default_workflow` (`task.rb:124-126`) returns the typo verbatim **without warning** (its rescue only catches `ConfigError`/`Psych`/IO, not a clean-but-unknown value), `Registry.fetch` raises `UnknownWorkflow` → re-raised as `InvalidTaskPath` (`task.rb:120-121`), and since the default applies to every field-less task, `Status#collect_rows` (`status.rb:525-527`) `next`s past all of them — the daemon then sees an empty/healthy project and silently stops dispatching, with no row, warn, or diagnostic. This violates the plan's own "fail loud" (Q4a) and "preserve status-listability" (R-2/D2) intents; the validator is plan-deferred, but a one-line `warn` in `project_default_workflow` when the resolved name isn't in `Registry::WORKFLOWS.keys` (no Config↔Registry coupling) would restore observability in-scope.

## Medium
- [ ] [silent-failure/type-design] `resolve_workflow` re-raises `UnknownWorkflow` as `InvalidTaskPath` (`task.rb:120-121`), conflating "malformed folder path" with "valid path, bad workflow selector"; every existing `rescue InvalidTaskPath` swallow site (status.rb:525, tui/bubble_model.rb:1731/1861/1909, approve.rb, stage_action.rb) was written for the former and now silently absorbs a config/selector misconfiguration too — D1 chose this deliberately, but the residual blast radius beyond `hive run` is broader than the plan's "skips one task" rationale claims.
- [ ] [tests] `project_default_workflow`'s 4-class rescue (`ConfigError, Psych::Exception, SystemCallError, IOError`, `task.rb:127`) is only exercised via `ConfigError` (`task_test.rb:207` feeds valid-YAML-that-fails-validation); a genuinely unparseable `config.yml` or an EACCES read goes through the same arm untested — a future narrowing of the rescue would crash the ctor with no test catching it. Add one malformed-YAML-in-config.yml case.
- [ ] [tests] No test drives `validate_workflow_stage!`'s unknown-stage or index-mismatch raise under a **non-coding** workflow (`task.rb:133`); both mismatch tests run against the default coding descriptor, so a bug comparing against `Registry.default` instead of the resolved `workflow` would still pass every test. A `research`-workflow task at an index-mismatched/unknown stage asserting the message names `workflow :research` would close it.

## Nit
- [ ] [silent-failure/type-design] `Task#state_file` → `workflow.state_file_for` raises a bare Ruby `KeyError` (`workflow.rb:23`) that is not a `Hive::Error`, so it bypasses the structured error envelope; unreachable for any constructed Task today (ctor `validate_workflow_stage!` guarantees membership, fields are read-only), so pure defense-in-depth — but raising a `Hive::Error` subclass would keep the taxonomy consistent for future non-coding workflows.
- [ ] [code] Latent round-trip data-loss: `TaskMeta.write`/`update_display_name` (`task_meta.rb:42`,`:59`) don't preserve the new `workflow:` field, so once any writer emits `workflow: research`, a later `update_display_name` (called by the live pipeline via `display_name/generator.rb`) erases it and the task reverts to the default. Plan IU-2 deliberately defers the selector *writer* to a later unit (read-only in U3), so this is correct-for-now — flag it so that downstream unit preserves the field (a regression test pinning the intentional drop today would document the seam).
- [ ] [tests] `test_state_file_per_stage` (`task_test.rb:54`) first assertion compares two values both derived from `Registry.default` (self-referential); harmless because the adjacent literal-basename assertion (`:55`) and the golden test provide the independent anchor.
- [ ] [tests] No precedence test for the case where a per-task `meta.workflow:` and a project `default_workflow:` are both set to *different* known workflows (task selector must win via `selector ||= project_default_workflow`, `task.rb:118`); the one precedence combination left unpinned.
- [ ] [type-design] public `attr_reader :workflow` (`task.rb:19`) exposes the full descriptor (budget_usd/timeout_sec/skill/advance_verb) when Task's own surface only needs the `state_file`/`stage_names` delegators — plan-required (Q5/Q6 pass `task.workflow` to the resolver), so intentional as the U6 migration seam; noted so the over-exposure is an explicit trade-off, not an oversight.
- [ ] [type-design] Once U6 registers more workflows, adding a `meta.workflow:` selector whose stage set lacks the on-disk `@stage_name` makes an existing folder un-constructable (fail-closed `InvalidTaskPath`); correct safety behavior today (only `coding` is registered), but a latent foot-gun worth a wiki/guard note when the selector writer lands.
- [ ] [comments] Milestone-coupled comments will rot: `task.rb:10-13` ("full consumer migration is U6") and `agent.rb:12-19` ("U3 seam (remaining gap)") have no in-file anchor and go stale on renumber / when workflow #2 is registered (the agent.rb one is self-announcing — it tells the maintainer to act exactly when it becomes false). Accurate today; optionally anchor to a plan doc.

ivankuznetsov:

# ce-code-review — pass 03

Scope: `git diff origin/main..HEAD` (stacked worktree; U1/U2 already reviewed — findings focus on U3: `task.rb`, `task_meta.rb`, `config.rb`, `commands/run.rb`).
Verdict: **Ready.** U3 is correct and plan-faithful. The byte-identical guarantee for field-less coding tasks holds (`stage_named` + `stage.dir == stage_dir` + `state_file_for` ≡ old `STAGE_NAMES.include? && Stages.parse`); `Task.new` is total for field-less tasks (every `Config.load`/`TaskMeta.read` error is rescued or re-raised as `InvalidTaskPath`, which `Status#collect_rows` skips at status.rb:525); `UnknownWorkflow` correctly re-raises as `InvalidTaskPath`/exit 64; `pick_runner` is behavior-preserving for coding. No in-scope High/Medium defect found. Items below are plan-deferred risks and test/style nits.

## High
_None._

## Medium
_None._

## Nit
- [ ] Latent selector data-loss in existing write paths (`task_meta.rb:42-69`, migrate id-backfill): U3 makes `meta.yml` `workflow:` load-bearing for reads, but `TaskMeta.write`/`update_display_name` don't persist it — harmless today (only `:coding` registered, so a dropped selector falls back to the identical coding default), but once a 2nd workflow lands the daemon's display-name backfill will silently wipe a hand-tagged selector; plan defers write-support (IU-2) but its "U3 only reads" rationale overlooks the automatic round-trip, so the later tagging unit must add `workflow:` to `write`.
- [ ] Vacuous test `test_stage_names_returns_descriptor_stage_names` (`workflow_test.rb:74-78`): asserts `workflow.stages.map(&:name) == workflow.stage_names`, but `stage_names` *is* `stages.map(&:name)` (`workflow.rb:32`) — always true regardless of values; anchor to an independent literal (real coverage already exists at `task_test.rb:56-57`).
- [ ] Dead `require "hive/stages"` (`task.rb:3`): `Hive::Stages` has zero references after the ctor dropped `Hive::Stages.parse`; the plan kept it for "other consumers" but none remain in this file — low-risk removal (confirm no transitive loader relies on it).
- [ ] stderr-only signal for a typo'd project `default_workflow` (`task.rb:124-149` + `status.rb:522-527`): an unregistered default drops *every* field-less task from `hive status`, so `--json` reports `ok:true, tasks:[]` and the daemon sees an empty project with only a stderr warn; this is the documented consequence of the deliberate no-validator decision (IU-3), flagged so the option of a config-load validator / status error-key is on record.
- [ ] Double `Registry.fetch` + split warn/raise in `project_default_workflow` (`task.rb:124-149`): `warn_if_unregistered_project_default` fetches-and-discards purely for the warn, then `resolve_workflow` fetches again to raise; functional and documented, but inlining the warn would make the control flow linear.
- [ ] stderr warn leak in `test_meta_readers_fall_back_when_sidecar_absent_or_malformed` (`task_test.rb:349-367`): the malformed-`meta.yml` `Task.new` isn't wrapped in `capture_io`, leaking the U3-updated "depends_on, workflow dropped" warn to test output — wrap and assert the message as `task_meta_test.rb` already does.

---
Verified out of scope / correctly deferred (not findings): the agent runner re-resolving its stage from `Registry.default` instead of `task.workflow` (`stages/agent.rb`) is unreachable in U3 (only `:coding` registered) and documented as a U6 seam; the ~20 `STAGE_NAMES`/`STATE_FILES` consumers (e.g. `stages/open_pr.rb`) are explicitly deferred to U6; declining a `default_workflow` config validator is an explicit IU-3 decision.

ivankuznetsov:

# pr-review-toolkit — pass 03

Union of 6 personas (code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier) on `git diff origin/main..HEAD`, scoped to U3 (`task.rb`, `task_meta.rb`, `config.rb`, `commands/run.rb`; U1/U2 already reviewed).

Verdict: **Ready.** code-reviewer found no high-confidence defect; pr-test-analyzer confirmed 100% line **and** branch coverage on every U3 file; all Q1–Q7 requirements are implemented. Every "Critical/Important" persona finding maps to a deliberate plan scope-boundary and is demoted to Nit with rationale: the `validate_default_workflow!` validator is **explicitly declined** by IU-3; the `open_pr.rb` `STATE_FILES` consumer and agent-runner descriptor-threading are **deferred to U6**; the malformed-meta selector fallback is the **Q4b** intended behavior. (Note: a `StopHookInstallerTest` failure under `rake coverage` is environmental — the harness exports `HIVE_TASK_STAGE_DIR`; the file is untouched by U3 — not a finding.)

## High
_None._

## Medium
_None._

## Nit
- [ ] [silent-failure-hunter] Typo'd project `default_workflow` makes every field-less task raise `InvalidTaskPath`, which `Status#collect_rows` skips (status.rb:524-526), so `hive status --json` reports the whole project as `tasks:[]` with no error key and the daemon (JSON consumer) is blind to the stderr-only warn (task.rb:124-149): plan-acknowledged — IU-3 deliberately declined a `validate_default_workflow!` validator and chose warn + fail-at-resolution; flagged as an accepted tradeoff whose documented future hardening is the config-load validator.
- [ ] [type-design-analyzer] Coding-pinned `Task::STAGE_NAMES`/`STATE_FILES` coexist with the per-task instance `stage_names`/`state_file`, and the one production consumer still on the constant (`stages/open_pr.rb:353,364` `STATE_FILES.fetch("review"/"artifacts")`) becomes a wrong-state-file bug once a non-coding workflow registers: consumer migration is explicitly U6 and the constants are intentionally kept unchanged, so this is a forward-looking trap, not a U3 defect.
- [ ] [silent-failure-hunter] A malformed/unreadable `meta.yml` silently drops the `workflow:` selector and falls back to project-default/coding (task_meta.rb:30-39) — the Q4b graceful-fallback intent; harmless today (only `:coding` registered) but fail-open (runs the wrong pipeline) the day a 2nd workflow lands, so consider refusing on "meta present but unreadable" or sharpening the warn to name the pipeline-substitution risk.
- [ ] [code-simplifier] Dead `require "hive/stages"` (task.rb:3): `Hive::Stages` is no longer referenced after the ctor dropped `Hive::Stages.parse` — low-risk removal (the plan assumed "other consumers" but none remain in this file; verify no transitive loader depends on it).
- [ ] [code-simplifier] Double `Registry.fetch` on the project-default path (task.rb:119,128,144): `warn_if_unregistered_project_default` fetches purely for the warn, then `resolve_workflow` fetches again — fold the warn into `resolve_workflow`'s rescue (guarded by "selector came from project default") for one linear fetch, preserving the exact warn text and scope.
- [ ] [pr-test-analyzer] `test_meta_readers_fall_back_when_sidecar_absent_or_malformed` (task_test.rb:361) constructs a Task with malformed `meta.yml` without `capture_io`, leaking the U3-updated "depends_on, workflow dropped" warn to stderr (U3-introduced: ctor now reads meta eagerly) — wrap in `capture_io` like the sibling tests at lines 266/281/300.
- [ ] [pr-test-analyzer] Tautological assertions: `workflow_test.rb:77` asserts `workflow.stages.map(&:name) == workflow.stage_names` but `stage_names` *is* `stages.map(&:name)` (cannot fail); `task_test.rb:58` self-references `STATE_FILES` (both sides derive from `Registry.default`) — anchor each to an independent literal (real coverage already exists at task_test.rb:56-59/155).
- [ ] [pr-test-analyzer] `test_pick_runner_uses_on_disk_task_workflow_selector` (run_test.rb:97-113) and `resolver_test.rb:48-54` assert the agent runner *resolves* for a research `gather` stage, but *calling* it raises `no agent stage gather` (agent.rb:20 re-resolves from `Registry.default`) — the selection-only tests step around the documented U6 runtime gap; when U6 threads the descriptor, add a test that `.call`s the resolved runner.
- [ ] [comment-analyzer] `stages/agent.rb:12-13` "the resolver's `descriptor:` param is production-used" overstates: the param is *passed* per-task but its value is inert in production (CODING_RUNNERS name-first early-return covers all 9 coding stages, so `descriptor:` is never consequentially read until a 2nd workflow registers) — soften the opening clause to match the comment's own "dormant today" tail.
- [ ] [type-design-analyzer] `validate_workflow_stage!` (task.rb:151-162) implicitly depends on `@workflow` being set first (ctor line 33 before 34) — pass the workflow in explicitly so the construction-order invariant isn't reorder-fragile.
- [ ] [type-design-analyzer] `Workflow#stage_names` (workflow.rb:30-32) returns a fresh **unfrozen** array each call while the parallel U1 constants are frozen — `freeze` (or memoize) it for a uniform immutability contract across both sources of truth.
- [ ] [type-design-analyzer] The soft (`stage_named` → nil) vs hard (`state_file_for` → raises `KeyError`) dual lookup contract on `Workflow` is coherent but not self-evident from the names — lead each method's doc comment with the return-vs-raise contract (renaming `state_file_for` is discouraged: it's the brainstorm-named API).

ivankuznetsov:
Superseded by #562, which consolidates the full U1–U10 workflow-as-data inversion into one branch off current main (all unit fixes integrated, gates green) for single-feature evaluation. Closing u3's standalone PR.

PR #540, opened 2026-06-20
feat(workflow): resolve generic task advances from the descriptor

ivankuznetsov:
# feat(workflow): resolve generic task advances from the descriptor

## Summary

A non-coding (generic) task could not advance correctly. `hive approve` and `hive run --json` resolved the next stage through the global coding stage table (`Hive::Stages.next_dir`/`resolve`/`parse`), so a generic task either crashed (`parse` returns `nil` on a dir the coding table doesn't know) or landed in a coding directory. This PR makes verb and sequence resolution per-workflow, derived from the task's active descriptor, so a generic task advances to its own workflow's next stage. Coding behavior is byte-identical, and a golden test pins it.

This is the physical-forward-move half of the workflow-as-data inversion. U4 (#539) shipped the classification half: `TaskAction` routes non-coding tasks through the generic path behind a `coding_workflow?` gate. U5 rewires the two command sites that still reached the global coding table so they resolve through `task.workflow`.

**What changed** — descriptor-scoped resolution moves onto `Hive::Workflow`, and the advance sites consume it:

| Method on `Hive::Workflow` | Purpose | Replaces |
|---|---|---|
| `next_stage_after(name)` | next `Stage` by array order, `nil` at terminal | `Hive::Stages.next_dir` |
| `advance_verb_for(name)` | the stage's incoming verb, `nil` ⇒ mv-only | `Workflows.verb_arriving_at` |
| `stage_for_dir(dir)` | `Stage` for a dir, `nil` if unknown | `Hive::Stages.parse` |
| `resolve_stage_ref(ref)` | canonical dir for a full dir or short name | `Hive::Stages.resolve` |
| `stage_dirs` | frozen dir list for error messages | `Hive::Stages::DIRS` |

- `Commands::Approve` resolves `--to`/`--from`, forward destinations, move validation, success-payload stage fields, and the `next:` hint through `task.workflow`. An unknown destination now raises a typed `InvalidTaskPath` (listing the descriptor's own dirs) instead of a `NoMethodError` on a `nil` parse.
- `Commands::Run#next_stage_dir` emits its `approve` next-action for the descriptor's actual next stage.
- "mv-only" is now a property of a stage with no declared `advance_verb`, not of `id == :coding`. Such a stage's `next:` hint falls back to `hive run <slug>`, never a raw `mv`.

**Scope boundaries**

[...]

PR #539, opened 2026-06-20
feat(task-action): classify non-coding workflows from the descriptor

ivankuznetsov:
## Summary

After U1–U3, a registered non-coding workflow task resolves a valid `Hive::Workflow` descriptor, but `TaskAction#action` was still a hardcoded `case task.stage_name` over the nine **coding** stages whose `else` fell to `ACTIONS[:error]`. So the task rendered ERROR in `hive status` and the daemon — which advances only when `action` is in `ADVANCE_ACTIONS` and dispatches `TaskAction#command` — never advanced it. This is the load-bearing unit that unblocks status classification and daemon auto-advance for non-coding workflows.

U4 branches `#action` and `#command` on **workflow identity**: both short-circuit to descriptor-driven generic helpers `unless coding_workflow?`, inserted **after** the existing marker-keyed pre-`case` short-circuits (agent-working, error, manual-steering), so generic stages inherit those unchanged. The coding workflow reaches its existing `case` byte-for-byte unedited — making zero-regression true by construction rather than by review.

**What changed:**

- New `:ready_to_advance` action kind (`Hive::Schemas::TaskActionKind`, both status JSON schemas, and the `ACTIONS` table) carries a generic non-terminal COMPLETE into the daemon's advance path.
- `generic_action` classifies by marker **plus descriptor position** — first/terminal stages are detected as `stages.first`/`stages.last`, never by name:

  | Marker @ stage | Action | Command |
  |---|---|---|
  | COMPLETE @ non-terminal | `ready_to_advance` | `hive approve <slug> --from <dir>` |
  | COMPLETE @ terminal | `archived` (`:done`) | — |
  | WAITING | `needs_input` | `hive run <slug>` |
  | none @ **inert** entry | `ready_to_advance` | `hive approve <slug> --from <dir>` |
  | none @ non-inert / non-entry / other | `needs_input` (ready to run) | `hive run <slug>` |

  The markerless entry advance is gated on `entry && !terminal && stage.kind == :inert`, so a future workflow whose entry stage runs an agent (`kind: :agent`) is never auto-advanced past its own work.

- `generic_command` builds the command string directly rather than through the coding verb table (whose `--from`/`--stage` logic keys on `Workflows.workflow_verb?`, false for `approve`/`run`). `approve` is the move-forward primitive; the plan's original `hive run … --from` was invalid (`hive run` has no `--from` and re-runs the current stage).
[...]

PR #543, opened 2026-06-20
feat(workflow): run non-coding workflows end-to-end under the daemon

ivankuznetsov:
## Summary

Before this change a non-coding workflow could be *described* by a `Hive::Workflow` descriptor but could not *run*: roughly 20 consumers embedded raw `"N-stage"` literals, and the daemon/CLI advance path was pinned to the coding pipeline, so a generic task was silently stranded (it never surfaced in `status`, couldn't be found by slug, and never auto-dispatched). U6 is the capstone of the U1–U5 descriptor stack: a registered generic workflow now advances end-to-end under the real daemon path, every hardcoded stage literal is either routed through the descriptor or explicitly annotated, and a durable guard test makes "no un-routed literal" a CI invariant.

> **Stacked work.** This branch carries the cumulative U1–U5 descriptor foundation (already in flight as #529, #530, #538, #539, #540) and opens against `main`; the new work in this task is U6. The isolated U6 delta is ~75 files / ~890 lines.

### The end-to-end path U6 unblocks

```mermaid
flowchart TB
  A["hive status --json<br/>(row carries workflow id)"] --> B["Daemon::Policy.decide<br/>ready_to_run / ready_to_advance → :dispatch"]
  B --> C["hive approve &lt;slug&gt; --from &lt;dir&gt;<br/>(enum relaxed, validated in-command)"]
  C --> D["TaskResolver<br/>finds task by bare slug at a generic dir"]
  D --> E["Stages::Resolver<br/>routes generic stage to the Agent runner"]
  E --> F["Stages::Agent<br/>resolves stage from task.workflow"]
  F --> A
```

Each hop previously assumed the coding workflow; U6 swaps the coding-pinned source for a descriptor lookup or the registered-workflow union, while keeping coding byte-identical (pinned by characterization tests written *before* the refactor).

### Key design decisions

- **Relax-and-validate the stage enum, don't recompute it.** `cli.rb` dropped the static Thor `enum: APPROVE_TO_ENUM` from every `--from/--to/--stage` option and validates the stage ref inside the command against the resolved task's workflow. A class-load-time enum can never see a runtime-registered workflow, so a static union would still break the generic path. Coding's exit-code contract (exit 4 `WRONG_STAGE`, unknown-stage usage error) is preserved, and the default-workflow stage list is embedded in each option `desc:` so `hive help` keeps stage discoverability.
- **Un-collapse first-run dispatch.** `generic_ready_to_run` previously shared the `NEEDS_INPUT` action key with `generic_needs_input`, so the daemon could not distinguish "fresh stage, never run" from "agent ran, now waiting on the user" and never auto-dispatched a never-run generic stage. It now maps to a distinct `READY_TO_RUN` action key (additive schema enum value), which `Daemon::Policy` routes to `:dispatch` — gated through the same marker/mtime debounce brake as edit-resume so a markerless agent can't re-dispatch unboundedly.
- **Gate name-first runner precedence on the coding descriptor.** `Stages::Resolver` keeps coding's bespoke `plan`/`brainstorm` runners via a name-first lookup, but only when `descriptor.id == :coding`; a generic stage named `plan` routes to the generic `Agent` runner by its `kind: :agent`, so a name collision can't hijack it.
- **Scan the registered-workflow union, not the coding dirs.** `status.rb` (`collect_rows` + `detect_legacy_stage_dirs`), `task_resolver.rb`, `drop.rb`, and the daemon's post-advance state-file lookup + `stage_rank` scan `Workflows.all_stage_dirs` so a generic `2-gather` task surfaces, is findable by slug, is never mis-flagged legacy, and auto-dispatches its next stage under a live tick. The union equals the coding dirs today, so coding is unchanged.
- **Coding-specific concepts stay coding-scoped, by the bright-line test.** A literal is coding-scoped only when it names a concept that exists *only* in coding (PR / review / finalize / unpushed-commits / bench-submit) with no sensible generic analog. The daemon's `3-plan` plan-auto-approval, for example, is now guarded on `coding_workflow?` so a generic plan pause is an annotated, tested skip — never a misfire.

### How every stage literal is accounted for (NO-SILENT-CAPS)

`test/unit/stage_literal_guard_test.rb` greps all of `lib/` for stage literals (double-quote, single-quote, `%w[...]`, and interpolation forms) and fails unless each hit is gone or carries `# coding-scoped: <reason>` / `# not-a-stage-ref: <reason>` with a non-empty reason. The guard self-tests its own regex against fixtures so it can't silently match nothing, and includes a positive control proving it fails on a planted un-annotated literal.

| Treatment | What it means |
|---|---|
| **Migrated** | Literal replaced by a `task.workflow` / `row[:workflow]` descriptor lookup (status union, resolver, agent runner, policy threading, healer re-dispatch, daemon post-advance lookup) |
| **`# coding-scoped:`** | References a coding-only concept (PR/review/finalize/unpushed/bench-submit, TUI per-stage affordances, plan auto-approval); guarded so a generic task skips rather than misfires |
| **`# not-a-stage-ref:`** | Historical/legacy stage names or doc-comment examples (migrate tables, recovery remap, comments) |

### New shared affordances (additive, no behavior change to existing methods)

- `Workflow#has_stage?(ref)` — presence predicate accepting a dir (`"6-review"`) or a name (`"review"`).
- `Registry.all` / `Registry.ids` — live enumeration read at call time so a test-registered workflow is visible.
- `Workflows.all_stage_dirs` / `all_stage_names` and a shared `coding_row?` / `coding_id?` predicate — the union the gates scan and the single source of the "what counts as coding" rule; `VERBS` / `Stages::DIRS` stay coding-pinned.
- `workflow` id on the status row + the additive optional `workflow` property on `hive-status.v4.json` (no version bump, mirroring the `live_task_lock` / `unanswered_questions` precedent) so row-based consumers (bot/daemon/digest) can route. The `stage`/`dependency_stage` fields on the status/approve/run schemas were relaxed from a closed 9-value enum to a `^[0-9]+-…$` pattern so generic dirs validate.
- `task_meta.rb` now persists and round-trips the `workflow:` selector through both `update_display_name` and `hive migrate` id-backfill, so a mid-flight rename or id self-heal never drops it.

## Test plan

Characterization/golden tests (`title_formatter`, `stale_agent_healer`, `status`, `policy`) were written *before* the refactor and stay byte-identical green for coding. `test/integration/generic_workflow_daemon_e2e_test.rb` registers a dispatch-capable generic workflow and drives the **product** path — `status --json` → `Policy.decide` → `hive approve --from <dir>` → resolver-by-slug → next-stage auto-dispatch — through a real `Dispatcher#tick`, asserting two real stage hops with an explicit "no coding-only branch fired" negative assertion, plus a parallel coding task advancing unchanged as a regression cross-check.

Verification (recorded during the work):

- `bundle exec rake test` → `5801 runs, 22929 assertions, 0 failures, 0 errors`
- `bundle exec rake coverage` → 100% line coverage on new/changed lines
- `bundle exec rubocop` → clean (648 files)
- `test/unit/stage_literal_guard_test.rb` → green (every `lib/` stage literal migrated or annotated)

## Review summary

Three review passes ran over `git diff origin/main..HEAD` using `claude-ce-code-review`, `codex-ce-code-review`, and the `pr-review-toolkit` persona suite. The coding paths held up as byte-identical-green throughout; every finding was auto-fixed or resolved from plan/brainstorm context, and no question was escalated to the user across the three passes.

Resolved **High** findings (all on the generic daemon-advance surface this PR introduces):

- Markerless generic stage could re-dispatch `hive run` every tick until the per-project daily cap drained — `ready_to_run` dispatch is now gated through the marker/mtime debounce brake (coding was already immune).
- Branch was 1 commit behind `origin/main` and missing PR #542 (`TaskIdBackfiller`) — merged with the task-id self-heal preserved (resolved in pass 2, re-reviewed in pass 3).
- Generic task stalled at its 2nd stage under a live daemon — `find_post_advance_state_file` and `stage_rank` swapped from coding `Stages::DIRS` to `Workflows.all_stage_dirs`, and the e2e now drives a real `Dispatcher#tick`.
- Generic `hive approve/run --json` envelopes violated the published `hive-approve.v2`/`hive-run.v2` schemas — enums relaxed to patterns matching the `hive-status.v4` precedent.
- Generic `Stages::Agent` swallowed `spawn_agent`'s `{status: :error}` envelope — it now writes an attributed `:error` marker (NO-SILENT-CAPS).

Resolved **Medium/Nit** themes: a typo'd `default_workflow` now surfaces a synthetic Error row instead of silently emptying a project; `Config.load` memoized per project to avoid a per-task `Task.new` regression; the duplicated `coding_workflow?` predicate consolidated; generic-side `stale_agent_healer` paths covered; descriptor `status_mode`/`timeout_sec` threaded into the generic runner; per-file rescue in `prior_artifacts`; the dead `capability`/`ACTIVE_STAGE_DIRS` removed; and a sweep of stale wiki/module-comment anchors corrected to name the new workflow/union seam. The auto-commit scope check correctly held back the two `schemas/*.json` edits in pass 1 for operator inspection (outside the fix-agent allowed paths); they were applied as part of the schema-relax work.

## Linked task

Hive task **#1458** — *Route Stage Consumers Through Descriptor* (slug `u6-route-the-20-hardcoded-260618-33f7`), depends on `u5-per-workflow-verbs-and-260618-36fc`. U6 is the capstone of the U1–U5 descriptor stack tracked in `wiki/gaps.md` (#139, #141).

---


ivankuznetsov:

## High
- [ ] `lib/hive/daemon/dispatcher.rb:821` / `lib/hive/commands/status.rb:283` still apply the coding brainstorm Q&A gate to generic rows: both checks key only on `2-brainstorm` + `needs_input`, so a non-coding workflow that reuses that dir can be held as `answers_pending` instead of taking the guarded generic path.

## Medium
- [ ] `lib/hive/bot/notification_builders.rb:14` and `lib/hive/web/dispatcher.rb:14` omit `ready_to_run` / `ready_to_advance`: generic ready rows get no Telegram notification/button and the web dispatcher rejects the new actions as unknown.
- [ ] `lib/hive/task_resolver.rb:168` resolves `--stage` short names against `Registry.all` before loading the task: `hive run <slug> --stage plan` can fail as globally ambiguous even when the resolved task workflow would make the stage unambiguous.
- [ ] `lib/hive/bot/handlers/recovery_sequence.rb:91` still derives retry verbs from the coding workflow table: generic recovery rows ignore `TaskAction`'s descriptor-safe retry command and either get "No retry verb" or a coding verb on colliding dirs.
- [ ] `lib/hive/commands/drop.rb:123` still scans only `Hive::Stages::DIRS`: generic task folders are invisible to `hive drop <slug>` and generic `--from` values are rejected despite the CLI now accepting them.

## Nit

ivankuznetsov:

# Code Review — U6: Route hardcoded stage-literal consumers through the descriptor

Scope: `git diff origin/main..HEAD` (focused on the U6 commits `787be6e4..HEAD`; U1–U5 were reviewed in their own passes). Mode: report-only. Reviewers: correctness + adversarial (Opus), plus orchestrator verification. U6-touched test suite re-run locally: **432 runs, 0 failures** — coding paths are byte-identical-green and the generic e2e passes. Findings below are edge cases the tests do not cover.

## High
- [ ] Generic `ready_to_run` re-dispatch cascade drains the per-project daily cap and then halts ALL dispatch (`lib/hive/daemon/policy.rb` `ADVANCE_ACTIONS` + `lib/hive/task_action.rb` `generic_ready_to_run`): moving `generic_ready_to_run` from `NEEDS_INPUT` to the new `READY_TO_RUN` key and adding `ready_to_run` to `ADVANCE_ACTIONS` makes `decide` return `:dispatch` with **no mtime/marker-progress debounce** (unlike the `edit_resume`/`needs_input` path that previously braked it). A fresh or post-run generic stage whose agent exits 0 without writing a `WAITING`/`COMPLETE` marker (`stages/agent.rb` leaves marker `:none`, and `run.rb` does not raise → no cooldown) recomputes to `ready_to_run` and re-dispatches `hive run` every tick until `max_runs_per_day_per_project` is exhausted, after which the daemon stops dispatching for the whole project. Coding is immune because a markerless post-run coding stage classifies as `needs_input` (debounced), not an advance action — this asymmetry is the defect. Cross-corroborated by both Opus reviewers (adversarial P1 / correctness P2). Latent until a real non-coding workflow registers (U7/U9), but the wiring is live now and the no-marker case is untested. Fix: gate `ready_to_run` dispatch on marker/mtime progress (route a never-progressed generic stage through a `record_baseline`/debounce brake), and add a test asserting a markerless generic run does not re-dispatch unboundedly.

## Medium
- [ ] `hive migrate` id-backfill silently drops the `workflow` selector (`lib/hive/commands/migrate.rb:189-196`): `backfill_task_ids` reads `meta = Hive::TaskMeta.read(folder)` (which now returns `:workflow`) but the following `TaskMeta.write(...)` forwards only `id/slug/display_name/depends_on` — omitting `workflow:`. An id-less task carrying `workflow: <non-coding>` loses its selector and reverts to the project default workflow. This is the exact selector-preserve gap the plan called out (A6) and that the sibling `update_display_name` site **was** fixed for (`task_meta.rb:69-70`); this write site was missed. Trivial, certain (confidence 100). Fix: add `workflow: meta[:workflow]` to the backfill `TaskMeta.write` call and pin it with a round-trip test.

## Nit
- [ ] Guard test enforces only annotation *presence*, not correctness, and its regex under-matches (`test/unit/stage_literal_guard_test.rb`): `STAGE_LITERAL` matches only single-digit `"\d-..."` prefixes, so a future two-digit stage literal (e.g. `"10-deploy"`) is silently unguarded; and a literal that *should* have been routed through the descriptor passes the net as long as it carries any `# coding-scoped:`/`# not-a-stage-ref:` comment. The durable net is weaker than the triage table implies — note the limitation or broaden the regex.
- [ ] `coding_workflow?` predicate is duplicated across five files in two slightly different shapes (`bot/notification_builders.rb`, `bot/notification_dispatcher.rb`, `bot/supervisor.rb`, `daemon/stale_agent_healer.rb` take a `row`; `daemon/policy.rb` takes a `workflow` value): five copies of the nil/empty/`"coding"` defaulting rule is a maintainability trap if the "what counts as coding" definition ever changes. Consider one shared helper (e.g. on the row/Workflows module) the sites delegate to.
- [ ] Generic inert middle stage strands the task (`lib/hive/task_action.rb:337` / `lib/hive/stages/resolver.rb`): an inert, non-entry, non-terminal stage with marker `:none` falls to `generic_ready_to_run` → `hive run`, but `Resolver.resolve` only routes `kind: :agent` and raises `StageError` for `:inert`, so the task can neither run nor advance. Latent (needs an unusual descriptor shape; the test workflow is all-`:agent`). Route an inert non-entry stage to `ready_to_advance`, or document that inert stages must be entry-only.
- [ ] Schema `stage`/`dependency_stage` pattern relax is purely syntactic (`schemas/hive-status.v4.json`): replacing the closed 9-entry enum with `^[0-9]+-[a-z0-9][a-z0-9-]*$` is the deliberate plan choice (R3) so generic dirs validate, but it now also admits unregistered dirs like `"0-x"`/`"99-bogus"`. Consumers degrade by close-defaulting (`Stages::DIRS.index(stage) || -1`) rather than erroring; worth a one-line note that downstream readers must close-default on unknown stage dirs. (`null` stays valid for `dependency_stage` — `pattern` is ignored for non-strings — so no bug there.)
- [ ] Pre-existing: `update_display_name`'s unlocked read-modify-write can drop the selector on a read failure (`lib/hive/task_meta.rb:61-72`): the detached name-generator runs ~60s with no task lock; if `TaskMeta.read` hits corrupt YAML / a permission error it returns `empty()` (workflow nil) and the subsequent full-file rewrite drops `workflow` (and `depends_on`). Pre-existing shape, but U6 widened the dropped set by adding `workflow`. Consider locking or in-place-merging rather than rewriting from a possibly-empty read.

ivankuznetsov:

# PR Review (pr-review-toolkit) — U6: Route hardcoded stage literals through the descriptor

Scope: `git diff origin/main..HEAD`, focused on the U6 commits `787be6e4..HEAD` (U1–U5 reviewed previously). Personas run: code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer, code-simplifier. Union of findings below. The U6-touched test suite is green (432 runs, 0 failures); findings are edge cases / quality issues the tests don't cover.

## High
- [ ] [silent-failure-hunter, pr-test-analyzer] Markerless generic stage → unbounded daemon re-dispatch of `hive run` (`lib/hive/stages/agent.rb:31` + `lib/hive/daemon/policy.rb:70` + `lib/hive/task_action.rb:341`): a generic agent that exits 0 without writing a `WAITING`/`COMPLETE` marker leaves the stage `:none` → `generic_action` returns `ready_to_run` → `ready_to_run` is now an `ADVANCE_ACTION` so `Policy.decide` returns `:dispatch` with no mtime/marker-progress debounce, and exit 0 applies no cooldown → the daemon re-spawns `hive run` every tick until the per-project daily cap drains and then halts all dispatch. Coding is immune (markerless post-run coding stages classify as debounced `needs_input`). Both personas rated Critical; latent until a real non-coding workflow registers (U7/U9) but the wiring is live and untested. Fix: raise `StageError` when a `kind: :agent` stage produces no marker, or gate `ready_to_run` re-dispatch on an mtime/dispatch baseline like edit-resume rows.

## Medium
- [ ] [code-reviewer, silent-failure-hunter, pr-test-analyzer] `migrate.rb backfill_task_ids` drops the `workflow` selector (`lib/hive/commands/migrate.rb:189-196`): it reads `meta` (which now includes `:workflow`) but the `TaskMeta.write` omits `workflow:`, so an id-less task carrying a non-coding selector silently reverts to the project default workflow on `hive migrate` — the exact sibling of the `update_display_name` fix (`task_meta.rb:69-70`) that was applied. Reproduced (`{workflow:"research"}` → `{workflow:nil}`); untested. Fix: add `workflow: meta[:workflow]` to the write and pin it with a round-trip test.
- [ ] [silent-failure-hunter] `status.rb collect_rows` silently drops a task whose workflow fails to resolve/validate (`lib/hive/commands/status.rb:528`, `rescue Hive::InvalidTaskPath; next`): U6 widened this rescue's blast radius — a typo'd `meta.yml workflow:` or a stage-dir/workflow mismatch now raises in `Task.new` and the task vanishes from `hive status` (and from the daemon) with zero diagnostic, while the same diff added *logged* rescues elsewhere (`task_meta.rb:30-39`, `task.rb`). Fix: `warn` on the skip (or emit a diagnostic-bearing placeholder row) so a misconfigured workflow is observable, matching the diff's own warn-then-continue convention.
- [ ] [pr-test-analyzer] `stage_literal_guard_test.rb` is the durable backstop for the whole "route ~20 literals" intent but checks only annotation *presence*, not classification correctness, and its regex `"\d-(...)"` matches single-digit prefixes only (`test/unit/stage_literal_guard_test.rb:4`): a generic literal mis-annotated `# coding-scoped:` passes, and a future two-digit stage literal (`"10-review"`) is silently unguarded. Fix: broaden the regex to `\d+-` (+ a two-digit positive assertion), and optionally pin an expected annotated-literal inventory count so new literals force a deliberate test update.
- [ ] [type-design-analyzer, code-reviewer, code-simplifier] `coding_workflow?` invariant duplicated 6× across 3 incompatible shapes (`daemon/policy.rb:195`, `task_action.rb:311`, and byte-identical row copies in `bot/notification_builders.rb:373`, `bot/notification_dispatcher.rb:113`, `bot/supervisor.rb:1468`, `daemon/stale_agent_healer.rb:351`), with un-named `:coding`/`"coding"` magic at ~8 sites: this is the most load-bearing branch U6 adds (gates every coding-only daemon/bot behavior) yet the "nil/blank/coding ⟹ coding" rule lives nowhere central and the three encodings can drift independently. Fix: extract one predicate (e.g. `Workflows.coding_row?`/`coding_id?`) + a `CODING_ID = :coding` constant; consolidate the four identical row copies (leave policy/task_action shapes delegating).

## Nit
- [ ] [code-simplifier] Dead method `Hive::Workflow#has_stage?` (`lib/hive/workflow.rb:62-64`): zero production call sites (only its own test references it), yet `wiki/modules/workflows.md:22` claims it is "used by coding-scoped consumers to skip absent-stage behavior" — the sweep used `coding_workflow?`/`== :coding` guards instead. Wire it into the inline `resolve_stage_ref(...)`-truthiness sites or delete it and correct the wiki claim.
- [ ] [comment-analyzer] `lib/hive/agent.rb:491` annotation "documentation of legacy constructor coercion" is factually wrong: `Reviewers::SyntheticTask` is current/live (still sets `stage_name: "6-review"` today) and stores the value unchanged (no coercion). Reword to e.g. "documents the SyntheticTask stage_name label, not a routing literal" (and note `synthetic_task.rb` annotates the same fact as `coding-scoped`).
- [ ] [comment-analyzer] `lib/hive/daemon/policy.rb:181-185` — the U6 `# coding-scoped:` annotations were appended to a comment block with stale facts: the `bubble_model.rb:1833 when "3-plan"` cross-reference is dead (now lines 1738/2399) and `Hive::Stages::DIRS` is called a "closed 8-entry enum" when it has 9 entries. Fix the line anchor (or drop it) and "8-entry" → "9-entry".
- [ ] [comment-analyzer] `lib/hive/bot/handlers/recovery_sequence.rb:95` — `# not-a-stage-ref: legacy pre-open-pr retry mapping` is imprecise for the `"4-execute"` key: `4-execute` is a *current* coding dir, and the entry is actually dead (`verb_arriving_at("4-execute")` resolves first). Reword to "fallback retry table for legacy/renamed stage dirs".
- [ ] [type-design-analyzer] `schemas/hive-status.v4.json` `workflow` property uses `"type":"string"` while every other optional Task string (`display_name`, `depends_on`, `blocked_by`, `pr_url`) uses `["string","null"]`: a future producer mirroring the sibling idiom and emitting `"workflow": null` would fail `additionalProperties:false`. Fix: widen to `["string","null"]` or document the deliberate "omit, never null" choice inline.
- [ ] [type-design-analyzer] `schemas/hive-status.v4.json` `stage`/`dependency_stage` pattern (`^[0-9]+-[a-z0-9][a-z0-9-]*$`) advertises a weaker guarantee than the code upholds (the producer only emits dirs that pass `validate_workflow_stage!`): correct relaxation for runtime-registered dirs, but add a one-line `description` noting the real set is the runtime union `Workflows.all_stage_dirs` so SDK consumers don't assume any `N-slug` is valid input.
- [ ] [code-reviewer] `verb_for_action` has no `ready_to_run` mapping (`lib/hive/bot/supervisor.rb:1457`, `notification_builders.rb`): a generic first-run `ready_to_run` row produces no Telegram approve/run button. Not a regression (the `if verb` guard suppresses a malformed callback and the daemon auto-dispatches these rows), but a daemon-disabled project can't run a generic row from Telegram. Map `ready_to_run` to its `run` verb if generic Telegram control is in scope.

ivankuznetsov:

## High
- [ ] `lib/hive/commands/new.rb:99` creates coding-inbox tasks without pinning `workflow: coding`: when a project default_workflow is non-coding, the new `1-inbox` folder resolves against that descriptor and disappears from status as an invalid task.

## Medium
- [ ] `web/app/helpers/application_helper.rb:117` still derives web run actions from coding stage numbers: a generic `ready_to_run` row such as `1-intake` is queued as a coding workflow verb instead of `hive run --stage`, and `lib/hive/web/dispatcher.rb:260` similarly rejects generic reject/backstage moves through `Hive::Stages.parse`.
- [ ] `lib/hive/bot/notification_builders.rb:235` drops workflow from Autofix callback data: generic recovery rows can render an Autofix button, but `CallbackHandlers#autofix` calls `RecoverySequence.build` with workflow nil and falls back to coding retry verbs.
- [ ] `lib/hive/commands/status.rb:698` applies the coding PR metadata gate to every workflow by stage index: non-coding stages at index >= 5 can have `pr.md` read and exposed as `pr_url` instead of taking a guarded coding-only skip.
- [ ] `lib/hive/commands/approve.rb:413` prints descriptor advance verbs as CLI commands: non-JSON generic approvals can suggest nonexistent commands such as `hive gather ...` even though the runnable generic next step is `hive run`.

## Nit

ivankuznetsov:

# Code Review — U6: Route the ~20 hardcoded stage-literal consumers through the descriptor (pass 2)

**Scope:** `git diff origin/main..HEAD` (worktree `u6-route-the-20-hardcoded-260618-33f7`).
**Method:** `/ce-code-review` — parallel persona reviewers (correctness, adversarial, testing, maintainability, cli-readiness) + direct verification of the highest-stakes findings.
**Overall:** The refactor is disciplined and adheres to its guiding rule (constant-swap/descriptor-method where generic, annotated coding-scope where specific, no speculative role taxonomy). Coding paths are largely byte-identical (DIRS/NAMES/VERBS derive from the coding descriptor; healer argv, drop sweep, dest-dir math, policy outcomes preserved), and the generic e2e drives the real product path. The findings below are the exceptions worth addressing — led by one merge/staleness issue that blocks a clean review.

## High
- [ ] Branch is 1 commit behind `origin/main` and is missing PR #542 (`TaskIdBackfiller`): verified `task_id_backfiller.rb` is absent from the merge-base (2a8ab4b0), present on `origin/main`, absent on HEAD — so the diff renders #542's file + `dispatcher.rb`/`task_meta.rb`/`TaskMeta.update_id` wiring as **spurious deletions** that are not U6 work. U6 independently edits the same files #542 touched, so the merge must be resolved carefully to avoid reverting the daemon's task-id self-heal; rebase/merge `origin/main` and re-review before finalizing so U6 is assessed against a clean base.

## Medium
- [ ] Stage-literal guard matches **double-quoted literals only** (`test/unit/stage_literal_guard_test.rb:6`): single-quoted, `%w[...]`, and interpolated stage refs evade it — e.g. the live routing constant `WORKTREE_OWNING_STAGES = %w[4-execute 6-review 8-finalize]` (`lib/hive/stages/base.rb:125`) is unannotated yet passes the "durable net," weakening the central NO-SILENT-CAPS deliverable (R6). Broaden the regex (single-quote + `%w` token forms) and add self-test fixtures for those vectors.
- [ ] Coding notification behavior changed but unpinned (`lib/hive/bot/notification_builders.rb:139`): a coding `waiting` marker at any non-brainstorm/non-plan stage (e.g. `5-open-pr`, `8-finalize`) now yields neutral "Needs input"/"Show details" instead of the prior `brainstorm_waiting` "Answer in chat"/`/answer` affordance — an observable coding change with no characterization test on those stages, contradicting the byte-identical-coding mandate (A7). Decide deliberately and pin it (or restore the old routing).
- [ ] A `default_workflow` config typo silently empties a whole project (`lib/hive/task.rb:130` + `commands/status.rb` `collect_rows` `InvalidTaskPath` skip): one unregistered id makes every field-less task raise→get skipped, so `hive status` shows zero rows and the daemon dispatches nothing, with signal only on stderr — a silent-failure footgun U6 newly introduces (reproduced end-to-end). Validate `default_workflow` loudly at `Config.load`, or fall back to coding with a visible warning.
- [ ] Generic heal branches in `stale_agent_healer` are untested (`lib/hive/daemon/stale_agent_healer.rb`): five new `coding_workflow?(row) &&` gates (`auto_recoverable_error?`, `should_clear_error_marker?` 6-review exclusion, `error_heal_label`, `error_recovery_remediation`, plan-rerun gating) create a divergent non-coding path, but every healer test passes no `workflow:` so only the coding arm runs — the generic side of a safety-critical daemon component is uncovered. Add `make_row(workflow: "research", stage: "6-review"/"8-finalize")` cases.
- [ ] Generic `Stages::Agent` runner ignores the descriptor `status_mode` (`lib/hive/stages/agent.rb:27`): it hardcodes `status_mode: :state_file_marker` while `Stage` carries a `status_mode` field (coding sets `:exit_code_only`), so a generic descriptor declaring a different mode is silently overridden — descriptor-field-vs-behavior drift on the exact seam U6 threads. Consume `stage.status_mode || :state_file_marker`, or drop the field from the generic path.
- [ ] Dropping the Thor `enum:` removes `--help` stage discoverability (`lib/hive/cli.rb`): ~14 `--from`/`--to`/`--stage` options lose Thor's auto-generated "Possible values: 1-inbox, ..." list, so an agent reading `hive help approve|run|status` no longer sees the stage vocabulary. The static enum genuinely can't hold runtime workflows, but embed the default-workflow stage list in each option `desc:` (or add a `hive stages` subcommand) to restore agent discoverability.

## Nit
- [ ] `approve --from` unknown-stage error omits the short-name list its siblings include (`lib/hive/commands/approve.rb:126`): `--to`/`run`/`drop` all append "or short names ..."; `--from` lists only dirs, so an agent may wrongly conclude short names are unsupported.
- [ ] `approve --to/--from <bad-stage>` against a nonexistent slug now reports "no task folder" instead of "unknown stage" (`lib/hive/commands/approve.rb:99`): exit code is preserved (64) but the more useful diagnostic is masked in the combined case (task resolves before the stage is validated).
- [ ] Status `workflow` field is undocumented in the `hive help status` DESC (`lib/hive/commands/status.rb`): the JSON schema describes it well, but agents reading `--help` rather than the schema file won't discover the new nullable row field.
- [ ] `coding_workflow?` name collision (`lib/hive/daemon/policy.rb:226` vs 4 identical private row-wrappers in `stale_agent_healer`/`notification_builders`/`supervisor`/`notification_dispatcher`): `Policy.coding_workflow?(value)` and the bot modules' `coding_workflow?(row)` share a name with incompatible arg shapes; collapse the thunks to `Hive::Workflows.coding_row?` and rename the value predicate (`coding_workflow_id?`).
- [ ] Dead constant `ACTIVE_STAGE_DIRS` in `lib/hive/commands/drop.rb:30`: production switched to `active_stage_dirs`/`from_stages`; only a comment + test still reference it, misleading a reader into thinking drop sweeps a fixed coding list.
- [ ] Unused speculative `capability` field on `Workflow::Stage` (`lib/hive/workflow.rb:105`): no descriptor sets it and no consumer reads it (only nil-asserting tests), which sits against the "no speculative descriptor abstraction" rule — likely U1-owned, drop it or defer until a real producer exists.
- [ ] Stale comment in the `agent_entry_workflow` fixture (`test/support/workflow_helpers.rb:22`): it describes the dropped `entry &&` guard and now contradicts both `task_action.rb:340` and the test's own (correct) comment.
- [ ] `recovery_sequence.retry_verb_for_stage` returns `"run"` for a generic **terminal** stage (`lib/hive/bot/handlers/recovery_sequence.rb:91`): the `9-done` terminal guard is gated behind `coding_id?`, so a generic terminal (inert) stage would be offered a re-run that raises `StageError`. Coding's terminal guard is correctly preserved; the generic side needs a terminal check.
- [ ] Coding-scoped annotation reasons are largely copy-pasted boilerplate (e.g. `lib/hive/bot/title_formatter.rb`, `lib/hive/tui/bubble_model.rb`): the guard only enforces non-empty, so low-information reasons pass and future annotation drift can hide under identical text.
- [ ] `Stages::Agent#prior_artifacts` reads every `*.md` without per-file rescue (`lib/hive/stages/agent.rb:53`): a file deleted/unreadable between glob and `File.read` aborts the generic stage run (TOCTOU); minor, generic-path only.
- [ ] Generic e2e lacks an explicit "no coding-only branch fired" negative assertion (`test/integration/generic_workflow_daemon_e2e_test.rb`): the stated intent calls for it — assert the generic rows never emit a coding verb / hit a bespoke runner across the two hops.
- [ ] Schema `stage`/`dependency_stage` relaxed from a closed 9-value enum to a `^[0-9]+-...$` pattern (`schemas/hive-status.v4.json`): additive and documented, but any downstream consumer that relied on the closed enum to whitelist-validate now accepts arbitrary `N-name` dirs — worth a one-line audit note for consumers.
- [ ] Pre-existing/unrelated: `StopHookInstallerTest#test_stop_hook_requires_stage_dir_env` fails under the review harness because `HIVE_TASK_STAGE_DIR` is exported in the shell env (the test's `Open3.capture3` doesn't scrub it) — not a U6 regression, but the test should clear that env var to stay hermetic.

ivankuznetsov:

# PR Review (pr-review-toolkit) — U6: Route the ~20 hardcoded stage-literal consumers through the descriptor (pass 2)

**Scope:** `git diff origin/main..HEAD` (worktree `u6-route-the-20-hardcoded-260618-33f7`).
**Personas:** code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier (run in parallel; union of findings below).
**Overall:** Disciplined, test-first refactor; coding parity is pinned by golden tests and held up under scrutiny. Findings below are the exceptions — most "generic path" gaps are test-only-impact today (no production non-coding workflow yet) but matter for the "generic runs end-to-end" goal.

## High
- [ ] [orchestrator] Branch is 1 commit behind `origin/main` and missing PR #542 (`TaskIdBackfiller`): the diff renders #542's file + `dispatcher.rb`/`task_meta.rb` wiring as spurious deletions (verified: absent from merge-base, present on `origin/main`, absent on HEAD); rebase/merge `origin/main` before finalize so #542 isn't dropped and U6 is reviewed against a clean base — also recorded in `claude-ce-code-review-02.md` (personas were told to ignore it).

## Medium
- [ ] [code-reviewer] `Hive::Task.new` now triggers a full `Config.load` (file read + YAML parse + `validate!`) for every field-less task (`lib/hive/task.rb` `resolve_workflow`→`project_default_workflow`): unmemoized and called once per task-folder per stage on every `hive status` (i.e. every daemon tick), a real perf regression where `Task.new` previously read no config — memoize `Config.load` per `project_root` or short-circuit to `Registry.default` when no meta selector is present.
- [ ] [silent-failure] A typo'd project `default_workflow:` (unregistered id) makes every field-less task raise `InvalidTaskPath`, which `collect_rows` rescues/`warn`s/skips, so `hive status` shows zero rows for the whole project and the daemon dispatches nothing with only a stderr signal (`lib/hive/task.rb` + `lib/hive/commands/status.rb:531`): surface a synthetic per-project/per-task Error row in the status payload instead of relying on a stderr `warn`.
- [ ] [silent-failure] A generic `kind: :agent` stage whose agent exits 0 without writing a `WAITING`/`COMPLETE` marker stays marker `:none`→reclassified `ready_to_run` forever (`lib/hive/daemon/policy.rb:230-243`): if the agent rewrote the file, mtime advances past `last_dispatched` and `hive run` re-dispatches every debounce window until the daily cap drains; if unchanged, it silently stalls with only a `:skipped` log — record an explicit "needs marker/stalled" diagnostic that surfaces as a status Error row.
- [ ] [code-reviewer] Web "Run stage" button still derives a coding action for generic rows (`web/app/helpers/application_helper.rb` `STAGE_DISPATCH_ACTIONS`, untouched): a generic `1-intake` row renders "Run brainstorm" and POSTs `ready_to_brainstorm`→`hive brainstorm --from 1-intake` (wrong verb), and the new `web/dispatcher.rb` `ready_to_run` path is unreachable from the UI — derive the action from the row's `workflow`/`action`, not the coding index map.
- [ ] [silent-failure] Inline Telegram Autofix/clear-retry buttons don't thread `workflow` (callback_data carries no workflow token), so a generic task in `:error` falls to the coding verb table in `retry_verb_for_stage` and replies "No retry verb for stage …" instead of recovering via `hive run` (`lib/hive/bot/handlers/callback_handlers.rb` + `recovery_sequence.rb`): encode the workflow id into the callback_data (slash `/autofix` and web `recover` already thread it).
- [ ] [pr-test] All five new `coding_workflow?(row) &&` gates in `lib/hive/daemon/stale_agent_healer.rb` (plan-rerun requeue, finalize unpushed-commits, 6-review exclusion, finalize heal label/remediation) are untested on the generic side — the healer test file is untouched and `make_row` never sets `workflow`, so a generic task reusing `6-review`/`8-finalize` dirs takes a divergent, uncovered heal path in a safety-critical component.
- [ ] [pr-test] Stage-literal guard regex matches only double-quoted literals (`test/unit/stage_literal_guard_test.rb:6`): a future `'3-plan'`, `%w[3-plan]`, or interpolated `"#{n}-plan"` slips past un-annotated with the suite green, and these evasion gaps are neither tested nor documented (the live `WORKTREE_OWNING_STAGES = %w[4-execute 6-review 8-finalize]` in `lib/hive/stages/base.rb:125` already proves the hole).
- [ ] [type-design] `Stage#status_mode` advertises a per-stage knob the only generic consumer ignores: `Stages::Agent.run!` hardcodes `status_mode: :state_file_marker` (`lib/hive/stages/agent.rb:27`) while the descriptor carries `status_mode` (coding sets `:exit_code_only`), so a generic stage declaring a different mode is silently overridden — forward `stage.status_mode || :state_file_marker` or drop the field from the generic path.
- [ ] [comment-analyzer] `agent_entry_workflow` fixture comment (`test/support/workflow_helpers.rb:20-27`) still describes the dropped `entry &&` guard and asserts the inert middle stage must NOT advance, directly contradicting the live two-conjunct gate at `task_action.rb:340` and the test that asserts it DOES advance (`task_action_generic_test.rb:222-245`) — rewrite to match the implemented behavior.

## Nit
- [ ] [code-reviewer] `lib/hive/stages/agent.rb` passes `timeout_sec: nil` when a generic `:agent` stage omits `timeout_sec` and cfg has no override; `Hive::Agent#run` does `Time.now + @timeout_sec` with no nil guard (budget is guarded, timeout isn't), crashing with `TypeError` at spawn for a misconfigured generic descriptor — default or validate it.
- [ ] [code-reviewer] Derived `VERBS` uses `stages.fetch(index - 1)` (`lib/hive/workflows.rb`); safe today (coding's first stage has no advance_verb) but `fetch(-1)` would silently pick the terminal stage as the source if a default workflow's stage[0] ever gets a verb — guard the first-stage case.
- [ ] [silent-failure] `Stages::Agent#prior_artifacts` (`lib/hive/stages/agent.rb:52-56`) `File.read`s each glob hit with no per-file rescue, so one unreadable/race-deleted `.md` aborts the whole generic run (TOCTOU) — rescue per-file to a placeholder.
- [ ] [type-design/simplifier/maintainability] `Stage#capability` (`lib/hive/workflow.rb:99`) is a speculative field with zero producers and zero consumers (only nil-asserting tests) — drop it until a real user exists (against the plan's no-speculative-abstraction rule).
- [ ] [type-design/maintainability/simplifier] `coding_workflow?` names three incompatible argument shapes — `Policy.coding_workflow?(value)`, four private `coding_workflow?(row)`, and the arity-0 `task_action.rb` form — a contract hazard since both a row and an id respond to `to_s`/`workflow`; rename the row wrappers (`coding_row?`) and inline the rest.
- [ ] [simplifier] Inner `else` arm of the `needs_input` "waiting" branch (`lib/hive/bot/notification_builders.rb:143-149`) is byte-identical to the case's outer `else` (153-159); collapse the two coding special-cases to guards and let everything else fall through to the single default. (Note: that inner branch also changed coding behavior for non-brainstorm/non-plan `waiting` rows — see `claude-ce-code-review-02.md`.)
- [ ] [simplifier/maintainability] `ACTIVE_STAGE_DIRS` (`lib/hive/commands/drop.rb:30`) is dead in production after the switch to `active_stage_dirs`/`from_stages`; only a comment + one test reference it — remove and repoint the test.
- [ ] [type-design/simplifier] `Workflow#has_stage?` (`lib/hive/workflow.rb:62`) has zero production callers — the sweep used `coding_id?`/`== :coding` instead; drop the dead additive API or give it a real call site.
- [ ] [type-design] `Stage` (`Data.define`, `lib/hive/workflow.rb:89-108`) accepts arbitrary unvalidated `kind`/`status_mode`/`index`, so a malformed descriptor constructs fine and only fails later at dispatch — add a constructor guard against the closed `kind`/`status_mode` sets (reuse `Hive::AgentProfile::STATUS_DETECTION_MODES`).
- [ ] [type-design] The status row exists as a builder Hash (`row[:workflow]`), a JSON string key (`task["workflow"]`), and a `StatusWatcher::Row` Data object (`row.workflow`); correct today but a future consumer can pick the wrong accessor and get a silent `nil` — route new reads through `coding_row?`/`coding_id?` and document the dual representation.
- [ ] [comment-analyzer] `STAGE_LABELS` (`lib/hive/bot/title_formatter.rb:7-15`) carries the same boilerplate reason on ~7 of 9 lines, and `bubble_model.rb:2399-2403` repeats one reason verbatim across three arms; the guard only checks non-emptiness, so replace with a single block annotation each.
- [ ] [comment-analyzer] `recovery_sequence.rb:104` annotates the fallback hash "legacy/renamed stage dirs," but `4-execute` is a current coding dir (and the entry is dead since `verb_arriving_at("4-execute")` already returns "develop") — reword to "defensive fallback reached only when `verb_arriving_at` returns nil."
- [ ] [comment-analyzer] `lib/hive/commands/approve.rb:148` says "unlike its three `stage_for_dest!` siblings," but there is one helper called from three sites — change to "the three `stage_for_dest!` call sites."
- [ ] [comment-analyzer] Stale doc anchors/labels in the wiki: `wiki/state-model.md` still stamps `workflow:` read-only "in U3," and `wiki/modules/workflows.md`/`wiki/stages/agent.md` cite `agent.rb:21`/`:34` for lines that are now 14/27 (claims correct, anchors drifted).
- [ ] [pr-test] `lib/hive/commands/status.rb:531` widened `InvalidTaskPath` rescue (warn arm + silent non-slug skip) has no test — seed a slug folder with an unregisterable workflow and assert the warning reaches stderr while a non-slug dir stays silent.
- [ ] [pr-test] `web/dispatcher.rb` `recover(..., workflow:)` is never exercised with a non-coding workflow (all four tests use coding), and `recovery_sequence` generic terminal-stage retry (`return "run"` before the `9-done` nil guard) is asserted nowhere — add the generic cases.
- [ ] [pr-test] The generic e2e (`test/integration/generic_workflow_daemon_e2e_test.rb`) drives status→`Policy.decide`→CLI manually rather than through an actual `Dispatcher.tick`, and never explicitly asserts "no coding-only branch fired" beyond `ran == %w[intake gather]` — route one hop through `Dispatcher.tick` or add the negative assertion.
- [ ] [code-reviewer] Pre-existing/unrelated: `StopHookInstallerTest#test_stop_hook_requires_stage_dir_env` fails under the review harness because `HIVE_TASK_STAGE_DIR` leaks from the surrounding env into `Open3.capture3` (test file untouched by this diff) — scrub the env var to keep the test hermetic.

ivankuznetsov:

## High
- [ ] lib/hive/daemon/dispatcher.rb:705 keeps post-advance state-file lookup on Hive::Stages::DIRS: after a daemon-dispatched generic approve moves a task into a non-coding dir like 2-gather, the controller cannot refresh the moved task baseline and the next ready_to_run stage can stall instead of auto-dispatching.
- [ ] lib/hive/stages/agent.rb:24 ignores the error result from Hive::Stages::Base.spawn_agent: preflight/version failures return before any marker is written, but the generic runner rereads an empty/stale marker and reports none instead of surfacing an error.
- [ ] schemas/hive-approve.v2.json:43 and schemas/hive-run.v2.json:46 still publish coding-only stage enums: generic approve/run JSON now emits descriptor stages such as gather or 3-report, so valid generic command payloads fail the published schemas.

## Medium

## Nit
- [ ] lib/hive/stages.rb:3 and wiki/modules/stages.md:10 still say Status, Run#next_stage_dir, and Approve delegate through Hive::Stages: this diff routes those generic paths through Hive::Workflows.all_stage_dirs and task.workflow, so the new boundary is misdocumented.
- [ ] wiki/stages/agent.md:13 says Hive::Stages::Agent reads the stage from Registry.default: the implementation resolves via task.workflow.stage_named, which is the core U6 generic-workflow behavior.
- [ ] lib/hive/workflows.rb:7 says Hive::Commands::Approve derives next-action commands from Workflows::VERBS: approve now uses task.workflow.advance_verb_for, so the module comment points future changes at the wrong source.

ivankuznetsov:

# Code Review — U6: Route the ~20 hardcoded stage-literal consumers through the descriptor (pass 3)

**Scope:** `git diff origin/main..HEAD` (worktree `u6-route-the-20-hardcoded-260618-33f7`).
**Method:** `/ce-code-review` — direct verification of the highest-stakes generic-path seams (daemon dispatch/baseline cycle, CLI/JSON contracts, generic agent runner, resolver/policy precedence) plus cross-check of the three unconfirmed High findings raised by codex-03 and the carry-over items from passes 1–2.
**Overall:** The consumer sweep, guard test, characterization net, and coding byte-identical posture are solid, and the claude-02 fixes landed (e.g. `collect_rows` now surfaces a synthetic Error row for an unloadable `default_workflow`, the generic heal arms are covered, the guard regex was broadened). The remaining defects all live on the *generic daemon advance* surface — the exact path U6 claims to prove — which the U6.6 e2e exercises only via direct CLI calls with `last_dispatched: nil`, so the real controller/baseline machinery is never traversed. The three codex-03 High leads are all genuine; #1 in particular means a live daemon cannot auto-run a generic task's second stage.

## High
- [ ] Generic task stalls at its 2nd stage under the live daemon (`lib/hive/daemon/dispatcher.rb:705` + `lib/hive/daemon/policy.rb:251`): `find_post_advance_state_file` iterates `Hive::Stages::DIRS` (coding-only), so after a daemon-dispatched `hive approve` moves a generic task into a non-coding dir (`2-gather`), the post-completion mtime refresh can't locate the new folder and the dispatch baseline recorded at the approve dispatch (`concurrency_controller.rb:209`) stays non-nil; on the next tick the fresh stage's `ready_to_run` hits `decide_run`, whose first-sight `:dispatch` is gated out (`last_dispatched` non-nil) and whose state-file is absent (`state_file_mtime` nil → `:skip`/`:markerless_stalled`) — so `hive run` is never auto-dispatched. Coding is immune because its advance actions dispatch unconditionally; this is generic-only and defeats U6's central "auto-dispatch on advance" goal.
- [ ] Published `schemas/hive-approve.v2.json:44` and `schemas/hive-run.v2.json` still close `from_stage`/`to_stage`/`stage` (and `*_stage_dir`, `stage_index` max 9) to the coding 9-value enums, but U6 now emits descriptor stages on the generic `--json` path — `test/integration/run_approve_test.rb:173` asserts `hive approve --json` yields `from "2-gather" → to "3-report"` (`to_stage:"report"`, `to_stage_dir:"3-report"`), which violate those enums: any consumer validating a generic envelope against the published contract rejects valid output, and the sibling `hive-status.v4.json` enum was deliberately relaxed to a pattern while these were missed.

## Medium
- [ ] U6.6 e2e never traverses the real daemon dispatch/baseline cycle (`test/integration/generic_workflow_daemon_e2e_test.rb:130`): `policy_decision` hardcodes `last_dispatched_state_file_mtime: nil` and the test drives `Hive::CLI.start(["approve"...])`/`["run"...]` directly rather than `Dispatcher#tick`, so the controller's cross-advance baseline carry + `find_post_advance_state_file` refresh (the High #1 stall) is untested — the "real end-to-end daemon test ≥2 stages" deliverable (A6) does not actually exercise the daemon's controller, masking the stall it was meant to prove against.
- [ ] Generic `Stages::Agent` swallows the `spawn_agent` error envelope (`lib/hive/stages/agent.rb:24`): `Base.spawn_agent` returns `{status: :error}` *without writing a marker* on a preflight/version failure (`base.rb:417`), but the runner ignores the return and rereads the (absent) marker → `Markers.current` is `:none` → `run!` reports `{commit:"none", status::none}` instead of an ERROR marker, so a credentials/version failure silently re-classifies as `ready_to_run` and loops/stalls. Unlike `artifacts.rb` (which routes claude through `spawn_claude_with_tmux_marker!`), the generic runner uses plain `spawn_agent` for *all* profiles, so even a claude generic stage has no error surface — consume the returned `{status: :error}` and write an attributed `:error` marker.

## Nit
- [ ] `wiki/stages/agent.md:14` documents the opposite of the implemented behavior: "the runner reads the active stage from `Hive::Workflows::Registry.default`" — but `agent.rb:18` resolves via `task.workflow.stage_named(task.stage_name)`, which *is* the core U6 generic behavior; the wiki would lead a future reader to reintroduce the coding-pinned lookup.
- [ ] `stage_rank` (`lib/hive/daemon/dispatcher.rb:978`) still indexes `Hive::Stages::DIRS`, so any generic stage dir gets rank `-1` and is sorted behind every coding row in `dispatch_priority_order` — a generic task is always deprioritized under slot scarcity; low impact but inconsistent with the union-scan applied to the sibling gates.
- [ ] Module-comment drift: `lib/hive/stages.rb:4` ("Run#next_stage_dir, Approve resolution … read from here") and `lib/hive/workflows.rb:10` ("`Hive::Commands::Approve` uses it to derive the next-action command") name `Stages::DIRS`/`VERBS` as the source, but those generic paths now route via `task.workflow` / `Workflows.all_stage_dirs` — the comments point future edits at the wrong seam.

ivankuznetsov:

# PR Review Toolkit — U6: Route hardcoded stage-literal consumers through the descriptor (pass 3)

**Scope:** `git diff origin/main..HEAD` (worktree `u6-route-the-20-hardcoded-260618-33f7`).
**Method:** `/pr-review-toolkit:review-pr` — union of six persona agents: `[code-reviewer]`, `[silent-failure]`, `[tests]`, `[comments]`, `[type-design]`, `[simplify]`. Overlapping findings are merged with all surfacing personas tagged.

## High
- [ ] [code-reviewer][silent-failure][tests] Daemon post-advance state-file refresh scans only coding `Hive::Stages::DIRS` (`lib/hive/daemon/dispatcher.rb:705` `find_post_advance_state_file`): after a daemon-dispatched `hive approve` moves a generic task into a non-coding dir (`2-gather`), `refresh_post_completion_mtime` can't locate the new folder, so the controller's `[project,slug]` baseline stays stale; the next-stage `ready_to_run` then mis-debounces in `Policy.decide_run` (`policy.rb:251`) — at best a 30s `:wait_for_debounce` delay, at worst `:skip`/`:markerless_stalled` when the new stage's state file is absent — so generic auto-dispatch after advance degrades or stalls. Sibling gates (`task_resolver.rb:112,126`, `status.rb:520`) were migrated to `Workflows.all_stage_dirs`; this call site was missed and is untested (`dispatcher_test.rb` only uses coding `6-review`).
- [ ] [silent-failure][code-reviewer][tests] Generic `Stages::Agent` runner discards `spawn_agent`'s `{status: :error}` return (`lib/hive/stages/agent.rb:24-40`): on a preflight/version failure `Base.spawn_agent` returns the error envelope WITHOUT writing a marker (`base.rb:417-420`), so the reread yields `:none` → `run!` returns `{commit:"none", status::none}`, `hive run` exits 0, no `:error` marker/event, and the row re-classifies as `ready_to_run` forever. Coding's claude path avoids this via `spawn_claude_with_tmux_marker!` (writes an attributed `:error`); the generic runner uses bare `spawn_agent` for all profiles — a NO-SILENT-CAPS violation. Fix: on `status: :error` with no marker, write `Markers.set(..., :error, reason:"agent_preflight_failed")` before the reread.
- [ ] [tests] Generic `hive approve --json` / `hive run --json` envelopes violate the published schemas (`schemas/hive-approve.v2.json:45-66`, `schemas/hive-run.v2.json`): both still close `from_stage`/`to_stage`/`stage` to the coding 9-value enum and `*_stage_dir` to `1-inbox..9-done` (`additionalProperties:false`), but `Approve#success_payload` now emits `to_stage:"report"`/`to_stage_dir:"3-report"`. `run_approve_test.rb:173` asserts the field values but never runs `JSONSchemer…validate` (the repo's near-universal idiom), masking that valid generic output fails the v2 contract bot/daemon consumers validate against. The sibling `hive-status.v4.json` enum was relaxed to a pattern; approve/run were missed.
- [ ] [tests] U6.6 "real end-to-end daemon test" never exercises the daemon (`test/integration/generic_workflow_daemon_e2e_test.rb:21-83`): despite the name it never instantiates `Daemon::Dispatcher`/`#tick`/reap/`ConcurrencyController`; it drives `Hive::CLI.start(["run"/"approve"…])` directly and calls `Policy.decide` with `last_dispatched_state_file_mtime: nil` hardcoded (`:130`), so only the first-sight `:dispatch` branch is ever hit. The headline deliverable — "next-stage auto-dispatch after advance under the daemon" — is unproven, and this is exactly why the High #1 stall is uncaught.

## Medium
- [ ] [type-design] Terminal-inert generic stage emits an unrunnable command (`lib/hive/task_action.rb:340-342`): a `:none` marker on a stage that is both terminal AND `kind: :inert` falls through to `:generic_ready_to_run` (`hive run`), but `Resolver.resolve` raises `StageError` for `kind: :inert` — so a generic single-stage / terminal-inert descriptor produces a runnable command that cannot run (coding is masked because it never takes the generic path).
- [ ] [type-design][simplify] `Hive::Workflow#has_stage?` has no production consumer (`lib/hive/workflow.rb:62`) — every live site uses `resolve_stage_ref`/`all_stage_dirs`; wiki itself notes "no production call sites yet." Speculative against A2 (add a predicate only when a real consumer reads it); wire it into `approve.rb:112` `known_stage_ref?` or drop it.
- [ ] [type-design] `Hive::Workflows::Registry.ids` has no production consumer (`lib/hive/workflows/registry.rb:31`) — only its own unit test reads it; speculative descriptor affordance (`Registry.all` IS consumed).
- [ ] [type-design][simplify] `Policy.coding_workflow_id?` is a zero-logic pass-through to `Workflows.coding_id?` sharing the identical value-arg shape (`lib/hive/daemon/policy.rb:241`); the doc comment justifies it against the row-shaped `coding_row?`, but the net is three spellings of one concept — inline `coding_id?` at the two call sites or drop the delegate.
- [ ] [simplify] Byte-identical `resolve_stage_filter` duplicated across `lib/hive/commands/drop.rb:209-227` and `lib/hive/task_resolver.rb:165-180` (same `Registry.all.filter_map(&:resolve_stage_ref)` branch + two error strings that must stay in lockstep; `project_hint` also duplicated) — extract a shared `Workflows.resolve_stage_ref_across_workflows`.
- [ ] [simplify] Three near-identical `hive <verb> <slug> [--project] [--from|--stage]` builders in `lib/hive/task_action.rb` (`command:238`, `generic_command:348`, `workflow_command:980`) — fold the shared prefix + `@project_count > 1` guard into one private builder.
- [ ] [comments] `lib/hive/stages.rb:4` module header still lists "Run#next_stage_dir, Approve resolution" as consumers that read `Stages::DIRS`, but both now route via `task.workflow` (`run.rb:465`, `approve.rb` resolve_destination) — overstates the consumer set with exactly the two paths U6 rerouted.
- [ ] [comments] `lib/hive/workflows.rb:10` module header claims `Hive::Commands::Approve` derives the next-action command from the `VERBS` map, but Approve now uses `task.workflow.advance_verb_for` (`approve.rb:451-459`) — points future edits at the wrong seam.
- [ ] [code-reviewer][comments] Stale comment in `lib/hive/commands/status.rb:199`: `detect_legacy_stage_dirs` says "collect_rows walks only canonical DIRS," but `collect_rows` now scans `Workflows.all_stage_dirs` (`:520`) — the premise justifying the legacy detector is now inaccurate.
- [ ] [comments] `wiki/modules/task.md:160` documents `TaskMeta.write` without the new `workflow:` kwarg (actual: `write(..., depends_on: nil, workflow: nil)`, `task_meta.rb:42`) and omits that `workflow` is written only when present.
- [ ] [tests] The generic stall scenario (a `ready_to_run` row with a NON-nil dispatch baseline) is untested through a real `tick`: `dispatcher_test.rb` has zero `markerless_stalled`/`ready_to_run` coverage, so the `handle_row → :markerless_stalled` wiring (`dispatcher.rb:817-825`) and the "does not re-spawn" guarantee are unproven end-to-end.
- [ ] [tests] No dispatcher tick test feeds a row with `workflow: "research"` + `action: "ready_to_run"` and asserts a spawn (`dispatcher_test.rb` `row` helper defaults `workflow: nil`) — the generic workflow's dispatch is never observed firing through the real daemon loop.

## Nit
- [ ] [comments] `wiki/stages/agent.md:11` TLDR says the runner "reads the active stage from `Hive::Workflows::Registry.default`," but the implementation resolves via `task.workflow.stage_named` (`agent.rb:18`) — documents the opposite of the central U6 generic behavior; a reader could reintroduce the coding-pinned lookup. (comment-analyzer rated this High.)
- [ ] [type-design] `TaskAction#coding_workflow?` is a fourth spelling of the coding predicate (`task_action.rb:311`) alongside `coding_id?`/`coding_row?`/`coding_workflow_id?` — drift risk; mirror the established pair or document why.
- [ ] [type-design] `Workflow#next_stage_after` collapses "unknown name" and "terminal stage" into the same `nil` (`workflow.rb:34`) — documented and currently unreachable (only validated names reach it), but constructible into a misleading `FinalStageReached` by any future unvalidated caller.
- [ ] [comments] `wiki/log.d/20260619T130000Z-generic-agent-runner.md` repeats the agent.md error ("resolves the current stage from the workflow registry") — historical log fragment, same wrong mental model.
- [ ] [comments] `wiki/modules/task_action.md` "Workflow-aware branch" bullet labels the generic `WAITING` action `needs_input`, but the code key is `generic_needs_input` (`task_action.rb:329`) — prose/table inconsistency (wire key `NEEDS_INPUT` is unaffected).
- [ ] [tests] `test/unit/stage_literal_guard_test.rb:4` hardcodes the `STAGE_NAME` alternation to today's coding stage names instead of deriving from `Stages::DIRS`/the descriptor — a coding stage rename silently shrinks the guard's net with no failing test.
- [ ] [tests] The guard's lib-wide sweep (`stage_literal_guard_test.rb:70-82`) has no positive control proving it can FAIL on a planted unrouted+unannotated literal, so a future `stage_literal_hits` regression (e.g. bad glob root) would let it pass vacuously.
- [ ] [simplify] `generic_command` re-derives `verb`/`stage` already known to its only caller `command` (`task_action.rb:348`) — folds away with the shared builder above.
- [ ] [simplify] `generic_action` `:none` (non-inert) and `else` branches both fetch `:generic_ready_to_run` (`task_action.rb:325-345`) — collapsible unless the `else` is intentionally reserved for forward-compat unknown markers (the comment suggests it is).

ivankuznetsov:
Superseded by #562, which consolidates the full U1–U10 workflow-as-data inversion into one branch off current main (all unit fixes integrated, gates green) for single-feature evaluation. Closing u6's standalone PR.

PR #544, opened 2026-06-20
feat(workflows): wire workflow selection into init and new

ivankuznetsov:
## Summary

The workflow-as-data foundation (U1–U6) could already *resolve* a task's workflow from a descriptor, but nothing let anyone *choose* one: every project and every task was implicitly coding. This PR (U7) wires the user-facing selection onto that foundation and proves the foundation composes end-to-end.

`hive init --workflow <name>` now sets a project's default workflow and `hive new --workflow <name>` pins a workflow onto a single task. A non-coding workflow can be created entirely from the CLI, renders correctly in `hive status`, the TUI, and the web feed, and auto-advances under the daemon through all of its stages, on the real `init → new → daemon → approve` path. The dominant coding path stays byte-identical: no new config field, no `meta.yml` change, identical output.

It is stacked on the open U6 PR (ivankuznetsov/hive#543), not on `main`, so the diff here is only the selection write-paths and the proofs. It introduces no new workflow abstraction (that is U1–U6) and ships no production content workflow (that is U8); the `1-inbox → 2-research → 3-draft → 4-done` fixture used to prove the seam is test-only and never enters the production registry.

### Selection write-paths

Both commands validate the name against the runtime workflow registry and fail fast (USAGE exit 64) with the list of valid names on an unknown value (`Hive::WorkflowSelection.fetch!`). `init` resolves the default in precedence order: explicit `--workflow` flag, then an interactive TTY prompt (numbered list, `coding` pre-selected, skipped when only one workflow is registered), then silent `coding`.

`hive new` resolves the effective workflow as override → project default → coding, seeds the task into that workflow's *first* stage (not a hardcoded `1-inbox`), and writes the workflow's first-stage state file. Whether it pins the resolved workflow into `meta.yml` follows one rule:

| `hive new` invocation | Project default | `meta.yml workflow:` |
|---|---|---|
| no `--workflow` | coding | *(absent — byte-identical to today)* |
| `--workflow <wf>` | any | pinned to `<wf>` |
| no `--workflow` | non-coding | pinned to the project default |

Pinning at creation time is deliberate: it freezes a non-coding task's workflow into `meta.yml` so a later change to the project default cannot retroactively re-resolve an in-flight task to a different stage set and make it vanish from status.

### Notable behavior

- **Coding stays byte-identical.** `default_workflow:` is written into the project `config.yml` only when the resolved default is non-coding; a coding init (or explicit `--workflow coding`) writes no new field. Same for `meta.yml`. Golden tests pin both.
- **`init` is idempotent and warns before rebinding.** Re-running `init` on an initialized project sets/confirms the default and writes it atomically under the commit lock. Before changing an existing default it scans the project's stage dirs and warns when in-flight field-less tasks would be re-resolved by the change (it warns, it does not block). A TTY re-run defaults the prompt to the current default so a bare Enter keeps it.
- **Generic agent error markers are preserved.** The agent stage runner re-reads the current marker after writing a preflight-failure `:error` marker instead of trusting the just-set value, so a marker the agent wrote itself is never clobbered on a generic (non-coding) stage.
- **Agent-native parity.** `--workflow` help enumerates the registered workflow ids; `hive init --json` (fresh and already-initialized) reports the bound `workflow`; `UnknownWorkflow` carries structured data and a USAGE exit code.

## Test plan

The integration tests are the substance of U7, driving the real CLI and daemon rather than shortcutting the advance:

- `test/integration/content_workflow_daemon_e2e_test.rb` — full `init → new → daemon → approve` run of the fixture, asserting all 3 signals: terminal dir reached (`4-done`), per-stage state-file artifacts present, and one advance commit per transition.
- `test/integration/status_test.rb`, `test/unit/tui/snapshot_test.rb`, `test/unit/web/status_feed_test.rb` — status, TUI, and web render a non-coding workflow's stages and a mixed coding+content project side-by-side, each task by its own descriptor.
- `test/integration/init_test.rb`, `test/integration/new_test.rb`, `test/unit/commands/init_test.rb` — flag/prompt/default precedence, fail-fast validation, idempotent re-run + warning, and the byte-identical coding goldens.
- `test/support/workflow_helpers.rb`, `test/unit/content_workflow_fixture_test.rb` — the deterministic, test-only fixture and its scoped registration.

Gates:

- `bundle exec rake coverage` — green: 5890 runs, 23300 assertions, 0 failures, 0 errors, 2 skips, 100.00% line coverage on new code.
- `bundle exec rubocop` — clean, no offenses.

## Review summary

Two automated review passes ran against the full U1–U7 diff (`origin/main..HEAD`); only `:coding` is registered in production, so most generic-path findings are real-but-latent defects that bite when a second/real workflow registers (U8).

- **No High findings** in either pass. Every plan requirement (IU1–IU5) is implemented; the daemon e2e asserts all three A7 signals and coding `init`/`new` stay byte-identical (golden tests).
- **Pass 1** (claude-ce-code-review, codex-ce-code-review, pr-review-toolkit): Medium + Nit findings — re-init idempotency/atomicity, typed exit codes, generic `drop` archive guard, TUI/web generic render literals, agent-native JSON/help gaps, SSOT constant cleanups, and test-coverage gaps — all auto-fixed or resolved from plan/brainstorm context.
- **Pass 2** (claude-ce-code-review, pr-review-toolkit; the codex reviewer hit a provider usage limit and was skipped): re-reviewed the pass-1 fixes plus inert-stage daemon edge, corrupt-config `hive new` hardening, `Workflow` construction-time validation, `VERBS` first-stage hazard, and further test/doc parity — all auto-fixed or resolved from context.
- **Escalations:** none. All findings were auto-fixed or resolved from the plan, brainstorm, prior decisions, or project context, with `fix-success` recorded for both passes.

## Linked task

- Hive task `1459` — **Workflow Selection and Rendering** (`u7-hive-init-new-workflow-260618-f152`).
- Depends on U6 (`u6-route-the-20-hardcoded-260618-33f7`), PR ivankuznetsov/hive#543 — the base branch of this stacked PR.

---


ivankuznetsov:

## High
- [ ] Web task detail does not render non-coding artifacts: `web/app/controllers/tasks_controller.rb:178` enumerates only the coding `ARTIFACT_ORDER`, so `content_fixture` outputs like `research.md`, `draft.md`, and `done.md` never appear in the web view required by U7's generic surface proof.

## Medium
- [ ] Web manual run control queues coding verbs for non-coding stages: `web/app/helpers/application_helper.rb:117` maps stage index `2` to `ready_to_brainstorm`, so a content `2-research` row posts `hive brainstorm ... --from 2-research` instead of the dispatcher's generic `ready_to_run` / `hive run --stage` path.

## Nit

ivankuznetsov:

# Code Review — U7 `hive init`/`hive new` workflow selection + render/daemon proofs

Scope: `git diff origin/main..HEAD` (full U1–U7 "workflow-as-data" stack; only `coding`
is registered in production — most generic-path findings are real defects in shipped
code that bite the moment a second/real workflow registers, i.e. U8). Plan requirements
IU1–IU5 are all implemented; the daemon e2e (`content_workflow_daemon_e2e_test.rb`)
asserts all three A7 signals (terminal dir + per-stage artifacts + commit-per-advance),
and coding `init`/`new` stay byte-identical (golden tests). No plan-required item is
missing, so there are no High findings.

## High

(none)

## Medium
- [ ] Interactive re-`init` silently resets a non-coding `default_workflow` back to `coding` (`lib/hive/commands/init.rb:620,626`): the prompt is hardcoded `Default workflow [coding]:` and an empty answer returns `fetch!(CODING_ID)`, so an operator re-running `hive init` in a TTY and pressing Enter deletes the existing `default_workflow:` line — re-binding every field-less in-flight task. Default the prompt to the project's current default (via `current_default_workflow(ops)`), or treat a bare re-init as `:implicit`. Violates A8's idempotent-re-run intent.
- [ ] `hive new` lets `Workflows::UnknownWorkflow` escape its rescue contract (`lib/hive/commands/new.rb:82-83,150-156`): when a project's `default_workflow` is unregistered (hand-edit, or a later-removed workflow), `resolve_workflow` → `WorkflowSelection.fetch!` raises `UnknownWorkflow`, which is absent from `New#call`'s rescue list and does not name config.yml — blocking all task creation for that project with an opaque error. Add `UnknownWorkflow` to the rescue list and/or re-raise a typed error naming config.yml (mirror `Task#warn_if_unregistered_project_default`).
- [ ] `hive drop` ARCHIVE_STAGE guard is still coding-only while the active-stage set was generalized (`lib/hive/commands/drop.rb:27,210,437`): `ARCHIVE_STAGE = Hive::Stages::DIRS.last` (9-done) but `active_stage_dirs` now uses `Hive::Workflows.all_stage_dirs`, so a completed generic-workflow terminal task (e.g. `3-report`/`4-done`) bypasses `AlreadyArchived` and is hard-deleted. The line-24 "lockstep" comment is now false. Use the union of every workflow's terminal dir (`Registry.all.map { |w| w.stages.last.dir }`, as `init.rb:173` already does).
- [ ] `Workflows::UnknownWorkflow` exits GENERIC(1) instead of USAGE(64) (`lib/hive/workflows/registry.rb:6` vs `lib/hive.rb:366`/`InvalidTaskPath:371`): every other bad-argument error (`InvalidTaskPath`, `WrongStage`, `AlreadyInitialized`) overrides `exit_code` to USAGE(64), so an automated caller cannot classify `--workflow bogus` as a do-not-retry usage error. Add `def exit_code = Hive::ExitCodes::USAGE` to `UnknownWorkflow`.
- [ ] `hive init --workflow X --json` on an already-initialized project writes nothing to stdout (`lib/hive/commands/init.rb:110`): `handle_existing_project` suppresses `write_existing_summary` under `@json` and emits no JSON, so an agent gets exit 0 with empty stdout and no confirmation of the re-bind/default update. Emit a minimal `hive-init` JSON payload (`ok:true`, `workflow:`, `already_initialized:true`).
- [ ] TUI icon map missing the new generic action keys (`lib/hive/tui/views/tasks_pane.rb:44`): `ICONS` enumerates only coding `ready_to_*` actions, so generic-workflow rows (`ready_to_run`/`ready_to_advance`, new in this diff) hit `DEFAULT_ICON` and render with a blank icon — an IU5 render-path coding-literal the proof didn't surface. Add `"ready_to_run" => "▶ "` and `"ready_to_advance" => "▶ "`.
- [ ] Re-`init` config write is non-atomic and never committed (`lib/hive/commands/init.rb:121-127,140-157`): `write_default_workflow!` does a bare `File.write` with no tmp+rename and no commit-lock, so the hive/state worktree is left dirty (unlike every other hive mutation, which commits through `Hive::Lock.with_commit_lock`), and an interrupted write can truncate config.yml — silently masked by `Config.load`'s Psych-rescue → coding fallback. Write via tmp+rename under the commit lock and commit the change.
- [ ] Re-`init --workflow X` rebinding warning understates impact (`lib/hive/commands/init.rb:159-169`): field-less in-flight tasks at a stage the new workflow doesn't define become unloadable Error rows (Policy maps `error → :skip`, so they never advance again), not merely "re-resolved" as the message implies. Escalate the wording for those tasks and consider auto-pinning `workflow: coding` into their meta.yml. (Latent until a second workflow registers — U8.)

## Nit
- [ ] Markerless generic agent has no terminal park (`lib/hive/daemon/dispatcher.rb`/`policy.rb`): an `:agent` stage whose agent exits 0 without writing a marker stays `ready_to_run`, re-dispatching under only the per-project daily cap with no error marker or operator signal. The deterministic U7 fixture always writes `COMPLETE`, so this is latent until a real (non-deterministic) workflow ships (U8); consider escalating to an `:error` marker after N markerless ticks.
- [ ] `generic_workflow_daemon_e2e_test.rb` under-asserts vs its sibling: it checks the terminal dir and which agents ran but not per-stage artifacts or the commit-per-advance trail; `content_workflow_daemon_e2e_test.rb` already covers all three A7 signals, so this is optional hardening (not a plan gap). Mirror the artifact/commit-trail assertions for the descriptor's stage shape.
- [ ] `Stages::Resolver` hardcodes the bare `:coding` symbol (`lib/hive/stages/resolver.rb:53`) instead of `Hive::Workflows::CODING_ID` / `coding_id?` — the only new file in the diff to re-spell the literal the single-source-of-truth constant was created to prevent.
- [ ] Redundant one-line passthroughs add indirection: `WorkflowSelection.coding?` (`lib/hive/workflow_selection.rb:20`) just delegates to `Hive::Workflows.coding_id?`, and `Policy.coding_workflow_id?` (`lib/hive/daemon/policy.rb:241`) wraps the same for a single caller that already requires `hive/workflows`. Inline both.
- [ ] Schema relaxations landed without a `schema_version` bump (`schemas/hive-approve.v2.json`, `hive-run.v2.json`, `hive-status.v4.json`): enum→pattern stage fields and new `ready_to_run`/`ready_to_advance` action enum values, plus a bare stage-name pattern `^[a-z0-9][a-z0-9-]*$` broader than the producer's `^[a-z]…`. In-place pre-1.0 edits are policy (ADR-028) but the carve-out isn't noted in the schema files and the over-broad pattern accepts digit-prefixed names the producer can never emit.
- [ ] `generic_command` dereferences `stage.dir` without the `return … unless stage` guard `generic_action` has (`lib/hive/task_action.rb:347-355`): unreachable for a real `Hive::Task` (construction-time `validate_workflow_stage!`) and guarded in practice by the `:error` action's nil command, but the asymmetry is a latent NoMethodError for test-double callers — add the symmetric guard.
- [ ] `write_default_workflow!` fallback insert index `[…, 1].max` (`lib/hive/commands/init.rb:152`) can place `default_workflow:` mid-document on a config.yml missing the `hive_state_path:` line; only reachable on a hand-mangled config, value is still read correctly, so cosmetic.
- [ ] `--workflow` option help text doesn't enumerate valid workflow names (`lib/hive/cli.rb`): unlike the `STAGE_VOCABULARY` pattern this PR added for `--stage`, agents discover valid workflows only via a failed call. Inject `Workflows::Registry.ids` into both `--workflow` descriptions.

ivankuznetsov:

# PR Review Toolkit — U7 `hive init`/`hive new` workflow selection + render/daemon proofs

Scope: `git diff origin/main..HEAD` (full U1–U7 "workflow-as-data" stack). Personas run:
code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter,
type-design-analyzer, code-simplifier. code-reviewer found no above-threshold defects
("ship it") and confirmed coding `init`/`new` stay byte-identical, fixture leak-scoping
holds, and the daemon e2e asserts all three A7 signals. No plan-required item is missing,
so there are no High findings. Note: only `coding` is registered in production, so most
generic-path findings are latent until a real workflow registers (U8) but are real defects
in code U7 ships/proves.

## High

(none)

## Medium
- [ ] [tests] Workflow-prompt EOF/abort path is untested (`lib/hive/commands/init.rb:623`): branch coverage shows the `raise Aborted, "input closed" if answer.nil?` arm never fires, so the closed-stdin → USAGE-exit behavior is unverified. Add an init test feeding a tty-flagged `StringIO` whose `gets` returns nil; assert exit `USAGE` and zero disk state.
- [ ] [tests] ">5 fieldless tasks" truncation arm untested (`lib/hive/commands/init.rb:168`): the `"... and N more"` line never executes because the re-init warning test seeds only one fieldless task. Seed 6+ fieldless in-flight tasks and assert the overflow line plus that only the first 5 are listed.
- [ ] [tests] `write_default_workflow!` fresh-file and fallback-insert arms untested (`lib/hive/commands/init.rb:142,152`): every test starts from a config that already has `hive_state_path:`, so the new-file (`"---\n"`) arm and the no-`hive_state_path:` insert-position fallback never run. Add unit cases for a nonexistent cfg_path and a config lacking `hive_state_path:`.
- [ ] [comments] `drop.rb` ARCHIVE_STAGE "lockstep" comment is now false (`lib/hive/commands/drop.rb:24-26`): it claims the active-stage list and archive-guard update in lockstep, but `active_stage_dirs` derives from `Hive::Workflows.all_stage_dirs` (the workflow union) while `ARCHIVE_STAGE`/`guard_archived!` recognize only coding's `9-done` — a non-coding terminal stage is treated as active (droppable) and never archive-guarded. Reword to state the guard is coding-only and cross-workflow archive guarding is unfinished.
- [ ] [comments] `wiki/state-model.md` mis-states `hive new` pinning: it says `hive new` pins `workflow: coding` in the seeded meta, but `New#resolve_workflow` pins only when `--workflow` is given OR the project default is non-coding (and then pins the non-coding id); a plain coding capture stays field-less. Correct it to match `wiki/commands/new.md`.
- [ ] [comments] `wiki/modules/task_action.md` "inert **entry** stage" wording is stale: `generic_action` gates on `!terminal && stage.kind == :inert` with no entry-stage condition (the entry-only restriction was intentionally dropped so inert middle stages advance). Reword to "any non-terminal inert stage."
- [ ] [errors] `hive init --workflow X` on an existing project never commits the config change (`lib/hive/commands/init.rb:104-157`): `update_existing_default_workflow!` → `write_default_workflow!` does a bare `File.write` on the tracked `.hive-state/config.yml` with no `hive_commit`, leaving the hive/state worktree permanently dirty; a later reset/clean silently reverts the default. Commit it under `Hive::Lock.with_commit_lock`.
- [ ] [errors] Foreground `hive run` on a generic `:agent` stage that exits 0 without a marker silently succeeds (`lib/hive/agent.rb:395-413`, `lib/hive/stages/agent.rb:38-50`): the clean-exit-no-marker case sets `status: :none` (only nil exit_code stamps `:error`), so `hive run` exits 0 and the task re-classifies to `ready_to_run` with no operator signal (daemon side is braked via `:markerless_stalled`, foreground isn't). Stamp an attributed `:error` marker (`reason: agent_exited_without_marker`) on clean-exit-no-marker.
- [ ] [type-design] `Workflow.new` does no construction-time validation (`lib/hive/workflow.rb:2-9`): empty stage list, duplicate `dir`/`name`, missing terminal stage, and bad indices are all representable and fail lazily at point-of-use (KeyError / nil-collapse / StageError) far from the descriptor. Add a private `validate!` raising `ArgumentError` at construction.
- [ ] [type-design] `Stage#index` can disagree with array position (`lib/hive/workflow.rb:104-106`, `lib/hive/workflows.rb:38-49`): `Stage#dir` is built from `index`, but the `VERBS` derivation computes a verb's source via `stages.fetch(array_position - 1).dir` — the two coincide only because the coding descriptor numbers 1..9 in order, and nothing enforces it. Derive `index` from array position, or assert `index == position+1` in `validate!`.
- [ ] [type-design] `Stage#kind`/`#status_mode` are open symbols, not closed sets (`lib/hive/workflow.rb:89-102`): a typo (`kind: :agnet`) only fails at dispatch via `Resolver` "no runner", and an `:agent` stage missing `advance_verb` runs but can never advance (a representable "un-advanceable agent stage" bug). Validate `kind`/`status_mode` against constants and assert `:agent ⇒ advance_verb`.

## Nit
- [ ] [tests] Same-workflow re-init no-op is uncovered (`lib/hive/commands/init.rb:125`): re-running `hive init --workflow X` when X already equals the current default (skipping the rebind warning) is untested; add a case asserting no warning.
- [ ] [tests] `write_existing_summary` `:implicit` arm is dead/uncovered (`lib/hive/commands/init.rb:114`): implicit re-init raises `AlreadyInitialized` at `init.rb:75` before reaching this method, so the `:implicit` ternary arm never runs — drop the dead branch or add a directed test.
- [ ] [tests] `task-`-prefix slug fallback uncovered (`lib/hive/commands/new.rb:183`): the `unless candidate.match?(/\A[a-z]/)` body never fires in the current suite; add a case where the derived prefix leaves a digit/dash leading char.
- [ ] [errors] `Init#current_default_workflow` silently swallows corrupt config (`lib/hive/commands/init.rb:129-138`): `rescue Psych::Exception, SystemCallError` returns `coding` with no warning (unlike `Task#load_project_default_workflow`), so the rebind warning under-reports and the line-edit proceeds on an unparseable file. `warn` in the rescue.
- [ ] [type-design] `next_stage_after`/`advance_verb_for` collapse "unknown name" and "terminal" into one nil (`lib/hive/workflow.rb:34-46`): a typo'd stage name is misattributed to `FinalStageReached` by `Approve#resolve_destination`. Add a `terminal?(name)` predicate and raise on an unknown name.
- [ ] [type-design] `WorkflowChoice.source` is a free-form symbol on a mutable Struct with a nullable descriptor (`lib/hive/commands/init.rb:48`): the load-bearing `:flag`/`:prompt`/`:implicit` distinction is unconstrained. Use `Data.define` with a `SOURCES` whitelist, or a boolean `implicit:`.
- [ ] [type-design][simplify] `Stages::Resolver` hardcodes the bare `:coding` symbol (`lib/hive/stages/resolver.rb:53`) instead of `Hive::Workflows::CODING_ID` — the single-source-of-truth constant the rest of the diff uses, and not caught by the stage-literal guard test (which only guards stage-dir literals).
- [ ] [type-design] `Registry.default` re-spells `fetch(:coding)` (`lib/hive/workflows/registry.rb:23-25`) instead of `fetch(CODING_ID)` once the constant is reachable.
- [ ] [simplify] `WorkflowSelection.coding?` is a pure passthrough to `Hive::Workflows.coding_id?` (`lib/hive/workflow_selection.rb:20-22`); its 3 callers (init.rb:146, new.rb:155/232) could call `coding_id?` directly, matching the dominant convention. Inline and delete.
- [ ] [simplify] `Policy.coding_workflow_id?` is a wrapper reached through two hops (`lib/hive/daemon/policy.rb:241-243`; `dispatcher.rb:861` imports `Policy` only to reach `Workflows.coding_id?`). Inline `Hive::Workflows.coding_id?` at both call sites and delete the wrapper.
- [ ] [simplify] Triplicated `default_workflow` reader with divergent fallback constants (`lib/hive/commands/init.rb:129`, `lib/hive/commands/new.rb:158`, `lib/hive/task.rb:147`): `New` falls back to `CODING_ID.to_s` while `Task` uses `Config::DEFAULTS["default_workflow"]`. Reference one constant so they can't drift if the default workflow ever changes.
- [ ] [simplify] `all_stage_dirs`/`all_stage_names` rebuild a frozen array on every call (`lib/hive/workflows.rb`); the registry is frozen at load, so memoize (not an eager constant — mind the require cycle). Intent-signaling, negligible perf today.
- [ ] [simplify] `init.rb:613,626` round-trip `fetch!(Hive::Workflows::CODING_ID)` where `Hive::Workflows::Registry.default` says the same thing more directly (`fetch!` only earns its keep on operator input that may be blank/unknown).
- [ ] [simplify] `Stages.next_dir` (`lib/hive/stages.rb:36`) appears to have no `lib/`/`bin/`/`web/` caller after the refactor (only `stages_test.rb`); likely prunable coding-only helper (low confidence — still a tested public API).
- [ ] [comments] `wiki/index.md` `Updated: 2026-06-19` is stale (newest branch changes are 2026-06-20) and `Page count: 82` is off by one (83 entries; pre-existing upstream skew, not introduced here).
- [ ] [comments] `wiki/log.d/20260620165502-u7-workflow-selection.md` uses a non-ISO `YYYYMMDDHHMMSS` timestamp while every other fragment on the branch uses the `…T…Z` form; cosmetic convention break.

ivankuznetsov:

# Code Review (pass 02) — U7 `hive init`/`hive new` workflow selection + render/daemon proofs

Scope: `git diff origin/main..HEAD` (clean tree; committed files = review state). Full U1–U7
"workflow-as-data" stack, 96 commits. Only `:coding` is registered in production — generic-path
findings are real shipped defects that bite when a second/real workflow registers (U8), not the
"only coding registered" anti-finding. Pass-1 findings were all auto-fixed; this pass re-reviews
the current state and the pass-1 fixes for completeness/asymmetry.

Reviewer team: correctness, adversarial, testing, maintainability, cli-readiness, reliability,
project-standards, api-contract. All plan requirements IU1–IU5 are implemented; the content daemon
e2e (`content_workflow_daemon_e2e_test.rb`) drives the REAL init→new→daemon→approve path and asserts
all three A7 signals (terminal dir + per-stage artifacts + one-commit-per-advance), and coding
init/new stay byte-identical (golden tests). No plan-required item is missing or undeferred, so
there are no High findings.

## High

(none)

## Medium
- [ ] `hive new --workflow bogus` exits 1 (GENERIC) instead of 64 (USAGE) (`lib/hive/commands/new.rb:88-93`): `New#call`'s rescue list now catches `Hive::Workflows::UnknownWorkflow` but does a hardcoded `exit 1`, discarding the `USAGE` exit code pass-1 deliberately added to that class — so an agent gets the same code for "typo'd workflow, fix the flag" as for a transient failure, while `hive init --workflow bogus` correctly exits 64. Change to `exit(e.respond_to?(:exit_code) ? e.exit_code : 1)`. 
- [ ] Inert non-terminal generic stage stalls and drains the project dispatch cap (`lib/hive/task_action.rb:339`): `generic_action` classifies a markerless (`:none`) inert middle stage as `ready_to_advance`, so the daemon dispatches `hive approve <slug> --from <inert>` — but `Approve#validate_move!` (`lib/hive/commands/approve.rb:169-178`) rejects it because `:none ∉ VALID_TERMINAL_MARKERS` and there is no `--force`, so the move raises `WrongStage` every tick and each failed dispatch counts toward the per-project daily cap. Coding is immune (its inert stages are entry/terminal only) and the U7 content fixture has no inert middle stage, so it is latent until U8 — but the `agent_entry_workflow` unit test asserts only the classification (`ready_to_advance`), giving false confidence that the deliberately-added inert-advance path works end-to-end. Fix: write a `:complete` marker on arrival into an inert stage, or have the generic `ready_to_advance` command append `--force` for inert stages. 
- [ ] Fresh `hive init --workflow X --json` omits the bound `workflow` from its payload (`lib/hive/commands/init.rb:568-592`, `schemas/hive-init.v1.json:217-330`): `SuccessPayload` has `additionalProperties:false` and no `workflow` field, while the re-init `AlreadyInitializedPayload` *requires* `workflow` — so an agent initializing a fresh project with `--workflow content --json` gets exit 0 but cannot confirm which workflow was bound, and must read config.yml out of band. Add `"workflow"` to `success_payload` and to the `SuccessPayload` schema (mirroring the AlreadyInitialized arm) for parity. 
- [ ] `Hive::Workflows::UnknownWorkflow` exit code is not pinned in the exit-code contract test (`test/unit/exit_codes_test.rb:19-42`): `test_error_subclasses_map_to_their_contract_code` pins every `Hive::Error` subclass that overrides `exit_code`, but the newly-added `UnknownWorkflow` (`exit_code → USAGE`, `lib/hive/workflows/registry.rb:6-13`) is absent — the file's own header calls these values "a breaking change for the CLI contract," so a future refactor to GENERIC would silently slip through. Add `assert_equal Hive::ExitCodes::USAGE, Hive::Workflows::UnknownWorkflow.new("x").exit_code`. 

## Nit
- [ ] Terminal-stage-dirs expression duplicated verbatim (`lib/hive/commands/drop.rb:443`, `lib/hive/commands/init.rb:252`): both compute `Hive::Workflows::Registry.all.map { |w| w.stages.last.dir }.uniq` with no shared helper, despite the parallel `Hive::Workflows.all_stage_dirs`/`all_stage_names` SSOT pattern. Since one site is drop's hard-delete archive guard (pass-1's High fix), drift on a stage renumber re-opens that bug. Add `Hive::Workflows.all_terminal_stage_dirs` and call it from both. 
- [ ] Generic recovery offers an un-runnable `hive run --stage <stage>` for non-`:agent` non-terminal stages (`lib/hive/bot/handlers/recovery_sequence.rb:99-105`): `retry_verb_for_stage` only guards the generic terminal stage, but `Stages::Resolver.resolve` raises `StageError` for any `kind != :agent` (`lib/hive/stages/resolver.rb:65`), so an inert/marker middle stage in error state queues a command that always fails. Return nil (no retry verb) when the descriptor stage's kind is not `:agent`. Latent (needs a non-agent middle stage). 
- [ ] `VERBS` construction wraps a first-stage `advance_verb` to the last stage (`lib/hive/workflows.rb:38-49`): `stages.fetch(index - 1)` at `index == 0` is `stages.fetch(-1)` → the terminal stage, silently producing a wrong `source`. Guarded today only because coding's `inbox` has no `advance_verb`. Use `each_cons(2)` or raise on a first-stage verb so a future descriptor fails loudly instead of silently. 
- [ ] Repairing a corrupt config.yml via `hive init --workflow coding` is a silent no-op (`lib/hive/commands/init.rb:155-159`): when `current_default_workflow` hit its Psych/SystemCall rescue and returned the coding default, `update_existing_default_workflow!` short-circuits on `current == next_value` *before* rewriting, so the operator's explicit repair gesture leaves the unparseable file untouched (Config.load keeps masking it as coding). Force the rewrite when the read came from the corrupt-config rescue. 
- [ ] Stale comment on `Workflow#next_stage_after` (`lib/hive/workflow.rb:29-33`): claims "The only current caller (Approve#resolve_destination)…" but U7 added three more (`approve.rb:430` json_next_action, `run.rb`, `new.rb:149`). Update the comment so the unreachable-unknown-name claim isn't read as still single-caller. 
- [ ] `hive init` TTY prompt fires even when only one workflow is registered (`lib/hive/commands/init.rb:688-695`): with `coding` as the sole registry entry the numbered prompt offers no real choice yet still blocks on stdin, which can hang a PTY-allocating wrapper (tmux/`script`/expect). Skip the prompt (treat as `:implicit`) when `Registry.ids.size == 1`. 
- [ ] `known_stage_ref?`/`known_stage_list` re-implement `resolve_stage_ref_across_workflows` (`lib/hive/commands/approve.rb:112-118`): the bespoke predicate also silently accepts an ambiguous short name that `resolve_stage_ref_across_workflows` would reject, and duplicates the valid-stage-list string. Route the early `validate_stage_refs!` check through the shared resolver (rescue to re-prefix the message). 
- [ ] `update_existing_default_workflow!` leaves the hive/state worktree dirty if `hive_commit` raises after `atomic_write` (`lib/hive/commands/init.rb:167-175`): the rename has already landed, but there is no rollback of the file on commit failure — a later `git reset`/`hive state` could discard the change. Idempotent re-run recovers, but a snapshot/restore of the pre-write content in a rescue would make the mutation atomic. 
- [ ] `action_for(:none)` returns the string `"none"` as a commit action (`lib/hive/stages/agent.rb:91-98`): safe only because `hive_commit`'s empty-diff guard no-ops on the markerless path; an explicit `when :none then nil` (and a nil-guard at the commit call site) removes the implicit dependence on that guard. 
- [ ] `hive new` has no `--json` mode at all (`lib/hive/cli.rb:332-344`, pre-existing): the entry point of every agent task pipeline emits only prose (`hive: captured …` + a raw `mv … && hive run` hint), so an agent must regex stdout to recover the task folder/id and now the resolved workflow. Pre-existing, but `--workflow` makes a structured `hive-new` envelope (with `workflow` + `next_action`) materially more useful. 
- [ ] content daemon e2e could assert the daemon drove the advance explicitly (`test/integration/content_workflow_daemon_e2e_test.rb:99-144`): its `InlineSupervisor` doesn't expose `spawned`, so unlike the generic sibling it never asserts `hive run`/`hive approve` went through `supervisor.spawn`. The `ran` array + approve-commit assertions already transitively prove this, so it's defense-in-depth parity hardening, not a coverage gap. 
- [ ] `project_default_workflow` cache is keyed by mtime alone (`lib/hive/task.rb:140-152`): on coarse-mtime filesystems (FAT, some NFS) a same-second config rewrite can collide, so a long-lived reader (daemon/TUI) serves the stale default until restart. `hive new` is immune (reads fresh). Key on `(mtime, size)` to close the window; negligible on ns-mtime ext4/xfs.

ivankuznetsov:

# PR Review Toolkit (pass 02) — U7 `hive init`/`hive new` workflow selection + render/daemon proofs

Scope: `git diff origin/main..HEAD` (clean tree = committed files), full U1–U7 "workflow-as-data"
stack, 96 commits. Multi-persona review: code-reviewer, pr-test-analyzer, silent-failure-hunter,
type-design-analyzer, comment-analyzer, code-simplifier. Findings are the union of all personas,
prefixed by persona. Only `:coding` ships in production, so generic-path defects are real-but-latent
(they bite when a real second workflow registers in U8), not the "only coding registered"
anti-finding. Personas verified: coding init/new stay byte-identical, the content/generic daemon
e2e tests drive the REAL init→new→daemon→approve path (no shortcut) and assert all three A7 signals,
golden tests are real, rubocop clean, suite green. No plan-required item is missing/undeferred →
no High findings.

## High

(none)

## Medium
- [ ] [code] Inert non-terminal generic stage dead-loops the daemon (`lib/hive/task_action.rb:339`, `lib/hive/commands/approve.rb:169-178`): a markerless (`:none`) inert MIDDLE stage is classified `ready_to_advance` → suggested `hive approve <slug> --from <stage>` (no `--force`), but `Approve#validate_move!` rejects a forward move on `:none` (not in `VALID_TERMINAL_MARKERS`) with `WrongStage`, so the row re-classifies and the daemon re-dispatches a command that always fails until the per-project daily cap drains. Latent (no shipped non-coding workflow has an inert middle stage; the `agent_entry_workflow` test asserts only the policy decision, not that `hive approve` succeeds). Fix: stamp `:complete` on arrival into an inert stage, or append `--force` for the inert advance. 
- [ ] [code] Generic `:marker`-kind stage is classified `generic_ready_to_run` but has no runner (`lib/hive/task_action.rb:334`, `lib/hive/stages/resolver.rb:59-65`): `generic_action`'s `else` branch routes a `:marker` stage to `hive run`, yet `Resolver.resolve` only handles `kind == :agent` and raises `StageError "no runner for stage"` for `:marker` — so the run exits non-zero instead of advancing. The comment claims "any non-inert kind (`:agent`, `:marker`…) must run," but `:marker` cannot run. Latent (no `:marker` stage ships). 
- [ ] [silent-failure] `hive new` crashes with a raw backtrace on a corrupt project `config.yml` (`lib/hive/commands/new.rb:186-189`): `project_default_workflow` calls `Hive::Config.load` (unconditionally, even with `--workflow`), which `YAML.safe_load`s the project config with no Psych rescue on that path — a syntactically-corrupt config raises `Psych::SyntaxError`, which is neither `Hive::Error` nor `Thor::Error`, so it escapes both `New#call`'s rescue list and `bin/hive`'s rescues (exit 1, ugly trace). Every sibling (`Task#load_project_default_workflow`, `Init#current_default_workflow`) degrades a corrupt config to a warning + coding fallback; `hive new` is the lone path that crashes. Wrap the read to rescue `Psych::Exception`/`ConfigError`/`SystemCallError`/`IOError` and raise a typed error naming config.yml. 
- [ ] [type-design] `Hive::Workflow.new` enforces no structural invariants (`lib/hive/workflow.rb:2-9`): empty `stages`, duplicate/gapped/non-monotonic `index` values, and duplicate `name`/`dir` are all constructible — every "ordered, unique pipeline" assumption lives in consumers (`next_stage_after`, VERBS, `validate_workflow_stage!`), not the type. A typo'd descriptor loads fine then mis-routes at runtime far from the cause. (Consumed U1 descriptor layer, but present in the diff.) Add a post-freeze validation pass (non-empty, `indices == (1..n)`, unique names/dirs, `kind` ∈ known set, `stages.first.advance_verb.nil?`). 
- [ ] [type-design/simplify] `VERBS` derivation can bind a verb's `source` to the wrong stage (`lib/hive/workflows.rb:38-48`): `stages.fetch(index - 1).dir` at `index 0` is `fetch(-1)` → the terminal stage, silently producing a wrong `source` if a first stage ever carries an `advance_verb` (guarded today only by coding's verb-less inbox). Use `stages.each_cons(2)` to express source→target directly and drop the negative-index hazard. 
- [ ] [test] Inert middle-stage advance is never exercised end-to-end through a real `hive approve` (`test/integration/run_approve_test.rb`, only classification at `test/unit/task_action_generic_test.rb:228` + policy layer): the PR's stated reason for dropping the `entry &&` conjunct is exactly this stranding case, but no test seeds a task at an inert middle stage (`agent_entry_workflow`'s `2-hold`) and runs `Approve.new(folder, from: "2-hold").call` to confirm it actually moves `2-hold → 3-ship`. A regression in `resolve_destination`/`validate_move!` for an inert middle would ship green (this is the test-side of the inert-stall finding above). 
- [ ] [test] `project_default_workflow` mtime-cache invalidation branch is untested (`lib/hive/task.rb:140-152`): no test resolves a field-less task, rewrites `config.yml` with a new mtime, then re-resolves in the same process to prove the cache picks up the new default — the exact "long-lived reader (TUI/daemon) stays correct mid-run" invariant the cache exists for. A stale-cache regression would silently serve the old default to every field-less task with the suite green. 
- [ ] [simplify] Terminal-stage-dirs expression duplicated verbatim (`lib/hive/commands/drop.rb:443`, `lib/hive/commands/init.rb:252`): both inline `Hive::Workflows::Registry.all.map { |w| w.stages.last.dir }.uniq`; one is drop's hard-delete archive guard, so drift on a stage renumber re-opens a delete bug. Add a memoized `Hive::Workflows.all_terminal_stage_dirs` (sibling to `all_stage_dirs`/`all_stage_names`) and call it from both (the `stages.last` pattern also recurs at task_action.rb:322 / recovery_sequence.rb). 

## Nit
- [ ] [code] Fresh `hive init --workflow X --json` omits the bound `workflow` from its payload (`lib/hive/commands/init.rb:568`, `schemas/hive-init.v1.json`): `SuccessPayload` (`additionalProperties:false`) has no `workflow` field, while the re-init `AlreadyInitializedPayload` requires it — an agent initializing a fresh non-coding project can't confirm the binding from JSON. Add `workflow` to producer + SuccessPayload schema for parity. 
- [ ] [silent-failure] `ReviewHandoff#write_meta` mints a coding `6-review` task without pinning `workflow: coding` (`lib/hive/patrol/review_handoff.rb:36-43`): the field defaults nil → the task inherits the project default, so on a non-coding-default project this hard-coded coding handoff produces a `6-review/<slug>` that fails `validate_workflow_stage!` and surfaces only as a synthetic Error row. Latent (patrol is coding-only, no non-coding workflow ships) but exactly the silent-drop class the diff hardened elsewhere (TaskMeta.update_id/update_display_name now preserve `workflow`). Pin `workflow: Hive::Workflows::CODING_ID.to_s`. 
- [ ] [type-design] `Workflow::Stage#kind` is an untyped symbol enum (`lib/hive/workflow.rb:89-99`): `{:agent,:inert,:marker,nil}` is behaviorally significant (selects runner + auto-advance gate) but `Stage.new(kind: :agnet)` constructs cleanly; a typo strands the task at run time (`StageError`) rather than failing at descriptor definition. Validate `kind` in the construction pass. 
- [ ] [type-design] Soft-lookup nil contract folds two outcomes (`lib/hive/workflow.rb:34-46`): `next_stage_after` and `advance_verb_for` return `nil` for BOTH "unknown name" and "terminal/no-inbound-verb"; safe today only because every caller passes a `validate_workflow_stage!`-validated name (an invariant the type can't see). A future caller that forgets gets a silently wrong answer. Documented as deliberate, but worth a hard-lookup variant for unvalidated call sites. 
- [ ] [type-design] `WorkflowChoice` is a mutable, unvalidated `Struct` (`lib/hive/commands/init.rb:49`): its `source` is an effective enum (`:flag|:prompt|:implicit`) but unconstrained, and it diverges from the new `Data.define` value-object convention (`Workflow`/`Stage`/`AdvanceVerb`). Convert to `Data.define` for free immutability. 
- [ ] [type-design] `Workflows::UnknownWorkflow` carries no structured data (`lib/hive/workflows/registry.rb:6-13`): the bad id and valid-names list exist only in the interpolated message, so an agent/TUI must regex-parse it — unlike the sibling `UnregisteredProjectWorkflow` which exposes `attr_reader :value`. It's the most agent-facing error here (USAGE, `--workflow`). Expose `value`/valid-set as fields. 
- [ ] [comment] `next_stage_after` doc claims "The only current caller (Approve#resolve_destination)" (`lib/hive/workflow.rb:29-33`): there are now four callers (approve.rb:140 + approve.rb:430, run.rb, new.rb:149). The unknown-name-unreachable invariant still holds, but the enumeration is false and the nil-fold framing ("attributes it to FinalStageReached") is incomplete (run/new treat terminal-nil as a `hive run` fallback). 
- [ ] [comment] Stale line reference (`lib/hive/commands/status.rb:737`): "collect_rows (:422) is the sole caller" — `collect_rows` is now ~520 and its call site ~526 after this diff reworked it. Drop the `(:422)`. 
- [ ] [comment] Coding-only assumption in rollback comment (`lib/hive/commands/new.rb:132-135`): says an orphan "would surface as a broken `1-inbox/` entry," but `task_dir` is now built from `entry_stage.dir` (not `1-inbox` for a non-coding workflow). The rollback is correct; the comment drifted. 
- [ ] [comment] Rot-prone cross-file citation (`lib/hive/daemon/policy.rb:223-224`): `plan_approval?` cites `bubble_model.rb:1738 when "3-plan"` (accurate now, but there are two such sites and it was already re-pointed once this diff). Prefer naming the method over the line number. 
- [ ] [test] `json_next_action` degrade-and-warn branch untested (`lib/hive/commands/approve.rb:438`): the path where a post-move task fails to re-parse (`InvalidTaskPath`, e.g. unregisterable `meta.yml workflow:`) and degrades to `hive run <folder>` + warning has no coverage; a refactor could drop the warn (silent degradation) undetected. 
- [ ] [test] `write_default_workflow!` hand-mangled-config paths untested (`lib/hive/commands/init.rb:171`): neither the coding-rebind line-DELETION branch (revert a project to coding) nor the `hive_state_path:`-absent / marker-less insert-index fallbacks are exercised; the coding-rebind deletion is a plausible real operation. 
- [ ] [test] `current_default_workflow` corrupt-config warn arm untested (`lib/hive/commands/init.rb:131`): the re-init Psych/SystemCallError rescue (warn + assume coding) has no coverage, a parity gap vs the well-tested `Task#load_project_default_workflow` rescue. 
- [ ] [simplify] `stages.rb` re-derives short names by string-splitting `DIRS` (`lib/hive/stages.rb:13-14`): `NAMES`/`SHORT_TO_FULL` do `d.split("-",2).last` even though `DIRS` is built from the descriptor, which exposes structured `name`/`dir` (and `Workflow#stage_names`). Read the canonical short name from the descriptor instead of round-tripping the formatted dir. 
- [ ] [simplify] Union-scope valid-stages hint duplicated (`lib/hive/commands/approve.rb:117`, `lib/hive/workflows.rb:165`): the "`<dirs> or short names <names>`" tail is byte-identical in both; extract `Hive::Workflows.stage_ref_hint` so they can't drift. 
- [ ] [simplify] `Hive::Workflows.verb_advancing_from` is a dead exported method (`lib/hive/workflows.rb:68-70`): no callers in `lib/`, only its own unit tests (pre-existing, PR #6). Drop it (and its tests) or document it as intentional public surface. 
- [ ] [simplify] Two provably-equal stage-dir representations in `task_action.rb` (`lib/hive/task_action.rb:1181-1186`): coding path uses `stage_dir` (`"#{index}-#{name}"`), generic path uses `workflow_stage.dir`; `validate_workflow_stage!` guarantees they're equal. Add a one-line note so a future reader doesn't "consolidate" them and add a needless nil-guard to the hot coding path.

ivankuznetsov:
Superseded by #562, which consolidates the full U1–U10 workflow-as-data inversion into one branch off current main (all unit fixes integrated, gates green) for single-feature evaluation. Closing u7's standalone PR.

PR #546, opened 2026-06-21
feat(scope): add opt-in per-stage permission scopes

ivankuznetsov:
## Summary

Hive-launched agents previously ran every stage with full `bypassPermissions` ("yolo") access, with no way to narrow what a stage's agent could touch. This PR adds opt-in, per-stage permission scoping: a stage (or the whole project) can declare a `permissions:` preset that Hive translates into the spawned Claude agent's tool allowlist, denylist, permission mode, and add-dirs.

The default is unchanged. Absent any `permissions:` block, every stage spawns with byte-identical arguments to today, per stage and per mode. Scoping is strictly additive and opt-in.

> Permission scoping is tool/MCP-level, enforced by the agent's allowed tool set. It is **not an OS sandbox** — the agent runs as the same OS user. `read-only` limits accidental over-reach; it does not contain a determined or mis-prompted agent. For real isolation, run hive under a sandboxed user or container (e.g. hivebox).

## Presets

| Preset | Effect |
|---|---|
| `yolo` | Current default. Bypass permissions, no tool allow/deny lists from Hive. |
| `read-only` | Allows only `Read`, `LS`, `Grep`, `Glob`. Denies `Write`, `Edit`, `MultiEdit`, `NotebookEdit`, `Bash`. No network or MCP tools. |
| `scoped` | Custom: provide `tools:` (authoritative) and optional `dirs:`, or `bash: true/false` as sugar on the read-only base set. |

Config accepts a scalar shorthand (`permissions: read-only`) or a map (`permissions: {preset: scoped, tools: [...], dirs: [...]}`), both at the top level (project default) and on any stage block including `review.*` substages and reviewers. A per-stage value fully replaces the project default — no field merge.

## Design decisions

- **Resolver keyed by spec, not location.** `Hive::PermissionScope` validates and resolves a spec independent of where it came from, so when the planned U1 generic workflow descriptor lands it carries the same field into the same resolver with no rework. Today the field lives on the existing `config.yml` per-stage blocks plus a new top-level default.
- **Fail closed.** Unknown presets, unknown keys, malformed maps, a present-but-blank value, and `bash:` combined with `tools:` are hard errors at config load. Non-yolo on a non-Claude runner (Codex/Pi) is a hard error at spawn time, where the resolved profile is known — Hive never silently degrades scoping to a no-op, and the gate fires for `codex_review` reviewers regardless of the declared agent profile.
- **Enforcement built into both spawn paths.** The headless agent path emitted no tool flags and dropped the `allowed_tools` kwarg; the tmux path emitted only `--allowedTools`. This PR wires `--allowedTools`/`--disallowedTools` through both paths (claude-guarded), so a scope is actually reachable. Flags are emitted only when a non-yolo scope supplies them.
- **`scoped` grants what it lists.** Explicit `tools:` are authoritative — the deny list subtracts the granted set, so an allowed tool is never re-denied (`allowed ∩ disallowed == ∅`). The `Scope` value object is built through a single validating, deep-frozen factory so the disjointness and immutability invariants can't drift.
- **`dirs:` extends, never replaces.** Relative paths resolve from the task folder, absolute paths are honored as written, and the task folder is always retained.

## The load-bearing proof

The feature's correctness rests on disallowed tools being unreachable headless without a blocking prompt. `test/smoke/permission_scope_headless_smoke_test.rb` spawns the real `claude` binary and proves it against the live CLI:

- **read-only**: an agent told to `Write` a file auto-denies and the run completes within a bounded timeout (no hang), with the file never created.
- **yolo control**: the same prompt writes the file, confirming the mechanism is the tool list, not a broken spawn.

This pins `permission_mode: "default"` as the non-prompting mode. The smoke test is excluded from `rake test` and runs via `rake smoke`; it was run live against `/usr/bin/claude` and passed.

## Test plan

`rake test` (5960 runs, 0 failures, 2 skips) and `rubocop` (662 files) pass. New unit coverage spans the resolver (`test/unit/permission_scope_test.rb`), config validation (`test/unit/config_test.rb`), both spawn paths (`test/unit/agent_test.rb`, `test/unit/claude_launcher_test.rb`, `test/unit/interactive_wrapper_script_test.rb`), and stage wiring (`test/unit/spawn_agent_test.rb`). The absent-permissions path is pinned by golden spawn-arg assertions (per stage and per mode) so building headless flag support does not silently start enforcing tools on the default path.

User-facing reference: `docs/permissions.md` (with the OS-sandbox caveat), linked from `docs/concepts.md`; commented examples in `templates/project_config.yml.erb` and `config.example.yml`.

Reviewer note: this PR is stacked on `u8-ship-the-content-research-260618-bb97` and targets that branch, so the diff shows only the 23 U10 commits (resolver/config/wiring + three rounds of review fixes).

## Review summary

Three rounds of multi-agent code review ran (a Claude fleet, a Codex fleet, and the pr-review-toolkit personas in each round). Highlights:

- **Pass 1** found and auto-fixed the load-bearing bug: `scoped` re-denied the tools it explicitly granted (`Write`/`Edit`/`Bash` landed in both allow and deny lists, and Claude's deny rules win). Fixed by subtracting the granted set from the deny list, guarded by an `allowed ∩ disallowed == ∅` test plus a scoped-grant smoke proof. Also attributed A8 runner-gate `ConfigError`s to a stage `:error` marker instead of an uncaught exit.
- **Pass 2** closed fail-open gaps at the edges: `codex_review` reviewers and `patrol.review.reviewers` now run the A8 runner gate and load-time validation; `patrol.review.permissions` is rejected as an unsupported location; the `Scope` struct got a validating, deep-frozen factory.
- **Pass 3** hardened fail-closed validation (null-byte / comma / whitespace guards on `tools:` and `dirs:`, `~user` expand errors re-raised as `ConfigError`), de-binarized a test file that embedded a raw NUL, and added shared-reviewer-grouping, A8-propagation, and brainstorm golden coverage.

Two items were escalated for human scope decisions and resolved:
1. **Present-but-blank `permissions:`** → fail closed (rejected at load as malformed) rather than silently resolving to yolo.
2. **Write-capable spawns outside the stage pipeline** (rebase conflict-resolution, babysitter PR-fixer) → left unscoped by design, with the per-stage boundary documented: U10's `permissions:` applies to pipeline stage spawns and review reviewers, not as a global floor. A restrictive project default would break those write-capable agents.

All confirmed findings are auto-fixed or resolved in the current tree; the remaining open items are the two escalations above, settled as above.

## Linked task

Hive task `u10-per-stage-permission-scopes` (id 1467, "Per Stage Permission Scopes"), depends on `u8-ship-the-content-research-260618-bb97`. Plan and brainstorm artefacts live in `hive-private` (`docs/plans/2026-06-18-001-feat-hive-workflow-as-data-inversion-plan.md`). Default stays yolo; scoping is strictly opt-in and forward-compatible with the planned U1 generic workflow descriptor.

---


ivankuznetsov:

## High
- [ ] `scoped` permissions deny tools they explicitly allow at `lib/hive/permission_scope.rb:203`: `tools: [Read, Write, Edit]` and `bash: true` still emit `--disallowedTools Write,Edit,...,Bash`, so the custom scope cannot actually grant those tools despite the plan requiring `tools:` to be authoritative.
- [ ] Explicit blank `permissions:` values fail open to yolo at `lib/hive/permission_scope.rb:95`: config validation calls the same nil-as-absent parser from `lib/hive/config.rb:1707`, so `permissions:` or `plan.permissions:` with no value is accepted instead of hard-erroring as malformed config.
- [ ] Non-Claude non-yolo stage scopes can leave stale task markers at `lib/hive/stages/execute.rb:222`: `stage_permission_scope` raises before `spawn_agent`, while `4-execute` has already written `AGENT_WORKING` at `lib/hive/stages/execute.rb:104`, and `with_stage_events` only emits events on rescue, so the required attributed `:error` marker is not written.

## Medium

## Nit

ivankuznetsov:

# Code review — U10: Per-stage permission scopes (pass 1)

Scope reviewed: the U10 commits (`15f979a1..HEAD`, the top 6 commits) of
`git diff origin/main..HEAD`. The rest of the branch (U1–U9 generic-workflow
work) was reviewed under prior tasks and is treated as out of scope here.

## High
- [ ] `scoped` with explicit `tools:` (and `bash: true`) emits the granted tool in BOTH `--allowedTools` and `--disallowedTools`, so Claude denies it — `scoped` is broken for its primary use case: `lib/hive/permission_scope.rb:199-205` (`scoped_scope`) hardcodes `disallowed_tools: READ_ONLY_DISALLOWED.dup` (`Write Edit MultiEdit NotebookEdit Bash`) regardless of `allowed_tools`. For `tools: [Read, Write, Edit]` the resolver returns `allowed=[Read,Write,Edit]` *and* `disallowed=[Write,Edit,…]` (confirmed by the test at `test/unit/permission_scope_test.rb:76-77`), and `bash: true` puts `Bash` in both lists. Claude Code's deny rules take precedence over allow rules, so Write/Edit/Bash are denied even though they were explicitly granted — contradicting `docs/permissions.md:88-89` ("`tools:` is authoritative … exactly those tools") and `:103` ("`bash: true` allows … `Bash`"), and the shipped `templates/project_config.yml.erb` example (`tools: [Read, Write, Edit]`). Fix: subtract the allowed set from the deny list (e.g. `READ_ONLY_DISALLOWED - allowed`) or emit no deny list when explicit `tools:` are given.

## Medium
- [ ] The `scoped`/`bash:` resolver tests assert only `allowed_tools`, never `disallowed_tools`, and the live smoke covers only `read-only`/`yolo` — so the High finding is completely unguarded: `test/unit/permission_scope_test.rb:60-100` and `test/unit/spawn_agent_test.rb:506-531` check the allow list and dirs but never the deny list; `test/smoke/permission_scope_headless_smoke_test.rb` only proves `read-only` denies a write and `yolo` permits it. There is no unit or live proof that a `scoped`/`bash:true` agent can actually *use* the tool it was granted, which is exactly why the contradiction slipped through.
- [ ] The A8 runner gate (non-yolo on codex/pi) is verified only at the helper level and escapes as an uncaught `Hive::ConfigError` (exit 78), not the plan's stated "attributed :error marker": `lib/hive/permission_scope.rb:58-62` raises, but the stage runners call `stage_permission_scope` outside any rescue (`lib/hive/stages/plan.rb:32`, `execute.rb`, reviewers, etc.), and `with_stage_events` re-raises (`lib/hive/stages/base.rb:212-214`), so `hive run` exits 78 with no task.md marker. This is fail-closed/loud (the core A8 safety intent is met, and it matches the existing template-`ConfigError` pattern), but the U10-4 scenario ("surfaced as an attributed `:error` marker, not a hang") is not what happens and there is no end-to-end test of the codex/pi path — only `test_stage_permission_scope_rejects_non_claude_non_yolo` at the helper boundary.

## Nit
- [ ] `lib/hive/stages/brainstorm.rb` keeps the comment "add_dirs is intentionally limited to the task folder … must not have write access to the project source," but `add_dirs` is now `scope.fetch(:add_dirs)`, so a `scoped` brainstorm with `dirs:` widens it beyond the task folder — the comment now mis-describes the non-yolo path.
- [ ] `lib/hive/permission_scope.rb:212-213` re-validates `dirs` inside `resolve_dirs` (with a synthetic `original: { "dirs" => dirs }`), but `parse_spec`→`validate_scoped!` already validated it on the resolve path, so this branch can never raise — dead defensive code with an inconsistent error context.
- [ ] A7's "golden spawn-arg" pinning is realized as scope-dict assertions (`test_stage_permission_scope_yolo_preserves_*`) plus separate emission tests, not a true per-stage byte-identical argv golden for the absent-permissions path; a per-stage argv golden would more directly lock "no block ⇒ identical argv to today."

ivankuznetsov:

# PR review (pr-review-toolkit, multi-persona) — U10: Per-stage permission scopes (pass 1)

Scope: the U10 commits only (`git diff 6d9a5d42..HEAD`, top 6 commits). Personas run:
code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer,
comment-analyzer, code-simplifier. Findings are the deduped union; the persona that
surfaced each is prefixed. Confirmed-correct by personas (not defects): the A9
OS-sandbox caveat is present/prominent; yolo is byte-identical per stage and mode;
`read-only` (disjoint allow/deny) works and is smoke-proven; non-yolo→codex/pi is a
hard error at resolve; full-replace (no field merge) is correct.

## High
- [ ] [code-review][type-design][tests][comment] `scoped` re-denies the tools it grants: `lib/hive/permission_scope.rb:199-205` (`scoped_scope`) sets `disallowed_tools: READ_ONLY_DISALLOWED.dup` (`Write Edit MultiEdit NotebookEdit Bash`) unconditionally, even when `tools:`/`bash:` grant those tools. Claude deny rules win over allow rules, so `scoped tools: [Read,Write,Edit]` denies Write/Edit and `bash: true` denies Bash (reproduced: overlap `["Write","Edit"]` / `["Bash"]`) — `scoped`'s core purpose is broken. Fix: `disallowed = READ_ONLY_DISALLOWED - allowed` (or omit the deny list when explicit `tools:` are given).
- [ ] [comment][code-review] Docs and the shipped init template present the broken behavior as working: `docs/permissions.md:88-89` ("`tools:` is authoritative … exactly those tools") and the `tools: [Read, Write, Edit]` / `bash: true` examples (`:39-44,63-67,103`), plus `templates/project_config.yml.erb:60-69` (written into every new project), all promise write/edit/bash that the resolver denies — an operator copying the flagship example gets a read-only-effective stage. (Same root cause as above; cured by fixing the resolver, but currently ships misleading guidance.)
- [ ] [silent-failure][tests] A8 gate `ConfigError` escapes single-agent stages with no attributed marker: `stage_permission_scope` raises (`lib/hive/stages/base.rb:130` → `permission_scope.rb:58-62`) as the first line of `plan.rb:32`, `execute.rb:222`, `open_pr.rb:137`, `finalize.rb:105`, `artifacts.rb:47`, `agent.rb:23`, `brainstorm.rb:53/90`, `brainstorm_tmux.rb` — none of which `rescue`, so it propagates through `with_stage_events` (`base.rb:212-215`, event-only re-raise) leaving the state file's prior marker untouched. `review.rb:627-639` rescues the identical error into an attributed `:review_error`, proving the intended contract; single-agent stages don't apply it, diverging from the plan's required "attributed `:error` marker, not a hang" and bypassing `agent.rb`'s own no-silent-caps marker conversion (`agent.rb:45-62`).

## Medium
- [ ] [code-review] A top-level `review: { permissions: … }` block is validated (`config.rb:1718` collects it) but never resolved — `permission_spec` is only ever called with `review.ci`/`review.triage`/etc., never `"review"` — so it passes load and is silently ignored at runtime (operator gets `yolo`). Honor a `review`-level default or reject the key.
- [ ] [tests] The suite locks in the High bug: `test/unit/permission_scope_test.rb:76-77` asserts the contradictory `allowed=[Read,Write,Edit]` + `disallowed=[Write,Edit,…]` as correct, and `:82-100` (bash sugar) asserts only `allowed_tools`, never that Bash is absent from the deny list. No test asserts `allowed ∩ disallowed == ∅` for scoped.
- [ ] [tests] No live or unit proof that `scoped` actually permits a granted tool: `test/smoke/permission_scope_headless_smoke_test.rb` covers only `read-only` (deny) and `yolo` (allow); the scoped-write path — exactly the one broken above — is never exercised end-to-end.
- [ ] [tests] A7 "absent ⇒ byte-identical argv" is asserted only as scope-dicts for one stage (`spawn_agent_test.rb:453-484`), not as a real per-stage, per-mode argv golden; a stage that forgot to thread `default_allowed_tools` (tmux) or flipped a list would still pass.
- [ ] [tests] The per-reviewer `review.reviewers[i].permissions` path is validated (`config.rb:1728-1730`) but has zero resolution/spawn coverage — validated-only, never resolved through a reviewer runner.
- [ ] [comment][code-review] `lib/hive/stages/brainstorm.rb:57-67` keeps the comment "add_dirs is intentionally limited to the task folder … must not have write access to the project source," but the line is now `add_dirs: scope.fetch(:add_dirs)`, which a `scoped` brainstorm with `dirs:` widens — the comment asserts a safety invariant a config can now override.
- [ ] [type-design] Dual representation: `stage_permission_scope` (`base.rb:118-152`) flattens the `Scope` struct into a differently-shaped bare Hash (`add_dirs` vs `add_dirs_extra`, struct→Hash) consumed by ~16 call sites via untyped `.fetch(:key)`; the smoke test re-derives `[task.folder] + scope.add_dirs_extra` itself. Two parallel "permission result" shapes that can drift; returning the struct (or a second struct) would unify them.
- [ ] [type-design] `scoped` is split half in the `PRESETS` table (`permission_scope.rb:24-27`, only `permission_mode`) and half hardcoded in `scoped_scope` (`:191-206`) — two construction paths with divergent rules, which is exactly what lets the deny-overlap slip through. Make the table fully describe each preset or derive the scoped deny list from the allow list.
- [ ] [comment] Maintainer comments don't disclose that `scoped` force-applies the read-only deny list: `permission_scope.rb:3-5` (`READ_ONLY_DISALLOWED`) frames it as the read-only preset's safety net, and `:9-10` (`NON_PROMPTING_PERMISSION_MODE`) says scopes "pre-approve the intended set" — neither notes the deny list overrides overlapping `tools:`/`bash:` entries.

## Nit
- [ ] [code-review][simplify] `permission_scope.rb:213` (`resolve_dirs`) re-runs `validate_dirs!` already performed by `parse_spec`→`validate_scoped!` (`:142`); dead defensive branch with a synthetic `original:` context that can never fire on the resolve path.
- [ ] [code-review] `Hive::Stages::BrainstormTmux.wrapper_command` (`brainstorm_tmux.rb:69-80`) is dead (run! uses `launch!` directly) yet the diff spent edits threading the scope through it — consider deleting rather than maintaining.
- [ ] [simplify] `config.rb:1718` repeats the "guard `is_a?(Hash) && key?('permissions')`, push `[label, h['permissions']]`" shape three times; a small local lambda collapses all three with identical output.
- [ ] [simplify] `permission_spec` (`config.rb:642`) sits ~700–1100 lines from its collaborators `permission_at` (`:1742`) and `MISSING_PERMISSION` (`:1351`); co-locating the unit aids readability (pure move).
- [ ] [type-design] `yolo?` keyed on `preset == "yolo"` (`permission_scope.rb:43-44`) restates a string literal also used in `resolve` (`:58`) and `config.rb` DEFAULTS; frozen constants/symbols would localize a rename.
- [ ] [type-design] `Scope` nullability is asymmetric (tool lists nil under yolo, `add_dirs_extra` always `[]`) (`permission_scope.rb:35-41`) and undocumented; a one-line struct comment would make the mixed contract self-evident.
- [ ] [tests] `tool_csv` order/dedup is unspecified (`permission_scope_test.rb:157-160` checks only blank-stripping); duplicates can arise from the allow/deny overlap and would pass through to argv.
- [ ] [tests] The `Config.load` fail-closed table (`config_test.rb:556-574`) omits the "unknown key for preset" and non-hash/`42` malformed shapes that the resolver table covers — operator-facing load path isn't pinned for those two.
- [ ] [silent-failure] `permission_entries` (`config.rb:1714-1734`) validates a `permissions` key on any top-level Hash, broader than the stage set `permission_spec` actually reads — fail-safe (validates more), but could emit a confusing "stage X permissions invalid" for a non-stage key.

ivankuznetsov:

## High
- [ ] Explicit blank `permissions:` fails open to `yolo` (`lib/hive/permission_scope.rb:117`): YAML `permissions:` / `stage.permissions:` becomes nil, passes validation, and disables scoping instead of failing closed per A3.
- [ ] Non-yolo Codex/Pi reviewer scopes are not always hard errors (`lib/hive/stages/review.rb:1339`): the A8 `Hive::ConfigError` from reviewer scope resolution is rescued as a per-reviewer infra failure, so a mixed reviewer pass can continue after silently dropping the unenforceable reviewer.

## Medium

## Nit

ivankuznetsov:

# Code review — U10: Per-stage permission scopes (pass 2)

Scope reviewed: the U10 commits (`15f979a1^..HEAD`) of `git diff origin/main..HEAD`.
The rest of the branch (U1–U9 generic-workflow work) was reviewed under prior tasks
and is treated as out of scope here. Pass-1 auto-fixes verified as landed: `scoped`
now subtracts granted tools from the deny list (`permission_scope.rb:230`), single-agent
stages attribute the A8 runner-gate failure to an `:error` marker via
`stage_permission_scope_or_mark!`, and the A9 OS-sandbox caveat is present
(`docs/permissions.md:9-13`). All U10 unit suites pass (permission_scope, config,
spawn_agent, agent, reviewers/agent, stages/execute).

## High
- [ ] Blank/empty `permissions:` value fails open to `yolo` (`lib/hive/permission_scope.rb:117-118`, `lib/hive/config.rb` `permission_spec`/`validate_permission_spec!`): a present-but-blank `permissions:` (or `plan.permissions:`) parses to `nil`, `normalize_spec` maps `nil → {"preset" => "yolo"}`, so it passes `Config.load` and silently resolves to full permissions instead of hard-erroring per A3's fail-closed-on-broken-config intent — a fail-open in a security feature, and there is no test pinning either reading. (Already raised as escalation Q1 / codex-02 #1; two defensible readings — "blank == absent default" vs "blank == present-but-broken → fail closed".)

## Medium
- [ ] Non-yolo scope on a non-claude (codex/pi) reviewer is swallowed as a per-reviewer infra error, not the A8 hard error (`lib/hive/stages/review.rb:1339` + `lib/hive/reviewers/agent.rb:27-33`): `Reviewers::Agent#run!` resolves the scope (calling `stage_permission_scope`, which raises `Hive::ConfigError` for non-claude non-yolo) *inside* `run_reviewer_spec`'s `rescue StandardError`, recording a per-reviewer `:error` and letting the pass continue (`:ok` if other reviewers pass) — so A8's "hard error, never silent-degrade" is not honored for reviewers; the scoped reviewer silently vanishes from every pass. No test covers this path. (codex-02 #2.)
- [ ] `patrol.review.reviewers[]` permission specs escape A3's fail-closed-at-load guarantee (`lib/hive/config.rb` `permission_entries`): the validator scans only `review.{ci,triage,fix,browser_test}` and `review.reviewers[]`, never the patrol equivalents, yet patrol reviewers DO resolve `permissions:` at spawn (via `Reviewers::Agent` `explicit_permission_spec`), so a malformed/unsupported patrol-reviewer spec passes load and only fails at spawn — where it is swallowed per the finding above.

## Nit
- [ ] Misleading maintainer comment in `lib/hive/stages/base.rb` (`stage_permission_scope_or_mark!`): it claims review sub-stages route the same `ConfigError` to the outer `Review.run!` rescue → `:review_error`, but for the only runtime `ConfigError` a reviewer can raise (A8 non-claude scope, resolved inside `Reviewers::Agent#run!`) the inner `run_reviewer_spec` `rescue StandardError` (review.rb:1339) intercepts it first, so the documented mapping never occurs — the comment overstates the guarantee (documentation half of the Medium reviewer finding).
- [ ] `Config#permission_entries` validates `permissions:` on *every* top-level Hash block (`cfg.each`), which is a superset of the locations `permission_spec` actually resolves (stage blocks + review roles/reviewers) — harmless (errs toward more validation) but inconsistent with the method's own comment ("Validating exactly the resolved set"); a stray `permissions:` under a never-resolved block (e.g. `claude:`) is validated rather than rejected as unsupported, unlike the `review.permissions` case which is explicitly rejected.

ivankuznetsov:

# pr-review-toolkit multi-persona review — U10: Per-stage permission scopes (pass 2)

Scope: `git diff 6d9a5d42..HEAD` (the 11 U10 commits). U1–U9 work excluded.
type-design-analyzer, comment-analyzer, code-simplifier. Union below, deduped.

Context for triage: the **code-reviewer** persona found **no High/Medium**
correctness bugs in the core resolver/wiring and verified the load-bearing
properties hold (byte-identity on the yolo/absent path, allow∩deny disjoint,
A8 ConfigError attributed on every *pipeline* path, blank fails closed). The
High/Medium items below are mostly **fail-open coverage gaps at the edges**
(reviewer kinds, patrol/non-stage config locations, ancillary spawns) and
**test/maintainability** concerns — not core-logic defects.

## High
- [ ] [silent-failure] `codex_review` reviewers never resolve/gate their `permissions:` — `lib/hive/reviewers/codex_review.rb` `run!` launches `codex review` directly without `PermissionScope.resolve`, so a non-yolo scope (which load-validation accepts as a valid reviewer entry) runs unscoped on codex with NO A8 hard error; `codex_review` is the patrol-default reviewer kind. Violates A8 "non-yolo on codex = hard error, never silent-degrade." (Real impact bounded by codex review being read-only, but the contract is broken on the validated reviewer surface.)
- [ ] [silent-failure] `patrol.review.reviewers[].permissions` is resolved at runtime but never validated at load — `permission_entries` (lib/hive/config.rb:1715-1735) scans only `review.reviewers`, not `patrol.review.reviewers` (dispatched at review.rb:1216). A malformed/non-yolo patrol-reviewer spec passes `Config.load` (A3 "a typo fails the load" missed) and, combined with the codex_review-default not resolving at all, yields a fully silent fail-open on the patrol review path.

## Medium
- [ ] [silent-failure] Ancillary execute-profile spawns bypass scoping: babysitter PR-fixer (`lib/hive/babysitter/pr_fixer.rb:219`) and rebase conflict agent (`lib/hive/rebase.rb:313`) spawn the configured `execute.agent` profile but never call `stage_permission_scope`, so a non-yolo project-default scope is silently unenforced and never hard-errors on codex/pi. NOTE: the code-reviewer persona judged these out-of-pipeline and "unaffected (not findings)" — flagged here for an explicit scope decision, since A8's "never silent-degrade" invariant does not hold on these write-capable spawns.
- [ ] [silent-failure / comment-analyzer] A `permissions:` key under any non-stage top-level block (`babysitter`, `daemon`, `digest`, `web`, `patrol`, `bot`, `rebase`, `update`, …) shape-validates at load (config.rb:1723 `cfg.each` collects from every top-level Hash) but is never resolved at runtime → accepted-but-silently-ignored, the same fail-open class `reject_unsupported_review_permissions!` deliberately blocks for `review.permissions`. The `permission_entries` doc comment (config.rb:1706-1714) wrongly claims it validates "exactly the resolved set."
- [ ] [type-design] `Scope` struct enforces zero invariants — `lib/hive/permission_scope.rb:51` — `Scope.new` accepts allow∩disallow non-empty, yolo-with-tool-lists, or non-yolo-without-permission_mode; the non-overlap invariant lives only in the `scoped_scope`/`build_scope` factories, so a future second construction site can silently reintroduce the deny-wins-over-grant bug just fixed in b6a0c72b. Consider a `.build` factory that validates+freezes.
- [ ] [type-design] Dual representation of tool lists — `lib/hive/permission_scope.rb:15` vs `lib/hive/claude_launcher.rb:76` — historical `*_ALLOWED_TOOLS` are CSV **strings** while `Scope#allowed_tools` is an **Array**, and both feed the same `allowed_tools:` param; it works only because `tool_csv` is incidentally idempotent on a pre-joined CSV string (`Array("a,b")→["a,b"]→"a,b"`). Latent: the day a constant is made an Array (or an Array is passed where a CSV string was expected), the contract breaks silently. Make the constants Arrays, or document the param as "CSV String | Array | nil".
- [ ] [pr-test-analyzer] A8 reviewer codex/pi propagation is pinned only against a hand-written stub that itself raises (`test/unit/stages/review/run_reviewers_test.rb:1841/1855`), never the real `Reviewers::Agent#run!`→`stage_permission_scope` path; a regression making the real adapter swallow/skip the gate would still pass this test.
- [ ] [pr-test-analyzer] The tmux half of the A7 golden is NOT byte-identical — `test/unit/spawn_agent_test.rb:575-592` asserts only `--allowedTools` presence, `--dangerously-skip-permissions` presence, and `--disallowedTools` absence (the headless branch IS byte-for-byte). Plan calls for "byte-identical … in BOTH headless AND tmux modes"; a reordered/extra tmux wrapper flag on the yolo path would go uncaught.
- [ ] [comment-analyzer] Template `execute` scoped example would fail closed under the default dev agent — `templates/project_config.yml.erb:65-69` renders `execute.agent` to the default development agent (codex); uncommenting the scoped `permissions:` without switching the agent to claude hard-errors (non-yolo is claude-only), and the per-stage example carries no Claude-only caveat.
- [ ] [code-simplifier] The 3-line claude scope-kwargs triplet (`permission_mode/allowed_tools/disallowed_tools: scope.fetch(...)`) is copy-pasted verbatim across ~13 spawn sites (plan/execute/open_pr/finalize/artifacts/agent/brainstorm×2/brainstorm_tmux/review.rb/triage/ci_fix/browser_test/reviewers/agent). A small splattable `Base.tool_scope_kwargs(scope)` would remove the drift trap without merging the spawn structure.

## Nit
- [ ] [code-reviewer / comment-analyzer] Commented template examples suggest a stage scope that breaks the stage's own contract: `read-only` on `brainstorm`/`plan` (`templates/project_config.yml.erb:55,60`) denies Write/Edit, but those stages must write `brainstorm.md`/`plan.md` to emit their marker — if uncommented the stage cannot complete. Use a `scoped` (tools: [Read,Write,Edit]) example for artifact-producing stages.
- [ ] [code-reviewer / type-design] `shared_reviewer_groups` (lib/hive/stages/review.rb:1222) keys on `spec["permissions"]` vs a `:stage_default` symbol, so a reviewer with explicit `permissions: yolo` and one omitting the key (project default yolo) get separate tmux sessions despite identical effective scope — correct but mild extra session cost (and a mixed string/symbol key space); could normalize on the resolved scope.
- [ ] [pr-test-analyzer] The A7 absent-permissions argv golden covers only the `plan` stage, not `execute` (different default allowlist, `IMPLEMENTER_ALLOWED_TOOLS`) — `test/unit/spawn_agent_test.rb:539`; the plan's "per stage" claim is literally proven only for plan.
- [ ] [pr-test-analyzer] `test_yolo_is_a_noop_scope` (test/unit/permission_scope_test.rb:45) proves bypassPermissions + nil tool lists only at the resolver level; the spawned-argv no-flags guarantee is backstopped only indirectly (headless golden) — the named test is resolver-only.
- [ ] [comment-analyzer] The `NON_PROMPTING_PERMISSION_MODE` comment (lib/hive/permission_scope.rb:18-21) says "The disallow list is always READ_ONLY_DISALLOWED with the granted tools subtracted out" — that subtraction only happens for `scoped` (line 243); the `read-only` preset uses the static set, so the claim is inaccurate for read-only.
- [ ] [comment-analyzer] `docs/permissions.md:88` "passes exactly those tools" ignores that `tool_csv` dedups and strips blanks — minor overstatement.
- [ ] [comment-analyzer] `docs/permissions.md:115-116` says runner-support is checked "when a stage resolves its agent profile"; it is actually checked inside `PermissionScope.resolve` — identical effect, harmless wording drift. (Same imprecision in `wiki/log.d/…review-fixes.md:18`.)
- [ ] [type-design] `normalize_tools` (lib/hive/permission_scope.rb:248) is a public `module_function` that strips but neither dedups nor rejects blank-after-strip; safe on the resolve path (validated upstream, deduped at `tool_csv`) but a direct caller gets unguarded entries — mirror the `resolve_dirs` "already validated" note.
- [ ] [type-design] `MAP_KEYS["scoped"]` (lib/hive/permission_scope.rb:44) and `validate_scoped!` (171-183) declare the scoped shape in two places; they agree today, but a future key added to one won't be enforced by the other.
- [ ] [code-simplifier] `permission_at` (lib/hive/config.rb:1779) branches `parts.length == 1 ? cfg[parts.first] : cfg.dig(*parts)`; `cfg.dig(*parts)` already handles the single-element case identically — the special-case can collapse.
- [ ] [code-simplifier] `kwargs[:explicit_permission_spec] = spec["permissions"] if spec.key?("permissions")` is duplicated at lib/hive/stages/review.rb:1236 and lib/hive/reviewers/agent.rb:29 — extractable to a one-line helper.
- [ ] [code-simplifier] `resolve` (lib/hive/permission_scope.rb:81-84) hard-codes the preset string twice per pass-through arm; `when YOLO, "read-only" then build_scope(parsed.fetch("preset"))` would de-dup (optional — current form is clear).

---
Out-of-diff follow-up (not a U10 finding, noted by comment-analyzer): `wiki/decisions.md:249` (ADR-008) still states every stage's add-dir is narrowed to `task.folder` only and brainstorm/plan never add the project root — U10's `scoped dirs:` can now append extra add-dirs, so that ADR is slightly stale. The file is not in the U10 diff; flag for whoever owns the decisions log.

ivankuznetsov:

## High

## Medium
- [ ] `test/unit/permission_scope_test.rb:201` embeds a literal NUL byte: Git treats the Ruby test file as binary, hiding the new permission-scope tests from normal diff review.

## Nit
- [ ] `templates/project_config.yml.erb:14` overstates `claude.permission_mode`: non-yolo `permissions:` scopes force `permission_mode: "default"`, so the inline config comment is misleading.
- [ ] `docs/permissions.md:104` says blank `tools:` entries are dropped: config validation rejects blank tool entries, so the reference page describes the wrong behavior.

ivankuznetsov:

# Code Review — U10 Per-stage permission scopes (pass 3)

Scope reviewed: the U10 commits only (`git diff 15f979a1^..HEAD`), isolated from
the earlier U1–U8 workflow-generalization commits on the same stacked branch.
Reviewers: own deep read + ce-correctness (no P0–P3 bugs found) + ce-adversarial.
Plan invariants A7 (byte-identical yolo argv per stage and per mode) and A8
(non-yolo on codex/pi = hard ConfigError, attributed marker) were verified to
hold on every spawn path; the A7 goldens in `test/unit/spawn_agent_test.rb` pin
the argv byte-for-byte. Findings below are gaps at the edges, not core defects.

## High

_No High-severity findings. The resolver, fail-closed validation, the A7/A8
invariants, and the deny-wins disjointness guarantee are correctly implemented
and well-tested._

## Medium
- [ ] `test/unit/permission_scope_test.rb:201` embeds a **raw NUL byte** in the `dirs: ["a\0b"]` test value: git/`file` classify the entire 300-line test file as binary (`Bin 0 -> 11129`), so it is unreviewable in textual diff and may be skipped by linters/tooling. Replace the raw NUL with the Ruby escape `"ab"` — identical runtime value, file stays text.
- [ ] `lib/hive/reviewers/codex_review.rb:202` + `lib/hive/config.rb:1847` — A8 gate is bypassable: `validate_agent_name!` accepts `agent: claude` on a `kind: codex_review` reviewer, so `enforce_permission_scope_gate!` resolves with the **claude** profile, the A8 guard never fires, and the declared non-yolo reviewer scope is silently discarded (codex_review passes no tool args). Contradicts the plan's A8 "never silent-degrade." Fix: reject `kind: codex_review` with a non-codex agent at load, or fail-closed on any non-yolo spec regardless of profile.
- [ ] `lib/hive/permission_scope.rb:263` — `validate_tools!` lacks the null-byte guard that `validate_dirs!:276` has: a `scoped tools:` entry containing a NUL passes load validation, resolves cleanly, then crashes `Process.spawn` (`ArgumentError: string contains null byte`); for `:state_file_marker` stages `stage_permission_scope_or_mark!` only rescues `Hive::ConfigError`, so `AGENT_WORKING` is stranded (looks like a hang). Mirror `validate_dirs!`'s `&& !tool.to_s.include?("\0")` guard.

## Nit
- [ ] `lib/hive/permission_scope.rb:331` — `resolve_dirs`'s `File.expand_path` raises `ArgumentError` on `dirs: ["~unknownuser/x"]`; this escapes `stage_permission_scope_or_mark!`'s ConfigError-only rescue and surfaces as a bare crash with a stale marker instead of the attributed `:error` the U10-4 contract promises. Re-raise expand_path failures as `Hive::ConfigError`, or broaden the rescue.
- [ ] `lib/hive/permission_scope.rb:263` — `validate_tools!` is fully string-trusting: a CSV-in-one-element (`tools: ["Read,Write"]`) or case/typo (`"write"`) silently yields deny-over-grant — the disjointness invariant sees one opaque element while Claude splits/ignores it, re-denying the intended tool, with no operator-visible error. Reject commas/whitespace in tool entries (and consider validating against Claude's tool vocabulary or documenting case-sensitivity).
- [ ] Testing gaps: no end-to-end test that a non-claude / `codex_review` reviewer A8 `ConfigError` reaches `Stages::Review.run!`'s rescue and stamps `:review_error reason=config_error` on the real task; no shared-session mixed-group test (one explicit `permissions:` + one inheriting an equal default landing in the same session); no reviewer-path test that `scoped dirs:` extends `[ctx.task_folder]` (relative + absolute) rather than replacing it.
- [ ] `test/smoke/permission_scope_headless_smoke_test.rb:120` — `log_label: "permission-scope-smoke-#{permissions}"` interpolates a Hash for the `scoped` case (renders the full hash literal in the label). Cosmetic; use a stable string token (e.g. `permissions.is_a?(Hash) ? permissions["preset"] : permissions`).

ivankuznetsov:

# PR Review Toolkit — U10 Per-stage permission scopes (pass 3)

Scope: U10 commits only (`git diff 15f979a1^..HEAD`), isolated from the earlier
U1–U8 workflow-generalization work stacked on the same branch. Six personas ran:
code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter,
type-design-analyzer, code-simplifier. code-reviewer found **no** production
bugs (rubocop/`ruby -c`/touched unit suites all green; no OpenStruct; A7/A8
hold). Findings below are the union; severities are synthesized into hive's
High/Medium/Nit with the originating persona's rating noted where it differs.

## High
- [ ] [errors] `resolve_dirs`'s `File.expand_path` raises `ArgumentError`/`TypeError` on a `dirs: ["~baduser/x"]` (or non-string) entry — `validate_dirs!` guards blanks/null-bytes but not un-expandable `~user`, and `stage_permission_scope_or_mark!` rescues **only** `Hive::ConfigError`, so it escapes as a bare crash that leaves a stale `AGENT_WORKING`/`REVIEW_WORKING` marker (looks like a hang), defeating U10-4's "attributed `:error` marker, not a hang" contract (`lib/hive/permission_scope.rb:331` + `lib/hive/stages/base.rb:199`). Fix: reject `~`/un-expandable dirs in `validate_dirs!` and/or rescue `ArgumentError`/`TypeError` in `resolve_dirs` → `ConfigError`. [silent-failure-hunter: Critical/blocking; edge-case trigger but fail-closed]

## Medium
- [ ] [tests] `codex_review`'s `enforce_permission_scope_gate!` raise path is behaviorally untested — no test asserts a non-yolo `codex_review` reviewer raises (the yolo path covers its lines without exercising A8); the plan explicitly requires verifying A8 "via agent.rb AND codex_review.rb" (`lib/hive/reviewers/codex_review.rb:202`). [pr-test-analyzer: Critical]
- [ ] [tests] No end-to-end test that an A8 `ConfigError` from a review sub-stage (`ci`/`triage`/`fix`/`browser_test`, which use plain `stage_permission_scope`) reaches `Review.run!`'s rescue and stamps `:review_error reason=config_error` on the real task (`lib/hive/stages/review.rb:627`). [pr-test-analyzer: Critical]
- [ ] [tests] Multi-group shared-session reviewer grouping is untested — `shared_reviewer_groups`/`shared_reviewer_session_name`/`shared_reviewer_permission_scope` have zero test references; differing-scope reviewers splitting into `-scope2` sessions and the same-scope dedup (the `group.first` invariant) are never exercised (`lib/hive/stages/review.rb:1219`). [pr-test-analyzer: Critical]
- [ ] [tests] `test/unit/permission_scope_test.rb:201` embeds a **raw NUL byte** (`"a\0b"`) → git/`file` treat the whole 300-line file as binary (`Bin 0 -> 11129`), so it is unreviewable in diff and tooling may skip/mangle it; use the `""` escape (identical runtime value).
- [ ] [tests] Brainstorm's A7 headless-vs-tmux argv split is not golden-pinned — only `plan`/`execute` have byte-identical goldens, but brainstorm has a distinct dual-path impl (`brainstorm.rb` headless vs `brainstorm_tmux.rb`) whose old `wrapper_command` was deleted here.
- [ ] [tests] `default_allowed_tools_for_mode` guard clauses are only tested on the yolo branch — the non-claude-profile-in-tmux→nil and nil-default→nil paths are unexercised (`lib/hive/stages/base.rb`).
- [ ] [type-design] `Scope.new` is a public **unvalidated** constructor: `Struct.new` exposes it, so `Scope.new(preset: "yolo", allowed_tools: ["Bash"], …)` builds an invariant-violating scope with no error — the `build` factory is convention, not a gate. Make `new` private and route `build` through it (`lib/hive/permission_scope.rb:59`). [type-design-analyzer: Important]
- [ ] [type-design] `Scope` `freeze` is shallow — the `allowed_tools`/`disallowed_tools`/`add_dirs_extra` arrays stay mutable, so a built scope's lists can be mutated in place to re-introduce the deny/allow overlap; the comment at `:76-78` ("can't be mutated after the fact") overstates the guarantee. Deep-freeze the member arrays in `build` (`lib/hive/permission_scope.rb:87`). [type-design-analyzer: Important]
- [ ] [errors/code] `validate_tools!` accepts tool entries with embedded commas/whitespace/unknown names (only "non-empty String" is checked): `tools: ["Read,Write"]` is one element to the disjointness invariant but split by Claude → silent deny-over-grant; `"write"`/typos grant nothing while staying denied — both silently mis-scope with no operator error. Reject commas/whitespace in `validate_tools!` (`lib/hive/permission_scope.rb:263`). [code-reviewer + silent-failure-hunter]
- [ ] [comments] `lib/hive/config.rb:1724-1727` claims `review` and `patrol.review` are both "skipped in the top-level scan," but only `review` is `next`-skipped — a top-level `patrol` Hash IS collected and shape-validated (`collect.call("patrol", …)`); `patrol.review` is merely never visited because the scan is shallow. Misleading symmetry; correct the wording.
- [ ] [type-design] Dual representation: `stage_permission_scope` immediately destructures the validated `Scope` struct into an unnamed spawn-kwargs Hash that every consumer uses, so the struct's invariants don't travel past the resolver (the base.rb:159 "NOT the Scope struct" warning is the tell). Give the Hash a named schema or add `Scope#to_spawn_kwargs` (`lib/hive/stages/base.rb:121`). [type-design-analyzer: Important design smell]

## Nit
- [ ] [comments] `lib/hive/reviewers/codex_review.rb:144` "codex_review ALWAYS runs a non-claude runner" overclaims — the profile comes from `spec.fetch("agent")` and isn't constrained, so `kind: codex_review, agent: claude` passes the gate and silently discards the scope; tie the claim to the `<bin> review` invocation, not the profile class, and/or reject `codex_review` with a non-codex agent at load.
- [ ] [comments] `lib/hive/stages/base.rb:158` "across the ~14 spawn sites" is stale (15 sites now); drop the number so it can't rot.
- [ ] [comments/simplify] `lib/hive/permission_scope.rb:18-24` duplicates the deny-list-derivation explanation already given at `:8-14` and `:311-315`; trim this block to the permission-mode-only rationale to avoid three-way drift.
- [ ] [errors] `tool_csv` returns `nil` on an all-blank Array, so a `scoped` scope whose `allowed_tools` normalized to empty would emit no `--allowedTools` (safety rests entirely on `validate_tools!` upstream); add a "non-yolo scoped ⇒ non-empty allowed_tools" invariant to `Scope.build` (`lib/hive/permission_scope.rb:90`). [silent-failure-hunter: Suggestion]
- [ ] [type-design] `tool_csv`'s String-vs-Array dual input (CSV-string constants vs `Scope` Arrays) is an acknowledged inconsistency; converting the `*_ALLOWED_TOOLS` constants to frozen Arrays gives a single input type so dedup/blank-drop applies uniformly (`lib/hive/permission_scope.rb:145`).
- [ ] [simplify] `build_scope`'s `add_dirs_extra:` keyword is dead — its only caller never passes it; inline `[]` (`lib/hive/permission_scope.rb:288`).
- [ ] [simplify] `resolve`'s `case` on preset has no `else`, so an unexpected preset returns `nil` and fails far away (`NoMethodError` on `scope.yolo?`); add a defensive `raise` (unreachable today, fails locally if a 4th preset is added) (`lib/hive/permission_scope.rb:135`).
- [ ] [tests] `tool_csv`'s documented String-passthrough contract isn't pinned directly — add `assert_equal "Read,LS", tool_csv("Read,LS")` / `assert_nil tool_csv(nil)` (`test/unit/permission_scope_test.rb`). [pr-test-analyzer: Important]
- [ ] [tests] The A8 reviewer-propagation test uses a hand-rolled raising stub (`UnenforceableScopeReviewer`) rather than the real adapter raise — acceptable (real raise covered in `agent_test.rb`), noted so it isn't mistaken for full coverage (`run_reviewers_test.rb:1841`).
- [ ] [code] Two analogous sentinels `MISSING_PERMISSION` (config.rb) vs `MISSING_EXPLICIT_PERMISSION_SPEC` (base.rb) encode the same "absent vs explicit-nil" at different layers — intentional and documented, but the divergent names could momentarily confuse; informational only.

ivankuznetsov:
Superseded by #562, which consolidates the full U1–U10 workflow-as-data inversion into one branch off current main (all unit fixes integrated, gates green) for single-feature evaluation. Closing u10's standalone PR.

PR #547, opened 2026-06-21
feat(workflows): author and run per-project workflows

ivankuznetsov:
## Summary

Until now Hive could only run its two built-in workflows, `coding` and
`content`. A project owner could not author their own. This PR lets a project
author a workflow descriptor under its Hive state tree, and Hive discovers,
validates, and registers it at runtime so `hive new PROJECT --workflow <id>
"..."` runs it end-to-end through the existing stage spine. `hive workflow new
<id>` scaffolds a valid, immediately-runnable blank starter to edit.

This is the first slice of the workflow-generalization epic: discovery,
validation, and a blank starter. Council attach, save-as-template, cross-machine
sharing, and reduced-permission sandboxing for untrusted descriptors are
deliberately out of scope and specced as follow-ups.

> Stacked on `u10-per-stage-permission-scopes` (PR #546). Review that one first;
> this PR's diff is only the `U9-*` commits on top.

## What a descriptor looks like

```yaml
# .hive-state/workflows/my-flow.yml
id: my-flow
stages:
  - { name: inbox, kind: terminal, state_file: idea.md }
  - { name: work,  kind: agent, state_file: work.md,
      instruction: ./my-flow/work.md, permissions: read-only }
  - { name: done,  kind: terminal, state_file: done.md }
```

Stage indexes and directories (`1-inbox`, `2-work`, `3-done`) derive from array
order. `kind: terminal` maps to an inert stage; `kind: agent` runs the generic
stage runner; `kind: council` is reserved and rejected with a clear error. Each
agent stage must declare exactly one of `skill:` or `instruction:`.

## How it fits together

```mermaid
flowchart TB
  B["Boundary call sites<br/>(Task, selection, init/new, status, drop)"] --> P["Workflows::Project.load!<br/>(idempotent, memoized per root)"]
  P --> L["Loader → DescriptorParser<br/>(glob, parse, validate)"]
  L --> R["Registry overlay<br/>(project_registrations)"]
  R --> U["reset_union_cache!<br/>(stage dirs / names / verbs)"]
  U --> C["Every consumer Just Works<br/>(resolver, approve, status, daemon)"]
```

The built-in `Registry::WORKFLOWS` stays frozen. Project descriptors layer into a
separate `project_registrations` overlay, merged at read time, with the memoized
stage-dir union reset on registration. This promotes the exact pattern the test
suite already used (`with_registered_workflow`) into production, so no consumer
of the union needed to change.

## Design decisions

| Decision | Rationale |
|---|---|
| Descriptors live at `<hive_state_path>/workflows/<id>.yml` | Consistent with `config.yml`; on the same state tree the daemon already manages. |
| `Project.load!` called at every resolution boundary | Idempotent + memoized per root, so "load at each boundary" beats a single fragile chokepoint and guarantees the union sees descriptors for status/drop/approve/resolver. |
| Overlay load/resolve held under `Project::LOCK` | The overlay is process-global; in-process web threads (status poller, dispatcher Approve/Drop) swap the active root concurrently, so each load+resolve critical section is serialized to keep custom-workflow rows from resolving against the wrong overlay. |
| `instruction:` injected as the stage instruction, not as `user_supplied` task data | Descriptors are trusted project-owner config. The `user_supplied` data tag stays reserved for prior artifacts. |
| `permissions:` flows through U10's `explicit_permission_spec` hook | Reuses the per-stage scope resolver; invalid specs still fail closed with a marker. |
| `state_file` restricted to a bare basename; id collision rejected at registration | A descriptor can neither write outside its task folder nor silently shadow a built-in (`coding`/`content`). |
| Cross-workflow stage-dir/short-name overlap tolerated | Matches shipped `content` behavior (shares `1-inbox`, ambiguous bare `done`); only descriptor id and intra-descriptor structure are enforced. |

The coding pipeline is behaviorally untouched: the stage-literal golden guard
stays green, line coverage stays at 100%, and rubocop is clean.

## Test plan

- `bundle exec rake test` — 6044 runs, 24254 assertions, 0 failures, 0 errors, 2
  skips.
- `bundle exec rake coverage` — 100.00% line coverage (the plan's gate, restated
  in every U9 unit).
- `bundle exec rubocop` — no offenses.
- New unit coverage spans the descriptor parser, loader, project registration,
  the agent runner's `instruction:`/`permissions:` paths, and the `workflow new`
  command, including the negative branches (malformed/colliding descriptors,
  path-traversal `state_file`, instruction-read failure, `--json` error
  envelopes).
- End-to-end (`test/integration/user_workflow_e2e_test.rb`): scaffold a flow,
  create a task with `--workflow`, advance it through `work` to `done`, and
  assert the scaffolded instruction body reached the work prompt — proving the
  authoring round-trip while the coding fixtures stay green in the same run.

## Review summary

Three report-only review passes ran (claude-ce, codex-ce, and the
pr-review-toolkit personas), each scoped to the U9 slice. Every finding was
resolved in-branch — auto-fixed or closed against the plan/owner-trust model with
a recorded rationale; none were escalated to the user. Substantive fixes that
landed:

- **Fault tolerance.** A single malformed `workflows/*.yml` no longer bricks
  unrelated tasks/commands — `load_dir` isolates per-file errors and consumers
  degrade per-project. An explicitly named `--workflow <id>` that maps to a
  broken/colliding descriptor surfaces the real `ConfigError` instead of a
  misleading "unknown workflow."
- **Concurrency.** The process-global overlay load+resolve (and the irreversible
  Drop archive guard) are serialized under `Project::LOCK`, closing a data race
  on the in-process web tier (status poller vs. dispatcher Approve/Drop) that
  could blank or mis-resolve custom-workflow rows.
- **Path safety.** Descriptor `state_file` is restricted to a bare basename
  (no leading `/`, no `..`, no nested dirs), so an authoring typo can't make Hive
  write marker/output files outside the task folder.
- **Structural invariants.** Parser enforces a terminal/inert last stage (so
  custom-workflow tasks stay droppable) and rejects `skill`/`instruction`/
  `permissions` on non-agent stages.
- **Operability.** `hive workflow new` is atomic (commit inside the rollback),
  routes all failure paths through the shared `--json` error envelope (schema
  now declares the `value` extra), and the agent runner guards instruction reads
  + no longer masks a re-run preflight failure behind a stale marker.

## Linked task

- Hive task **#1466** — "Per Project Workflows"
  (`u9-per-project-workflows-built-260618-7114`).
- Depends on **u10-per-stage-permission-scopes** (PR #546), the base of this
  stacked branch.

---


ivankuznetsov:

## High
- [ ] `hive approve` validates custom `--from/--to` stages before loading the project: a fresh CLI process rejects scaffolded stages like `2-work`, so user-authored workflows cannot advance past their first custom stage.
- [ ] Blank workflow terminal stages remain markerless and runnable: the generated `done` stage points at unwritten `done.md`, so status/daemon classify completed scaffolded tasks as `ready_to_run` even though inert terminal stages have no runner.
- [ ] `hive drop` checks archive stages against the wrong project overlay: slug resolution loads every project but archive filtering/guards use the last-loaded registry, so a completed custom-workflow task in another project can be hard-deleted instead of refused.

## Medium

## Nit

ivankuznetsov:

# Code Review — u9-per-project-workflows (pass 1)

**Scope:** U9 slice only — `git diff 7400a72d..HEAD` (the per-project workflow
authoring work: U9-1…U9-6). The earlier U7/U8/U10 stack carried on this branch
was reviewed in its own stages and is excluded.

**Intent:** Discover/validate/register per-project user-authored workflow
descriptors (`<hive_state_path>/workflows/<id>.yml`), resolve them at every
consumption point, inject `instruction:` files into agent prompts, honor
per-stage `permissions:`, and scaffold a blank flow via `hive workflow new`.

**Reviewers (report-only):** correctness, reliability, security, adversarial,
cli-readiness, maintainability, testing, project-standards. Plan-deferred items
(CLI `--help`/`WORKFLOW_VOCABULARY` staleness; untrusted/shared-descriptor
governance & OS sandbox) were dropped per the plan's Scope Boundaries/Risks.

**Verdict:** Not ready — one High blast-radius defect that also misses a stated
U9-2 requirement, plus several Medium reliability/contract gaps.

## High
- [ ] One malformed `.hive-state/workflows/*.yml` bricks unrelated tasks/commands for the whole project (`lib/hive/workflows/loader.rb:19`, `lib/hive/workflows/project.rb:9`, `lib/hive/task.rb:50`): `load_dir` aborts on the first bad descriptor and `Project.load!` runs eagerly from `Task#initialize` and every boundary (text `status`, `new --workflow` even `coding`, `drop`, `task_resolver`, daemon `hive run`) with no per-file isolation or per-project rescue, so a single authoring typo raises `ConfigError` for coding tasks that never touch the new workflow — directly contradicting the plan's U9-2 loader requirement ("good one loads, broken one reported with its path"), which `test/unit/workflows/loader_test.rb:39` codifies the abort for instead of catching.

## Medium
- [ ] `Project.load!` is not exception-safe (`lib/hive/workflows/project.rb:13`): it clears the project overlay via `reset_project_registrations!` before loading but sets `@active_root` only on success, so a rescued failure (JSON `status`'s `project_payload_or_degraded`, or a long-running TUI) leaves `@active_root` pinned to a root whose overlay is now empty, and the next same-root `load!` short-circuits and serves a stale/empty registry.
- [ ] Cross-project stage-ref resolution can raise or misresolve (`lib/hive/task_resolver.rb:144`, `lib/hive/commands/drop.rb:186`): deferring `resolve_stage_ref_across_workflows` into the per-project loop makes a `--from`/`--stage` that names a project-specific stage raise `UnknownStage` while scanning an earlier project that lacks it, and the `drop` WrongStage message computes "expected" against the last-loaded project rather than the project where the slug was actually matched.
- [ ] Instruction file is read unguarded at run time (`lib/hive/stages/agent.rb:67` `File.read(stage.instruction)`): if the `.md` is renamed/deleted after parse (a normal authoring edit), the stage dies with an uncaught `Errno::ENOENT` and no error marker — unlike sibling `prior_artifacts`, which rescues `SystemCallError` for the same race.
- [ ] `loaded_workflows` caches parsed descriptors per root for the whole process with no mtime/size invalidation (`lib/hive/workflows/project.rb:35`), unlike Task's `project_default_workflow_cache` which stamps on (mtime,size); a long-running in-process consumer (TUI) never sees an edited/added descriptor until restart, since `reset!` is only called by the scaffold command.
- [ ] Scaffold is non-atomic (`lib/hive/commands/workflow.rb:73`): `commit_scaffold!` runs *after* the `begin/rescue` rollback block (lines 65–71), so a commit failure (`GitError`/`ConcurrentRunError`) leaves orphan descriptor + instruction files on disk that make the next `hive workflow new <id>` fail in `refuse_overwrite!` — a retryable command that isn't safely retryable.
- [ ] `hive workflow new` failure paths bypass the `--json` error envelope (`lib/hive/commands/workflow.rb:47`): `call` rescues only `UsageError`/`ConfigError`/`GitError`, so the commit-lock `ConcurrentRunError` (TEMPFAIL/75) and Thor-arity errors emit plain stderr instead of a JSON payload, hiding the retryable-vs-terminal distinction from agent callers.
- [ ] `stages_for_project` is duplicated byte-for-byte across `lib/hive/commands/drop.rb:207` and `lib/hive/task_resolver.rb:144` (same `load!` + resolve state protocol), inviting silent drift between the two cross-project resolution paths.
- [ ] A user workflow whose last stage is `kind: agent` yields undroppable tasks (`lib/hive/workflow.rb:33`, `lib/hive/workflows.rb:157`): `validate_structure!` does not require a terminal final stage, yet `all_terminal_stage_dirs` treats every workflow's last stage dir as the archive guard, so `drop` refuses a task sitting at a short flow's only working stage with no workaround.

## Nit
- [ ] `docs/workflows.md` says `instruction:` must be a readable `.md` file, but the parser (`lib/hive/workflows/descriptor_parser.rb:114`) only checks `File.file?`/`File.readable?` with no extension enforcement — tighten the doc or the check (the `wiki/` page already says only "readable file").
- [ ] `permissions:` (and `skill`/`instruction`) on a `terminal`/inert stage is validated, deep-frozen, and stored but never consulted (only the agent runner reads them — `lib/hive/workflows/descriptor_parser.rb:119`, `lib/hive/stages/agent.rb:23`): a silent no-op config trap; reject or warn at parse time.
- [ ] The `all_stage_dirs` memo comment (`lib/hive/workflows.rb:139`) still claims the union "never changes after boot," which this slice falsified by adding `reset_union_cache!` for post-boot project registrations — update the comment to describe the new invalidation contract.
- [ ] The registry's runtime overlay tier (`registrations`, `register!(project: false)`, `reset_runtime_registrations!` in `lib/hive/workflows/registry.rb:87`) is referenced only by tests; production uses only `project_registrations` — test-shaped abstraction living in production code.
- [ ] Instruction bodies are injected into the prompt unescaped (`templates/agent_prompt.md.erb`), with no size cap (unlike `prior_artifacts`' clamp) and via a path that can resolve outside the project: acceptable under the plan's owner-trust model (A3), but the trust boundary must be revisited before any shared/untrusted descriptor support (plan Risks).
- [ ] `hive workflow new` JSON payloads omit `schema`/`schema_version`/`error_kind` that the gem-wide `Hive::ErrorEnvelope.build` provides (`lib/hive/commands/workflow.rb:170`), so agents routing by those keys can't treat this command's output uniformly with the rest of the CLI.

ivankuznetsov:

# PR Review (pr-review-toolkit) — u9-per-project-workflows (pass 1)

**Scope:** U9 slice only — `git diff 7400a72d..HEAD` (per-project workflow
authoring; the U7/U8/U10 stack carried on this branch is reviewed elsewhere).

**Personas run (union below):** code-reviewer, silent-failure-hunter,
type-design-analyzer, comment-analyzer, pr-test-analyzer, code-simplifier.
Plan-deferred items (CLI `--help`/`WORKFLOW_VOCABULARY` staleness; untrusted /
shared-descriptor governance & OS sandbox) dropped per the plan's Scope
Boundaries/Risks. `[verified]` = reproduced directly during this review.

## High
- [ ] [code-reviewer] One malformed `<hive_state_path>/workflows/*.yml` breaks `Task.new` for every task in the project — including built-in coding tasks (`lib/hive/task.rb:50` eager `resolve_workflow` → `Project.load!`; rescue at :138 only catches `UnknownWorkflow`): `load_dir` raises `ConfigError`, propagating through text `status`, `drop`, `new --workflow` (even `coding`), and `task_resolver`, while JSON status degrades but text status aborts the whole fleet — contradicts the plan's U9-2 loader intent ("good one loads, broken one reported").
- [ ] [code-reviewer] Cross-project overlay corruption: `Project.load!` (`lib/hive/workflows/project.rb:13`) clears the project overlay before loading but sets `@active_root` only on success (:26), so a load that raises mid-way (bad descriptor or built-in-id collision in a later project) leaves `@active_root` stale while the overlay is empty, and the next same-root `load!` early-returns and never re-registers — that project's workflows silently vanish for the rest of a long-lived TUI session `[verified]`.

## Medium
- [ ] [silent-failure] `render_project` (text mode) calls `Project.load!` unguarded inside the per-project loop (`lib/hive/commands/status.rb:407`) while the JSON path wraps each project in `project_payload_or_degraded` — asymmetric handling means one bad descriptor blanks text `status` for all projects.
- [ ] [silent-failure] `workflow_dir_for` silently falls back to the default `.hive-state/workflows` when `config.yml` is broken (`lib/hive/workflows/project.rb:39-43`): a project with a custom `hive_state_path` then loads descriptors from the wrong directory with no warning, so the user's workflows go unregistered and later surface as a confusing "unknown workflow".
- [ ] [silent-failure] Internal asymmetry in `Project.load!` (`lib/hive/workflows/project.rb:14-16`): `workflow_dir_for` swallows `ConfigError`/`Psych`/`SystemCallError`/`IOError` (silent fallback) while `load_dir` propagates the same `ConfigError` class loudly — opposite handling of one error class depending only on which file is broken.
- [ ] [silent-failure] `WorkflowSelection.fetch!` now calls `Project.load!` (`lib/hive/workflow_selection.rb:8`), defeating `new.rb`'s stated invariant (`lib/hive/commands/new.rb:173-175`) that a fully-specified `--workflow` must not be blocked by a corrupt project config — a corrupt sibling descriptor now blocks even `hive new --workflow good "idea"`.
- [ ] [silent-failure] Unguarded `File.read(stage.instruction)` in `render_prompt` (`lib/hive/stages/agent.rb:67`): TOCTOU — if the instruction `.md` is deleted/chmod'd after parse, the stage dies with a raw `Errno::ENOENT` and no attributed `:error` marker, unlike the sibling `prior_artifacts` which rescues `SystemCallError`.
- [ ] [silent-failure] Scaffold is non-atomic: `commit_scaffold!` runs outside the rollback `begin/rescue` (`lib/hive/commands/workflow.rb:65-73`), so a `GitError`/`ConcurrentRunError` commit failure leaves orphan descriptor+instruction files that make the next `hive workflow new <id>` fail in `refuse_overwrite!`.
- [ ] [silent-failure] `hive workflow new` failure paths bypass the `--json` error envelope (`lib/hive/commands/workflow.rb:47`): `call` rescues only `UsageError`/`ConfigError`/`GitError`, so the commit-lock `ConcurrentRunError` (TEMPFAIL) and Thor-arity errors emit plain stderr instead of a JSON payload.
- [ ] [silent-failure] Cross-project loops call `Project.load!` per project with no per-project rescue (`lib/hive/commands/drop.rb:120,176`, `lib/hive/task_resolver.rb:144`), so one project's malformed/colliding descriptor aborts resolution for tasks living in healthy projects.
- [ ] [code-reviewer] `registry_test.rb` errors when run in isolation — `NoMethodError: undefined method 'reset_union_cache!' for Hive::Workflows` (`test/support/workflow_helpers.rb:170` delegates to it, but `registry_test.rb` requires only `hive/workflows/registry`); full `rake test` is green only by load-order luck `[verified]`.
- [ ] [type-design] `validate_structure!` does not require a terminal/inert final stage (`lib/hive/workflow.rb:33-66`); a user workflow ending in `kind: agent` is constructible, and since `all_terminal_stage_dirs` treats every workflow's last stage dir as the archive guard, tasks at that final stage become undroppable.
- [ ] [type-design] `Stage` permits meaningless field combinations (`lib/hive/workflow.rb:146-178`): the "agent stage has exactly one of skill/instruction" and "permissions/instruction only meaningful for agent stages" invariants live only in `DescriptorParser`, not in `Stage.new`/`validate_structure!`, so any non-parser construction site can build an invalid Stage and a `permissions:`/`instruction:` on a terminal stage is silently stored and ignored.
- [ ] [type-design] `permissions` is stored as a raw frozen Hash/String on the Stage and re-parsed by `PermissionScope.resolve` on every run rather than carried as a validated spec object (`lib/hive/workflows/descriptor_parser.rb:68-72`), so "validated" isn't expressed by the type and validation/resolution can drift.
- [ ] [pr-test] Descriptor cache `loaded_workflows` has no mtime/stat invalidation (`lib/hive/workflows/project.rb:35`); a long-running in-process consumer (TUI) never re-reads edited/added descriptors until `reset!`/restart (the sibling `project_default_workflow_cache` stamps on mtime/size) — untested staleness.
- [ ] [pr-test] `loader_test.rb:39` `test_load_dir_reports_broken_descriptor_path` asserts only that `ConfigError` propagates and never that the good descriptor still loaded — it codifies the all-or-nothing abort, the opposite of the plan's U9-2 "good one loads, broken one reported".
- [ ] [comments] `workflows.rb:139-143` memo comment ("the registry is frozen at load, so the union … never changes after boot") now contradicts the code — this slice added `reset_union_cache!` precisely because project registrations mutate the union at runtime.
- [ ] [simplifier] `stages_for_project` is byte-for-byte duplicated in `lib/hive/commands/drop.rb:207` and `lib/hive/task_resolver.rb:144` — extract to a shared `Hive::Workflows` function so the per-project resolution rule lives in one place.
- [ ] [simplifier] Registry's runtime overlay tier (`registrations`, `register!(project: false)`, `reset_runtime_registrations!` in `lib/hive/workflows/registry.rb:67-114`) is referenced only by tests; production uses only `project_registrations` — collapse to one overlay or document why two tiers exist (also flagged by type-design and comments).

## Nit
- [ ] [code-reviewer] `instruction:` path has no containment guard (`lib/hive/workflows/descriptor_parser.rb:113-116`): `instruction: ../../etc/passwd` resolves and its body is read into the agent prompt — low-risk under the plan's owner-trust model (A3, governance deferred), but a path-confinement check would harden it.
- [ ] [silent-failure] `rescue ArgumentError` in `parse_hash` (`lib/hive/workflows/descriptor_parser.rb:39`) wraps the whole body, so a future wrong-kwarg bug in `Stage.new`/`Workflow.new` would be mislabeled to the user as a descriptor error — narrow it to the two constructor calls.
- [ ] [type-design] `valid_names(project_root: nil)` silently skips loading project descriptors when root is nil (`lib/hive/workflow_selection.rb:21-24`); the rescue path in `fetch!` calls bare `valid_names`, so an error message can omit the very project workflow the user named.
- [ ] [type-design] `next_stage_after` collapses "unknown name" and "terminal stage" into one nil (`lib/hive/workflow.rb:79-94`); with user-authored descriptors now in play, the unvalidated-name arm is more reachable than the comment's "unreachable today" assumes.
- [ ] [comments] `docs/workflows.md:71` says `instruction:` must point to a readable `.md` file, but the parser only checks `File.file?`/`File.readable?` with no extension enforcement — tighten the doc or the check.
- [ ] [comments] The registry's three-tier precedence (project shadows runtime shadows built-in) in `lib/hive/workflows/registry.rb:67-114` is load-bearing for collision/override behavior but entirely undocumented.
- [ ] [comments] `Project.load!`'s memoization and the "must call `reset!` to pick up newly written descriptors" coupling (`lib/hive/workflows/project.rb:9-27`) is non-obvious and uncommented.
- [ ] [simplifier] `Project.load!` invalidates the union cache up to N+2 times per call; the trailing explicit `reset_union_cache!` (`lib/hive/workflows/project.rb:25`) is dead in every path and re-spells the `respond_to?` guard the registry already encapsulates.
- [ ] [simplifier] Third spelling of "the workflows dir" across `Workflow#workflow_dir`, `Loader.workflow_dir`, and `Project.workflow_dir_for` (`lib/hive/commands/workflow.rb:113-116`, `lib/hive/workflows/loader.rb:13-17`) — consolidate.
- [ ] [tests] `project_test.rb:32` memoization test asserts an exact stub call count (`== 1`), pinning the current caching strategy rather than the observable "no reparse on repeat" contract — it would break under a legitimate mtime-revalidation refactor (the same fix the staleness finding needs).
- [ ] [comments] `drop.rb:117-119` comment still references the coding `Hive::Stages::DIRS` and the now-duplicated `stages_for_project`; a note on the intentional duplication would help the next reader.

ivankuznetsov:

## High
- [ ] Project workflow validation fails open instead of failing fast (`lib/hive/workflows/loader.rb:35`, `lib/hive/workflows/project.rb:47`): malformed descriptors and built-in id collisions are only warned and skipped, so `hive new/init --workflow` can report an unrelated unknown workflow or run the built-in `coding` workflow instead of rejecting the owner-authored descriptor with `Hive::ConfigError` as required by U9-2/U9-3.

## Medium

## Nit

ivankuznetsov:

# Code Review — u9-per-project-workflows (pass 2)

**Scope:** U9 slice only — `git diff 7400a72d..HEAD` (commits U9-1…U9-6 plus the
pass-1 fix commits). The U7/U8/U10 stack carried below `7400a72d` was reviewed in
its own stages and is excluded.

**Intent:** Discover/validate/register per-project owner-authored workflow
descriptors (`<hive_state_path>/workflows/<id>.yml`), resolve them at every
consumption boundary (`Task`, `WorkflowSelection`, `init`, `new`, `drop`,
`status`, `task_resolver`, `approve`), inject per-stage `instruction:` `.md`
bodies into the agent prompt, honor per-stage `permissions:` (U10), and scaffold
a blank `inbox→work→done` flow via `hive workflow new`. Owner-trust model (A3):
instructions are trusted prompts, no sandbox. Coding/content must be
behaviorally untouched.

**Reviewers (report-only, parallel personas):** correctness, security,
adversarial (Opus tier); testing, maintainability, cli-readiness (mid tier).
High-stakes findings verified directly against the code by the orchestrator.
Plan-deferred items (CLI `--help`/`WORKFLOW_VOCABULARY` staleness;
untrusted/shared-descriptor governance & OS sandbox; live-reload of edited
descriptors; the instruction-injection trust boundary accepted under A3; richer
authoring UX) were dropped per the plan's Scope Boundaries/Risks.

**Verdict:** Not ready — three High findings (one new safety bug, one new
concurrency defect on a production path, one plan-requirement miss), plus two
Medium contract/operability gaps.

## High
- [ ] Unvalidated descriptor `state_file` allows path traversal out of the task folder (`lib/hive/workflows/descriptor_parser.rb:117` → `lib/hive/stages/agent.rb:20`, `lib/hive/markers.rb:86/219`): `id` and `name` are `SAFE_SLUG`-guarded but `state_file` is only `required_string`; it flows unmodified into `File.join(task.folder, stage.state_file)` and into `Markers.set` → `ensure_dir` (`FileUtils.mkdir_p`) + `write_atomic`, so a descriptor with `state_file: ../../escape.md` (even an authoring *typo*, not just malice) makes hive itself mkdir/write marker files outside the task folder — distinct from the plan-accepted A3 instruction boundary (that accepts trusting the instruction *body* and the read-only instruction path); built-in descriptors all use bare basenames, so basename-only is the intended-but-unenforced contract. 
- [ ] U9 wired unsynchronized global `Project.load!` overlay mutation onto the concurrent web read path (`lib/hive/commands/status.rb:182`, `lib/hive/workflows/project.rb:16`, `lib/hive/workflows/registry.rb:99`, `lib/hive/web/status_feed.rb:39/104`): `json_payload` now calls `Project.load!(path)` per project, mutating module-level `@active_root`/`@loaded_workflows`/`@project_registrations` + the union caches with no lock, and StatusFeed runs `json_payload` on both the background poller thread and the per-request `snapshot` path — so a concurrent `load!(B)` calls `reset_project_registrations!` (clearing A's overlay) while the other thread is mid-render of project A, making A's custom-workflow rows resolve against B's overlay → `UnknownWorkflow` → `project_payload_or_degraded` blanks the project; custom-workflow tasks intermittently vanish/degrade in the dashboard under normal multi-tab/multi-request load (read-path, self-healing next tick, but a genuine data race newly introduced by this slice). 
- [ ] Fail-open loader/registration contradicts U9-3's explicit "reject at `register!` with `ConfigError` naming the file — never silently shadow the coding pipeline" (`lib/hive/workflows/loader.rb:38`, `lib/hive/workflows/project.rb:47`): a malformed `my-flow.yml` and a built-in-colliding `coding.yml` are both rescued+warned-on-stderr and dropped from the registry, so a later `hive new --workflow my-flow` raises a misleading "unknown workflow" (the parse error is buried in a one-shot stderr line far upstream) and `hive new --workflow coding` silently runs the *built-in* coding while the owner's descriptor is ignored — load-time isolation (don't brick siblings) is correct, but the explicitly-requested broken/colliding id must surface its real error at the resolution boundary to satisfy the plan. 

## Medium
- [ ] `hive workflow new --json` disk-write errors escape the JSON error envelope (`lib/hive/commands/workflow.rb:48`, `bin/hive:185-192`): `call`'s rescue list is `UsageError, ConfigError, GitError, ConcurrentRunError`, so a `SystemCallError`/`IOError` from `write_scaffold!`'s `FileUtils.mkdir_p`/`File.write` is re-raised by the inner `rescue StandardError` and propagates past `call` to bin/hive, which only catches `Errno::EPIPE`/`Thor::Error`/`Hive::Error` — yielding a raw Ruby backtrace and exit 1 instead of the schema'd payload, breaking the `--json` agent contract exactly as the pass-1 `ConcurrentRunError` fix intended to prevent; widen the rescue to `SystemCallError, IOError`. 
- [ ] Collision/parse-skip warning is re-emitted on every `Project.load!`, not once per process (`lib/hive/workflows/project.rb:30-33,47-53`): `loaded_workflows` caches only the parse result, so `register_descriptor` re-runs on every `load!` that actually swaps `@active_root`; a project containing a built-in-colliding descriptor (`coding.yml`) therefore writes `hive: skipping … collides with registered workflow :coding` to stderr/daemon.log on every status tick in a multi-project daemon that alternates active roots — unbounded log flood. Cache the post-registration skip decision (or dedup the warn) per source_path. 

## Nit
- [ ] `Registry#workflows` allocates two fresh merged hashes on every `fetch`/`all`/`ids` call (`lib/hive/workflows/registry.rb:115-117`) — called repeatedly during a cross-project status scan and inside `register!`'s own `key?` check; memoize the merge in an ivar and clear it in `reset_union_cache`. 
- [ ] `Loader.load(project_root)` is dead/test-only production code (`lib/hive/workflows/loader.rb:9`): production goes through `Project.load! → load_dir`; only `loader_test.rb:9,35` call `load`. Remove it (and point the tests at `load_dir`) or document it as the no-fallback variant. 
- [ ] `validate_agent_instruction!` uses `skill.nil? ^ instruction.nil?` (`lib/hive/workflows/descriptor_parser.rb:199`): XOR on `nil?` results reads obscurely and is only correct because `optional_string` pre-normalizes blanks to nil; replace with `[skill, instruction].count { !_1.nil? } == 1` for legibility. 
- [ ] Bare `hive workflow` (no subcommand) with `--json` bypasses the JSON envelope (`bin/hive:19`): `JSON_USAGE_ERROR_CONTRACTS` registers `run/approve/drop/…` but not `workflow`, so a Thor-arity error on the new command emits plain stderr; add `"workflow" => { schema: "hive-workflow-new", error_kind: "usage" }`. (`hive workflow new` with no id is already enveloped via the command's own `UsageError`.) 
- [ ] `Registry.reset_union_cache`'s `respond_to?(:reset_union_cache!)` guard (`lib/hive/workflows/registry.rb:134`) is unreachable-false at every runtime call site (the require chain always defines it before any `register!`); collapse to a direct call now that the require-cycle analysis it guarded is settled. 
- [ ] `UsageError#value` (the rejected/colliding id) is not surfaced into the `--json` error payload (`lib/hive/commands/workflow.rb`, `Hive::Schemas::ErrorEnvelope.build`): agents must regex `message` to recover the bad id; either teach `ErrorEnvelope.build` to extract `error.value` generically or pass it via `extras`. 
- [ ] A hand-authored workflow whose **first** stage is `kind: agent` gets a `<!-- WAITING -->` initial marker, so `hive status` classifies the fresh task as "needs input" though `hive run` works fine (`lib/hive/task_action.rb`); cosmetic — the blank scaffold's inert first stage avoids it, but a custom agent-first flow looks stuck when it isn't. 
- [ ] `Project.@loaded_workflows` accumulates one parsed-descriptor entry per project-root for the daemon's whole lifetime (`lib/hive/workflows/project.rb:62`), evicted only by an all-or-nothing `reset!`; harmless at today's K projects, but document the bounded-K assumption at the definition site so a future live-reload/lazy-load author sees the `@loaded_workflows`↔`@active_root` coupling. 
- [ ] Test-assertion strengthening (line-covered by the 100% gate but behavior weakly pinned): the loader skip path asserts the file path is in stderr but not the `hive: skipping …` message contract (`test/unit/workflows/loader_test.rb:73`); the scoped-`permissions:` test asserts `.frozen?` but not the stored `preset`/`tools`/`dirs` values (`test/unit/workflows/descriptor_parser_test.rb:431`); the `workflow_dir_for` fallback warning is unasserted (`test/unit/workflows/project_test.rb:95`); and the `instruction_unreadable` marker, `stages_for_project` unknown/ambiguous-stage branches, and the git-failure→rollback path are only covered transitively.

ivankuznetsov:

# PR Review Toolkit — u9-per-project-workflows (pass 2)

**Scope:** U9 slice only — `git diff 7400a72d..HEAD` (per-project workflow
descriptor authoring: parser/loader/registry/project, agent runner
`instruction:`/`permissions:`, `hive workflow new` scaffold). The U7/U8/U10 stack
below `7400a72d` was reviewed in its own stages and is excluded.

**Personas run (union captured below):** code-reviewer, silent-failure-hunter,
type-design-analyzer, comment-analyzer, pr-test-analyzer, code-simplifier.
Plan-deferred items (CLI `--help`/`WORKFLOW_VOCABULARY` staleness;
untrusted/shared-descriptor governance & OS sandbox; live-reload; richer
authoring UX; the instruction-injection trust boundary accepted under A3) were
dropped per the plan's Scope Boundaries/Risks.

**Note:** `code-reviewer` found no bug above its clearing bar (it chased and
cleared several candidates: frozen-permissions `FrozenError`, coding stages
reaching the generic runner, inert-entry-stage stranding, approve-before-load,
auto-verb on terminal stage, scaffold ordering). `code-simplifier` found only
Suggestion-level items. The substantive findings come from
silent-failure-hunter, type-design-analyzer, comment-analyzer, and
pr-test-analyzer — several cross-corroborated by the orchestrator's own code
reads.

## High
- [ ] [pr-test-analyzer] The plan-mandated 100% line-coverage gate is currently UNMET — a fresh `bundle exec rake coverage` run reports **99.95%** (26089/26102) with every uncovered line a U9-introduced error branch: the agent `instruction_unreadable` marker path (`lib/hive/stages/agent.rb:33,99-108`), `hive workflow new`'s `error_kind_for` Git/Concurrent/Config/else arms + the git-failure rollback (`lib/hive/commands/workflow.rb:208-211,78-82`), the text-`status` per-project rescue (`lib/hive/commands/status.rb:126,128`), and drop's matched-project `expected_stage` degrade (`lib/hive/commands/drop.rb:210`): independently corroborated — orchestrator grep confirms no test makes the instruction read raise and none drives the workflow Git/Concurrent error_kind, so the pass-1 fix commits added error branches without covering tests; the runner must re-run `rake coverage` and cover these (or the gate the plan restates in every U9 unit's Verification is failing). 
- [ ] [type-design-analyzer] Descriptor `state_file` is an unconstrained path-shaped String — no basename/traversal guard while `id`/`name` are `SAFE_SLUG`-validated (`lib/hive/workflows/descriptor_parser.rb:117`): it flows verbatim into `File.join(task.folder, stage.state_file)` (`lib/hive/stages/agent.rb:20`, `lib/hive/task.rb:59`) and on to `Markers.set`→`mkdir_p`/`write_atomic`, so `state_file: ../../escape.md` (even an authoring typo) makes hive write marker/output files outside the task folder; built-ins use bare basenames, so basename-only is the intended-but-unenforced contract — a "no leading `/`, no `..` segment" guard closes it without breaking the dotted-basename convention. 
- [ ] [silent-failure-hunter] `hive workflow new --json` disk-write errors escape the JSON error envelope (`lib/hive/commands/workflow.rb:48-58`, `bin/hive:185-192`): the inner `rescue StandardError` re-raises and `call`'s rescue list omits `SystemCallError`/`IOError`, so an `ENOSPC`/`EROFS`/`EACCES` from `write_scaffold!` (or a missing packaged template) propagates past bin/hive's `Thor::Error`/`Hive::Error`-only handler — the persona reproduced it empirically (read-only workflows dir → `ESCAPED UNCAUGHT: Errno::EACCES … dir_s_mkdir`, exit 1, raw backtrace) instead of the promised `hive-workflow-new` error payload; add a `SystemCallError, IOError` arm. 

## Medium
- [ ] [silent-failure-hunter] Built-in-id collision warning re-emits unboundedly in long-lived processes (`lib/hive/workflows/project.rb:41-54`): `register_descriptor`'s `warn "hive: skipping …"` runs on every `load!` body (gated only by `@active_root`, NOT by the `loaded_workflows` cache, unlike the loader's cache-bounded malformed-descriptor warn), so a daemon/TUI flipping `@active_root` between projects each `status` tick re-prints the same `coding.yml` collision breadcrumb forever — drowning a genuinely new collision in repeats. 
- [ ] [silent-failure-hunter] The instruction-error handler's own `Markers.set` is unrescued (`lib/hive/stages/agent.rb:97-110`): `instruction_error_result` stamps the attributed `:error` marker via `Markers.set(output_path,…)`, which does its own `File.write`/`mkdir` under `task.folder`; if the instruction is unreadable because the *task folder itself* is gone/unwritable, `Markers.set` raises a `SystemCallError` that escapes the handler — relocating the exact raw-`Errno`-with-no-marker death the handler was added to prevent. 
- [ ] [comment-analyzer] Stale invariant re-introduced in a test-helper comment — the same one pass-1 removed from source (`test/support/workflow_helpers.rb:158-162`): it claims "the registry is frozen in production, so the cache is permanent there" and "This helper is the ONE place the registry mutates," both false after U9 (`Project.register_descriptor` mutates the registry in production via `Registry.register!(project: true)` and the union cache is explicitly invalidated) — a maintainer trusting it would mis-reason about the production registration path. 
- [ ] [comment-analyzer] `build_workflow` comment attributes the non-terminal-last-stage check to the wrong method (`lib/hive/workflows/descriptor_parser.rb:65-68`): it lists "a non-terminal last stage" among the rules `Hive::Workflow#validate_structure!` raises `ArgumentError` for, but that rule is enforced by the parser's own `validate_terminal_last_stage!` (`descriptor_parser.rb:52`, raising `ConfigError` before `build_workflow`); a maintainer trusting the comment could wrongly drop the parser's terminal guard believing `validate_structure!` covers it. 
- [ ] [pr-test-analyzer] Descriptor `permissions:` round-trip is line-covered but not value-asserted (`test/unit/workflows/descriptor_parser_test.rb:43-44`): the valid-descriptor test declares `permissions: {preset: scoped, tools: [Read, Write], dirs: [../extra]}` yet asserts only `.frozen?` on the hash and its `tools` — a regression in `stringify_hash`/`deep_freeze` that dropped `dirs:` or mangled `tools` would still pass (the corrupted hash is still frozen), and the structured `scoped` map's resolved behavior is pinned nowhere (agent_test only pins the `read-only` preset string). 

## Nit
- [ ] [silent-failure-hunter] The text-`status` `rescue StandardError` (`lib/hive/commands/status.rb:121-128`) swallows a wide surface (`Project.load!` + `collect_rows` + `annotate_actions`) so a genuine `NoMethodError`/logic bug degrades to a benign `"<name>: failed to load (…)"`; the stderr `warn` includes `#{e.class}` but the user-facing `puts` omits it — consider narrowing to the load/IO-shaped errors actually expected. 
- [ ] [silent-failure-hunter] `rollback_scaffold` is filesystem-only, not git-state-aware (`lib/hive/commands/workflow.rb:79-81`): `rm_f`/`rm_rf` swallow their own errors and unwind only the working-tree files, not any staging side-effects of a partially-run `commit_scaffold!`; harmless today (retry's `refuse_overwrite!` still works) but worth a note. 
- [ ] [type-design-analyzer] The skill-XOR-instruction invariant lives only in the parser, never on `Hive::Workflow::Stage` (`lib/hive/workflow.rb:146-178`): `Stage.new(kind: :agent, skill:"/x", instruction:"/y")` or `Stage.new(kind: :inert, permissions:{…})` constructs cleanly, so the type's construction-time guarantee is weaker than `Workflow`'s — the intentional U9-1 "authoring rules stay in the parser" tradeoff, flagged as the type-design consequence the agent runner trusts. 
- [ ] [type-design-analyzer] `permissions:` stores the raw deep-frozen YAML Hash on `Stage`, not the validated `PermissionScope::Scope` (`lib/hive/workflows/descriptor_parser.rb:155-163`): `validate!` builds and discards the rigorously-typed `Scope`, storing the loose value and re-resolving at run time — the field's validity is a property of *how it was built* (the parser), not the type; previously triaged as a deferred type-design suggestion, noted for continuity. 
- [ ] [type-design-analyzer] `AdvanceVerb#name` is an unconstrained String though it becomes a CLI subcommand token (`lib/hive/workflow.rb:180-184`, `descriptor_parser.rb:165-177`) — lower-stakes than `state_file` (not a filesystem path) but another field looser than its domain (a command token); the first-stage-no-verb rule correctly lives on the type. 
- [ ] [type-design-analyzer / code-simplifier] `Registry#workflows` rebuilds `WORKFLOWS.merge(registrations).merge(project_registrations)` un-memoized on every `fetch`/`all`/`ids` call (`lib/hive/workflows/registry.rb:113-117`), and its last-wins precedence is dead code because `register!` forbids any id already in a lower tier; memoize the merge and clear it in `reset_union_cache`. 
- [ ] [comment-analyzer] `Project` comment says `Commands::Workflow#call!` "resets before re-resolving" (`lib/hive/workflows/project.rb:12-15`), but `call!` calls `Project.reset!` once and never re-resolves in-process — the reset exists so a *later separate* invocation discovers the new descriptor; the "before re-resolving" tail points at a non-existent step. 
- [ ] [comment-analyzer] `Registry.default` comment cites `stages.rb:12 (DIRS)` but `DIRS` is now at `lib/hive/stages.rb:16` (`registry.rb:49-54`) — the hunk is unchanged by this slice (below the review boundary), flagged only for a future cleanup pass. 
- [ ] [pr-test-analyzer] `Project.load!` mid-load exception-safety (drop `@active_root` to nil before load, re-set only after a clean load) is documented at length but asserted by no test (`lib/hive/workflows/project.rb:16-39`); add a stub that raises mid-load and assert a subsequent same-root `load!` re-attempts rather than serving a stale/empty registry. 
- [ ] [pr-test-analyzer] `workflow_dir_for` fallback is line-hit but its contract unasserted (`test/unit/workflows/project_test.rb:95`): the test checks `Registry.ids` but never `capture_io`s the `"could not read hive_state_path … falling back to <fallback>"` warning nor proves descriptors load from the default dir (assertion passes trivially because the fallback dir is empty). 
- [ ] [pr-test-analyzer] `stages_for_project`'s ambiguous-stage re-raise is exercised only at the underlying `resolve_stage_ref_across_workflows` level, never through the new method with two overlay workflows colliding on a short name (`lib/hive/workflows.rb:218-228`); and `registry.rb:88,113` branch directions are line-covered-but-branch-uncovered (the gate is line-only, so these pass silently). 
- [ ] [pr-test-analyzer] The loader malformed-descriptor fixture is degenerate — `id: [` (a YAML syntax error) at `test/unit/workflows/loader_test.rb:65` exercises only the `Psych::Exception` rescue, not a semantically-invalid-but-parseable sibling (id/filename mismatch, non-terminal last stage) that routes through `DescriptorParser`→`ConfigError`, the more realistic authoring mistake. 
- [ ] [code-simplifier] `Loader.load(project_root)` is dead/test-only production surface (`lib/hive/workflows/loader.rb:9-11`) — only `loader_test.rb:9,35` call it; production composes `workflow_dir` + `load_dir` separately. Remove it and point the test at the two public methods. 
- [ ] [code-simplifier] The thin per-call-site `stages_for_project` wrappers in `lib/hive/commands/drop.rb:231-238` and `lib/hive/task_resolver.rb:144-150` are near-identical 1-line forwarders to `Hive::Workflows.stages_for_project`; callers could invoke the shared module method directly (or keep one wrapper) to drop the twin indirection. 
- [ ] [code-simplifier] `return if skill.nil? ^ instruction.nil?` (`lib/hive/workflows/descriptor_parser.rb:199`) — the "exactly one of skill/instruction" check reads clearer as `return if [skill, instruction].compact.one?`. 
- [ ] [code-simplifier] In drop's WrongStage path, `Project.load!` runs per project in the scan loop (`drop.rb:191`) then again on the matched project (`:205`); correct (it short-circuits on unchanged `@active_root`) but reads as a redundant twin — a one-line "no-op when matched project scanned last" note would mark it intentional. 
- [ ] [code-simplifier] The `permission_kwargs = stage.permissions.nil? ? {} : {…}` local + `**` splat (`lib/hive/stages/agent.rb`) could pass `explicit_permission_spec: stage.permissions` (nil when absent) directly if the callee tolerates a nil default, removing the conditional allocation.

ivankuznetsov:

## High
- [ ] Project workflow resolution is not held under the project overlay lock (`lib/hive/task.rb:132`): `Task#resolve_workflow` calls `Project.load!` and then fetches from the process-global registry after the lock is released, so concurrent web/status/daemon threads can swap the overlay and make custom-workflow tasks resolve as unknown or against the wrong descriptor.
- [ ] Cross-project stage scans use a mutable overlay outside the load lock (`lib/hive/workflows.rb:219`): `stages_for_project` loads one project and then reads `all_stage_dirs` or resolves the stage filter after another thread may have loaded a different project, so run/approve/drop slug scans can miss or misclassify project-authored stages.
- [ ] Drop can hard-delete archived custom-workflow tasks after an overlay swap (`lib/hive/commands/drop.rb:156`): the matched project is reloaded but `archive_stage_dirs` is computed later from the global overlay, so a concurrent load can hide that workflow's terminal dir and bypass the archive guard.

## Medium
- [ ] Parser accepts nested entry state files that `hive new` cannot create (`lib/hive/commands/new.rb:135`): `DescriptorParser` allows values like `sub/idea.md`, but task creation writes the entry state file without creating its parent directory, so a descriptor that passes validation fails before the first task is captured.
- [ ] Task artifact pages fetch project workflows from the wrong overlay (`web/app/controllers/tasks_controller.rb:206`): `generic_artifact_order` reads `Registry.fetch(row["workflow"])` without loading or synchronizing the row's project, so project-authored workflows can render fallback or unrelated artifacts whenever another project's overlay is active.
- [ ] Recovery retry classification does not load the row's project workflow (`lib/hive/bot/handlers/recovery_sequence.rb:127`): terminal or inert custom stages fall through to `hive run` when the active overlay lacks that descriptor, queueing retries that have no runner.

## Nit

ivankuznetsov:

# Code Review — u9-per-project-workflows (pass 3)

**Scope:** U9 slice — `git diff origin/main..HEAD`, focused on the U9 commits
(`ff634a1e`…`dec528f5`) layered on the U7/U8/U10 stack (reviewed in its own
stages). Intent: discover/validate/register per-project owner-authored workflow
descriptors, resolve them at every consumption boundary, inject per-stage
`instruction:`/`permissions:`, scaffold a blank flow via `hive workflow new`.
Coding/content must stay behaviorally untouched (golden guard + 100% coverage).

**Method:** report-only. Verified the pass-2 fixes landed (descriptor `state_file`
traversal guard, `json_payload` overlay lock, `workflow new --json` envelope,
registry union memoization, collision surfacing). Then traced the
`Project.load!` global-overlay race **past** the one path pass-2 hardened
(`Status#json_payload`) into the other in-process consumers. Independent of
codex pass-3, which flagged the same overlay surface; this pass pins the precise
concurrency source (the in-process `Web::Dispatcher`) that makes those findings
reachable, and separates the genuinely-racy web/bot paths from the CLI paths
(single-threaded, not racy on their own).

**Root cause (shared by H1/M1/M2):** `Hive::Workflows::Project.load!` mutates
process-global state (`@active_root`, `Registry.@project_registrations`, the
union cache). Pass-2 serialized only `Status#json_payload`
(`status.rb:188 Project.synchronize`). But `Hive::Web::Dispatcher` executes
`Approve`/`Drop`/`New` **in-process on Puma request threads**
(`web/dispatcher.rb:34-70,248-250`), and those threads run concurrently with the
`StatusFeed` poller thread (`web/status_feed.rb:104,108` → `json_payload`), which
swaps `@active_root` across every registered project each tick on a multi-project
box. None of the in-process command paths hold `Project::LOCK`.

**Verdict:** Not ready — one High (overlay race can hard-delete an archived
custom-workflow task on the web tier) plus four Medium correctness/operability
gaps on the U9 surfaces.

## High
- [ ] Web "Drop" can irreversibly hard-delete an archived custom-workflow task via the overlay race (`lib/hive/web/dispatcher.rb:68-70` → `lib/hive/commands/drop.rb:156`/`drop.rb:235` `guard_archived!`): `Drop` runs in-process on a Puma thread and computes its archived-guard from the global `Hive::Workflows.all_terminal_stage_dirs` union; the matched-project reload at `drop.rb:156` (and the sequential design the author relied on, per the `drop.rb:150-156` comment) is defeated because the `StatusFeed` poller thread can call `Project.load!(otherProject)` → `reset_project_registrations!` + union-cache reset between that reload and `guard_archived!`'s read at `drop.rb:235`, recomputing the union against another project's terminal dirs; when the custom workflow's terminal dir (e.g. `3-done`) is then absent from the union, `guard_archived!` no longer refuses and `cleanup_context` rm_rf's the completed task folder/logs/worktree with no undo. Pass-2 only wrapped `json_payload`, not the in-process command paths — wrap the load!+terminal-dir critical section in `Project.synchronize` (or capture `all_terminal_stage_dirs` under the lock). Trigger is narrow (multi-project box, dashboard open, terminal-stage drop, sub-millisecond window) but the operation is irreversible and the sole guard fails open.

## Medium
- [ ] In-process web `Approve`/`Drop` resolve workflows outside `Project::LOCK`, intermittently failing custom-workflow actions (`lib/hive/task.rb:132`→`task.rb:137`, `lib/hive/workflows.rb:219`→`workflows.rb:220-222`): `Task#resolve_workflow` does `Project.load!` then `Registry.fetch(selector)` after the lock is released, and `stages_for_project` does `load!` then reads `all_stage_dirs`/`resolve_stage_ref_across_workflows` unlocked; both run on a Puma thread (via `Web::Dispatcher#approve`/`#drop` → `TaskResolver`/`Drop`) concurrently with the poller's overlay swap, so a custom-workflow task's `fetch` can hit the wrong/empty overlay → `UnknownWorkflow` → `InvalidTaskPath`, surfacing the operator's Approve/Drop as a spurious error. Non-destructive (retry works) but it's the U9 headline path failing under the normal "dashboard open" condition. CLI invocations are single-threaded and unaffected — the web Dispatcher is the concurrency source.
- [ ] `TasksController#generic_artifact_order` fetches the descriptor from the wrong overlay (`web/app/controllers/tasks_controller.rb:206`): it calls `Hive::Workflows::Registry.fetch(row["workflow"].to_sym)` with no `Project.load!` for the row's project and no lock, so the result depends on whichever project the poller's overlay last left active. On a multi-project box the custom workflow is registered only for the single `@active_root` project; for every other project the fetch raises `UnknownWorkflow` and falls back to the coding `ARTIFACT_ORDER`, rendering coding artifact names (which don't exist) → an empty artifact list on the custom-workflow task page. Near-deterministic for N-1 of N projects, not just a rare race; load the row's project under `Project.synchronize` before the fetch.
- [ ] Bot/web recovery misclassifies custom terminal/inert stages and queues a failing retry (`lib/hive/bot/handlers/recovery_sequence.rb:127,142`): `generic_terminal_stage?`/`generic_non_agent_stage?` call `Registry.fetch(workflow)` without loading the row's project; when the descriptor isn't in the active overlay (the bot process never loads it; the web process may have swapped it) both rescue `UnknownWorkflow` → return `false`, so `retry_verb_for_stage` falls through to `return "run"` for a terminal/inert stage and dispatches `hive run --stage <terminal>`, which `Stages::Resolver` rejects with `StageError` — the operator gets a "Recovery queued" notice for a retry that can never succeed. The helpers receive only `(stage, workflow)`, never a project path, so they cannot load the overlay — recovery for custom non-:agent stages needs the descriptor resolved from the row's project.
- [ ] Parser accepts a nested entry `state_file` that `hive new` cannot write (`lib/hive/workflows/descriptor_parser.rb:143-152` vs `lib/hive/commands/new.rb:135`): the pass-2 traversal guard rejects leading `/` and `..` but still accepts a subdirectory like `sub/idea.md`; `new.rb` then does `File.write(File.join(task_dir, entry_stage.state_file), …)` having only `mkdir_p`'d `task_dir`, so the write raises `Errno::ENOENT`, the rescue rolls back, and a descriptor that passes validation can never capture its first task. Built-ins all use bare basenames — restrict `state_file` to a basename (reject any `/`), which also removes the nested-dir fragility in the agent runner's output path (`stages/agent.rb:20`).

## Nit
- [ ] `Task#resolve_workflow` owns its `Project.load!` but leaves the consistency of the subsequent `Registry.fetch` to the caller's lock (`lib/hive/task.rb:131-140`): the method is thread-safe only when invoked inside `Project.synchronize` (as `Status#json_payload` does), and silently racy otherwise (M1) — document this caller invariant at the method, since `Task.new` is a widely-reused constructor and the coupling is invisible at its other call sites.

ivankuznetsov:

# PR Review Toolkit — u9-per-project-workflows (pass 3)

**Scope:** U9 slice — `git diff origin/main..HEAD`, U9 commits `ff634a1e..HEAD`
(the U7/U8/U10 stack below `7400a72d` was reviewed in its own stages). Six
personas ran report-only and in parallel: code-reviewer, silent-failure-hunter,
type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier. This
is the deduped **union** of their findings; two were independently corroborated
across personas (noted). The overlay-race findings and the nested-`state_file`
`hive new` bug were excluded as already-captured in this pass's other review
file, so persona attention went to NEW issues. Deliberate plan boundaries
(council, save/sharing, sandbox governance, `--help` vocabulary staleness, the
skill/instruction XOR living in the parser per U9-1) were dropped.

## High
- [ ] [code-reviewer, pr-test-analyzer] `hive workflow new --json` error envelope emits a `value` key the published schema rejects (`lib/hive/commands/workflow.rb:226` `error_extras` vs `schemas/hive-workflow-new.v1.json:55-57`): the `ErrorPayload` sets `additionalProperties: false` and declares only `holder`/`lock_path`, not `value`, so every `UsageError` payload (the headline reserved/invalid/colliding-id path) fails JSONSchemer validation — pr-test-analyzer reproduced the rejection (`property at /value is a disallowed additional property`); a regression from the pass-2 fix (`b80907fa`) which added the `value` extra without updating the schema, and `hive-workflow-new` is the only enveloped command lacking a schema round-trip test so it shipped invisibly. Fix: add `value` to the schema's `ErrorPayload`.
- [ ] [silent-failure-hunter, pr-test-analyzer] Generic agent runner masks preflight/instruction failures on a re-run via the `marker.name == :none` guard (`lib/hive/stages/agent.rb:67` and `:100` `instruction_error_result`): both error paths stamp the attributed `:error` marker only when the existing marker is `:none`, so on a re-run of an already-markered stage (e.g. a `:waiting` agent stage the operator edits and reruns), a `spawn_agent` `{status: :error}` preflight/version/auth failure — or an unreadable instruction — leaves the stale `:waiting`/`:complete` in place, `hive run` exits 0 reporting that status, and the failure is unobservable (an operator-invisible retry loop). This re-defeats the NO-SILENT-CAPS guarantee the surrounding work added; the coding path (`spawn_claude_with_tmux_marker!`) overwrites the marker unconditionally. The non-`:none` branch is line-covered but behaviorally untested — dropping `&& marker.name == :none` fails no test.

## Medium
- [ ] [silent-failure-hunter] A per-task `workflow:` pin to a malformed/skipped descriptor surfaces a misleading "unknown workflow" instead of the real parse error (`lib/hive/task.rb:131-140` `resolve_workflow`): unlike the `--workflow`/`hive init` path, which routes through `Project.assert_descriptor_loadable!` (`project.rb:95`) to re-raise the descriptor's real `ConfigError`, `Task#resolve_workflow` calls `Registry.fetch(selector)` directly, so a task pinned to a descriptor that was skipped at load (bad YAML, invalid stage, collision) resolves to `UnknownWorkflow` → `InvalidTaskPath` "unknown workflow", masking the true cause that U9-3 deliberately surfaced for the named-selector path. Untested.
- [ ] [type-design-analyzer] Descriptor `permissions:` is validated incompletely at parse time, breaking the parser's fail-fast promise (`lib/hive/workflows/descriptor_parser.rb:176-184` → `lib/hive/permission_scope.rb`): `parse_permissions` calls `PermissionScope.validate!`, which runs only `parse_spec`, never `resolve`, so a `scoped` permission with an unresolvable `dirs:` (e.g. `~nouser/x` → `ArgumentError` at run) or a non-`claude` profile mismatch is accepted at load and only blows up into an attributed `:error` marker on every stage run — the parser advertises "surfaces at load rather than vanishing at run time" but `dirs:`/profile errors escape it.
- [ ] [comment-analyzer, pr-test-analyzer] `lib/hive/workflows/registry.rb:54` asserts a non-existent test guarantee: the `default` comment claims `:coding` is "equal to CODING_ID by construction (workflows_test pins them equal)", but no test references `CODING_ID` (zero hits across `test/`); the nearest real assertion (`registry_test.rb:13`) pins the literal `:coding` without tying it to the constant, so a maintainer is misled into thinking the `CODING_ID`-vs-`:coding` invariant is regression-protected when it is not.

## Nit
- [ ] [code-reviewer] An agent-FIRST entry stage clobbers the captured idea text (`lib/hive/commands/new.rb` `render_initial_state` ~286-292): `hive new` writes the user's idea into the first stage's `state_file`; if that stage is `kind: agent`, the first `hive run` overwrites the same file with agent output, losing the input — only reachable with a hand-authored agent-first descriptor (the blank scaffold's entry stage is `terminal`, and `validate_terminal_last_stage!` constrains only the last stage), so off the shipped happy path.
- [ ] [type-design-analyzer] `Workflow::Stage` is an anemic record whose cross-field consistency lives only in the YAML parser (`lib/hive/workflow.rb:146-177`): `Stage#initialize` enforces nothing, so the Ruby-built path (built-ins, future descriptors) can construct contradictory stages — agent-only fields on a non-agent kind (`kind: :inert` + `permissions:`), or a non-positive `index` yielding `dir == "0-x"`/`"-3-x"` — that the parser rejects but the type accepts; mirroring the relevant `validate_structure!`-style checks down onto `Stage` would close the asymmetry (the skill/instruction XOR is intentionally parser-side per U9-1 and excluded).
- [ ] [silent-failure-hunter] `hive drop --from <bogus-stage>` with an absent slug now reports "no task folder for slug 'X'" instead of "unknown stage '<x>'" (`lib/hive/commands/drop.rb:38` + `lib/hive/workflows.rb:218-228`): the eager constructor stage-ref resolution moved into `stages_for_project`, whose rescue returns `[]` for a non-ambiguous unknown ref, so the invalid-stage diagnostic is lost (still non-zero exit; message specificity only).
- [ ] [comment-analyzer] Stale citation `lib/hive/workflows/registry.rb:50`: "called at load time from `stages.rb:12` (DIRS)" — `DIRS` is at `lib/hive/stages.rb:16` (line 12 is `module Stages`); substance correct, line drifted +4.
- [ ] [comment-analyzer] Stale citation `wiki/modules/workflows.md:23`: the runner "honors the descriptor's `status_mode` (`agent.rb:35`)" — the actual `status_mode:` line is `lib/hive/stages/agent.rb:52` (line 35 is `render_prompt`); drifted ~17 lines after the instruction-read block was inserted.
- [ ] [comment-analyzer] `wiki/stages/agent.md:38-39` marker→action table contradicts `action_for` (`lib/hive/stages/agent.rb:137`): it omits the explicit `:none → nil` arm, implying `:none` maps to the `"none"` string under the "otherwise" rule.
- [ ] [pr-test-analyzer] The in-process web `Dispatcher#drop`/`#approve` overlay-registration path is unit-tested only on coding stages (`test/unit/web/dispatcher_test.rb:146,162`), never on a custom-workflow stage dir, so a web-tier regression in per-project overlay loading on a generic stage would slip the dispatcher unit suite (only indirectly covered by the CLI-driven e2e).
- [ ] [pr-test-analyzer] The headline e2e (`test/integration/user_workflow_e2e_test.rb:52-58`) asserts the `work` stage ran and reached `COMPLETE` but never asserts the spawned prompt carried the scaffolded `work.md` instruction body, so it cannot distinguish "ran the custom instruction" from "ran a generic fallback" (covered at the unit level `agent_test.rb:204`, so cosmetic).
- [ ] [code-simplifier] `lib/hive/commands/init.rb:208-210` `current_default_workflow` is a one-line wrapper over `current_default_workflow_with_status(ops).first` with a single caller (`:743`) — inlinable, though the name documents intent (borderline).
- [ ] [code-simplifier] `lib/hive/task_resolver.rb:87-92` and `lib/hive/commands/drop.rb:227-232` `project_hint` are near-verbatim duplicates (only `@stage_filter` vs `@from`, set equal in Drop's ctor) building the same `" in project '…' and stage '…'"` tail — a shared helper would keep them in lockstep (cross-class cleanup, marginal).

ivankuznetsov:
Superseded by #562, which consolidates the full U1–U10 workflow-as-data inversion into one branch off current main (all unit fixes integrated, gates green) for single-feature evaluation. Closing u9's standalone PR.

PR #557, opened 2026-06-22
fix(worktree): stop dependency stacking collapsing onto main

ivankuznetsov:
## Summary

Dependency-stacked tasks were silently collapsing onto `main`. When a dependent task already had an **empty placeholder branch** (a slug branch pre-created from the default branch with zero commits of its own), `Hive::Worktree#create!` attached the worktree to it as-is and ignored the stacked `base_override`. The 4-execute agent then found none of its unmerged prerequisite's code and either halted with `EXECUTE_WAITING reason=no_worktree_changes` or built on the wrong base. This recurred on u7–u10 during the U1–U10 inversion dogfood.

A second, latent half of the same bug: `freshest_override_base` resolved a prerequisite only via `origin/<prereq>` and fell straight back to the default branch when the prerequisite lived only in a **local** (never-pushed) branch — so local-only stacking collapsed onto `main` too.

This PR fixes both halves, narrowly gated so they activate only when a non-empty stacked `base_override` is present. The no-dependency path is untouched, and branches carrying real commits are always preserved as-is — work is never discarded.

### What changed

Both changes live in `lib/hive/worktree.rb`:

- **`create!` re-points empty placeholders.** With a stacked override present, an existing slug branch that is an empty placeholder is deleted and recreated through the existing first-creation path on the correctly-resolved base. Emptiness is `git rev-list --count <ref>..<branch>` == 0 measured against **both** the local `<default>` and `origin/<default>` refs (a branch is empty iff it carries no unique commits beyond *either*), never the override — so the check is robust whether the placeholder was cut from a local default that leads or lags a stale origin. The predicate is fail-closed: any git error preserves the branch. The delete is guarded — if `git branch -D` fails (e.g. the branch is checked out in another worktree), it raises a clear `Hive::WorktreeError` naming the branch instead of crashing.
- **`freshest_override_base` prefers a local prerequisite over default.** Resolution order is now `origin/<prereq>` → local `refs/heads/<prereq>` → default. The existing fallback warnings fire verbatim only when no local prerequisite exists, so a stacked request that genuinely can't be satisfied stays observable.

| Prerequisite location | Old base | New base |
|---|---|---|
| On origin | `origin/<prereq>` | `origin/<prereq>` |
| Local branch only (unpushed) | default (collapse) | local `refs/heads/<prereq>` |
| Empty placeholder + stacked override | attach as-is on default (collapse) | recreate on resolved stacked base |
| Branch with its own commits | attach as-is | attach as-is (preserved) |

### Root cause

No separate in-`lib/` creator pre-creates normal task slug branches — `create!`'s own `worktree add -b` is the only one. The leading cause is a stale placeholder left by an earlier collapsed execute attempt: a first run resolved the override to the default branch (the latent local-only bug, or a transient resolution miss), created `<slug>` from default, then halted `no_worktree_changes` and dropped the worktree, leaving `refs/heads/<slug>` behind on default. The next run hit the attach-as-is path — a self-reinforcing collapse. The `create!` hardening fixes this regardless of what pre-created the branch; details are in `wiki/gaps.md`.

## Test plan

`test/unit/worktree_test.rb` adds focused regressions covering both bug halves and both guards:

- re-point an empty placeholder onto an **origin** prerequisite;
- re-point an empty placeholder onto a **local-only** prerequisite (the path that previously collapsed to `main`);
- re-point when the placeholder's local default is **ahead of** a stale origin, and when it **lags** origin (both-ref emptiness measure);
- **preserve** a branch carrying its own real commits (attach as-is, never re-point);
- `branch -D` failure (branch checked out elsewhere) raises a named `Hive::WorktreeError`;
- fail-closed preserve when the emptiness check can't be measured;
- the local-stack resolution path through all three `override_local_or_default` fallback reasons, plus a tag-shadowing safeguard.

Verification:

- `ruby -Itest -Ilib test/unit/worktree_test.rb` → **33 runs, 84 assertions, 0 failures**.
- `env -u HIVE_TASK_STAGE_DIR bundle exec rake coverage` (full suite, the exact command CI runs) → **5700 runs, 22435 assertions, 0 failures, 2 pre-existing skips; coverage gate passed (100%)**.
- `rubocop lib/hive/worktree.rb test/unit/worktree_test.rb` → clean.

## Review summary

Three CE review passes ran (claude-ce, codex-ce, and the multi-persona pr-review-toolkit); all findings were auto-fixed or resolved from the plan/brainstorm — no user escalations.

- **Pass 1** — central High: `empty_placeholder?` measured emptiness against the **local** default while placeholders are cut from `origin/<default>`, so a local default lagging origin misclassified an empty placeholder as carrying work and collapsed. Fixed to measure against `origin/<default>` with a local fallback, plus an origin-ahead-of-local regression test. Doc/warning-text nits applied.
- **Pass 2** — Medium (the symmetric inverse): measuring *only* against origin reintroduced collapse for a placeholder cut from a local default **ahead** of a stale origin. Generalized to the **both-ref / "empty against either"** measure (strictly preserves any branch with unique work; keeps the fail-closed contract) with a local-ahead regression test. Tag-shadowing safeguard (`refs/heads/<branch>` start-point) applied.
- **Pass 3** — High: the both-ref refactor left three inner error branches uncovered, dropping `rake coverage` to 99.99% and failing the CI gate; added a covering test for a present-but-unmeasurable default ref and corrected stale `wiki/` docs so the both-ref basis is described consistently. Final state: gate passes at 100%.

One reviewer-infra timeout (pr-review-toolkit, pass 2) and one environmental review-harness note (a `HIVE_TASK_STAGE_DIR` leak unrelated to this branch) were recorded; neither reflects a defect in the diff.

## Linked task

Hive pipeline task `fix-worktree-dependency-stacking-hive-260622-81f3` (branch `fix-worktree-dependency-stacking-hive-260622-81f3`). Plan requirements R1–R9 are all implemented; the root-cause investigation (R7) is documented in `wiki/gaps.md`.

---

HARNESS-D97757?logo=claude&logoColor=white)


ivankuznetsov:

# Code Review — fix-worktree-dependency-stacking-hive-260622-81f3 (pass 1)

Scope: `git diff origin/main..HEAD` — `lib/hive/worktree.rb` (+wiki/tests).
Intent: re-point empty placeholder branches onto the stacked base and prefer a
local prereq over default, so dependency stacking stops collapsing onto main.
Verification run locally: `bundle exec ruby -Itest test/unit/worktree_test.rb`
green (25 runs / 50 assertions / 0 failures); `rubocop lib/hive/worktree.rb
test/unit/worktree_test.rb` clean.

## High
- [ ] `empty_placeholder?` measures against **local** `<default>` but placeholders are created from `origin/<default>` — silently defeats the fix when local default lags origin (`lib/hive/worktree.rb:308-312`): `git rev-list --count <default>..<branch>` uses local `refs/heads/<default>`, yet `create!`/`freshest_base` start new branches at `origin/<default>` (`worktree.rb:130`) and `freshest_base` deliberately *never* fast-forwards local `<default>` (comment at `worktree.rb:76-77`). Per Unit 4's own root-cause hypothesis the stale placeholder was left by a prior run that resolved the override to `freshest_base` → `origin/<default>`. So when local default trails origin (the exact drift the origin-first design exists for, and which the fetch in `freshest_base` makes *more* likely over time), the placeholder shows `count > 0`, is misclassified as carrying real commits, attaches as-is, and stacking collapses onto the stale base — R1/R8 unmet with no warning emitted. The plan's A1 note asserts local default "reads a placeholder as empty against either ref," which is false once the refs diverge. Fix: treat `count == 0` against `origin/<default>` (when the tracking ref exists, falling back to local) as empty — A1 explicitly permits this.

## Medium
- [ ] Re-point tests never exercise the origin-ahead-of-local placeholder path (`test/unit/worktree_test.rb:201-263`): both `test_repoints_empty_placeholder_onto_origin_prereq` and `_onto_local_prereq` build the placeholder with `git branch <slug> master` while local `master` == `origin/master`, so `empty_placeholder?` always sees `count == 0`. The file already has `with_origin_ahead_of_local` (`:93`) reproducing the divergence used by `test_create_branches_from_origin_default_when_origin_ahead_of_local`; a placeholder-recreation test built on that helper would have caught the High finding and locks in the intended behavior once fixed.

## Nit
- [ ] Misleading warning text on the no-origin path (`lib/hive/worktree.rb:325`): `override_local_or_default` always warns `origin/<branch> unavailable; stacking on local <branch>`, but it is also reached from `freshest_override_base`'s "no origin remote configured" arm (`:141-147`) where there is no `origin` at all — "origin/<branch> unavailable" overstates what was checked. Cosmetic; the stacking decision is correct.
- [ ] New defensive branch outcomes are unverified (`lib/hive/worktree.rb:311,277,318`): the coverage gate enforces line coverage at 100% but not branch coverage (`test/support/coverage.rb:417-421`), so `empty_placeholder?`'s git-failure (`status.success?` false → preserve) arm, `local_branch_ref_exists?`'s empty-name guard, and `delete_local_branch!`'s stdout-only detail fallback have untested outcomes. The fail-closed/preserve directions are safe, but given the `branch -D` data-safety stakes a direct assertion on the rev-list-failure-preserves path is cheap insurance.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — fix-worktree-dependency-stacking-hive-260622-81f3 (pass 1)

Scope: `git diff origin/main..HEAD` — `lib/hive/worktree.rb`, `test/unit/worktree_test.rb`, wiki docs.
Personas run: code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, code-simplifier
(type-design-analyzer skipped — no new types/classes introduced). Tests pass (25 runs, 0 failures);
rubocop clean. Findings are the union across personas; the central correctness item was raised
independently by four of the five personas.

## High
- [ ] [code-reviewer, silent-failure-hunter, comment-analyzer] `empty_placeholder?` measures `git rev-list --count <default>..<branch>` against the **local** default ref (`lib/hive/worktree.rb:308-312`), but placeholders are created from `origin/<default>` via `freshest_base` (`:120-131`), which never advances local default: when local default lags origin the placeholder reads non-zero (reproduced: `master..placeholder == 1` vs `origin/master..placeholder == 0`), is misclassified as carrying real commits, attaches as-is, and stacking silently collapses — the exact bug this PR fixes — with no warning. Fix: measure against `origin/<default>` when its tracking ref exists, falling back to local.
- [ ] [pr-test-analyzer] The realistic placeholder case (local default lags origin, placeholder based on `origin/<default>`) is untested (`test/unit/worktree_test.rb:201,228` both build the placeholder with local==origin default so the count is always 0); the file already has a `with_origin_ahead_of_local` helper (`:93`) that reproduces the divergence — a re-point test built on it would currently fail and lock in the High fix.

## Medium
- [ ] [silent-failure-hunter] `empty_placeholder?` folds a git failure (`status.success?` false) into "not a placeholder" and preserves silently (`lib/hive/worktree.rb:311`) — a dropped stacking request leaves no trace; capture `_err` and `warn` on the failure path while keeping the fail-closed preserve, matching the warn-on-fallback discipline of the rest of `freshest_override_base`.
- [ ] [comment-analyzer] `wiki/modules/worktree.md:48` states a zero count means "an empty placeholder from the default branch" — overstates a local-ref heuristic as fact and omits the local-vs-origin distinction; name the ref (`local-default`) and frame it as a heuristic differing from the origin-first recreate base.
- [ ] [comment-analyzer] `create!` comment (`lib/hive/worktree.rb:60-66`) calls the case "empty placeholders" without noting "empty" is measured vs the *local* default, which differs from the origin base the branch is recreated on — add a pointer to `empty_placeholder?`'s measurement basis.
- [ ] [comment-analyzer] `wiki/modules/task_dependencies.md:74-76` "zero commits relative to the default branch" conflates the local measurement ref with the origin-first recreate base; qualify as *local* default and "origin→local→default base resolution".
- [ ] [comment-analyzer] `override_local_or_default` warns `origin/<branch> unavailable; stacking on local <branch>` (`lib/hive/worktree.rb:325`) even on the no-origin-remote-configured arm (`:141-147`) where no origin exists — collapses the three distinct fallback reasons (no remote / fetch failed / ghost ref) into one generic line; carry the reason into the local-stack message.
- [ ] [pr-test-analyzer] `empty_placeholder?` fail-closed path (rev-list non-success → preserve) is untested (`lib/hive/worktree.rb:311`); add a `create!` call with a `base_override` and a nonexistent `default_branch`, asserting the branch survives and no `WorktreeError` is raised.
- [ ] [pr-test-analyzer] `test_repoint_delete_failure_raises_named_error` (`test/unit/worktree_test.rb:282`) asserts only the message prefix and exercises only the stderr arm of the `detail` ternary (`lib/hive/worktree.rb:318`); a regression dropping `: #{detail}` would still pass — assert the underlying git reason (e.g. `/used by worktree|cannot delete/`).
- [ ] [pr-test-analyzer] `self.local_branch_ref_exists?` empty/whitespace-name guard (`lib/hive/worktree.rb:275-283`) is untested though it gained a real caller this PR; add a direct `refute local_branch_ref_exists?(dir, "")`/`"  "` plus a true case.

## Nit
- [ ] [comment-analyzer] `empty_placeholder?` (`lib/hive/worktree.rb:308-312`) has no comment despite defining the load-bearing term referenced by `create!` and two wiki pages — document (a) it measures against local default and (b) any git error returns false to preserve.
- [ ] [comment-analyzer] `wiki/modules/worktree.md:119` "## Tests" bullet was left stale (still "attach-vs-new, remove, exists?, pointer, prefix") while the parallel `testing.md` row was updated for the four new tests — bring it in line.
- [ ] [silent-failure-hunter] `delete_local_branch!` (`lib/hive/worktree.rb:314-321`) raises with actionable detail (correct); optionally append a remediation hint like "(is `<branch>` checked out in another worktree?)" since that is the common cause the test exercises.
- [ ] [code-simplifier] The three `override_local_or_default` call sites repeat a byte-identical tail "falling back to the default branch base (origin/<default> when reachable)" (`lib/hive/worktree.rb:141-169`); move the shared tail into the helper and pass only the distinguishing reason (minor punctuation normalization needed on the no-origin arm).

ivankuznetsov:

## High

## Medium
- [ ] [lib/hive/worktree.rb:319] Empty-placeholder detection ignores local `<default>` whenever `origin/<default>` exists: a placeholder created during `freshest_base`'s fetch-failure fallback from a local default ahead of stale origin is counted as non-empty, so `create!` preserves it and dependency stacking collapses again.

## Nit

ivankuznetsov:

# Code review — fix-worktree-dependency-stacking (pass 2)

Scope reviewed: `git diff origin/main..HEAD` — `lib/hive/worktree.rb`,
`test/unit/worktree_test.rb`, and the `wiki/` documentation.

Verification performed during review:
- `ruby -Itest test/unit/worktree_test.rb` → 28 runs, 0 failures.
- `rubocop lib/hive/worktree.rb test/unit/worktree_test.rb` → clean.
- All new `lib/` lines traced to dedicated covering tests (re-point ×2,
  origin-ahead measure, fail-closed warn+preserve, real-commit preserve,
  delete-failure raise, both `override_local_or_default` arms, blank-name guard).
- Plan Requirements Trace R1–R8 all implemented; R7 documented in `wiki/gaps.md`.
- NOTE (not a diff defect): a full `rake coverage` run inside this review stage
  reports one failure — `StopHookInstallerTest#test_stop_hook_requires_stage_dir_env`.
  This is environmental: the review harness exports `HIVE_TASK_STAGE_DIR`, which
  leaks into the hook subprocess that test spawns, so the hook succeeds when the
  test expects it to fail. Re-running with `env -u HIVE_TASK_STAGE_DIR` makes the
  test pass. The branch does not touch `stop_hook_installer` or the hook.

## High
- [ ] (none)

## Medium
- [ ] (none)

## Nit
- [ ] `override_local_or_default` returns the bare local branch name as the worktree start-point (`lib/hive/worktree.rb:357`): git resolves a bare `<branch>` through gitrevisions precedence, where a same-named tag (`refs/tags/<branch>`) shadows `refs/heads/<branch>`. Returning `"refs/heads/#{branch}"` would be unambiguous. Impact is low — a tag colliding with a task slug is implausible, and the partial-ref style matches the pre-existing `origin/<branch>` / bare-`<default>` returns — but the fully-qualified form is strictly safer.
- [ ] `delete_local_branch!` appends the `(is <branch> checked out in another worktree?)` hint unconditionally (`lib/hive/worktree.rb:343`), even when `git branch -D` failed for an unrelated reason (e.g. ref-store I/O error). The true cause is preserved via `detail`, so this is only a potentially-misleading suffix, not lost information.

ivankuznetsov:

## High

## Medium

## Nit
- [ ] `wiki/modules/task_dependencies.md:76` still documents origin-else-local placeholder measurement: `empty_placeholder?` now checks both default refs, so the dependency docs contradict the changed behavior.

ivankuznetsov:

# Code review — fix-worktree-dependency-stacking (pass 3)

Scope reviewed: `git diff origin/main..HEAD` — `lib/hive/worktree.rb`,
`test/unit/worktree_test.rb`, and the `wiki/` documentation.

Verification performed during review (in this worktree):
- `ruby -Itest -Ilib test/unit/worktree_test.rb` → **29 runs, 0 failures**.
- `bundle exec rubocop lib/hive/worktree.rb test/unit/worktree_test.rb` → **clean**.
- `env -u HIVE_TASK_STAGE_DIR bundle exec rake coverage` (full suite) →
  **5696 runs, 0 failures, 2 pre-existing skips**, but the **100% coverage gate
  FAILS**: `line coverage 99.99% (25057/25060)`, `lib/hive/worktree.rb` 98.03%
  (149/152), uncovered lines **337, 339, 349**. CI runs this exact command
  (`.github/workflows/ci.yml:34`), so the branch would fail CI as-is.
- Logic of the both-ref `empty_placeholder?` re-point traced for data-loss
  safety: "empty against EITHER default ref ⇒ delete" only deletes a branch
  whose every commit is already contained in some default ref, so no unique
  committed work can be lost — consistent with plan R6 / brainstorm A1. The
  origin→local→default base resolution and the no-dependency path are correct.

## High
- [ ] Coverage gate regression introduced by commit `31c8a592` — `empty_placeholder?` inner error branches at `lib/hive/worktree.rb:337`, `:339`, and `:349` are uncovered across the **entire** suite, dropping `rake coverage` to 99.99% (25057/25060) below the enforced 100% minimum (`test/support/coverage.rb:7`, gated in CI at `.github/workflows/ci.yml:34`): this violates plan **R9** ("full suite green + 100% coverage") and Unit-1/Risks ("every new branch has a dedicated test"), and falsifies the PR's own `bundle exec rake coverage` verification claim. The pass-02 single-ref version routed a bad ref straight into `rev-list`, so `test_create_preserves_branch_when_placeholder_check_errors` (`default_branch: "no-such-default"`) covered the error/warn arm; the pass-03 `default_base_refs` now filters refs by existence first, so that same test hits the `base_refs.empty?` early-return (line 326) instead, leaving the per-ref `rev-list` failure path (337–339) and the all-refs-unmeasurable warn (349) with no covering test. Fix: add a test that makes a *present* default ref fail its `rev-list --count` measurement (one-ref-errors → 337/339, all-refs-error → 349), or restructure so the inner failure handling is reachable/removed.

## Medium

## Nit
- [ ] `wiki/modules/task_dependencies.md:76-77` still documents the superseded single-ref ("origin-else-local") emptiness basis — "the default branch base used for the emptiness check (`origin/<default>` when that tracking ref exists, else local `<default>`)" — but commit `31c8a592` changed `empty_placeholder?` to measure against **both** `origin/<default>` and local `<default>` and treat the branch as empty when it carries no unique commits beyond **either**. This contradicts both the code and the sibling page `wiki/modules/worktree.md` (which was updated to the both-ref/either phrasing in the same PR), so the delivered docs are internally inconsistent. (Same observation as codex-ce-code-review-03.)

ivankuznetsov:

# PR review (pr-review-toolkit, multi-persona) — fix-worktree-dependency-stacking (pass 3)

Scope: `git diff origin/main..HEAD` — `lib/hive/worktree.rb`, `test/unit/worktree_test.rb`, `wiki/*`.
Personas run: code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, code-simplifier.
type-design-analyzer: no applicable findings (no new types/signatures; methods added to the existing `Hive::Worktree`).

Verified during review: unit tests pass (29 runs, 0 failures); rubocop clean; full-suite `rake coverage` FAILS at 99.99% (worktree.rb:337,339,349 uncovered). Core both-ref "empty against EITHER ⇒ delete" logic confirmed data-loss-safe (deletes only commits already reachable from a default ref).

## High
- [ ] [pr-test-analyzer] `test_create_preserves_branch_when_placeholder_check_errors` (`test/unit/worktree_test.rb:336-340`) does not exercise the path it claims, leaving `lib/hive/worktree.rb:337,339,349` uncovered and failing the CI-enforced 100% gate (R9): with `default_branch: "no-such-default"`, `default_base_refs` returns `[]`, so control hits the `base_refs.empty?` early-return (line 326) — `rev-list` never runs — instead of the per-ref failure (337/339) + all-refs-unmeasurable (349) branches; fix needs a test driving a *present* default ref whose `rev-list` fails.

## Medium
- [ ] [comment-analyzer] `wiki/modules/task_dependencies.md:75-77` still describes a single-ref emptiness basis (`origin/<default>` "when that tracking ref exists, else local `<default>`"), directly contradicting the both-ref/either code in `empty_placeholder?` and the sibling page `wiki/modules/worktree.md:48` updated in the same PR (also flagged by codex-03): the delivered docs are internally inconsistent.
- [ ] [pr-test-analyzer] The local-stacking arm of `override_local_or_default` (`worktree.rb:385-391`, returning `refs/heads/<branch>`) is covered only for the no-origin reason; the fetch-failed (`worktree_test.rb:412`) and ref-missing-after-fetch (`:561`) tests have no local prereq, so two of the three documented fallback reasons never verify the local-prereq stack path.
- [ ] [pr-test-analyzer] The `refs/heads/<branch>` tag-shadowing safeguard (`worktree.rb:391`) is asserted only indirectly — no test creates a same-named `refs/tags/<branch>` pointing elsewhere, so a regression reverting to a bare `<branch>` start-point would still ship green.

## Nit
- [ ] [code-reviewer] `empty_placeholder?` measures with a bare `#{base_ref}..#{branch_name}` (`worktree.rb:335`), which resolves through gitrevisions precedence where a same-named `refs/tags/<branch_name>` shadows `refs/heads/<branch_name>`, while the delete (`git branch -D`, line 367) is branch-namespace only — so a colliding tag could make the measurement and the deletion target different objects; qualify as `refs/heads/#{branch_name}` for parity with the line-391 guard (exotic in an agent-controlled repo).
- [ ] [silent-failure-hunter] Delete-then-recreate (`worktree.rb:55-91`) is non-atomic — if `worktree add` fails after `delete_local_branch!`, the placeholder is gone with no rollback; safe *only* because emptiness was positively proven, so add a one-line comment tying the non-atomicity to that proof so a future loosening of the emptiness gate doesn't silently become data loss.
- [ ] [comment-analyzer] `wiki/log.d/20260622T140000Z-worktree-placeholder-origin-measure.md:13-15,27` states in present tense that the check "now measures against `origin/<default>` ... falling back to local only when absent" and refreshed the module pages for the "origin-first" basis — stale vs the later either-ref change (append-only fragment, so historical, but reads as current state; it also explains how the `task_dependencies.md` contradiction survived).
- [ ] [comment-analyzer] `wiki/testing.md:66` lists only the "origin-ahead-of-local placeholder" while the parallel table in `wiki/modules/worktree.md:119` (same diff) lists both origin-ahead **and** local-ahead; both tests exist, so `testing.md` undersells coverage and is inconsistent with its sibling.
- [ ] [pr-test-analyzer] The two-direction measurement tests (`worktree_test.rb:287,317`) assert only HEAD sha, not the prereq's code presence — strictly redundant (the sha is a unique commit) but inconsistent with the sibling re-point tests that also assert `prereq*.txt` exists; add the file assertion for self-documenting parity.
- [ ] [pr-test-analyzer] The docstring of `test_create_preserves_branch_when_placeholder_check_errors` (`worktree_test.rb:336-339`) inaccurately claims `rev-list` "exit[s] non-zero"; it actually exercises the no-measurable-ref (`base_refs.empty?`) preserve path — correct the comment to distinguish the two fail-closed paths.
- [ ] [code-simplifier] `local_branch_ref_exists?`'s blank-name guard (`worktree.rb:269-270`) has no reachable production caller (both callers pass an already-non-empty/`default_branch` value); it is exercised solely by its unit test — defensible defensive code, but dead in production.
- [ ] [code-simplifier] `default_base_refs` (`worktree.rb:359-364`) is a single-use private helper; inlining would keep the both-ref logic in one place — judgment call, fine to leave as a readability seam.
- [ ] [silent-failure-hunter] The per-ref skip (`worktree.rb:336-340`) cannot distinguish an orphan/unrelated branch from a transient git failure; under the either-ref rule a single transient error on one ref never blocks a delete another ref justifies — by design (R1/A1), informational only.

PR #558, opened 2026-06-22
feat(review): suppress triaged no-fix findings from re-raising

ivankuznetsov:
## Summary

A finding that triage classifies as `RESOLVED/NO-FIX` used to re-enter triage on every review pass: reviewers re-emit it, triage re-classifies it, and nothing converges any faster for the effort. That re-entry added latency and carried a non-determinism risk, since triage could re-classify the same finding as `AUTO-FIX` on a later pass. This change fingerprint-suppresses no-fix findings before they reach triage, so a pass that only re-raises already-deferred findings hits the existing all-clean convergence branch instead of churning.

This is a hardening, not a convergence-logic fix. The all-clean branch in `Review.run!` already converges a pure no-fix pass; what was missing was keeping a persistently re-emitted no-fix finding out of triage so that pass arrives sooner and deterministically. Wall-clock and pass-cap behavior is deliberately untouched (that concern is owned separately by #552).

### How it works

Two hooks wire into the existing per-pass loop, on either side of triage:

```mermaid
flowchart TB
  R[run_reviewers writes reviews/*-NN.md] --> S[strip_suppressed! rewrite matching - [ ] lines to - [x] SUPPRESSED]
  S --> T[run_triage_with_retries sees only unsuppressed - [ ] lines]
  T --> SEED[seed_from_triage! append RESOLVED/NO-FIX fingerprints to suppressed.md]
  SEED --> C{accepted empty and zero escalations?}
  C -->|yes| DONE[write_fix_success and converge]
  C -->|no| FIX[Phase 4 fix]
```

- **Strip (before triage).** Each `- [ ]` reviewer line whose fingerprint matches an active entry in `reviews/suppressed.md` is rewritten in place to `- [x] SUPPRESSED: <original> <!-- suppressed: fp=<key> first-pass=NN -->`. Triage only classifies `- [ ]` lines, so it never sees it; the fix agent never collects it (`auto_fix_finding_line?` now excludes `SUPPRESSED:` alongside the existing `RESOLVED/NO-FIX` precedent).
- **Seed (after triage).** Each `- [x] RESOLVED/NO-FIX:` line triage just wrote has its fingerprint appended to `reviews/suppressed.md` so the next pass strips it.

### Design decisions

| Decision | Choice | Why |
|---|---|---|
| Fingerprint key | `normalized_file + severity + normalized_title`, body excluded | A loose key catches a re-raised finding even when its justification text shifts. File refs are extracted from the title region only and stripped of `:line` suffixes (dotted and extensionless paths alike); severity is part of the key. Reworded titles intentionally fall through and loop again — under-suppression is the safer failure mode than silently hiding a genuinely new issue. |
| Lifetime binding | Reviewer compare-base SHA, not per-pass HEAD | Intra-episode fix commits move HEAD every pass; binding to HEAD would reset the list every pass and suppression would never take effect. The base is resolved once per `run!` and frozen across the in-memory pass loop; it only resets across separate `run!`/resume re-invocations or a material base change. |
| Artifact protection | `reviews/suppressed.md` is orchestrator-owned and in both the fix-phase and triage-phase protected sets | Operator-visible and hand-editable (uncheck a line for a durable tombstone, or delete it to let the finding reach triage again), but the fix/triage agents can't rewrite it. The orchestrator only writes it during strip/seed, both outside the agent snapshot windows, so its own writes never trip a tamper check. |
| Read safety | Distinguish ENOENT from a real read error; warn and skip the write | A transient read failure on a populated `reviews/suppressed.md` no longer clobbers operator-edited entries — the mutation is skipped with a warning instead of overwriting the list. |
| Rollback | Config gate `review.triage.suppress_no_fix`, default on | Single flag to disable the whole path if needed. |

Only triage `RESOLVED/NO-FIX` dispositions seed the list. Operator-rejected escalations are explicitly out of scope here.

## Test plan

- `test/unit/stages/review/suppression_test.rb` covers the fingerprint (body-insensitive, `:line`-suffix-insensitive for dotted and extensionless paths, title-sensitive, severity-discriminating, prefix/comment stripping), base-change reset, active-key reading (tombstones and operator-added lines), dedup on append, the seed↔strip cross-shape key identity, the fenced-code-block skip, and the warn-and-skip behavior on an unreadable doc.
- `test/unit/stages/review/triage_test.rb` and `test/integration/run_review_test.rb` lock the protection wiring: a triage- or fix-spawn edit to `reviews/suppressed.md` trips a tamper error, while the orchestrator's own strip/seed writes do not.
- `test/integration/run_review_test.rb` locks the end-to-end behavior from the plan's A7 set: converge-by-pass-2 on a re-emitted no-fix finding (via the all-clean branch, not a wall-clock path), a genuine High still blocking/fixing even when a same-file no-fix was previously seeded, and a different-titled finding surviving the strip into triage.
- `rake test` and `rake coverage` (100% line coverage) pass; `rubocop` is clean on all touched files.

## Review summary

Three CE review passes ran to convergence with no findings escalated for operator input — every finding resolved from the plan, brainstorm, or project context (`escalations-01..03` all empty). Reviewers per pass: `claude-ce-code-review`, `codex-ce-code-review`, and the multi-persona `pr-review-toolkit` (codex hit its usage limit on pass 3 and is recorded in `errors-03`).

Substantive fixes landed during review:

- **Fingerprint correctness.** Two converging findings showed the key collapsing *different* titles of the shape `file:line: description` (truncating at the first `": "` landed right after `file:line`) and leaking justification-body file refs. Fixed by extracting/stripping file refs from the title region *before* the title/justification split, so different-title findings re-loop as the plan guarantees and a reworded rationale keeps the same key.
- **Extensionless paths.** `Gemfile:12` / `Dockerfile:8` style refs weren't normalized; broadened the token regex so their `:line` suffix is stripped too.
- **Destructive-read safety.** A transient read failure on a populated `suppressed.md` could silently turn an append into an overwrite, losing operator edits. Now ENOENT (legit first run) is distinguished from a real read error, which warns and skips the mutation.
- **Triage-phase protection.** `reviews/suppressed.md` was only in the fix-phase protected set; added a triage-local protected snapshot so a triage agent can't clear the list either (plan U3 / A4).
- **Operator-facing accuracy.** Reworded the degraded-base signal and inline comments (the base is frozen across the in-memory pass loop, so it resets across `run!`/resume, not "each pass"), gated that warning on the suppression-enabled flag, clarified the preamble's uncheck-vs-delete semantics, and corrected wiki/state-model notes.
- **Maintainability nits.** Reused `Findings::SEVERITY_HEADING_RE`, derived the severity labels from one canonical order, documented the load-bearing title/justification split, and added the deferred symbol-enrichment breadcrumb.

Each fix was re-verified empirically by the following pass. Gate status reported by reviewers: 100.00% line coverage (gate PASSED), rubocop clean across all touched files. The sole `rake test` failure (`StopHookInstallerTest#test_stop_hook_requires_stage_dir_env`) is an unrelated, pre-existing test-isolation artifact of running the suite inside a hive stage (the reviewer spawn exports `HIVE_TASK_STAGE_DIR`, which that test expects unset); the diff touches none of the stop-hook files, so it is not a regression.

## Linked task

- **Task:** `review-convergence-suppress-a-triaged-260622-4b48`
- **Branch:** `review-convergence-suppress-a-triaged-260622-4b48` → `main`
- **Pipeline:** hive 8-finalize

---


ivankuznetsov:

## High
- [ ] `reviews/suppressed.md` is not protected during triage: `lib/hive/stages/review.rb:524` adds the file only to the fix-agent protected set, while `Hive::Stages::Review::Triage::PROTECTED_FILES` remains unchanged, so a triage agent can edit or clear the suppression list without `triage_tampered`, missing U3's required fix/triage protection.

## Medium

## Nit

ivankuznetsov:



## High

## Medium
- [ ] Fingerprint over-suppresses genuinely-different findings when a reviewer writes `file:line: description` (`lib/hive/stages/review/suppression.rb:142` `title_text` truncates at the first `": "`, which lands right after `file:line`, so PATH_TOKEN_RE then strips the line number and the whole title collapses to just the filename): verified `key_for("lib/foo.rb:42: null deref", severity:"high") == key_for("lib/foo.rb:99: off-by-one", severity:"high")`, so a NO-FIX seed for one finding silently suppresses an unrelated, possibly-High, real finding on the same file and falsely converges to REVIEW_COMPLETE — this is NOT the plan's accepted same-title/different-body tradeoff, it collapses *different* titles and contradicts the plan's stated "different-title re-loops" guarantee; fix by stripping `:line` from / extracting file refs before the title/justification split (or hash the cleaned title-without-files rather than the pre-`": "` prefix).
- [ ] A transient read failure on an existing `reviews/suppressed.md` silently destroys the suppression list with no warning (`lib/hive/stages/review/suppression.rb:157` `recorded_base` rescues `SystemCallError, IOError` → `nil`, so `reset_if_base_changed!` (`:76`) wipes the file; `read_entries` (`:167`) rescues to `[]`, so `append_entries!` (`:85`) turns an append into a destructive overwrite that loses prior entries): the failure direction is safe-ish (findings re-surface) but operator-edited tombstones are lost silently, inconsistent with how this codebase guards other operator-owned provenance files; distinguish ENOENT (legit first run) from a real read error and warn + skip the mutation instead of overwriting.

## Nit
- [ ] Operator-hand-added suppression lines without an `fp=` comment can compute a key that never matches the strip path (`lib/hive/stages/review/suppression.rb:276` `split_entry_severity` treats any leading `Word:` as the severity, so `- [x] Bug: ...` under `## High` keys with severity `"bug"` while `strip_suppressed!` keys it with the section severity `"high"`): the operator's manual suppression silently fails to take effect; only treat the leading word as a severity when `normalize_severity(word)` is in `Hive::Findings::KNOWN_SEVERITIES`/the rendered labels.
- [ ] `title_text` (`lib/hive/stages/review/suppression.rb:142`) is load-bearing for key stability (it is how body/justification is excluded per A3) yet has no comment, unlike its sibling `Hive::Findings::Document#split_title_justification`: add a doc comment explaining the `": "` title/justification split and its key-stability role so a future edit doesn't silently break suppression matching.
- [ ] Parsing logic is duplicated from `Hive::Findings` and will drift (`lib/hive/stages/review/suppression.rb:18` `SECTION_RE` is byte-identical to `Hive::Findings::SEVERITY_HEADING_RE`, and the `in_fence` toggle idiom is copied verbatim at `:215` and `:254`, a third copy of which already lives in `Findings::Document#parse_lines`): reuse the existing constant and extract a shared non-fence line iterator.
- [ ] The `"suppressed"` prefix added to `ORCHESTRATOR_OWNED_PREFIXES` (`lib/hive/stages/review/orchestrator_owned.rb:18`) lacks the trailing dash every sibling uses (`errors-`, `fix-success-`), so `start_with?` matches more broadly than the `basename-NN.md` probe convention documented for the other entries: harmless today (the artifact is `suppressed.md`, no pass suffix) but inconsistent; align to a dash-terminated prefix or document the exemption in the list comment.
- [ ] `SEVERITY_LABELS` (`lib/hive/stages/review/suppression.rb:38`) manually mirrors `SEVERITY_ORDER` (`:30`); adding a severity to one but not the other raises `KeyError` in `render_document` via `SEVERITY_LABELS.fetch` (`:316`): derive the labels from the order list (`SEVERITY_ORDER.to_h { |s| [s, s.capitalize] }`).
- [ ] When the reviewer compare ref fails to resolve, `reviewer_compare_base_sha` falls back to the worktree HEAD (`lib/hive/stages/review.rb:753-757`), which changes on every fix commit, so `reset_if_base_changed!` wipes the list every pass and suppression silently never accumulates: the warn at `:751` doesn't signal that suppression continuity is degraded — make the degraded-mode signal explicit so an operator can tell the feature is effectively disabled.
- [ ] Test-strengthening: the `seed_from_triage!` unit test (`test/unit/stages/review/suppression_test.rb:244`) asserts file content but not the seed↔strip key round-trip, and `test_high_escalation_is_not_suppressed_and_waits_for_operator` (`test/integration/run_review_test.rb:1264`) only checks the finding is absent from `suppressed.md`, not that its unchecked line survived the strip into triage input; add a same-key-different-line round-trip assertion and a `file:line:` non-collision assertion (the two ternary branch outcomes at `review.rb:752`/`:760` are also unexercised — branch coverage only, not gated).

ivankuznetsov:



## High

## Medium
- [ ] [tests] `test_high_escalation_is_not_suppressed_and_waits_for_operator` (`test/integration/run_review_test.rb:1264`) passes for the wrong reason: the triage stub ignores the reviewer file and always writes a fixed escalation, so a regression that wrongly stripped the High to `- [x] SUPPRESSED:` before triage would still escalate and the test stays green — capture the triage input and assert the `- [ ] lib/security.rb leaks token` line reached triage unstripped.
- [ ] [errors] A swallowed read error on an existing, populated `reviews/suppressed.md` silently turns an append into a destructive overwrite (`read_entries` rescues to `[]` at `lib/hive/stages/review/suppression.rb:182`, `recorded_base` to `nil` at `:157`, then `append_entries!` writes `existing([]) + additions`) with no `warn`: fail-safe direction (findings re-surface) but operator-edited state is lost silently — distinguish "file absent" from "file present but unreadable" and warn/skip the write rather than clobber.
- [ ] [type-design] Out-of-enum severities are silently dropped at render (`lib/hive/stages/review/suppression.rb:311` `render_document` iterates only `SEVERITY_ORDER` via `grouped.fetch(severity, [])`, while `split_entry_severity`/`normalize_severity` (`:147`) never clamp): an operator-hand-added `- [x] Critical: …` entry vanishes on the next seed rewrite — clamp `normalize_severity` to `SEVERITY_ORDER` (default `"unknown"`) so no entry is dropped.

## Nit
- [ ] [tests] `test_different_title_reaches_triage_after_prior_no_fix_seed` (`test/integration/run_review_test.rb:1308`) never reads `suppressed.md`, so it can't distinguish "key mismatch" from "seed silently vanished"; add `assert_includes File.read(suppressed.md), "lib/foo.rb leaks stale state"` to confirm the seed is live at pass 2.
- [ ] [tests] The seed↔strip cross-shape key identity (the whole feature's pivot) is proven only via one integration test; add a direct unit `assert_equal key_for(seed-shape RESOLVED/NO-FIX text) == key_for(strip-shape text, different line)` (`test/unit/stages/review/suppression_test.rb`).
- [ ] [tests] `reviewer_compare_base_sha` (`lib/hive/stages/review.rb:744`) has no behavioral test asserting the HEAD/`unresolved-` fallback contract (lines are hit only incidentally; the ternary branch outcomes at `:752`/`:760` are unexercised — branch coverage, not gated): add a test driving an unresolvable ref.
- [ ] [tests] The `in_fence` skip in `no_fix_entries`/`strip_file!` (`lib/hive/stages/review/suppression.rb:215`/`:254`) is untested — no fenced-code-block fixture, so a regression dropping fence-tracking would wrongly seed/suppress a finding shown as a code example.
- [ ] [tests] `first-pass` provenance is never differentiated from the current pass (strip tests all run at pass 1); seed an entry with `first_pass: 1`, strip it at `pass: 3`, and assert the rewritten line keeps `first-pass=01` (`suppression.rb:246`).
- [ ] [errors] Bare `File.readlines` in `strip_file!` (`:208`) and `no_fix_entries` (`:253`) lets one unreadable reviewer file abort the whole pass (top-level `REVIEW_ERROR`) instead of degrading — inconsistent with the deliberate `rescue` on the suppression doc 50 lines above; wrap to warn-and-skip that file.
- [ ] [errors] When `reviewer_compare_base_sha` falls back to worktree HEAD (`lib/hive/stages/review.rb:753`), the base changes on every fix commit so `reset_if_base_changed!` wipes the list each pass and suppression self-disables, signalled only by a one-time setup `warn`: surface the degraded mode per-pass or skip suppression instead of thrashing the file.
- [ ] [comments] `title_text`'s first-`": "` truncation (`lib/hive/stages/review/suppression.rb:142`) is the actual body-exclusion mechanism and load-bearing for key stability, yet is undocumented and diverges from the plan's "normalized_title = remaining text" description; add a comment explaining the title/justification split.
- [ ] [comments] The plan's U1-mandated "deferred symbol enrichment (A3 nice-to-have, not v1) — note it in a comment" breadcrumb is absent from `key_for` (`lib/hive/stages/review/suppression.rb:52`); add the one-line deferral note.
- [ ] [comments] The `reviews/suppressed.md` entry in `protected_set` (`lib/hive/stages/review.rb:524`) is added bare while every sibling (`errors-NN.md`, `fix-guardrail-NN.md`) carries an inline rationale; add a matching one-line comment.
- [ ] [comments] `wiki/state-model.md:228` implies the `suppressed` prefix guard is what excludes `suppressed.md` from the `Hive-Reviewer-Sources` trailer, but the `*-NN.md` glob already excludes it (the prefix is belt-and-suspenders); clarify or drop it from that enumeration.
- [ ] [comments] The pre-existing inline prefix list at `reviewer_sources_for` (`lib/hive/stages/review.rb:1686`) is stale/partial vs the single-source `ORCHESTRATOR_OWNED_PREFIXES`; point to the constant instead of re-listing.
- [ ] [type-design] The severity enum is duplicated across `SEVERITY_ORDER`/`SEVERITY_LABELS`/`SECTION_TITLES`/`Findings::KNOWN_SEVERITIES` and they disagree (`KNOWN_SEVERITIES` lacks `"unknown"`), so a `## Unknown` heading that `render_document` writes is not re-recognized by `parse_section` (`suppression.rb:191`) on read-back; collapse to one canonical list.
- [ ] [type-design] `Entry` (`lib/hive/stages/review/suppression.rb:11`) carries no construction-time validation/factory — `key` can be nil, `active` uses `!= false`, `text` may be unclean — its invariants hold only by `normalize_entry` discipline at the two write boundaries; an `Entry.build` factory would make a fresh entry valid by construction.
- [ ] [simplify] `SECTION_RE` (`lib/hive/stages/review/suppression.rb:18`) duplicates `Hive::Findings::SEVERITY_HEADING_RE` verbatim and the file already depends on `Hive::Findings`; reuse the shared constant.
- [ ] [simplify] The `in_fence`-toggle + section-skip loop is copied between `strip_file!` (`:215`) and `no_fix_entries` (`:254`) (a third copy lives in `Findings::Document#parse_lines`); extract a shared non-fence line iterator (at least for `no_fix_entries`).
- [ ] [simplify] `SEVERITY_LABELS` (`lib/hive/stages/review/suppression.rb:38`) is a verbatim `capitalize` table and `SECTION_TITLES` is derivable except the High suffix; derive both from one source to drop the hand-kept mirrors (output stays byte-identical).
- [ ] [code] `HTML_COMMENT_RE` (`lib/hive/stages/review/suppression.rb:22`, anchored `\z` with non-greedy `.*?` from the first `<!--`) swallows visible text between two HTML comments — harmless today (findings carry at most one trailing comment) but a latent footgun; strip all comments globally if a finding ever embeds a mid-text comment.
- [ ] [code] `execute_stale_artifacts` (`lib/hive/task_action.rb:545`) surfaces the last 3 sorted `reviews/*.md`; `suppressed.md` (late-alphabetical) could displace a reviewer file from that set — cosmetic only, and the file normally doesn't exist at the 4-execute EXECUTE_STALE point.

ivankuznetsov:

## High

## Medium
- [ ] `key_for` includes rationale-only file refs in the fingerprint: `lib/hive/stages/review/suppression.rb:78` scans path tokens before dropping the `: justification` body, so a reworded rationale that adds/removes a path changes the normalized file set and prevents the plan-required same-title no-fix re-emission from being suppressed.
- [ ] Extensionless file refs keep line numbers in the title: `lib/hive/stages/review/suppression.rb:29` only recognizes dotted filenames, so common refs like `Gemfile:12` or `Dockerfile:8` are not normalized and `Gemfile:12 ...` will not match `Gemfile:88 ...`, violating the line-number-stripping requirement.

## Nit

ivankuznetsov:



## High

## Medium
- [ ] Fingerprint leaks justification-body file refs, contradicting A3 "body excluded" and the function's own doc comment (`lib/hive/stages/review/suppression.rb:78` scans `PATH_TOKEN_RE` over the whole `cleaned` text — title AND justification — before the title/justification split, so a re-emitted finding whose *body* mentions a different path produces a different key): empirically `key_for("lib/foo.rb leaks stale state: caused by helper in lib/bar.rb")` != `key_for("…: caused by helper in lib/baz.rb")`, yet the comment at `:69` claims "a re-emitted finding with a reworded rationale keeps the same key (A3)"; safe-direction (under-suppression) and harmless when triage preserves the finding verbatim, but a reviewer that rewords the rationale's file mention across passes silently re-loops — fix by extracting file refs only from the title portion (post-split), so the body never feeds `normalized_file`.
- [ ] Extensionless file refs never get their `:line` suffix stripped, partially missing the A3/U1 "strip :line numbers from file refs" requirement (`lib/hive/stages/review/suppression.rb:29` `PATH_TOKEN_RE` requires a dotted `\.[ext]`, so `Gemfile:12`, `Dockerfile:8`, `Rakefile`, `Makefile`, `Procfile`, `.env` are not recognized as path tokens and keep their line numbers inside the title): empirically `key_for("Gemfile:12 pins insecure gem")` != `key_for("Gemfile:88 pins insecure gem")` while the dotted control matches, so a line-drifted re-emission on a common extensionless file is not suppressed; safe-direction but the integration suite only exercises dotted paths — broaden the token regex to also match extensionless basenames carrying a `:NN[:NN]` suffix.

## Nit
- [ ] Operator escape-hatch preamble conflates "delete" with "uncheck", but only unchecking durably un-suppresses (`lib/hive/stages/review/suppression.rb:422` writes "Uncheck or delete an entry to let the finding reach triage on the next pass", yet `append_entries!` builds `known_keys` from ALL existing entries including tombstones at `:120`, so a `- [ ]` tombstone is deduped and never re-seeded — durable — while a *deleted* entry is gone from `known_keys` and is re-seeded → re-suppressed on the very next NO-FIX pass): the documented hatch misleads an operator who deletes (rather than unchecks) expecting it to stay un-suppressed; reword the preamble to say uncheck (tombstone) for durable un-suppression, delete only un-suppresses until triage re-flags it.

ivankuznetsov:



## High

## Medium
- [ ] [code] Fingerprint leaks justification-body file refs, breaking the documented "reworded rationale keeps the same key" guarantee (`lib/hive/stages/review/suppression.rb:78` `cleaned.scan(PATH_TOKEN_RE)` runs over the whole cleaned text — title AND justification — before the `: ` title split, so a re-emitted finding whose body mentions a different path-shaped token gets a different key): confirmed empirically `key_for("lib/foo.rb leaks stale state: caused by helper in lib/bar.rb")` != `key_for("…: caused by helper in lib/baz.rb")`, contradicting the comment at `:58-63`/`:176-183`; safe-direction (under-suppression) but the stated robustness is not met — extract file refs from the post-split title only.
- [ ] [tests] Extensionless file refs never get their `:line` suffix stripped and have zero coverage (`lib/hive/stages/review/suppression.rb:29` `PATH_TOKEN_RE` requires a dotted `\.[ext]`, so `Gemfile:12`, `Dockerfile:8`, `Rakefile`, `Makefile`, `Procfile` keep their line numbers in the title): `key_for("Gemfile:12 …")` != `key_for("Gemfile:88 …")` while the dotted control matches, so a line-drifted re-emission on a common extensionless config file is never suppressed — partially misses the A3/U1 line-stripping requirement; add a `Gemfile:12 == Gemfile:88` test (currently failing) and broaden the regex.
- [ ] [errors] Read/mutate asymmetry can silently wipe operator-edited entries with no warning (`lib/hive/stages/review/suppression.rb`: `suppressed_doc_unreadable?` at `:238` probes only the first byte, but on the mutate path `recorded_base` at `:221` reads the whole first line and `read_entries` at `:258` reads the whole file, each rescuing to `nil`/`[]` SILENTLY — unlike sibling `reviewer_lines` at `:271` which warns): a corrupted/removed header line (or a partial read that slips past the 1-byte probe) makes `reset_if_base_changed!`/`append_entries!` overwrite the doc with an empty list, destroying suppressions unlogged — defeating the exact "don't clobber operator edits" intent the guard was added for; narrow trigger but add a `warn` on the lossy `recorded_base`/`read_entries` branches.
- [ ] [comments] The generated operator preamble conflates "uncheck" and "delete" (`lib/hive/stages/review/suppression.rb:422` writes "Uncheck or delete an entry to let the finding reach triage on the next pass"): because `append_entries!` builds `known_keys` from ALL existing entries including tombstones (`:120`), an UNCHECKED `[ ]` entry durably un-suppresses while a DELETED entry is re-seeded and re-suppressed on the very next NO-FIX pass — operator-facing guidance baked into the artifact that will mislead anyone trying to permanently release a finding; document uncheck as durable, delete as one-pass.
- [ ] [tests] `test_high_escalation_is_not_suppressed_and_waits_for_operator` cannot fail for the reason it claims (`test/integration/run_review_test.rb:1264-1314`): `suppressed.md` is never pre-populated, so `strip_suppressed!` is a guaranteed no-op regardless of High-handling — the test would still pass with all High-protection removed; the real A7 "genuine-High-still-blocks even when a same-file no-fix was previously seeded" risk is untested — seed a no-fix on the same file/severity first, then re-emit a different-title High and assert it reaches triage unstripped.
- [ ] [tests] Triage-phase tamper protection for `reviews/suppressed.md` is only asserted statically, never through a spawn (`test/unit/stages/review/triage_test.rb:355-365` checks the `protected_paths` list, and the parameterized tamper loop at `:237` iterates `ORCHESTRATOR_OWNED` which excludes suppressed.md): a regression reverting `run!`'s snapshot from `TRIAGE_PROTECTED_FILES` to `PROTECTED_FILES` would pass the static test but silently drop the U3/A4-required triage protection — add a triage-spawn-edits-suppressed.md → `:tampered` case (the fix-phase tamper IS covered end-to-end at `run_review_test.rb:1565`).

## Nit
- [ ] [tests] The convergence integration test asserts the no-fix IS recorded but never asserts the AUTO-FIX line is EXCLUDED from `suppressed.md` (`test/integration/run_review_test.rb:1259`): a regression seeding AUTO-FIX dispositions too would pass — add `refute_includes File.read(suppressed.md), "lib/fix.rb"` (the unit-level exclusion is covered at `suppression_test.rb:300`).
- [ ] [comments] `triage.rb` comment cites a non-existent shared consumer (`lib/hive/stages/review/triage.rb:49-50` says `ORCHESTRATOR_OWNED` is "also used by 4-execute and ci-fix"): 4-execute defines its own `PROTECTED_FILES = %w[plan.md worktree.yml]` (`execute.rb:38`) and does NOT use `ORCHESTRATOR_OWNED`; only ci-fix (`ci_fix.rb:43`) does — the conclusion is sound but the attribution is wrong.
- [ ] [comments] Degraded-base warning/comment says the list "resets on every fix commit … each pass" (`lib/hive/stages/review.rb:398-400`, `:84-87`), but the base SHA is resolved ONCE per `run!` (`:168`) and frozen across all in-memory passes, so the reset only fires across separate `run!`/resume invocations after HEAD advances — "each pass" is misleading for the in-memory loop.
- [ ] [code] `"suppressed"` in `ORCHESTRATOR_OWNED_PREFIXES` matches via `start_with?("suppressed")` (no trailing dash, by design — `lib/hive/stages/review/orchestrator_owned.rb:11-24`), so a reviewer whose `output_basename` begins with `suppressed` (e.g. `suppressed-checks-01.md`) would be silently misclassified as orchestrator-owned and dropped from `discover_reviewer_files`/`collect_accepted_findings` — a broader net than the dash-terminated siblings (note: `suppression-*` does NOT collide).
- [ ] [types] `Entry` and `ReviewerCompareBase` are anemic `Data` structs whose invariants (canonical severity, present 16-hex key, boolean `active`, non-negative `first_pass`; `degraded` ⟺ sentinel sha) are enforced only in `normalize_entry`/call-sites, not at construction (`lib/hive/stages/review/suppression.rb:11`, `lib/hive/stages/review.rb:89`): `render_document`'s `SECTION_TITLES.fetch`/`%02d` rest on that discipline — unreachable today (write_entries always normalizes first) but a future caller rendering un-normalized entries would `KeyError` mid-pass; a smart constructor (`Entry.build`) would move the guarantee into the type.
- [ ] [tests] `clean_finding_text` mid-text HTML-comment stripping is documented as a "latent footgun" (`lib/hive/stages/review/suppression.rb:21-28`, `:161-174`) but not tested — a one-line `key_for("foo <!-- a --> bar <!-- b --> baz")` preserves-"foo bar baz" regression test would lock the global-sub behavior cheaply.

ivankuznetsov:



## High

## Medium

## Nit
- [ ] Degraded-base warning and the `reviewer_compare_base_sha` inline comment overstate the failure, contradicting the code's own accurate `ReviewerCompareBase` doc: the per-pass warning says "will not accumulate for pass NN — no-fix suppression is **effectively disabled this run**" (`lib/hive/stages/review.rb:400-402`) and the method comment says `reset_if_base_changed!` "wipes the list **each pass**" (`lib/hive/stages/review.rb:778-781`), but the base SHA is resolved once at `review.rb:168` and frozen across the in-memory pass loop (`loop do` at `:229`), so within a single `run!` the recorded base always equals the frozen `base_sha` and the list is NOT wiped between passes — degraded suppression still strips/seeds correctly across passes of the same `run!` (the primary convergence path the integration test exercises). The reset only fires across separate `run!`/resume invocations, exactly as the `ReviewerCompareBase` comment at `:85-89` correctly states ("only moves between separate `run!`/resume invocations"); reword the warning/comment to say "resets across `run!`/resume re-invocations" (drop "each pass"/"disabled this run") so the operator signal matches the frozen-base behavior and the feature's own doc. Safe-direction (over-warns), low-reach (`reviewer_compare_ref` already validates the ref exists, so this fallback is near-unreachable), and wording-only — no functional impact.

ivankuznetsov:



## High

## Medium
- [ ] [silent-failure] Post-probe read failure on a present `suppressed.md` silently clobbers operator-edited entries (`lib/hive/stages/review/suppression.rb:137` `read_entries` and `:122-125` `recorded_base`): the `suppressed_doc_unreadable?` guard reads only byte 0, but `read_entries` does a full `File.readlines` (rescued to `[]`) and `recorded_base` a full `readline` (rescued to `nil`); a genuine error surfacing only on the full read (e.g. mid-file `EIO`) makes `append_entries!` write `existing(=[]) + additions` (`:150`) or `reset_if_base_changed!` write `base_sha, []` (`:124`), dropping operator tombstones and re-suppressing a finding the operator explicitly un-suppressed — the unsafe (clobber) direction, opposite the skip-don't-clobber intent the byte-0 probe enforces; the rescues warn but don't abort the write. Fix: have `read_entries`/`recorded_base` signal failure to the mutation callers so they skip the write (query paths keep fail-to-empty).
- [ ] [tests] Severity is not pinned as a fingerprint key discriminator (`test/unit/stages/review/suppression_test.rb` — the `refute_equal` cases at `:65`/`:77` only vary the title; none varies severity): mutation-verified by the persona that replacing the `severity_key` component in `key_for` (`lib/hive/stages/review/suppression.rb:107`) with a constant leaves all unit+integration tests green, so a regression collapsing severity from the key would be silent and let a Medium no-fix suppress a re-emitted same-title/same-file High (defeating A5a "suppress ANY severity" + genuine-High-still-blocks); add `refute_equal key_for(t, severity: "medium"), key_for(t, severity: "high")`.

## Nit
- [ ] [tests] The High-escalation and different-title integration tests can't distinguish "strip ran and correctly skipped" from "strip never ran" (`test/integration/run_review_test.rb:1259`, `:1410`): mutation-verified that neutering `strip_suppressed!` to a no-op leaves both green (only the convergence test fails); their `assert_includes suppressed, "…"` proves the seed persisted, not that strip executed against it — add a strip-executed assertion (e.g. the per-pass `"suppressed N no-fix finding(s)"` stderr is present/absent as expected).
- [ ] [tests] A7-4 "wall-clock untouched" is not genuinely asserted (`test/integration/run_review_test.rb:1252-1255`): the `review_complete`/`pass=2`/`assert_nil reason` marker checks are decorative — reviewers/triage are stubbed to return instantly, so `reason=wall_clock` could never fire regardless of suppression; the only load-bearing proof in that test is the SUPPRESSED input-capture at `:1256`, and the `assert_nil reason` comment claiming it proves "wall-clock untouched" overstates.
- [ ] [tests] `test_read_active_keys_returns_empty_for_unreadable_suppression_doc` (`test/unit/stages/review/suppression_test.rb:207`) asserts only `assert_empty`, not the data-loss warning the `read_entries` rescue emits (`suppression.rb:290-298`); a regression silencing that warn wouldn't be caught (cf. the exemplary `append_entries_skips_when_doc_present_but_unreadable` test at `:449`, which does assert the warn).
- [ ] [tests] `recorded_base`'s `SystemCallError`/`IOError` warn branch (`lib/hive/stages/review/suppression.rb:248-258`) is untested — no test drives a header read that fails after the byte-0 probe succeeds (the partial-read race the comment describes); acknowledged hard-to-reach, low value, pairs with the Medium clobber-window above.
- [ ] [comment] The inline comment "binding suppression to it … means `reset_if_base_changed!` wipes the list each pass" (`lib/hive/stages/review.rb:778-781`) is inaccurate: the base SHA is resolved once at `:170-171` and frozen across the pass `loop do`, so within a single `run!` the list is NOT wiped between passes — it resets only across separate `run!`/resume re-invocations, exactly as the sibling `ReviewerCompareBase` doc (`:83-90`) and the degraded warning (`:399-402`) correctly state; reword "each pass" → "on the next run!/resume re-invocation". (Same inaccuracy raised in claude-ce-code-review-03.)
- [ ] [comment] `wiki/state-model.md:231` claims `discover_reviewer_files` / `collect_accepted_findings` "glob `*.md` more broadly", justifying the `suppressed.` prefix as load-bearing there — but both glob the narrow `*-#{pass}.md` (`lib/hive/stages/review/triage.rb:181`, `lib/hive/stages/review.rb:1514`), so `suppressed.md` (no `-NN` suffix) is excluded by glob in those consumers too; the prefix is belt-and-suspenders for them as well, and a maintainer trusting this line would hunt for a broad `*.md` glob that doesn't exist.
- [ ] [code] The degraded-base warning (`lib/hive/stages/review.rb:399-403`) is gated only on `@suppression_base_degraded` inside `if triage_enabled?`, not on `Suppression.enabled?(cfg)` / the `suppress_no_fix` flag, so an operator who set `suppress_no_fix: false` and has an unresolved compare ref still gets a per-pass "no-fix suppression is effectively disabled this run" warning when suppression was already off by config — redundant/alarming; cosmetic only.
- [ ] [type-design] `Entry` couples the visible severity label and the persisted `fp=` without enforcing their relationship (`lib/hive/stages/review/suppression.rb:356`/`:395`): a hand-edit that changes the `## section`/inline `Label:` but keeps `fp=` permanently diverges (parse keeps the stored fp, strip recomputes from section severity → no match → silent un-suppress); outside the documented uncheck/delete workflow, but worth a one-line doc note that editing the severity label is not a supported un-suppress gesture.
- [ ] [type-design] `Entry#active` is normalized via `entry.active != false` (`lib/hive/stages/review/suppression.rb:445`), so an Entry built with `active: nil` coerces to active=true; harmless today (every constructor passes an explicit boolean) but a latent footgun if a future caller relies on a nil default.

PR #561, opened 2026-06-22
feat(screenote): connect via OAuth 2.1/MCP, replace api-key upload

ivankuznetsov:
# feat(screenote): connect via OAuth 2.1/MCP, replace api-key upload

## Summary

Screenshot publishing for the 7-artifacts stage moves off a static, operator-global api-key and onto OAuth 2.1. Previously hive uploaded stills itself with a `POST /api/v1/screenshots` REST call carrying `screenote.api_token`; that path is gone. Now an operator runs `hive connect screenote` once on their own machine to mint a revocable token via auth-code + PKCE + a loopback redirect, and the headless box hands Screenote to the 7-artifacts agent as an MCP server so the **agent** captures and uploads. The box runs non-interactively off the stored token and degrades softly when it is missing, expired, or revoked.

```mermaid
flowchart TB
  subgraph Operator[Operator machine -- interactive]
    C[hive connect screenote]
    B[Browser + loopback 127.0.0.1]
    C -->|discover, register, PKCE S256| SN[(Screenote OAuth/Doorkeeper)]
    C -->|open authorize_url| B
    B -->|code + state| C
    C -->|exchange_code| SN
    C -->|list_projects, pick default| MCP[(Screenote MCP)]
    C -->|save 0600| CRED[~/.config/hive/screenote.json]
  end
  subgraph Box[Headless box -- non-interactive]
    A[7-artifacts stage] -->|valid token?| CRED
    A -->|write ephemeral .mcp.json 0600| CFG[cache_home/mcp/...]
    A -->|--mcp-config --strict-mcp-config| AG[Claude artifacts agent]
    AG -->|create_screenshot_upload + PUT bytes| MCP
  end
```

### What changed

| Area | Change |
|---|---|
| `hive connect screenote` | New Thor verb. Discovers OAuth/MCP metadata (RFC 8414 + RFC 9728), dynamically registers a public client (RFC 7591) reusing a cached `client_id`, runs auth-code + PKCE S256 over a one-shot loopback server, exchanges the code, lists projects over MCP, and prompts for a default project before persisting. |
| `hive disconnect screenote` | Revokes the token at the discovered revocation endpoint, then clears the local credential. Idempotent and fail-soft when nothing is stored or revoke fails. |
| Credential store | `~/.config/hive/screenote.json` at `0o600` holding `access_token`, `expires_at`, `client_id`, `issuer`, `project_id`, `base_url`, `mcp_resource`, and an optional `refresh_token`. Honors `HIVE_HOME`. |
| MCP injection | When a valid credential exists, the stage writes a per-run `.mcp.json` (bearer in the `Authorization` header) under `cache_home`, threads `--mcp-config` / `--strict-mcp-config` and the Screenote tool names into the Claude launch, and deletes the file after the run. |
| Capture rewire | The agent calls `create_screenshot_upload` and PUTs the captured bytes, writing the resulting `screenote_url` into `media/manifest.json` itself. `Hive::ScreenoteUploader` and the Ruby upload loop are deleted. |
| Config rip-out | `screenote.api_token`, `HIVE_SCREENOTE_API_TOKEN`, and the env-scrub entries for it are removed. A config that still sets a non-blank `api_token` now raises a one-time migration error pointing at `hive connect screenote`. `base_url` stays configurable (default `https://screenote.ai`). |

### Design decisions

- **The agent uploads, not hive.** The OAuth bearer is intentionally exposed to the artifacts agent via its MCP config — the consciously relaxed invariant from the brainstorm. It is mitigated by an ephemeral `0o600` file under `cache_home` (never the committed task folder), deletion after the run, `--strict-mcp-config` so the operator's other MCP servers aren't inherited, narrow `mcp_read`/`mcp_write` scope, and a revocable token.
- **Discovery, not hard-coded URLs.** The MCP endpoint comes from the `resource` field of `/.well-known/oauth-protected-resource`, so self-hosted and staging `base_url`s work.
- **Fail-soft, two ways.** Not-connected / expired / incomplete credentials inject no MCP server at all, and the prompt instructs the agent to record a skip reason. 7-artifacts still captures and commits local media and completes `:complete`; it never hard-fails on Screenote. The degradation is also `warn`ed at run time.
- **`refresh_token` deferred.** The 1-year token is stored (and `refresh_token` kept if returned) but there is no auto-refresh; re-connect on expiry.
- **Read-side contract unchanged.** The web layer still reads `screenote_url` from the manifest; only who writes it changed. `screenote_skipped_reason` is an additive writer-side diagnostic.

## Test plan

`rake test` (5706 runs, 22565 assertions, 0 failures, 0 errors, 6 skips), `rake coverage` (100% line), and `rubocop` (651 files, no offenses) are green. Unit suites cover the OAuth client, the PKCE RFC 7636 vector, the loopback callback (including non-`/callback` prefetch tolerance, malformed-request-line handling, and read-deadline/byte caps), the credential store, MCP config shape and `0o600` placement, the connect/disconnect commands (including `--json` envelopes, the `needs_project_selection` two-step, and best-effort post-exchange revoke), and the rewired artifacts stage's fail-soft branches.

Live, env-gated integration tests hit Screenote directly with no stub seam:

- `test/integration/screenote_oauth_live_test.rb` exercises discovery, dynamic registration, and code exchange.
- `test/integration/screenote_capture_live_test.rb` round-trips real PNG bytes through `create_screenshot_upload` but is **skipped-with-reason until Screenote ships its non-interactive MCP test-token endpoint** — enable it by setting `HIVE_SCREENOTE_TEST_TOKEN_ENDPOINT` / `HIVE_SCREENOTE_TEST_TOKEN_SECRET` / `HIVE_SCREENOTE_TEST_PROJECT_ID`.

## Review summary

Three Compound Engineering review passes ran (`claude-ce-code-review` + the `pr-review-toolkit` personas: correctness, security, adversarial, reliability, testing, maintainability, cli-readiness, api-contract, agent-native, project-standards). The `codex-ce-code-review` reviewer hit a usage-limit infra error on all three passes and contributed no findings.

Auto-fixed across the passes:

- **Fail-soft hardening (A8).** Broadened `screenote_context` / `McpConfig#write!` rescues to `SystemCallError` + `Hive::ConfigError`, treated a corrupt local credential as "no credential", and made the degrade visible via `warn`.
- **OAuth/connect correctness.** Fixed the migration catch-22 (only a non-blank `api_token` now raises), the project-picker `"0"` misselect, a typed error for a missing `expires_in`, best-effort `revoke` of an abandoned token on post-exchange failure, and moved `emit_success` out of the revoke-scoped region so a closed `--json` pipe can't revoke the just-saved token.
- **Transport/MCP robustness.** Completed `NETWORK_ERRORS` (TLS, OS-level timeouts, `Net::WriteTimeout`), surfaced MCP tool-channel `isError` results as typed errors, added SSE `data:`-frame de-framing, and hardened the loopback server (validate-before-write, read deadline, header byte cap, malformed-URI tolerance).
- **CLI / API contract.** Added `long_desc`, `emit_error_envelope` on both commands, the `needs_project_selection` envelope with `error_kind`/`exit_code`, and shape-consistent disconnect envelopes.
- **Cleanup & DRY.** Removed orphaned `media_manifest_path`, extracted shared `Http` / `SecureFile` helpers, refreshed stale docs/comments and the changelog fragment, and added the branch-coverage tests the 100% line gate didn't catch.

Two findings were escalated and consciously deferred, each consistent with a deliberate plan decision:

- **Q1 — reconnect `redirect_uri` matching.** Reconnect reuses the cached `client_id` with a fresh ephemeral loopback port. The plan relies on the RFC 8252 loopback-port exemption to avoid the 10/hr DCR registration limit; whether re-registration is needed hinges on Screenote's (Doorkeeper) exact-match policy, which the env-gated/blocked live discovery test will settle. No code change.
- **Q2 — `Screenote::Credential` value object.** The plan models the credential as a JSON Hash by design. A typed value object with a redacting `inspect` (to stop a stray `p`/backtrace leaking the bearer) is a worthwhile follow-up but a new abstraction touching 4+ call sites; deferred.

## Linked task

- Task: **Connect Screenote via OAuth** (`#6110`, slug `connect-screenote-via-oauth-2-260622-4e18`)
- Pull request: https://github.com/ivankuznetsov/hive/pull/561

---


ivankuznetsov:

# Code Review — connect-screenote-via-oauth-2 (pass 1)

Scope: `git diff origin/main..HEAD` (7 commits, U1–U6). Replaces the api-key REST Screenote uploader with an OAuth 2.1 / MCP connect+capture path.
Reviewers: correctness, security, adversarial, reliability, testing, maintainability, cli-readiness, project-standards. Findings cross-verified against the code; plan-deferred scope (refresh_token auto-refresh, codex MCP injection, server-side-URL-capture default, the blocked live capture test, the consciously-accepted token-to-agent exposure) was excluded.

## High
- [ ] Migration guard is a catch-22 and over-broad (`lib/hive/config.rb:1852` `validate_screenote!`): it raises on `screenote.key?("api_token")` regardless of value, and `validate!` runs on **every** config load (`config.rb:1299`), so any operator who set `screenote.api_token` — **including the `api_token: null` leftover the previous `config.example.yml` shipped** — gets `ConfigError` on every command; the prescribed remedy `hive connect screenote` itself calls `load_global_screenote` (`commands/connect.rb:90`) → same error, so the documented A7 migration path is non-functional and the message misleads. Fix: only raise on a non-blank api_token, and/or have the message instruct removing the key, and/or let connect's base-url resolution tolerate it.
- [ ] Fail-soft contract (A8) has uncaught-exception escapes: `screenote_context` rescues only `Hive::ConfigError` (`lib/hive/stages/artifacts.rb`), but `CredentialStore#load`'s `File.read` raises `SystemCallError` (`Errno::EACCES`/`EISDIR`, TOCTOU `ENOENT`) and `McpConfig.new(...).write!` raises `Errno::*` on an unwritable `cache_home` — both escape `Artifacts.run!` and are re-raised by `with_stage_events` (`stages/base.rb:168`), hard-failing the stage instead of skipping upload + committing local media. Fix: broaden the rescue to `StandardError`/`SystemCallError` → `disconnected(...)` and wrap `write!` to fall back to the no-MCP path.

## Medium
- [ ] Project picker silently misselects on input `"0"` (`lib/hive/commands/connect.rb:118`): `projects[answer.to_i - 1]` makes `"0"` → `projects[-1]`, connecting the **last** project with no error (a typo/cancel becomes a wrong default). Fix: reject values outside `1..projects.size` before indexing. (correctness + adversarial, confidence 100)
- [ ] Reconnect reuses the stored `client_id` with a fresh ephemeral loopback `redirect_uri` (`lib/hive/commands/connect.rb:42`): `LoopbackServer` binds port `0` each run, so the redirect URI changes, but on reconnect `register` is skipped — the new URI was never registered for that client. An OAuth server enforcing exact `redirect_uri` matching (RFC 6749/8252) will reject authorize/exchange; the reconnect unit test masks this with a fixed-port fake. Verify against Screenote's matching policy; if exact, re-register or persist/reuse a stable redirect.
- [ ] `OpenSSL::SSL::SSLError` and `Errno::ETIMEDOUT` are missing from `NETWORK_ERRORS` (`lib/hive/screenote/oauth_client.rb:13`, reused by `mcp_client.rb`): TLS/cert and OS-level TCP-timeout failures escape as unmapped `StandardError` from every OAuth/MCP request — uncaught in `connect`, and in `disconnect`'s `revoke` (which rescues only `Hive::Error`). Pre-existing pattern (also in `web/github_auth.rb`) but first user-configured external HTTPS service makes it newly relevant. Fix: add both to the list.
- [ ] `credential_payload` raises a raw `KeyError` when the token response omits both `expires_at` and `expires_in` (`lib/hive/commands/connect.rb` ~122, `token.fetch("expires_in")`): a non-conformant token endpoint crashes `hive connect` with an unhandled exception instead of a typed `Hive::Error`. Fix: guard with a default/typed error. (correctness + reliability + adversarial)
- [ ] Orphaned production code: `media_manifest_path` (`lib/hive/stages/artifacts.rb:125`) has no `lib/` caller after `push_manifest_media_to_screenote` was removed — it is referenced only by the test. Move it into the test helper or delete it. (maintainability, confidence 100)

## Nit
- [ ] `McpClient` couples to `OAuthClient` only for transport constants (`lib/hive/screenote/mcp_client.rb`): it `require`s `oauth_client` and reads `NETWORK_ERRORS`/`OPEN_TIMEOUT_SEC`/`READ_TIMEOUT_SEC`. Extract a shared `Screenote` HTTP-defaults constant so a timeout change in one client doesn't silently bind the other.
- [ ] Duplicated 0600 write+chmod pattern in `CredentialStore#save` and `McpConfig#write!` — extract a shared secure-write helper so the umask-backstop rationale lives in one place.
- [ ] Loopback callback reader has no per-read socket timeout (`lib/hive/screenote/loopback_server.rb`): `IO.select` guards only `accept`; `read_callback_params`'s `socket.gets` loop can block indefinitely on a stalled/partial local request (local DoS of the connect flow). Add a read deadline and cap header bytes. (security + adversarial + reliability, all P3)
- [ ] Loopback writes the **success** page to the browser before validating `state`/`code` (`loopback_server.rb`): on a state-mismatch/missing-code callback the browser shows "Screenote connected" while the CLI aborts. Choose success/failure after validation.
- [ ] `http://` non-loopback `base_url` sends the bearer token in cleartext: `validate_screenote!` accepts any `http(s)://` and `http_options` derives `use_ssl` purely from the scheme. Default is `https://screenote.ai`, so this only bites a deliberately misconfigured self-host — consider restricting plaintext `http` to loopback hosts.
- [ ] `connect`/`disconnect` crash on a corrupt local `screenote.json`: `CredentialStore#load` raises `ConfigError` and neither command rescues it (`connect.rb:37`, `disconnect.rb` `call`), so an operator can neither reconnect nor clear a bad file via the CLI. Treat a corrupt file as "no credential".
- [ ] `screenote_context` connected branch uses an `unless A.empty? || B.empty?` double-negation guard (`lib/hive/stages/artifacts.rb`) — rewrite as a positive `if access_token.empty? || mcp_resource.empty?` fallthrough for readability.
- [ ] `spawn_agent`'s `cli_flags ||= []; if cli_flags.empty? && cfg ...` guard (`lib/hive/stages/base.rb`) conflates `cli_flags: nil` (defer to cfg) with `cli_flags: []` (explicitly none); safe only because no current caller passes both. Use `nil`-vs-`[]` distinctly or document the intent.
- [ ] Screenote default base URL is defined in two places — `OAuthClient::DEFAULT_BASE_URL` and `Config::DEFAULTS["screenote"]["base_url"]` (plus the `disconnected()` fallback). Make one reference the other so they can't drift.
- [ ] `connect`/`disconnect` lack `long_desc` (unlike peer mutating commands) and `connect` opens the browser but suppresses the authorize URL entirely under `--json` (`commands/connect.rb` `show_authorize_url`): an operator whose browser fails to auto-open gets no fallback URL in JSON mode. Add `long_desc` and emit the URL in a structured form under `--json`. (`connect`/`disconnect` emitting prose on `Hive::Error` under `--json` is consistent with peers `update`/`prune`, so not flagged separately.)
- [ ] Wrapper-script env-scrub comment lost its threat-model rationale (`lib/hive/scripts/interactive_claude_wrapper.sh`): the new comment explains the base-url is passed explicitly but no longer documents why the `unset` still matters (token now arrives via a 0600 temp MCP config, deleted post-run) — a future author may delete the `unset` thinking it's a no-op.
- [ ] Test-coverage (branch) gaps that the 100%-line gate won't catch: `OAuthClient#revoke` without `client_id`, the `ClaudeLauncher` headless+`mcp_config_path` branch, `McpClient#extract_projects` `structuredContent`-without-projects fallthrough, `prompt_for_project` numeric edge inputs (`"0"`, out-of-range), and `screenote_context` against a non-`ConfigError` load failure (the A8 escape above). Add behavior tests for these.

ivankuznetsov:

# PR Review (pr-review-toolkit) — connect-screenote-via-oauth-2 (pass 1)

Scope: `git diff origin/main..HEAD`. Personas run: code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier. Findings are the union across personas (deduped, attributed by `[persona]`). Plan-deferred/accepted scope (refresh_token auto-refresh, codex MCP injection, server-side-URL capture default, blocked live capture test, token-to-agent exposure) was excluded. Confirmed strengths: ephemeral 0600 MCP config is cleaned up on both success and error paths and lives outside the task tree; PKCE S256 uses the RFC 7636 reference vector; CSRF `state` is validated before the code is accepted; env-scrub is consistent across all three child paths; wiki log.d/index discipline followed; live tests hit real APIs and skip cleanly.

## High
- [ ] [silent-failure] `screenote_context` rescues only `Hive::ConfigError`, but `CredentialStore#load`'s `File.read` raises `SystemCallError` (`Errno::EACCES`/`EISDIR`, TOCTOU `ENOENT`) — it escapes, `with_stage_events` (`stages/base.rb:166-168`) re-raises, and the 7-artifacts stage hard-fails instead of degrading (violates A8 fail-soft; also flagged untested by [tests]). Fix: add `rescue SystemCallError => e` → `disconnected(...)`.
- [ ] [silent-failure] `McpConfig#write!` Errno on an unwritable/full `cache_home` escapes uncaught in `spawn_artifacts_agent` (`stages/artifacts.rb:42-43`) on the *connected* path → hard-fails the stage; the `ensure` only `rm_f`s. A connected-but-unwritable cache should degrade to a no-MCP skip run, not crash (A8).
- [ ] [silent-failure] `disconnect#revoke` rescues only `Hive::Error`; an unmapped network error (TLS/timeout) raised by `discover`/`revoke` escapes `Disconnect#call` *before* `CredentialStore.clear`, so `hive disconnect screenote` crashes **and** leaves the stale local credential — the opposite of the documented fail-soft. Fix: broaden the revoke rescue to `StandardError` and/or complete `NETWORK_ERRORS`.
- [ ] [silent-failure] `NETWORK_ERRORS` is incomplete (`oauth_client.rb:13`, reused by `mcp_client.rb`): `OpenSSL::SSL::SSLError`, `Errno::ETIMEDOUT`, `ENETUNREACH`, `EPIPE` are not caught, so TLS/cert and OS-level timeout failures escape unmapped from every OAuth/MCP request (surfacing as opaque `InternalError` instead of "could not reach Screenote"; root cause of the disconnect crash above). Fix: add `OpenSSL::SSL::SSLError` and `SystemCallError` to the list.

## Medium
- [ ] [silent-failure] `LoopbackServer#wait_for_callback` writes the **success** page to the browser before validating `state`/`code` (`loopback_server.rb:26-29`): a CSRF state-mismatch or missing-code callback shows "Screenote connected" while the CLI aborts — validate first, then choose the page.
- [ ] [silent-failure] `disconnect --json` swallows the revoke-failure reason to stderr via `warn`; the JSON envelope emits only `"revoked": false` with no `reason`, so an automation consumer can't distinguish already-revoked vs unreachable vs missing-endpoint. Thread the reason into the envelope.
- [ ] [silent-failure] `McpClient#extract_projects` returns `[]` on a `JSON::ParserError` of the `content[].text` payload (`mcp_client.rb:75-76`), which `connect` then reports as "Screenote returned no projects; create a project and reconnect" — a parse failure is presented as a benign empty list with wrong remediation. Raise instead of swallowing on the text branch.
- [ ] [silent-failure] Diagnostic loudness regressed vs the deleted `ScreenoteUploader`: a reachable-but-misconfigured Screenote (200 with blank/garbage URL) used to `warn` loudly; it now silently no-ops via the agent's `screenote_skipped_reason`. Intentional per architecture (B) — confirm acceptable, or add a hive-side warn when `connected: true` but the manifest shows a skip reason.
- [ ] [code] Reused `client_id` is not re-registered against the new ephemeral loopback port (`connect.rb:42-43`): `LoopbackServer` binds `port: 0` each run, so a reconnect presents a `redirect_uri` the stored client was never registered for — an OAuth server enforcing exact redirect-URI matching (without the RFC 8252 loopback-port exemption) will reject authorize/exchange. Pin the port, persist+reuse it, or re-register when it differs.
- [ ] [code] `McpClient` sends only `Accept: application/json` (`mcp_client.rb:28`); MCP Streamable-HTTP servers frequently require `Accept: application/json, text/event-stream`. If Screenote's MCP endpoint enforces this, connect-time `list_projects` fails. Verify against the live endpoint and add the SSE accept type if needed.
- [ ] [code][tests] `credential_payload` does `token.fetch("expires_in")` (`connect.rb:122`) — a token response lacking both `expires_at` and `expires_in` raises a bare `KeyError` instead of a typed `Hive::Error`; also untested. Guard with a default/typed error.
- [ ] [tests] `prompt_for_project` input `"0"` → `projects[answer.to_i - 1]` = `projects[-1]`, silently selecting the **last** project (`connect.rb:118`); untested and flows into every artifacts upload. Reject values outside `1..projects.size` and add a test.
- [ ] [type-design] `CredentialStore` persists and returns an unvalidated free-form `Hash` (`save` only checks `is_a?(Hash)`): `{}` round-trips at 0600 with no complaint, and the "a stored credential is complete/usable" invariant is expressed nowhere. Introduce a `Credential` value object (`Data.define` + validating `from_h` + `#expired?`).
- [ ] [type-design] Credential validity is enforced in three places with no single source of truth — `connect.rb:121` (`fetch("access_token")`), `artifacts.rb:92` (token+resource non-empty), `mcp_config.rb:55-60` (`required`). A `Credential.from_h` factory would collapse these into one validated construction point.
- [ ] [type-design] `screenote_context` returns a symbol-keyed Hash wrapping a string-keyed credential Hash (`artifacts.rb:93-99`); a `screenote["connected"]` or `credential[:access_token]` typo returns `nil` silently (and sibling accesses are inconsistently loud: `.fetch(:credential)` vs `[]`). A `ScreenoteContext` `Data` class would make typos raise.
- [ ] [comments] `docs/visual-artifacts.md` is now factually stale (outside the diff but part of this migration surface): it still documents the removed REST uploader — `push_to_screenote: true`, `api_token: null` (which now triggers the migration `ConfigError`), `HIVE_SCREENOTE_API_TOKEN`, and "Hive uploads after the agent completes / the agent never receives the token" (all false now). Rewrite to the OAuth/MCP model or delete in favor of `wiki/commands/screenote.md`.
- [ ] [comments] `wiki/commands/screenote.md:52-54` `--json` field list drifts from code: it claims `connected` and `expires_at`, but `Connect#emit_success` (`connect.rb:138-152`) emits `issuer` and `client_id` and omits `connected`/`expires_at`. Reconcile doc with code.

## Nit
- [ ] [type-design][simplify] `screenote_context` happy path is a double-negation `unless A.empty? || B.empty?` guard with a trailing `disconnected(...)` fallthrough (`artifacts.rb:92-102`) — flip to a positive `if ... return disconnected` so all four reasons read as a uniform guard list and the connected hash is the tail value.
- [ ] [type-design] `Discovery` (`oauth_client.rb:18-22`) doesn't enforce its own completeness — `Discovery.new(issuer: nil)` is legal; validity lives only in the sole caller. Consider `Data.define` + a `from_responses` factory.
- [ ] [type-design] `McpClient#extract_projects` returns untyped Hashes and the id-or-`project_id` identity rule is duplicated loosely in `connect.rb:154-156`; a `project[:id]` typo yields `""`. Low severity (projects are short-lived).
- [ ] [comments] `agent.rb` `SCRUBBED_CHILD_ENV` comment dropped the threat-model "why" — it now says *what* the base_url is, not *why* it must be unset (inheriting an operator-shell base_url would be a redundant, unvalidated second source for the child). Restore the rationale.
- [ ] [comments] `interactive_claude_wrapper.sh` `unset HIVE_SCREENOTE_BASE_URL` comment ("connection context is passed explicitly by hive") no longer conveys the threat model now that a token is involved (token rides the 0600 MCP file; the unset stops the shell from overriding hive's base_url). Reword.
- [ ] [comments] `claude_launcher.rb` `build_runner` env comment has the same "what, not why" gap; also trim "for the artifacts stage" — this tmux env block is built for every claude session, not just artifacts. Keep its cross-reference to the sibling scrub sites.
- [ ] [simplify] The base_url fallback is duplicated between `disconnected` and the connected branch (`artifacts.rb`); extract `config_base_url(cfg)` / `connected_base_url(credential, cfg)`. (Keep the `|| DEFAULT_BASE_URL` fallback — it's load-bearing for the `render_prompt` empty-cfg default-arg path.)
- [ ] [simplify] `OAuthClient` and `McpClient` duplicate HTTP plumbing and `McpClient` reaches into `OAuthClient::{NETWORK_ERRORS,OPEN_TIMEOUT_SEC,READ_TIMEOUT_SEC}`; extract a `Screenote::Http` module (constants + `options` + `request`) to remove the cross-class constant reach. Leave the two JSON parsers unless the empty-body MCP path is confirmed unused.
- [ ] [simplify] Duplicated 0600 `File.write(perm:)` + `File.chmod` in `CredentialStore#save` and `McpConfig#write!`; extract a `SecureFile.write_json` helper that co-locates the why-two-operations comment (the redundant chmod is intentional — keep it).
- [ ] [simplify] `connect.rb#project_id` (`:154-156`) evaluates `project["id"].to_s` twice — assign it once.
- [ ] [simplify] `claude_launcher.rb` computes `mcp_flags` unconditionally (`:132`) but consumes it only on the headless branch (the tmux path recomputes via `wrapper_command`); move the computation inside the headless `if`.
- [ ] [tests] The non-numeric `prompt_for_project` answer (raw-string → `choose_project` `find`-by-id) path is untested — add a literal project-id-string case and a no-match case.
- [ ] [tests] `OAuthClient#revoke` with a present token but **nil** `client_id` is untested (the `fields[:client_id] = client_id if client_id` omit-branch), as is `disconnect` with `access_token` present but `client_id` absent.
- [ ] [tests] The `launch!` headless branch with a non-nil `mcp_config_path` (`--mcp-config` + `--allowedTools` assembly) is untested — only the tmux `wrapper_command` path is covered.
- [ ] [tests] `Stages::Base.spawn_agent`'s explicit-`cli_flags`-suppresses-config-derivation short-circuit (commit 01841e12) isn't asserted at the `spawn_agent` seam — add a test passing explicit flags + a cfg with model/effort and assert the explicit flags win.
- [ ] [tests] `connect_test.rb:237` uses `instance_variable_get(:@mcp_client_factory)` — brittle coupling to private ivar/lambda shape; exercise the default factory through `call` with an injected http seam instead.

ivankuznetsov:

# Code Review — connect-screenote-via-oauth-2 (pass 2)

Scope: `git diff origin/main..HEAD` (base `563dd96`). New Screenote OAuth 2.1 +
MCP connect/capture path replacing the ripped-out REST uploader.
Reviewers: correctness, security, adversarial, reliability, testing,
maintainability, cli-readiness. Findings below survived cross-reviewer
dedup/agreement gating. The plan's blocked live-capture test (A4) and the
accepted token-exposed-to-agent invariant (A1) are deliberately not flagged.

## High
- [ ] MCP tool-level errors are swallowed into an empty project list — `lib/hive/screenote/mcp_client.rb:38-43,60-78`: `call_tool` only checks the JSON-RPC `body["error"]`; an MCP tool error comes back as `result:{isError:true, content:[{text:"..."}]}` (rate-limited 60/min, revoked token, insufficient scope). `extract_projects` finds no projects and returns `[]`, so `connect.rb:121` raises "Screenote returned no projects; create a project and reconnect" — the wrong remediation. Flagged independently by correctness, adversarial, and testing; the code's own comment at `mcp_client.rb:73-77` shows this exact wrong-remediation hazard was understood but the `isError` shape slips through. Add an `isError` guard that raises a typed `Hive::Error` with the joined content text.

## Medium
- [ ] Loopback callback is single-accept with no path filter — `lib/hive/screenote/loopback_server.rb:31-46,69-84`: `wait_for_callback` accepts exactly one TCP connection and never filters on `/callback`. A stray first request (browser preconnect/`GET /`/`GET /favicon.ico`/empty EOF) consumes the slot, fails `callback_error` as a state mismatch, closes the server, and the real callback hits a refused port — aborting connect. Standard loopback-OAuth practice loops accepting until a request carries a valid `code`+`state`.
- [ ] Post-exchange failure abandons an unrevoked token and burns the DCR rate limit — `lib/hive/commands/connect.rb:42-67,121-125`: `exchange_code` (line 56) runs before the single `save` (line 67). If `list_projects` raises/returns `[]` or the project prompt is cancelled, the live access token is never persisted and never revoked, and because `client_id` is read only from the saved credential (line 42), the next `hive connect` re-registers a fresh client — repeated retries exhaust the 10/hr/IP dynamic-registration limit. Persist `client_id` early and/or revoke on post-exchange failure.
- [ ] Ephemeral bearer-token MCP file orphans on signal/daemon-reap — `lib/hive/stages/artifacts.rb:66-68` + `lib/hive/screenote/mcp_config.rb:45-49`: the 0600 token file in `cache_home/mcp/` is removed only by an `ensure`, which does not run on SIGTERM/SIGKILL (the daemon SIGTERMs then SIGKILLs wedged children); no startup sweep of `cache_home/mcp/*` exists, so live tokens accumulate on the long-lived headless box. Add a TERM trap that unlinks the file, or a stale-file sweep at stage start.
- [ ] MCP client advertises SSE but cannot parse it — `lib/hive/screenote/mcp_client.rb:29,51-58`: `call_tool` sends `Accept: application/json, text/event-stream` but `parse_response` does a flat `JSON.parse` of the raw body. If Screenote answers the POST with a `text/event-stream` frame (which the advertised Accept invites), `connect`'s `list_projects` aborts as "unparseable response." Either negotiate JSON only or add minimal `data:` SSE-frame extraction. (Affects hive's own `McpClient` used at connect time, not the agent's separate MCP transport.)
- [ ] `connect`/`disconnect` `--json` omit the `schema`/`schema_version` envelope — `lib/hive/commands/connect.rb:110,183`, `lib/hive/commands/disconnect.rb:73,82`: every other agent-callable command tags its `--json` output with `schema`+`schema_version` from `Hive::Schemas::SCHEMA_VERSIONS` (e.g. `forget.rb:81-82`), and neither `hive-connect` nor `hive-disconnect` is registered in `lib/hive.rb:15`. Agent consumers can't use the standard version-detection contract on these commands.

## Nit
- [ ] HTTP 429 surfaces with no Retry-After guidance — `lib/hive/screenote/http.rb:42`: any non-2xx (including rate-limit 429) maps to a generic `"... failed (HTTP 429)"`; the `Retry-After` header is ignored, so the operator gets no hint when to retry `hive connect`.
- [ ] `write_response` socket write is unrescued — `lib/hive/screenote/loopback_server.rb:98-102`: a browser disconnect mid-write raises `Errno::EPIPE`/`IOError` (not `Hive::Error`), which escapes `bin/hive`'s `Hive::Error`/`Thor::Error` boundary as a raw backtrace. The code/state are already parsed by then; rescue the write.
- [ ] Loopback request-line read is unbounded before the byte cap — `lib/hive/screenote/loopback_server.rb:71,88-96`: `MAX_REQUEST_BYTES` is enforced only on header lines; the request line is read with `socket.gets` (no length limit), so a local client can stream an unbounded first line within the 5s deadline, contradicting the file's "unbounded header bytes" rationale. Read the request line with a byte cap.
- [ ] `extract_projects` raises bare `TypeError` on a non-Hash `result` — `lib/hive/screenote/mcp_client.rb:60-61`: `call_tool` returns `body.fetch("result", body)`; a non-conformant server returning a JSON array/string makes `result["projects"]` raise `TypeError` instead of a typed `Hive::Error`. Guard with `return [] unless result.is_a?(Hash)`.
- [ ] SecureFile has a brief umask-mode window — `lib/hive/screenote/secure_file.rb:20`: `File.write(perm: 0o600)` then `File.chmod(0o600)` leaves the credential file at umask-default mode (e.g. 0644) between create and chmod; the MCP file's random name defeats the race but the predictable `~/.config/hive/screenote.json` does not on a multi-user box. Set the mode on the open fd before writing.
- [ ] Duplicated JSON-parse helper — `lib/hive/screenote/oauth_client.rb:125-135` and `lib/hive/screenote/mcp_client.rb:51-58` are near-identical; the empty-body difference can be a keyword arg on a shared `Hive::Screenote::Http` helper so error wording/guards live in one place.
- [ ] Duplicated `ensure_screenote!` — `lib/hive/commands/connect.rb:80-84` and `lib/hive/commands/disconnect.rb:34-38` are verbatim copies differing only by verb; extract a shared guard before a second connect service is added.
- [ ] `project_id` dual-key fallback leaks into the command layer — `lib/hive/commands/connect.rb:197-200` silently falls back `project["id"]` → `project["project_id"]`; normalize to one canonical key inside `McpClient#extract_projects` instead of propagating the wire ambiguity to every caller.
- [ ] `resolved_base_url` does a full config disk read to recover a constant — `lib/hive/commands/connect.rb:97-102` (and `disconnect.rb:58`) call `load_global_screenote.fetch("base_url")` for the blank-URL fallback, which is just `OAuthClient::DEFAULT_BASE_URL`; a missing/malformed `config.yml` then fails with `ConfigError` at the moment the operator is trying to connect. Fall back to the constant unless an override is actually configured.
- [ ] Untested error-shape behaviors despite the 100% line gate — no test exercises the MCP `isError:true` response (`test/unit/screenote/mcp_client_test.rb`), a non-`/callback`/empty first loopback request (`test/unit/screenote/loopback_server_test.rb`), ephemeral-file cleanup under a signal, or `connect` failing after `exchange_code` (token-abandonment invariant); these are the behaviors behind the High/Medium findings above.

ivankuznetsov:

# PR Review Toolkit — connect-screenote-via-oauth-2 (pass 2)

Scope: `git diff origin/main..HEAD` (base `563dd96`). Union of findings from the
pr-review-toolkit personas: `[code]` code-reviewer, `[tests]` pr-test-analyzer,
`[errors]` silent-failure-hunter, `[type-design]` type-design-analyzer,
`[comments]` comment-analyzer, `[simplify]` code-simplifier. Combined prefixes
mark findings surfaced by more than one persona. The plan's blocked live-capture
test (A4) and the accepted token-exposed-to-agent invariant (A1) are not flagged.

## High
- [ ] [errors/tests] MCP tool-level errors masked as "no projects" — `lib/hive/screenote/mcp_client.rb:39`: `call_tool` only inspects the JSON-RPC protocol error (`body["error"]`), never the MCP tool-execution channel (a successful response whose `result` is `{isError:true, content:[{text:...}]}`). For `list_projects`, an `isError` result with plain error text (bad scope, rate-limit, 5xx) falls through `extract_projects` to `[]`, so `connect.rb:121` raises the misleading "Screenote returned no projects; create a project and reconnect" — the exact wrong-remedy masking the code's own comment at `mcp_client.rb:72-77` set out to avoid. Add an `isError` guard that raises a typed `Hive::Error` with the joined content text, and a test driving it.
- [ ] [errors/tests] `connect` failure after token exchange abandons a live token — `lib/hive/commands/connect.rb:56-67`: between `exchange_code` and the single `save`, any raise from `list_projects`/`choose_project`/`credential_payload`/`save` leaves the issued access token unrevoked and a freshly-registered `client_id` unpersisted (so the next run re-registers, burning the 10/hr/IP DCR limit). Worse, a `SystemCallError` from `save` (ENOSPC/EACCES) is not a `Hive::Error`, so it escapes `bin/hive`'s rescue as a raw backtrace. Wrap with rescue mapping `SystemCallError` to a typed error and best-effort `revoke` on post-exchange failure.

## Medium
- [ ] [code] `connect --json` is unusable non-interactively — `lib/hive/commands/connect.rb:120-149`: under `--json` with no `project_picker`, `prompt_for_project` writes human prompt text ("Select a default Screenote project:", numbered list) onto the same stdout carrying the JSON lines (corrupting the stream) and blocks on `@input.gets`, which returns EOF→`nil`→"selection is required". There is no `--project ID` flag and no single-project auto-select, so `--json` connect cannot complete project selection in automation. Skip the interactive prompt under `--json`; auto-select a lone project or emit a structured `needs_project_selection` envelope.
- [ ] [code] `connect`/`disconnect` emit no structured JSON error envelope on failure — `lib/hive/commands/connect.rb:34-69`, `lib/hive/commands/disconnect.rb:34-37`: sibling commands (`run`, `status`) call `emit_error_envelope` under `--json` so a consumer always gets a parseable document; here `Hive::Error` propagates to `bin/hive` and only `warn`s plain text to stderr, so a `--json` failure (discovery error, no projects, state mismatch, unsupported service) produces no JSON at all. Wrap `call` to emit a typed `{"ok":false,...}` envelope before the non-zero exit.
- [ ] [code/tests] MCP client advertises SSE but cannot parse it — `lib/hive/screenote/mcp_client.rb:29,51-58`: `call_tool` sends `Accept: application/json, text/event-stream` but `parse_response` does a flat `JSON.parse` of the raw body; an actual `text/event-stream` reply (`event: message\ndata: {…}`) fails as "unparseable response", breaking `connect`'s `list_projects`. The only check of real server behavior is the skip-blocked live test. Negotiate JSON only, or add minimal `data:`-frame extraction.
- [ ] [errors] `disconnect` revoke rescue is wider than the transport surface — `lib/hive/commands/disconnect.rb:62-69`: `OAuthClient`/`Http` already normalize network/TLS/HTTP failures to `Hive::Error`, so `rescue StandardError` only *adds* coverage of genuine bugs — a `KeyError` from `load_global_screenote.fetch("base_url")` (line 58) or any `NoMethodError`/`TypeError` from a future refactor is silently downgraded to `revoked:false, reason:"<bug>"` and the credential cleared anyway. Narrow to `Hive::Error` (or an explicit transport-class list).
- [ ] [errors] Loopback `write_response` is unrescued and can mask the real callback failure — `lib/hive/screenote/loopback_server.rb:39,98-102`: the page write runs before `raise Hive::Error, error`; an `Errno::EPIPE`/`ECONNRESET` from `socket.write` (browser closed the tab) pre-empts and masks the intended state-mismatch/missing-code error, escaping as an unmapped `SystemCallError` backtrace. Rescue `SystemCallError, IOError` around the cosmetic write so the pending `raise` still delivers the real reason.
- [ ] [tests] Non-`/callback`/empty first loopback request is untested and confusing — `lib/hive/screenote/loopback_server.rb:81-83`: `wait_for_callback` accepts exactly one socket and does not skip non-callback probes; a browser/OS prefetch of `/` or `/favicon.ico` (or an empty/EOF first line) yields `{}` params and surfaces as "state mismatch", aborting a legitimate connect with a misleading error. Ignore non-`/callback` requests and keep waiting; add a test pinning the behavior.
- [ ] [errors] `mcp_config` write rescue catches only `SystemCallError` — `lib/hive/stages/artifacts.rb:46-52`: `McpConfig#payload`→`required` raises `Hive::ConfigError` on a blank `access_token`/`mcp_resource`; it is unreachable only because `screenote_context` (artifacts.rb:100) pre-validates, so if those two completeness checks ever drift a `Hive::ConfigError` from `write!` escapes and hard-fails the stage, defeating the A8 fail-soft this block exists to provide. Broaden to `rescue SystemCallError, Hive::ConfigError`.
- [ ] [type-design] Credential is a free-form string-keyed Hash with the required-field invariant re-implemented ad hoc — `lib/hive/commands/connect.rb:151-166` (producer) vs `lib/hive/screenote/mcp_config.rb:53` (checks access_token/mcp_resource, omits project_id), `lib/hive/stages/artifacts.rb:96-102` (checks all three), `lib/hive/screenote/credential_store.rb:59` (expires_at). Three subtly different copies of "what makes a credential usable", any key typo silently degrades to a soft-fail. Introduce a `Screenote::Credential` value object (mirroring the existing `Discovery` struct) with one validating factory, `usable?`/`expired?`/`to_mcp_payload`, and a redacting `inspect`.
- [ ] [errors] `screenote_context` fail-soft degrades silently — `lib/hive/stages/artifacts.rb:111-118`: the `Hive::ConfigError`/`SystemCallError` rescues fold the reason into the agent prompt but never `warn` to stderr, so an operator watching a run sees a degraded no-upload artifacts stage with no visible cause (and a future unrelated `Hive::ConfigError` would be absorbed the same way). Add a `warn "[hive] Screenote disabled for artifacts: #{reason}"`.

## Nit
- [ ] [simplify] Duplicated JSON-object parse — `lib/hive/screenote/oauth_client.rb:125-135` and `lib/hive/screenote/mcp_client.rb:51-58` are near-identical (same "non-object"/"unparseable" messages); extract `Hive::Screenote::Http.parse_json_object`, keeping OAuth's empty-body→`{}` guard.
- [ ] [simplify] Duplicated `ensure_screenote!` guard — `lib/hive/commands/connect.rb:80-84` and `lib/hive/commands/disconnect.rb:34-38` differ only by the verb in the message; a verb-parameterized shared helper removes the copy.
- [ ] [simplify] Repeated global-config base_url fallback — `lib/hive/commands/disconnect.rb:58` vs `lib/hive/commands/connect.rb:97-102` both `load_global_screenote.fetch("base_url")`; share a `Screenote.global_base_url` helper (the `artifacts.rb` cfg/`DEFAULT_BASE_URL` variant is genuinely different and stays separate).
- [ ] [comments] `disconnect` `reason` overclaim — `lib/hive/commands/disconnect.rb:50-53` and the `cli.rb` disconnect `long_desc` claim `reason` distinguishes "already-revoked", but RFC 7009 idempotent revoke always reports `revoked:true`; actual `reason` values are only `no_token`/`unreadable_credential`/raw error message. Reword both to match emitted values.
- [ ] [comments] Loopback byte-cap comment overstates protection — `lib/hive/screenote/loopback_server.rb:9-14`: the cap counter also seeds the request line (not "header" bytes alone), and `read_line` uses `socket.gets` with no per-line limit, so a single unterminated line is read unbounded before the between-lines cap fires. Soften the comment; optionally cap the request-line read too.
- [ ] [comments] Imprecise inline reference — `lib/hive/config.rb:1869` names `commands/connect.rb#load_global_screenote` as if the method lived on `Connect`; it is `Hive::Config.load_global_screenote`, invoked from `Connect#resolved_base_url`.
- [ ] [tests] Loopback `MAX_REQUEST_BYTES` overflow branch has no behavioral test — `lib/hive/screenote/loopback_server.rb:77-79`: add a test writing >64 KB of header bytes and asserting the raised `Hive::Error`.
- [ ] [tests] `disconnect` broadened rescue only exercised by a `Hive::Error` subclass — `test/unit/commands/disconnect_test.rb`: a non-Hive transport error (`Errno::ETIMEDOUT`/`OpenSSL::SSL::SSLError`) from `discover` is untested, so narrowing the rescue would not fail any test; add one to lock the contract.
- [ ] [tests] Blocked capture test leaves the upload round-trip with zero live verification — `test/integration/screenote_capture_live_test.rb:28`: gating is correct (genuine skip-with-reason), but the `create_screenshot_upload`→signed-PUT path has no automated coverage; attach a tracking reference so it is enabled once the Screenote endpoint ships.
- [ ] [type-design] `CredentialStore#save` accepts any Hash — `lib/hive/screenote/credential_store.rb:38`: the store cannot guarantee a persisted blob is a usable credential, only that it is a Hash; a `Credential` value object's `to_h` would constrain this.
- [ ] [type-design] Credential secrets have no redacting `inspect` — `lib/hive/commands/connect.rb:151`: the hash carries `access_token`/`refresh_token`; a stray log or exception dump would print them in full. A value object with a redacting `inspect` is defense-in-depth.
- [ ] [type-design] `register`/`exchange_code` return raw response hashes — `lib/hive/screenote/oauth_client.rb:44,76`: `connect.rb:43,63` index `["client_id"]`/`.fetch("access_token")` on untyped bags; wrap into typed objects at the boundary.
- [ ] [type-design] `Discovery.new` is an unvalidated bare Struct — `lib/hive/screenote/oauth_client.rb:13`: the all-endpoints-present invariant holds only on the `discover()` path, not the type; a `from_responses` factory (or private constructor) would enforce it on the type itself.
- [ ] [type-design] `McpConfig#write!` creates a file whose cleanup the caller owns — `lib/hive/screenote/mcp_config.rb:45` + `lib/hive/stages/artifacts.rb:67`: the create/delete invariant is split across two files; a `McpConfig#with_temp_config { |path| ... }` block form makes it unbreakable from outside.

ivankuznetsov:

# Code Review — connect-screenote-via-oauth-2 (pass 3)

Scope: `git diff origin/main..HEAD` (merge-base `563dd96`, 13 commits). OAuth 2.1 / MCP connect+capture replacing the api-key REST uploader.
Reviewers: correctness, security, adversarial, reliability, testing, maintainability, cli-readiness, api-contract, project-standards, agent-native, learnings.

Pass-1/2 status re-verified against the current tree: MCP `isError` tool-channel surfacing, post-exchange best-effort revoke, the `--json` `needs_project_selection` envelope + lone-project auto-select, SSE `data:` de-framing, `emit_error_envelope` on both commands, the loopback `write_response` rescue, the fail-soft `warn` in `screenote_context`, the broadened `McpConfig` rescue, non-`/callback` prefetch tolerance, and the `expires_in` typed error are all present. Plan A1–A8 / U1–U6 remain implemented; **no plan-required-but-missing gaps**. Plan-deferred scope excluded (refresh_token auto-refresh, codex MCP injection, server-side-URL-capture default, the blocked live capture test, the consciously-accepted token→agent MCP-config exposure). Findings below are net-new for this pass, except the two clearly-marked carry-over escalations (Q1/Q2 from `escalations-02.md`, still unanswered).

## High
- [ ] `connect` revokes the just-saved token when the success line fails to write (`lib/hive/commands/connect.rb:107-119`, `emit_success` at :115): pass-02 wrapped the post-exchange block in `rescue StandardError => best_effort_revoke`, but `emit_success` is the **last statement inside that rescue** and does an unguarded `@output.puts JSON.generate(...)`. `hive connect screenote --json | head -1` (or any piped consumer) closes the pipe after the `authorize` line, so the success `puts` raises `Errno::EPIPE`; `best_effort_revoke` then revokes the bearer that `save_credential` already wrote to disk. Result: `screenote.json` advertises a connected credential whose token is server-side revoked — the operator believes they are connected while every headless artifacts run silently fail-soft-skips. Move `emit_success` out of the revoke-scoped block (run it after the protected section returns) or guard its `puts` against `Errno::EPIPE`/`IOError`. (correctness + testing, cross-confirmed; net-new regression from commit `bfea854b`)

## Medium
- [ ] `needs_project_selection` `--json` envelope omits `error_kind` and `exit_code` (`lib/hive/commands/connect.rb:210-220`): every other `{"ok":false}` line from connect/disconnect carries `error_kind` and `exit_code`; this one (added in pass-02) does not, and the exit code is the generic `1`. Automation that branches on those fields — or on exit code to tell "retry with a project selection" from an unrecoverable auth/network failure — gets `nil`/ambiguity. Add `error_kind`/`exit_code` to the payload (and consider a distinct usage-class exit code). (api-contract + cli-readiness)
- [ ] `disconnect` not-connected envelope is shape-inconsistent with success and with its own docs (`lib/hive/commands/disconnect.rb:90-97`): `emit_absent` emits `disconnected:false` + `reason` but **no `revoked` field**, while `emit_disconnected` always includes `revoked` (and the `cli.rb` `long_desc` documents `{ "disconnected", "revoked", "reason" }` as the shape). A consumer that unconditionally reads `result["revoked"]` gets `nil` on the not-connected path. Add `"revoked" => false` to `emit_absent` (or carve out the case in `long_desc`). (api-contract + cli-readiness)
- [ ] Malformed loopback request path crashes `connect` with a raw backtrace (`lib/hive/screenote/loopback_server.rb:99-101`): `read_callback_request` calls `URI("http://#{host}#{path}")` **before** the `callback_path?`/404 check, so a local prefetch/scanner sending `GET /x%ZZ` (or a path with a raw `[`/control byte) raises `URI::InvalidURIError`. That class is neither `Hive::Error` nor `SystemCallError`, so it escapes `Connect#call`'s rescues (no `--json` envelope, `@error_emitted` never set) and `bin/hive` (Thor/`Hive::Error` only) as an uncaught Ruby backtrace — defeating pass-02's explicit prefetch-tolerance design. Rescue `URI::InvalidURIError` in `read_callback_request` and treat it as a non-callback request (404 + keep waiting). (adversarial, conf 75)
- [ ] `Net::WriteTimeout` is not mapped to `Hive::Error` (`lib/hive/screenote/http.rb:26`): `NETWORK_ERRORS` lists `Net::OpenTimeout`/`Net::ReadTimeout` but not `Net::WriteTimeout` (a `Timeout::Error` < `RuntimeError`, and `write_timeout` is left at the 60s default). A stalled POST-body write to any Screenote endpoint escapes `Http.request` and the connect rescues as an unmapped `RuntimeError` backtrace with no `--json` error document. Low practical likelihood (bodies are tiny) but the fix is one constant; add `Net::WriteTimeout` to `NETWORK_ERRORS`. (reliability, conf 100)
- [ ] Changelog fragment uses YAML frontmatter instead of the heading convention (`wiki/log.d/20260622T120521Z-screenote-oauth-mcp.md:1-5`): every other fragment (and the sibling `20260622T140000Z-…` one) starts with `## [ISO-8601Z] scope — title`; the shell compiler concatenates fragments verbatim, so `hive wiki compile-log` will splice raw `---\ndate:…\n---` YAML into `wiki/log.md`. Replace the frontmatter with a `## [...]` heading entry. (project-standards)
- [ ] (carry-over Q1, unanswered across passes 1–2) Reconnect reuses the cached `client_id` with a fresh ephemeral loopback `redirect_uri` (`lib/hive/commands/connect.rb:83-84` + `loopback_server.rb` `127.0.0.1:0`): `register` is skipped when `client_id` is cached, but the redirect port changes every run, so an OAuth/Doorkeeper server doing exact `redirect_uri` matching (RFC 6749, without the RFC 8252 loopback-port exemption) would reject authorize/exchange on every reconnect. The unit test masks this with a fixed-port fake; the live discovery test that would settle Screenote's policy is env-gated/blocked. Escalate, do not auto-fix (every fix trades against the plan's 10/hr DCR rate-limit mitigation).
- [ ] (carry-over Q2 security sub-point, unanswered) Credential Hash carries the bearer/`refresh_token` with no redacting `inspect` (`lib/hive/commands/connect.rb` `credential_payload` + `credential_store.rb`): a stray `p credential`, an interpolated `raise`, or a backtrace capturing locals would print the secrets in full. A minimal `Screenote::Credential` wrapper (or a redacting `inspect`/`to_s`) is the cheap hardening; the broader value-object refactor stays deferred. Escalate.

## Nit
- [ ] `emit_error_envelope` JSON payload is duplicated near-identically across `lib/hive/commands/connect.rb:136-150` and `lib/hive/commands/disconnect.rb:114-125` (only the `@error_emitted` guard differs); extract one shared builder so a future field rename happens once. (maintainability)
- [ ] `cli_flags` nil-vs-`[]` sentinel is a hidden contract (`lib/hive/stages/base.rb:425-431`): `nil` means "derive from cfg", `[]` means "explicitly none — do not derive". A caller passing `[]` expecting "empty list" would silently get derivation. An explicit `derive_flags:` keyword would name the two intents instead of encoding them in nil-vs-empty. (maintainability)
- [ ] `disabled` vs `disconnected` naming hides the warn-and-delegate vs silent-builder split (`lib/hive/stages/artifacts.rb:137-150`); a name like `warn_and_build_unavailable` / `build_unavailable_context` makes the `render_prompt` default-arg path (which deliberately skips the warn) self-evident. (maintainability)
- [ ] `spawn_artifacts_agent` default arg `screenote: screenote_context(cfg)` (`lib/hive/stages/artifacts.rb:27`) performs a real `CredentialStore.new.load` against the dev machine for any test that omits the kwarg — ambient disk coupling. The sole production caller already passes `screenote:`; make the kwarg required and have those tests pass an explicit stub. (maintainability)
- [ ] `McpConfig::TOOL_NAMES` whitelists `mcp__screenote__create_screenshot` and `…create_multi_viewport_screenshot` (`lib/hive/screenote/mcp_config.rb:12-17`) but the artifacts prompt never mentions them (only `create_screenshot_upload`); either document when to use them or trim them from the allow-list until the prompt covers them, so the agent doesn't reach for an undescribed/unavailable tool. (agent-native)
- [ ] `unframe_sse` joins the `data:` lines of *all* SSE events with newlines (`lib/hive/screenote/mcp_client.rb:83-92`): a spec-compliant Streamable-HTTP server that emits a `notifications/*` event before the result on the same `tools/call` stream would concatenate two JSON objects into an unparseable body, failing every `list_projects`. Select the result event (by JSON-RPC `id`) rather than blind-joining. Confidence ~50 — depends on whether Screenote multiplexes notifications; verify against the live server. (correctness + adversarial, needs-verification)
- [ ] `connect` `long_desc` documents only the `authorize` + success `--json` lines, not the `needs_project_selection` envelope or the fixed 300s loopback callback timeout (`lib/hive/cli.rb`); an agent driving the command can't discover the multi-project two-step or set an appropriate outer deadline. Document both. (cli-readiness)
- [ ] Fail-soft branches reachable for line-coverage but not behavior-asserted: the `Hive::ConfigError` degrade arm of `spawn_artifacts_agent`'s rescue (only `SystemCallError` is exercised), `best_effort_revoke` firing when `credential_payload` raises, and the `disabled()` `warn` output across the disconnected sub-paths. Add targeted assertions so an A8 regression fails a test rather than the gate passing on shared line coverage. (testing)
- [ ] `web/test/integration/tasks_test.rb:589,596` and `web/test/system/pipeline_flow_test.rb:398,405` still set the removed `push_to_screenote` item field, and no fixture exercises the new `screenote_skipped_reason` — inert at runtime (the web reader reads only `screenote_url`) but the only concrete manifest examples now mislead future authors. Drop `push_to_screenote`; add a `screenote_skipped_reason` fixture. (api-contract, pre-existing fixtures)

ivankuznetsov:

# PR Review (pr-review-toolkit) — connect-screenote-via-oauth-2 (pass 3)

Scope: `git diff origin/main..HEAD` (merge-base `563dd96`). OAuth 2.1 / MCP connect+capture replacing the api-key REST uploader.
Personas run: code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier. Union of findings below.

Pass-1/2 fixes verified present and NOT re-reported: MCP `isError` tool-channel surfacing, post-exchange best-effort revoke, the `--json` `needs_project_selection` envelope + lone-project auto-select, SSE `data:` de-framing, `emit_error_envelope` on both commands, the loopback `write_response` rescue, the fail-soft `warn`, the broadened `McpConfig` rescue, and non-`/callback` prefetch tolerance. Plan A1–A8 implemented; no plan-required-but-missing gaps. Plan-deferred scope excluded (refresh_token auto-refresh, codex MCP injection, server-side-URL default, blocked live capture test, accepted token→agent MCP-config exposure). The carry-over `Screenote::Credential` value-object item (escalation Q2, still unanswered) is re-affirmed below, not re-escalated.

## High
- [ ] [code-reviewer + silent-failure-hunter + comment-analyzer] `finalize_connection`'s broad `rescue StandardError` revokes the just-saved token when only `emit_success` fails (`lib/hive/commands/connect.rb:107-119`, `emit_success` at :115): `save_credential` already wrote `screenote.json`, but `emit_success` does an unguarded `@output.puts JSON.generate(...)`; a closed pipe (`hive connect screenote --json | head -1`) raises `Errno::EPIPE` after the save, so `best_effort_revoke` kills the persisted bearer and the command reports failure. The on-disk credential then looks connected while its token is server-revoked — every headless artifacts run silently fail-soft-skips. Trigger is narrow (broken stdout between save and print) but the outcome is a silent wrong-state. Move `emit_success` out of the revoke-scoped region and guard its `puts` against `Errno::EPIPE`/`JSON::GeneratorError` (as `emit_error_envelope` already does). Net-new regression from pass-02's revoke hardening. (The `finalize_connection` comment at :101-106 also omits `emit_success` from the revoke-triggering scope it enumerates.)

## Medium
- [ ] [code-reviewer + silent-failure-hunter] Malformed loopback callback request line escapes unmapped and crashes connect (`lib/hive/screenote/loopback_server.rb:99-101`): `read_callback_request` builds `URI("http://#{host}#{path}")` **before** the `callback_path?`/404 check, so a local request with an invalid percent-escape (`GET /callback?code=%ZZ HTTP/1.1`) or a raw space raises `URI::InvalidURIError` (an `ArgumentError`). It is neither `Hive::Error` nor `SystemCallError`, so it escapes `Connect#call`'s rescues and `bin/hive`'s top-level rescue as a raw backtrace with no `--json` envelope — defeating the deliberate "tolerate junk local requests" design. Rescue `URI::Error`/`ArgumentError` in `read_callback_request` and treat it as a non-`/callback` request (404 + keep waiting), or map to a typed `Hive::Error`. (Confirmed on Ruby 3.4.)
- [ ] [silent-failure-hunter] `Net::WriteTimeout` is not in `Http::NETWORK_ERRORS` (`lib/hive/screenote/http.rb:26`): it is `Timeout::Error < RuntimeError` (not `SystemCallError`/`SocketError`/`IOError`), and `write_timeout` defaults to 60s (not disabled in `Http.options`), so a stalled POST-body write to any Screenote endpoint (registration, token exchange, revoke, MCP `tools/call`) escapes `Http.request`'s `rescue *NETWORK_ERRORS` as a raw `Net::WriteTimeout` backtrace instead of the typed "could not reach Screenote" + `--json` envelope. Add `Net::WriteTimeout` to `NETWORK_ERRORS`. (Low practical likelihood — bodies are tiny — but a one-constant fix.)
- [ ] [comment-analyzer] `validate_screenote!` migration-guard comment inaccurately claims connect "stays reachable" (`lib/hive/config.rb:1872-1880`): a non-blank `api_token` still raises `ConfigError`, and bare `hive connect screenote` loads exactly this validation (`resolved_base_url` → `global_screenote_base_url` → `load_global_screenote` → `validate_screenote!`), so the real migration case (an operator with a leftover REST token, which is non-blank — the only case the guard fires on) still bricks bare connect; only `--base-url` or manually removing the key escapes. The A7 migration error itself is intended, but the comment (and the mirrored test comment at `test/unit/config_test.rb:2799-2803`) overstates the fix and will mislead a maintainer into thinking the catch-22 is fully resolved. Reword: blank/null leftovers are now inert, but a non-blank token still hard-fails bare connect.
- [ ] [type-design] (carry-over Q2, re-affirmed — still unanswered) Credential is a free-form string-keyed Hash whose "usable?" invariant is re-derived with three divergent key subsets across `connect.rb:243-257`, `stages/artifacts.rb:104-122`, `screenote/mcp_config.rb:30-58`, and `credential_store.rb:57-64` (`{access_token, mcp_resource}` vs `{…, project_id, not-expired}` vs `{expires_at}`), and the credential plus the `register`/`exchange_code` return Hashes carry `access_token`/`refresh_token` with no redacting `inspect`, so a stray `p`, exception `#inspect`, or backtrace capturing those locals (e.g. the broad `rescue StandardError` in connect/disconnect) prints the live bearer. Fix: a `Screenote::Credential` plain value class (validating factory, `usable?`/`expired?`/`to_mcp_payload`, redacting `inspect`) — a typed view over the on-disk JSON Hash, no OpenStruct, Hash-by-plan preserved.
- [ ] [type-design] `Discovery` struct enforces no required fields (`lib/hive/screenote/oauth_client.rb:13-17`): `keyword_init: true` only rejects unknown keys, so `Discovery.new(issuer: "x")` succeeds with `mcp_resource`/`token_endpoint` nil; only the single `discover` call site validates via `required(...)`, and a nil would propagate into `URI(metadata.authorization_endpoint)` (ArgumentError) or the stored `mcp_resource`. Add a validating `Discovery.from_endpoints(...)` factory (or raise-on-blank `initialize`) so the one new value type can't be built incomplete.
- [ ] [type-design] `expires_at` coercion is split and the pass-through branch isn't normalized (`lib/hive/commands/connect.rb:244,264-271` vs `lib/hive/screenote/credential_store.rb:59-62`): the computed (`expires_in`) branch yields an ISO8601 string, but the server's raw `token["expires_at"]` is stored verbatim. If Screenote ever returns it as a Unix-epoch integer (a common OAuth shape), `expired?` does `Time.parse("1234567890")` → `ArgumentError` → rescued to `true`, so a freshly minted token reads as already-expired and artifacts silently disables Screenote. Funnel both branches through one normalizer returning a guaranteed ISO8601 string.

## Nit
- [ ] [silent-failure-hunter] A `chmod` failure during MCP-config write orphans the bearer file with no cleanup and no path in the warning (`lib/hive/screenote/secure_file.rb:18-23` + `lib/hive/stages/artifacts.rb:44-79`): `write_json` creates the 0600 file then `File.chmod`; if chmod raises, `write_json` never returns the path, so `mcp_config_path` stays nil, the `ensure` `rm_f` is skipped, and the just-written `Authorization: Bearer …` file is left on disk while the fail-soft warning prints only the exception message (not the leaked path). Have `write_json` unlink its partial file on chmod failure (or capture the path so `ensure` can `rm_f` it) and include the path in any leak warning. (Narrow: chmod rarely fails right after a same-path create.)
- [ ] [silent-failure-hunter + pr-test-analyzer] A non-Hash JSON-RPC `error` degrades `list_projects` to the misleading "no projects" (`lib/hive/screenote/mcp_client.rb:38-44`): the guard only fires when `body["error"].is_a?(Hash)`; a non-conformant `"error":"<string>"` reply skips the raise, falls through `extract_projects` to `[]`, and connect surfaces the wrong "create a project and reconnect" remedy — the same misleading class the `isError`/unparseable guards were added to close. Treat any truthy `body["error"]` as a protocol error, and add a string-error test.
- [ ] [code-simplifier] The 6-key `emit_error_envelope` payload is duplicated near-identically in `lib/hive/commands/connect.rb:136-150` and `lib/hive/commands/disconnect.rb:114-125` (only connect's `@error_emitted` guard differs); extract one shared builder (cf. existing `Hive::ErrorEnvelope.build` at `lib/hive.rb:86`) returning the hash, keeping each command's distinct emission control flow. Byte-identical output, coverage preserved.
- [ ] [type-design] `project_id` extraction (`project["id"]` || `project["project_id"]`) is duplicated over the untyped raw-MCP-project Hash in `lib/hive/commands/connect.rb:289-291` and `lib/hive/stages/artifacts.rb:107-108`; normalize once at the `McpClient` seam (e.g. `list_projects` returns `{id:, name:}`) so a Screenote key change touches one place. Lowest-priority type item.
- [ ] [pr-test-analyzer] Net-new error-mapping branches reachable for line coverage but behavior-unasserted — add targeted tests: the top-level `SystemCallError`→typed-error mapping in `connect.rb:40-47` (inject a non-save `Errno::ENOSPC`) and `disconnect.rb:25-32` (`credential_store.clear` raising `Errno::EROFS`); the `@error_emitted` idempotency guard (assert exactly one `ok:false` line in the `needs_project_selection` test); the `http.rb` `OpenSSL::SSL::SSLError` branch; and the loopback empty/EOF first-request-line path. So a regression fails a test rather than passing on incidental coverage.
- [ ] [pr-test-analyzer] `test/unit/screenote/loopback_server_test.rb:97` drives `write_response` via `.send(:private_method)`, coupling the test to the private method name — acceptable (a real half-closed socket is flakier) but it will break on a rename with no behavior change; note as a brittleness risk.
- [ ] [code-simplifier] (optional micro) `lib/hive/commands/connect.rb:50-55` `open_browser` could pick `program = browser.empty? ? "xdg-open" : browser` then call `system(program, url, out:…, err:…)` once to de-duplicate the redirection options; marginal — current form already reads clearly.

PR #565, opened 2026-06-22
fix(new): lift recognized options out of task text in bin/hive

ivankuznetsov:
## Summary

`hive new PROJECT --workflow ID "<idea>"` silently swallowed `--workflow` (and `--depends-on`, `--json`) into the task text whenever the option sat between PROJECT and the text. The task then ran on the project's default workflow instead of the pinned one, with no error. The cruel part: `hive workflow new` prints exactly that command as its "next" hint, so dogfooding a custom workflow silently ran the coding `2-brainstorm` stage instead of the workflow's own second stage.

The bug lived entirely in the `bin/hive` wrapper. It anchored a `--` immediately after PROJECT, so any recognized option placed after PROJECT landed behind the `--` and splatted into the task text. This replaces that with a lift-and-rebuild: standalone allow-listed options are lifted out of argv wherever they appear (before PROJECT, between PROJECT and text, or after the text), and only the remaining `PROJECT TEXT...` tail is protected with `--`.

The allow-list is closed and single-sourced from the wrapper's existing constants, so it stays a deliberate set rather than a catch-all. `--json` is lifted so it stops polluting the text, but `hive new` still emits its plain-text surface (no JSON envelope here). Everything outside the allow-list stays literal task text. A value-option whose value is missing or itself option-like (e.g. trailing `--workflow`) is demoted back to literal text rather than eating the real PROJECT name.

### What gets lifted vs. kept literal

| argv element | Behavior |
|---|---|
| `--workflow ID`, `--workflow=ID` | lifted to Thor |
| `--depends-on N`, `--depends-on=N` | lifted to Thor |
| `--json`, `--no-json`, `--json=false` | lifted (but `new` ignores it) |
| `--foo`, `--help`, `--json=yes` after PROJECT | literal task text |
| `"... --workflow ..."` inside one quoted arg | literal task text |
| a trailing/value-less value-option | demoted to literal text |
| an explicit `--` tail | preserved as-is |

Docs follow the fix: the `new` long_desc, `--depends-on` examples, and the `hive workflow new` hint now all teach the one canonical ordering, `hive new PROJECT --workflow ID "<idea>"`.

## Test plan

New `test/integration/new_wrapper_argv_test.rb` drives the real `bin/hive` subprocess against on-disk project workflows — a Thor-only, in-process test cannot catch this bug, since the failure is upstream in the wrapper and an in-process registry overlay is invisible to a subprocess.

- **Centerpiece (red→green):** scaffold a real workflow via `hive workflow new`, run the exact "next" hint command, and assert the task binds to the workflow whose second stage is its own — not coding's `2-brainstorm` — and that `idea.md` keeps no literal `--workflow` token.
- Options before / between / after PROJECT all lift; `--workflow=ID` equals-form; combined `--depends-on` + `--workflow` (both orderings); `--depends-on=42` equals-form.
- A literal `--workflow` substring inside a quoted idea stays in `idea.md`; unrecognized `--foo` stays literal task text.
- Trailing / value-less and option-like-valued value-options demote to literal text instead of eating PROJECT; explicit `--` tail stays literal.
- `--json` is lifted but ignored (plain-text surface unchanged); missing-text guard fires with no empty-slug folder created.
- The pre-existing wrapper test that asserted the old swallow behavior was flipped to assert the lift, plus a `hive help new` assertion for the canonical option-before-text example.

Verification: `rubocop` clean on all changed Ruby; `new_wrapper_argv_test.rb`, `cli_version_test.rb`, `cli_usage_error_json_test.rb` (R7 regression guard), and `new_test.rb` all green; `hive help new` renders the canonical examples without Thor errors.

## Review summary

Three review passes (claude-ce-code-review, pr-review-toolkit multi-persona, codex-ce-code-review). All 11 plan requirements (R1–R11) implemented and exercised; no High-severity findings in any pass.

- **Pass 1** surfaced one real regression (Medium): a trailing value-option with no value consumed PROJECT as its value. Fixed by the value-presence/option-like guard (`995369dd`) and pinned with a trailing-valueless test. Also added the missing explicit-`--` coverage test and `--depends-on=42` case.
- **Pass 2** found no High/Medium defects; auto-fixed test gaps (option-like-value sub-branch, positional-outcome assertion, reverse ordering) and tightened comments.
- **Pass 3** found no High/Medium defects; documentation/test polish only — cross-linking comment for the intentional `new_text_start_index` vs `lift_new_options!` boundary divergence, single-sourced the `name=`-form predicate, realigned a stale `wiki/gaps.md` row, and strengthened surviving-text assertions.

All findings were auto-fixed or resolved from plan/brainstorm/project context; no open escalations.

## Linked task

- Hive task: `bug-hive-new-project-workflow-260622-f0d8`
- Branch: `bug-hive-new-project-workflow-260622-f0d8`
- PR: https://github.com/ivankuznetsov/hive/pull/565

---

Claude Code


ivankuznetsov:

# Code review — bug-hive-new-project-workflow-260622-f0d8 (pass 1)

Scope: `git diff origin/main..HEAD` — `bin/hive` lift-and-rebuild for `hive new`
(`lift_new_options!` replacing `terminate_new_text_options!`), `lib/hive/cli.rb`
long_desc/examples, new `test/integration/new_wrapper_argv_test.rb`, the flipped
`cli_version_test.rb` assertion, and wiki updates.

Verdict: Ready with fixes. The core fix is correct and all 11 plan requirements
(R1–R11) are implemented and tested; the canonical centerpiece is a sound
red→green. One real behavior regression on a value-option edge the new feature
introduced, plus a plan-named test gap, are worth addressing.

Reviewers: ce-correctness, ce-testing, ce-cli-readiness, ce-adversarial
(report-only). The two Medium findings below carry cross-reviewer agreement.

## High

(none)

## Medium
- [ ] **Trailing value-option without a value consumes PROJECT as its value** (`bin/hive:137-140`): a trailing `--workflow`/`--depends-on` that lacks a value is still lifted as a bare option; on rebuild Thor then eats the real project name as that option's value, mis-binding PROJECT to the idea text. Confirmed against the binary: `hive new proj idea1 idea2 --depends-on` → `project not initialized: idea1` (PROJECT became the idea text "idea1"), and `hive new proj "fix bug" --workflow` → `missing task text`. The old wrapper kept the trailing token as literal text (`project not initialized: proj`), so this is a **new regression** from the trailing-lift behavior. It fails safe in the common (quoted-idea) case, but the error misattributes and an unquoted idea whose first word collides with a registered project could create a task in the wrong project/workflow. Fix: in the `NEW_VALUE_OPTIONS` branch only consume the next token when it exists and is not option-like (`--`/`-`-prefixed); otherwise push the bare option to `rest` so it stays literal text. The same guard also hardens the pre-existing value-eats-option cases (`--workflow --json`, `--depends-on --`, `--depends-on --workflow flow`), which produce identical corrupt argv today.
- [ ] **Untested explicit `--` / `protected_tail` branch** (`bin/hive:131-134`; `test/integration/new_wrapper_argv_test.rb`): the plan's Risks section explicitly required carrying the pre-existing-`--` guard into the rebuild ("`--` already present … carry that guard into the rebuild"), and the code does — but no test anywhere in `test/` passes a literal `--` to `hive new` (flagged independently by the testing and correctness reviewers; verified by grep). This branch is the sole consumer of a user-supplied `--`; a future refactor of the rebuild could silently break literal-`--` handling with a green suite. Add a case (e.g. `hive new proj -- --workflow id text`) asserting text after `--` stays literal — `--workflow` is NOT pinned (`meta[:workflow]` nil) and PROJECT still resolves.

## Nit
- [ ] **Arity error leaks the internal Thor method name `new_task`** (`bin/hive` rebuild / `lib/hive/cli.rb:369` `map "new" => :new_task`): `hive new proj --workflow` (no value, no text) now reaches Thor's arity path and prints `ERROR: "hive new_task" was called with no arguments` (exit 64), leaking the mapped method name instead of `new` and misdirecting an agent's correction. Pre-existing Thor `map` quirk, newly reachable via the lift; the value-presence guard from the first Medium gives a cleaner message for this same input.
- [ ] **`hive new --json` is silently ignored but advertised under `new` in help** (`lib/hive/cli.rb:33`, `:357`): `--json` is a global `class_option`, so it lifts and parses cleanly, yet `new_task` never reads `options[:json]` and always emits the plain-text surface; `hive help new` still lists `[--json] # emit a single JSON document…`. A real `new` JSON envelope is explicitly deferred (out of scope), so this is not a missing requirement — but the silent-ignore is undocumented on the CLI surface. Consider a one-line note in the `new` long_desc that it does not yet emit a JSON envelope (the wiki already documents this).
- [ ] **`--depends-on=VALUE` equals-form lift is untested** (`test/integration/new_wrapper_argv_test.rb`): the equals branch (`bin/hive:141`) is exercised only by `--workflow=ID`; `--depends-on=42` is documented (wiki/commands/new.md) and handled but not asserted. Add a `--depends-on=42` case for symmetry.

ivankuznetsov:

# PR review (pr-review-toolkit) — bug-hive-new-project-workflow-260622-f0d8 (pass 1)

Scope: `git diff origin/main..HEAD` — `bin/hive` lift-and-rebuild for `hive new`
(`lift_new_options!` replacing `terminate_new_text_options!`), `lib/hive/cli.rb`
docs, new `test/integration/new_wrapper_argv_test.rb`, flipped `cli_version_test.rb`
assertion, and wiki updates.

Personas run (read-only): code-reviewer, pr-test-analyzer, silent-failure-hunter,
comment-analyzer. (type-design-analyzer N/A — no new types.) Union of findings below.

Two persona escalations were **dropped** as deliberate plan scope boundaries:
silent-failure-hunter's mid-text *unquoted* `--workflow`/`--depends-on` corruption
(plan accepts the unquoted-idea edge per A2 / Risks; its "only lift before PROJECT"
fix would contradict R1/R8 and reintroduce the original bug for the canonical
`PROJECT --workflow ID "text"` form), and its "warn when `--json` is ignored"
(contradicts R5 "no behavior change"). All 11 plan requirements (R1–R11) are
implemented; the centerpiece is a sound red→green and tests run 9/9 green.

## High
- [ ] [code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer] **A value-taking option (`--workflow`/`--depends-on`) with no usable value consumes PROJECT (or the `--` sentinel) as its value** (`bin/hive:137-140`): the value branch does `lifted << arg; idx += 1; lifted << argv[idx] if idx < argv.length` and lifts the option even when no value follows. Confirmed regression vs the old wrapper: `hive new proj idea1 idea2 --depends-on` → `project not initialized: idea1` (the real project "proj" was eaten as the dependency value and the idea text "idea1" became PROJECT); `hive new proj "fix bug" --workflow` → `missing task text`. The idea is correctly quoted, so this is NOT the accepted unquoted-idea edge — it's a genuine break of the wrapper's "first remaining positional is PROJECT" invariant on the in-scope trailing-option feature (R1, IU3 case 5). Same root cause also corrupts `hive new proj --workflow -- "real text"` (the `--` sentinel is swallowed as the workflow value, `workflow="--"`). Fails loudly with a *misleading* message in the common case; can silently create a wrong-project/wrong-dependency task in the narrower case where the leftover positionals coincide with a registered project. Fix: in the `NEW_VALUE_OPTIONS` branch, only consume the next token when it exists and is not `--`/`-`-prefixed; otherwise push the bare option to `rest` so it stays literal text (or raise a clear "`--workflow` requires a value"). This single guard fixes all sub-cases. Add the missing trailing-valueless test (`new_wrapper_argv_test.rb` only covers the trailing-WITH-value form).

## Medium
- [ ] [pr-test-analyzer, code-reviewer] **Explicit user-supplied `--` (`protected_tail`) branch is untested** (`bin/hive:131-134`; `test/integration/new_wrapper_argv_test.rb`): no test anywhere in `test/` passes a literal `--` to `hive new` (verified by grep). The plan's Risks section explicitly required carrying this guard into the rebuild; it ships with zero coverage and a refactor could silently break literal-`--` handling with a green suite. Add e.g. `run_hive(home, "new", project, "--", "--workflow", WORKFLOW_ID, "ship it")` asserting `meta[:workflow]` is nil and `idea.md` keeps `--workflow … ship it` literal.

## Nit
- [ ] [comment-analyzer] **`lift_new_options!` has no explanatory comment** unlike its three sibling argv helpers (`bin/hive:121`): `normalize_leading_json_options!` and `rewrite_help_flag!` each document the non-obvious Thor "why"; the most intricate helper (allow-list lift + value consumption + `--` rebuild) has none. Add a one-line note that Thor stops parsing options at the first positional (PROJECT), so allow-listed options after PROJECT must be lifted ahead of it.
- [ ] [comment-analyzer] **`wiki/cli.md` "including `--name=value` forms where applicable" is vague**: state concretely that `--workflow=VALUE`/`--depends-on=VALUE` lift via their `name=` prefix and JSON booleans lift only in their exact accepted forms (`--json`, `--json=true`, `--no-json`, …).
- [ ] [comment-analyzer] **`wiki/gaps.md:79` (gap #29) is now stale**: it still describes the removed `terminate_new_text_options!` "insert one `--` after PROJECT" model rather than the new lift-and-rebuild contract. The diff changed the behavior but didn't update this page (CLAUDE.md requires updating affected wiki pages). One-line refresh pointing at the lift contract / `new_wrapper_argv_test.rb`.
- [ ] [pr-test-analyzer] **`--depends-on=VALUE` equals-form lift is untested** (`test/integration/new_wrapper_argv_test.rb`): the shared equals branch (`bin/hive:141`) is exercised only by `--workflow=ID`; add a one-line `--depends-on=42` assertion to close the per-option matrix.

ivankuznetsov:

# Code Review — bug-hive-new-project-workflow-260622-f0d8 (pass 2)

Diff: `git diff origin/main..HEAD` — `bin/hive` lift-and-rebuild for `hive new`,
doc/hint standardization, and an end-to-end wrapper test matrix.

**Verification performed (read-only):**
- `rubocop bin/hive lib/hive/cli.rb test/integration/new_wrapper_argv_test.rb test/integration/cli_version_test.rb` → clean (no offenses).
- `new_wrapper_argv_test.rb` → 12 runs, 96 assertions, 0 failures.
- `cli_version_test.rb` (refreshed) → 10 runs, 60 assertions, 0 failures.
- `cli_usage_error_json_test.rb` (R7 regression guard) → 3 runs, 100 assertions, 0 failures.
- `new_test.rb` → 33 runs, 144 assertions, 0 failures.
- `hive help new` renders cleanly with the canonical examples + single parenthetical.
- Traced lift/reject/help/normalize ordering across before/between/after/`--`/value-less/JSON-boolean/`name=`/unrecognized cases — all consistent.

Plan requirements R1–R11 are all implemented and exercised. No High or Medium
defects found.

## High

(none)

## Medium

(none)

## Nit
- [ ] Test gap: no case exercises a value-option immediately followed by another option-like token (e.g. `hive new proj --workflow --depends-on 42 text`), i.e. the `value && value.start_with?("-")` sub-branch of `lift_new_options!` (bin/hive:144). The trailing/`value==nil` side is covered by `test_trailing_valueless_value_option_stays_literal_text`, so the line is covered, but the "present-but-option-like value" sub-case is not directly pinned — a one-line scenario would lock it.
- [ ] Doc consistency: Thor's short `desc "new PROJECT TEXT"` (cli.rb:331) makes `hive help new` print `Usage: hive new PROJECT TEXT`, which doesn't reflect the now-canonical option-between-PROJECT-and-TEXT placement taught in the long_desc, examples, and wiki. The long_desc parenthetical + examples compensate, and the short desc was outside the plan's three named sites — optional polish only.
- [ ] Latent divergence (pre-existing, behavior unchanged — note only): `new_text_start_index` (consumed by `reject_unsupported_json_assignments!` and `rewrite_help_flag!`, which run before `lift_new_options!`) still advances `idx += 2` unconditionally for a value-option, whereas `lift_new_options!` now keeps a value-less option literal. They only disagree on pathological inputs like `--workflow --json=1` (`--json=1` is still rejected in option position, as before this PR). Acceptable to leave; worth a comment if these two scanners drift further.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — bug-hive-new-project-workflow-260622-f0d8 (pass 2)

Diff: `git diff origin/main..HEAD`. Personas run: code-reviewer, pr-test-analyzer,
comment-analyzer, silent-failure-hunter. (type-design N/A — no new types.)
Union of findings below; duplicates consolidated, plan-deferred items dropped per
scope (e.g. the lifted-but-ignored `--json` no-op is an explicit follow-up).

Reviewer verification: rubocop clean on all changed Ruby; new + modified test
files green (22 runs); centerpiece confirmed red→green; headline silent-failure
reproduced via real `bin/hive` subprocess (result quoted in the finding).

## High

(none)

## Medium
- [ ] [silent-failure] Forgotten/option-like value to `--workflow`/`--depends-on` silently swallows the bare option into task text and pins NO workflow (exit 0, no warning) — `bin/hive:139-149`: the guard `value && !value.start_with?("-")` (bin/hive:144) demotes a value-option to literal text when its next token is itself option-like. Reproduced: `hive new proj --workflow --depends-on 42 "real text"` → `EXIT=0`, `workflow=nil`, `depends_on="42"`, `idea.md` original_text = `--workflow real text`. This is a recurrence of the exact silent-swallow class the PR exists to kill (vs. the loud `unknown workflow "--depends-on"` error you'd get without the guard); the guard correctly fixes the *trailing* PROJECT-eat but trades a loud error for silence in the *mid-stream forgot-value* case. Typo-triggered/rare, and the canonical paths all work — surfacing for a decision, not a blocker.
- [ ] [tests] The option-like-value branch of the value-option guard is never exercised — `bin/hive:142-147` / `test/integration/new_wrapper_argv_test.rb`: both value-less tests place the option at end-of-argv (`value == nil` sub-branch only); none passes a value-option followed by another `-`-prefixed token, so a regression dropping `!value.start_with?("-")` would pass the whole suite. (Test twin of the silent-failure finding above.)
- [ ] [comments] `lift_new_options!` header comment under-describes scope — `bin/hive:121-122`: says options "appearing after PROJECT must be lifted ahead of it," but the function lifts allow-listed options from anywhere after `new` (before PROJECT and trailing too — pinned by `test_workflow_option_before_project_still_pins_task` and `test_trailing_workflow_option_after_text_is_lifted`). The wiki already says "before the project, between project and text, or after text"; the code comment should match that breadth.

## Nit
- [ ] [tests] Value-less scenarios don't assert the positional outcome — `test/integration/new_wrapper_argv_test.rb:88-90`: the core regression was PROJECT mis-binding to idea text, but the test verifies only `idea.md` content + nil `depends_on`, relying on `only_task_folder` as an indirect proxy. An explicit assertion that the slug derives from the real PROJECT (not `idea1`) would pin the regression more directly.
- [ ] [tests] Combined-options test covers only one ordering — `test/integration/new_wrapper_argv_test.rb:59-69`: only `--depends-on` then `--workflow`; back-to-back value-options / reverse interleave not exercised (overlaps with the Medium gap; cheap to fold in).
- [ ] [comments] Inline "not option-like" gloss is looser than the literal predicate — `bin/hive:140-142`: the check is exactly `!value.start_with?("-")`; harmless in practice (workflow ids / positive-int-or-letter slugs never start with `-`) but "does not begin with `-`" is the precise wording.
- [ ] [comments] `long_desc` parenthetical "(Options may also follow the text.)" omits the allow-list caveat — `lib/hive/cli.rb:335`: true (trailing options lift) but silently implies any flag lifts; surrounding prose + wiki cover the closed-allow-list distinction, so acceptable as help-text summary.

ivankuznetsov:

# Code Review — bug-hive-new-project-workflow-260622-f0d8 (pass 3)

Scope reviewed: `git diff origin/main..HEAD` — `bin/hive` (`lift_new_options!`
replacing `terminate_new_text_options!`), `lib/hive/cli.rb` long_desc/examples,
`test/integration/new_wrapper_argv_test.rb` (new), `test/integration/cli_version_test.rb`
(refreshed), and wiki doc updates.

Verification performed (read-only): ran `new_wrapper_argv_test.rb` (14/14),
`cli_version_test.rb` (10/10), `cli_usage_error_json_test.rb` (3/3, R7 regression
guard), `new_test.rb` (33/33), `user_workflow_e2e_test.rb` (1/1); `rubocop` clean
on all four Ruby files; `hive help new` renders the canonical examples without
Thor errors; confirmed no dangling `terminate_new_text_options!` references and
`bin/hive-e2e` has no parallel `new` path needing alignment.

Assessment: the fix is correct and plan-compliant (R1–R11 all satisfied). The
lift-and-rebuild lifts allow-listed options from before/between/after PROJECT,
keeps the literal-text splat behind a fresh `--`, preserves the missing-text
guard, and the value-option guard correctly demotes trailing/option-like values
(workflow/dependency ids never start with `-`, so legitimate values are never
demoted). No High or Medium defects found.

## High

(none)

## Medium

(none)

## Nit

- [ ] Two sources of truth for the "new option region" boundary: `new_text_start_index` (`bin/hive:67`) still consumes value-options unconditionally (`idx += 2`), while `lift_new_options!` (`bin/hive:141-151`) now consumes the value only when present and non-`-`-prefixed. The two only diverge for the pathological "value-option before PROJECT with a missing/option-like value" case (out of scope per the plan), so it is harmless today — but a future edit to the allow-list/guard in one helper can silently drift the boundary that `reject_unsupported_json_assignments!`/`rewrite_help_flag!` rely on. Consider a shared helper or a comment cross-linking the two.
- [ ] IU2's soft test scenario ("a doc/string assertion that the long_desc examples render with the option before the quoted text") is not automated: only the `hive workflow new` hint is asserted (`new_wrapper_argv_test.rb:202`); the `cli.rb` example block is verified manually only. Low value — such a test would be a brittle change-detector on static help text — so leaving it is defensible; noted for completeness against the plan scenario.

ivankuznetsov:

# PR Review (pr-review-toolkit) — bug-hive-new-project-workflow-260622-f0d8 (pass 3)

Multi-persona review of `git diff origin/main..HEAD`. Personas run: code-reviewer,
pr-test-analyzer, comment-analyzer, silent-failure-hunter, code-simplifier.
type-design-analyzer skipped — no new types/classes (the change is argv array
manipulation only).

All personas independently verified the fix is correct: ran the new
`new_wrapper_argv_test.rb` (14/114), `cli_version_test.rb` (10/60),
`cli_usage_error_json_test.rb` (3/100), `new_test.rb` (33/144) — all green;
`rubocop` clean; hand-traced the four-helper pipeline across all edge cases.
silent-failure-hunter found NO issues (the early returns defer cleanly to Thor's
arity error / `missing task text`, and value-option demotion is the documented,
user-visible contract — nothing swallowed). No High-severity findings from any
persona.

## High

(none)

## Medium

- [ ] [comment-analyzer] `wiki/gaps.md:16` is a stale coverage-table row still describing the superseded model ("the `hive new PROJECT` text-tail boundary that preserves post-project flag-looking tokens as task text"); this PR rewrote gap #29 (line 79) to note the lift-and-rebuild supersession but left line 16 inconsistent (pre-existing line made inaccurate by this diff; `wiki/log.md`/older log fragment are append-only history and correctly left alone).

## Nit

- [ ] [code-reviewer] `bin/hive` keeps two divergent boundary computations — `new_text_start_index` (lines 67-97) advances value-options with an unconditional `idx += 2`, while `lift_new_options!` (line 146) consumes the value only when it exists and is not `-`-prefixed; harmless today (the only affected consumer, `reject_unsupported_json_assignments!`, runs first and inspects only `--json=` tokens) but a two-source-of-truth tech-debt risk.
- [ ] [code-simplifier] `bin/hive:83` and `bin/hive:152` duplicate the `name=`-form predicate `NEW_VALUE_OPTIONS.any? { |name| arg.start_with?("#{name}=") }` verbatim; extract a tiny pure helper (e.g. `new_value_assignment?(arg)`) to single-source that one branch.
- [ ] [code-simplifier] `bin/hive:160-164` uses a mutate-then-replace (`rebuilt = …; rebuilt.concat([ "--", *rest ]) unless rest.empty?; argv.replace(rebuilt)`); could assemble the protected tail once (`tail = rest.empty? ? [] : [ "--", *rest ]`) into a single `argv.replace` — keep the `--`-only-when-nonempty behavior.
- [ ] [pr-test-analyzer] `test/integration/new_wrapper_argv_test.rb:33-34` (centerpiece) asserts `idea.md` lacks the literal `"--workflow"` but does not assert the body still contains the task text `"custom workflow task"`; sibling tests do assert surviving text — add `assert_includes` to close a theoretical "text dropped entirely while flag stripped" hole.
- [ ] [pr-test-analyzer] `test/integration/cli_version_test.rb:113` resolves the folder via `Dir[...].first` without asserting exactly one match (unlike `only_task_folder` in the new file); a nil match would yield a confusing `TaskMeta.read(nil)` failure instead of a clear count assertion.
- [ ] [pr-test-analyzer] negative `idea.md` assertions in `new_wrapper_argv_test.rb` (e.g. lines 34, 108, 115, 135, 146, 170) use bare flag substrings (`"--workflow"`, `"--json"`, `"--depends-on"`); robust only because the chosen `WORKFLOW_ID`/task texts avoid those substrings — an implicit coupling a future edit could silently weaken.
- [ ] [pr-test-analyzer] `test_depends_on_and_workflow_after_project_are_both_lifted` and `test_workflow_then_depends_on_after_project_are_both_lifted` (lines 59-81) are near-duplicates differing only in flag order — intentional ordering coverage, noted as redundancy only.
- [ ] [comment-analyzer] `wiki/cli.md:29` and `wiki/commands/new.md:20` use "standalone" for the allow-listed options, which is slightly loose for the single-token `--workflow=VALUE` form; both clarify the `=`-prefix forms in the following sentence, so cosmetic only.

PR #571, opened 2026-06-23
feat(init): scaffold and bind a custom workflow with --new-workflow

ivankuznetsov:
## Summary

Putting a project on a *custom, user-authored* workflow used to take four steps: `hive init` (creates `.hive-state`, binds `coding`), `hive workflow new ID` (needs `.hive-state` to exist), hand-edit the descriptor, then `hive init --workflow ID` again to rebind the default. The chicken-and-egg was structural: a custom descriptor lives under `.hive-state/workflows/`, which only exists after init, so `hive init --workflow ID` could never reference a not-yet-created custom workflow on a fresh project.

`hive init --new-workflow ID [PROJECT_PATH]` collapses that to **one bootstrap command + one edit**. It scaffolds a blank descriptor (`<hive_state>/workflows/ID.yml` + `ID/work.md`), binds `default_workflow: "ID"`, and prints both paths to edit. After editing, `hive new <project> '<idea>'` with **no** `--workflow` routes through the custom workflow because the default is already bound.

**How it works.** The flag reuses `hive workflow new`'s scaffolding rather than re-implementing it, so the reserved-id rejection and file-writing are literally the same code path. Two scenarios, both atomic:

- **Fresh project** — the descriptor scaffold runs *inside* init's existing bootstrap rollback window (via an `after_bootstrap` hook on `initialize_project_state`). If anything fails, the orphan branch, worktree, config, registration, and descriptor files are all unwound — zero footprint. The `default_workflow` binding is committed on `hive/state` so it survives a `reset --hard`/`clean`.
- **Already-initialized project** — the init half is a no-op. The descriptor write and the `default_workflow` rebind commit together under one commit lock, so a descriptor-without-binding or a default pointing at a missing descriptor can never persist. On commit failure the rescue resets the hive-state index and restores the config snapshot — both *inside* the lock — so a half-rolled-back rebind can never ride an unrelated commit.

`--new-workflow` is mutually exclusive with `--workflow` (exit 64), and built-in ids like `coding` / `content` are rejected through the same reserved-id path as `hive workflow new`, before any disk write. Keyword-like ids (`no`, `on`, `null`, …) are quoted in the emitted YAML so `YAML.safe_load` round-trips them back to the bound string.

**JSON contract.** The `hive-init.v1` schema is extended additively — `schema_version` stays at 1. Two optional fields are emitted only with `--new-workflow`, on both the fresh and already-initialized payload arms:

| Field | Meaning |
|---|---|
| `descriptor_path` | path to the scaffolded `ID.yml` |
| `instruction_path` | path to the stub `ID/work.md` to edit |
| `workflow` | the newly bound default id (already present, now carries `ID`) |

## Test plan

`bundle exec rake` and `bundle exec rake coverage` both green (100% lines). Coverage spans the new branches:

- Fresh happy-path: config binds `default_workflow: "ID"`, descriptor + `work.md` scaffolded and parse, descriptor + config committed on `hive/state`, project registered.
- Already-initialized rebind: init half no-op, single combined commit touches both `config.yml` and the descriptor; behavioral `reset --hard`/`clean` durability re-read.
- Reserved-id rejection (`coding`) and `--workflow`/`--new-workflow` mutual exclusivity — both exit 64, zero footprint.
- Mid-flow rollback (fresh and existing), including the inline `rollback_scaffold` path and the existing-path index-reset + config-restore.
- Descriptor collision refusal; keyword-like id (`no`/`yes`) round-trip through `YAML.safe_load`; fieldless-task rebind warning on the `--new-workflow` existing path.
- Best-effort warn branches (`reset_hive_state_index` failure, `warn_failed_scaffold_cleanup`) forced via stubs.
- Dogfood: flag-less `hive new` binds the task to `workflow: ID` and advances into the new workflow's second stage.

## Review summary

Three Compound Engineering review passes (claude `ce-code-review` ×3, codex `ce-code-review` ×3, `pr-review-toolkit` multi-persona ×3). All findings were auto-fixed or resolved against the plan/brainstorm/project context — no user escalations across any pass. Resolved highlights:

- **Existing-path index leak (High, convergent):** a `git commit` failure after staging left `config.yml` + descriptor staged in the long-lived `.hive-state` worktree, where the next bare commit would sweep the half-rolled-back rebind. Fixed by resetting the hive-state index in the rescue, moved *inside* the commit lock to close the concurrent-committer window; the corresponding test now fails at the git layer (after staging) and asserts a clean index.
- **Fresh-path durable binding (High):** the fresh path now commits the `default_workflow` rebind too, symmetric with the existing path.
- **Keyword-id YAML coercion (Medium):** `default_workflow` and descriptor `id:` are now quoted in both render paths, with a cross-referenced sync comment and a round-trip test.
- **100% coverage gate:** new best-effort warn branches were covered to clear the hard gate.
- Plus nits: shell-escaped next-step hints, simplified scaffold helpers (true class methods, shared commit helper), schema/doc/comment accuracy, and added fieldless-rebind + inline-rollback coverage.

The `hive workflow new` extract-method refactor was verified behavior-preserving (its full regression suite retained).

## Linked task

- Hive task: `ux-friction-you-cannot-initialize-260622-a219`
- Branch: `ux-friction-you-cannot-initialize-260622-a219`

---


ivankuznetsov:

## High
- [ ] Fresh `--new-workflow` does not commit the default-workflow binding: `lib/hive/commands/init.rb:165` commits only workflow files even though `config.yml` is written earlier at `lib/hive/commands/init.rb:419`, so the scaffolded descriptor can be committed without the `default_workflow` rebind required for durable flag-less `hive new`.
- [ ] Existing-project rollback can leave staged hive-state changes after a real commit failure: `lib/hive/commands/init.rb:201` restores files but never cleans the index after `GitOps#hive_commit` stages pathspecs, so a failed `--new-workflow` can leak rolled-back config/workflow additions into a later commit.

## Medium
- [ ] YAML-keyword workflow IDs bind to the wrong default workflow: `lib/hive/commands/init.rb:139` and `lib/hive/commands/init.rb:351` write unquoted `default_workflow` values, so valid IDs like `yes`, `on`, or `null` are coerced by `YAML.safe_load` and flag-less `hive new` no longer resolves the scaffolded workflow.

## Nit
- [ ] New human next-step hints are not shell-escaped: `lib/hive/commands/init.rb:236` and `lib/hive/commands/init.rb:782` interpolate the project basename directly, so projects with spaces or shell metacharacters get a broken `hive new ... '<idea>'` command.

ivankuznetsov:

# Code review — `hive init --new-workflow ID` (pass 1)

Scope: `git diff origin/main..HEAD`. Reviewed against the plan's Goals / Scope
Boundaries / Requirements Trace. The `workflow.rb` extract-method refactor was
verified behavior-preserving for `hive workflow new` (ordering and commit-failure
rollback coverage unchanged; `workflow_new_test.rb` regression suite retained), so
no findings there.

## High
- [ ] Existing-project rollback leaves the hive-state git **index** dirty after a post-staging commit failure (`lib/hive/commands/init.rb:201-205`): the rescue restores the working-tree `config.yml` and `rm`s the scaffold files but never `git reset`s the already-staged `workflows/<id>.yml`, `workflows/<id>/`, and `config.yml`. Because `GitOps#hive_commit` runs a **bare** `git commit` that commits the entire index (`lib/hive/git_ops.rb:199-202`), the next hive operation (e.g. `hive new`) sweeps the phantom descriptor *and* the supposedly-rolled-back `default_workflow: <id>` rebind into an unrelated commit — silently re-binding the field-less in-flight tasks that `warn_on_fieldless_tasks_rebinding` exists to protect, and resurrecting a descriptor the owner never accepted. Trigger (a `git commit` that fails *after* staging) is narrow, but the code already anticipates commit failure (that is what the snapshot/restore is for), so the rollback should reset the index too; consider `git reset -q -- <pathspecs>` in the rescue, or scope `hive_commit` to commit only its pathspecs.

## Medium
- [ ] The existing-path commit-failure test does not exercise the leak above (`test/integration/init_test.rb:369-390`): the stub redefines `hive_commit` to `raise` on entry (line 373-374), *before* `stage_hive_state_pathspec` ever runs, so the index is never polluted in the test; the assertions (386-390) only check working-tree state and never assert `git status --porcelain` / `git diff --cached --quiet` cleanliness — unlike the fresh-path test which calls `assert_clean_failed_init`. Fail at the git layer (e.g. a failing pre-commit hook) so staging runs first, then assert a clean index.
- [ ] `restore_config_snapshot` is warn-only on failure (`lib/hive/commands/init.rb:304-315`) and, combined with the unreset index from the High finding, means a failed working-tree restore leaves **both** the worktree and the staged index carrying `default_workflow: <id>` while the original `GitError` is re-raised as if nothing changed — so the "all-or-nothing" guarantee the analogous code claims is not actually achieved on the `--new-workflow` existing path.

## Nit
- [ ] Fresh `--new-workflow` path leaves `default_workflow: <id>` **uncommitted** in the hive-state worktree (`lib/hive/commands/init.rb:155-176` commits only the descriptor, not `config.yml`): the binding works (read from the working tree) but is not durable against a hive-state `reset --hard`/`clean`, asymmetric with the existing-project path that commits `config.yml`. This matches pre-existing standard `hive init` behavior (config.yml is never committed on fresh init) and is **not a regression** — flagging only because durable binding is the feature's value proposition; fix both paths together if desired.
- [ ] `Workflow.rollback_scaffold` / `normalize_and_validate_id!` / `scaffold_files!` instantiate a throwaway `new("new", …)` just to reach stateless private helpers (`lib/hive/commands/workflow.rb:39-65`); `rollback_scaffold` even passes a `nil` id. Harmless, but the path/validation/rollback logic could be true module functions to avoid the indirection.

ivankuznetsov:

# PR review (pr-review-toolkit, multi-persona) — `hive init --new-workflow ID` (pass 1)

Scope: `git diff origin/main..HEAD`. Personas run: code-reviewer, pr-test-analyzer,
silent-failure-hunter, comment-analyzer, type-design-analyzer, code-simplifier.
Union of findings below. Note: the `--json` reserved-id error printing plain stderr
instead of a JSON envelope is intentionally **not** flagged — it is a deliberate
plan-level scope boundary (key design decision #4; matches existing `--workflow`
behavior). Three personas independently converged on the existing-path dirty-index
leak; the code-reviewer's lone "false positive" verdict came from reproducing via the
flawed test stub, which is itself the test-gap finding below.

## High
- [ ] [silent-failure] Existing-path rollback never resets the hive-state git **index** (`lib/hive/commands/init.rb:201-205`): on a `git commit` failure *after* `hive_commit` staged `config.yml` + descriptor + instruction dir via `git add -A` (`lib/hive/git_ops.rb:186-205`), the rescue restores/removes only working-tree files, leaving `AM config.yml` + `AD workflows/<id>.yml` staged; the next bare `git commit` in the still-attached worktree silently sweeps the half-rolled-back `default_workflow` rebind (and a phantom descriptor) into an unrelated commit. Fix: `git reset -q -- <pathspecs>` in the rescue, or scope `hive_commit` to commit only its pathspecs.
- [ ] [tests] The existing-path commit-failure test cannot catch that leak (`test/integration/init_test.rb:360-393`): the stub redefines `hive_commit` to `raise` on entry (line 372-376), *before* the real `git add` staging runs, and the assertions (386-390) never check the hive-state index (`git -C <hive_state> diff --cached --quiet`); it exercises a fail-before-staging ordering that doesn't match production, so the data-corruption regression above is invisible. Fail at the git layer (e.g. a rejecting pre-commit hook) so staging runs first, then assert a clean index.

## Medium
- [ ] [silent-failure] `restore_config_snapshot` is warn-only on failure (`lib/hive/commands/init.rb:304-315`) and, with the index never reset, a failed/partial restore silently persists the `default_workflow` rebind the operator was told failed — the re-raised `GitError` makes "nothing changed" an illusion while the staged rebind survives for the next bare commit.
- [ ] [comments] `rollback_scaffold`'s "filesystem-only rollback is sufficient" rationale (`lib/hive/commands/workflow.rb:208-215`) no longer holds for the new shared-worktree caller: it assumes only a same-command retry re-stages and re-checks via `refuse_overwrite!`, but an unrelated `hive_commit` in the long-lived `.hive-state` worktree can now sweep the stale staged index; the comment should acknowledge the index is left dirty on commit failure.
- [ ] [tests] No test covers the existing-path rollback where `restore_config_snapshot` *itself* fails (`init.rb:304-315`): the rescue-within-rescue's non-masking-of-the-original-error and its warning text guard a config data-loss scenario yet are entirely unverified.
- [ ] [comments] Schema `description` is stale (`schemas/hive-init.v1.json:5`): it states AlreadyInitializedPayload is emitted only "with --workflow" (now also `--new-workflow`) and never mentions the new optional `descriptor_path`/`instruction_path` fields added to both `oneOf` arms — the property lists were updated but the prose contract was not.
- [ ] [type-design] `WorkflowChoice`'s documented invariant ("descriptor is a fully resolved `Hive::Workflow`", `init.rb:51-54`) is silently violated on the `--new-workflow` existing path (`init.rb:208`): it carries a one-field `WorkflowDescriptorRef` stand-in, and `:source == :flag` no longer lets a consumer distinguish it from a real descriptor, so any future `.descriptor` call expecting `.stages`/`.each` would `NoMethodError`. Safe today (only `.id` is read) but the type no longer makes the illegal state unrepresentable.

## Nit
- [ ] [silent-failure] `rescue StandardError` on the scaffold/commit blocks (`lib/hive/commands/init.rb:172, 201`) is broader than the expected `GitError`/`ConcurrentRunError`/`SystemCallError` set, so a programming bug (`NoMethodError`/`KeyError`) triggers destructive `rm_rf` rollback before re-raising — surfacing the bug with on-disk evidence already deleted. Narrow to the enumerated set like `workflow.rb#call` (lines 77-78).
- [ ] [silent-failure] `rollback_scaffold` `rm_f`/`rm_rf` swallow their own errors with no log (`lib/hive/commands/workflow.rb:216-219`): a failed cleanup (permission/busy) surfaces later as a confusing "already exists" on the next `--new-workflow` run; a single `write_warn` on failure would localize the signal.
- [ ] [type-design] `WorkflowDescriptorRef` (`init.rb:55`) is a lossy formatter-only shim wrapped at `init.rb:208` only to be unwrapped one layer later; since `write_existing_summary`/`existing_payload` read only `.descriptor.id`, passing the plain `id` (or fetching the now-on-disk real descriptor) would delete the type and the wrap/unwrap. (Also raised by simplify.)
- [ ] [comments] Fresh-path docs/help asymmetry (`lib/hive/cli.rb` help text, `wiki/commands/init.md`): the existing-path doc says it "commits the descriptor and config.yml rebind together" while the fresh-path prose is silent that `config.yml` (carrying `default_workflow: <id>`) is written but **not** committed to `hive/state` — a reader could infer fresh also commits it. Matches pre-existing fresh-init behavior; not a regression, but a one-clause note would remove the ambiguity.
- [ ] [code] Mutual-exclusivity check runs after `validate_clean_tree!` (`lib/hive/commands/init.rb:80-82`): on a dirty repo, passing both `--workflow` and `--new-workflow` reports the clean-tree error instead of the clearer "mutually exclusive" usage error (both exit non-zero; cosmetic ordering).
- [ ] [simplify] The three new `Workflow` class methods reach private instance methods via repeated `command.send(:...)`, and `rollback_scaffold` builds a throwaway `new("new")` with a nil id (`lib/hive/commands/workflow.rb:39-65`): promote `rollback_scaffold` to a true class/module method (it touches only its `paths` arg) and collapse the send-chain.
- [ ] [simplify] `init_with_new_workflow` validates the id twice (`init.rb:129` `normalize_and_validate_id!`, then `scaffold_files!` re-normalizes/re-validates the same raw id): derive `id` from the single `scaffold_files!` result to drop the redundant pass.
- [ ] [simplify] Three identical `hive_commit(stage_name: "workflows", slug: id, action: "created", pathspecs: […])`-under-`with_commit_lock` blocks (`init.rb:155-176, 185-200`; `workflow.rb:185-198`): extract a shared `commit_workflow_scaffold` helper so the `"workflows"`/`"created"` contract and relative-path mapping live in one place.
- [ ] [simplify] `before = :not_captured` sentinel (`init.rb:183, 202`) adds a third state for an extremely narrow pre-snapshot raise window; the sibling `update_existing_default_workflow!` (line 288) handles the same snapshot/restore entirely inside the lock with no sentinel — align with that pattern to drop the `unless before == :not_captured` special case.

ivankuznetsov:

# Code review — `hive init --new-workflow ID` (pass 2)

Scope: `git diff origin/main..HEAD`. Verified locally: `workflow_new_test` (15/0),
`init_test` integration (65/0), `schema_files_test` (111/0), `cli_test` (20/0) all green;
`rubocop lib/hive/commands/init.rb lib/hive/commands/workflow.rb lib/hive/cli.rb` clean.
Plan requirements (fresh, existing, reserved, mutual-exclusivity, rollback fresh/existing,
collision, dogfood routing, JSON/schema, docs) are all implemented and covered by tests.

## High
- [ ] None. No plan-required item is missing or unimplemented, and no happy-path correctness defect was found.

## Medium
- [ ] Existing-project rollback runs the index-reset and config-restore *outside* the commit lock (`lib/hive/commands/init.rb:187-206`): the `rescue` that calls `reset_hive_state_index` + `restore_config_snapshot` sits after the `Hive::Lock.with_commit_lock` block has already released the lock, so a concurrent `hive_commit` (e.g. the daemon committing a task) acquiring the lock in that window can still sweep the staged `config.yml` + descriptor into an unrelated bare `git commit` — the exact "ride the next unrelated commit" corruption the reset claims to eliminate ("so the rollback is total"). The sibling path `update_existing_default_workflow!:288-306` deliberately unwinds *inside* the lock; mirror that by performing the reset+restore before the lock is released (e.g. an inner `begin/rescue` around the commit). Low probability (requires a hive-state commit failure plus a concurrent committer in a sub-ms window), but the failure mode is a silent, durable `default_workflow` rebind on a foreign commit.

## Nit
- [ ] The keyword-id YAML-quoting invariant now lives in two places — `templates/project_config.yml.erb:7` and `Init#default_workflow_line` (`lib/hive/commands/init.rb:369-371`). Both are commented, but they are independent sources of truth for the same safety rule; a future change to one (e.g. switching to single quotes or YAML-escaping) can silently desync the fresh-render path from the rebind path. Consider routing the template through the same helper, or cross-referencing each comment to the other.
- [ ] Two new best-effort warning branches are unexercised: `Init#reset_hive_state_index`'s reset-failure `write_warn` (`lib/hive/commands/init.rb:951-955`) and `Workflow.warn_failed_scaffold_cleanup`'s leftover-warn (`lib/hive/commands/workflow.rb:158-166`). The success paths are covered, but the failure-signal branches have no direct test, so they may trip the repo coverage gate and the defensive messages are unverified. A stub forcing `git reset` non-zero / leaving a file behind would close both.

ivankuznetsov:

# pr-review-toolkit multi-persona review — `hive init --new-workflow ID` (pass 2)

Scope: `git diff origin/main..HEAD`. Personas run: code-reviewer, pr-test-analyzer,
comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier.
Verification by reviewer: full `bundle exec rake coverage` run completed (6100 runs,
0 failures, 0 errors, 2 pre-existing skips, 468s) — **coverage gate FAILED at 99.99%**
(authoritative; the on-disk `coverage/coverage.json` was a stale 27% partial artifact).
`rubocop` clean on changed lib files. All plan requirements are implemented.

## High
- [ ] [tests] Coverage gate fails: 99.99% (26311/26314), below the repo's hard 100% minimum — uncovered lines `lib/hive/commands/init.rb:953` (the `write_warn` in `reset_hive_state_index`'s reset-failure branch) and `lib/hive/commands/workflow.rb:162,165` (the `warn` body + `rescue Errno::EPIPE` in `warn_failed_scaffold_cleanup`). Confirmed by a clean full-suite run; no `:nocov:` mechanism exists, so these new best-effort warn branches (added by the hardening commit `d8052951`) are a merge blocker until a test forces a `git reset`/`rm` failure.

## Medium
- [ ] [code-review/silent-failure] Existing-project rollback runs *outside* the commit lock (`lib/hive/commands/init.rb:194-206`): the `rescue` calling `reset_hive_state_index` + `restore_config_snapshot` + `rollback_scaffold` sits after the `with_commit_lock` block, whose `flock` releases the instant `commit_workflow_scaffold` raises — so a concurrent `hive_commit` acquiring the lock in that window can still sweep the staged `config.yml` rebind + phantom descriptor into an unrelated bare `git commit`, exactly the corruption the reset claims to prevent; the sibling `update_existing_default_workflow!:288-306` correctly restores *inside* the lock. (silent-failure-hunter rated this High.)
- [ ] [silent-failure] Best-effort rollback degradation masks `.hive-state` inconsistency (`lib/hive/commands/init.rb:947-956` and `:309-320`): a non-zero `git reset` (init.rb:953) and a failed `restore_config_snapshot` (init.rb:313) are each downgraded to a `write_warn` while the original error is re-raised as if rollback succeeded — leaving a mixed index or a half-rebound `default_workflow` on disk in the long-lived worktree, surfaced only as a soft warning, not as state-requiring-repair.

## Nit
- [ ] [comment/silent-failure] `lib/hive/commands/init.rb:195` comment says `hive_commit` "stages … via `git add -A`" (whole-tree), but the explicit-pathspec path runs a *per-pathspec* `git add -A -- <rel>` (`git_ops.rb:189-194,252-259`); the scoped reset is correct, but the wording overstates what is staged and could mislead a future maintainer about whether the reset is sufficient.
- [ ] [comment] `lib/hive/commands/init.rb:199-201` overclaims that resetting the index "before restoring the working tree" is load-bearing "so the rollback is total" — `git reset` (no `--hard`) only touches the index and `restore_config_snapshot` only the working tree, so their ordering is independent and has no effect on the end state.
- [ ] [comment] `lib/hive/cli.rb:97` long_desc and `wiki/commands/init.md` render the binding unquoted (`default_workflow: writing` / `ID`) while this same PR made the written form mandatorily quoted (`default_workflow: "ID"`); the prose drifts from the new contract.
- [ ] [tests] The YAML keyword-coercion guard (`default_workflow_line` quoting) is only exercised with the literal id `content_fixture`; no test binds a coercion-prone id (e.g. `--new-workflow no`/`yes`) and asserts it round-trips back to the string via `YAML.safe_load` and resolves under flag-less `hive new` — the exact regression the quoting prevents is unpinned (`test/integration/init_test.rb:174-179,746,757`).
- [ ] [tests] The fresh-path durability claim ("survives a hive-state `reset --hard`/`clean`", init.rb:163-167) is verified only indirectly via commit membership (`git show --name-only`), not behaviorally by resetting the worktree and re-reading `default_workflow` (`test/integration/init_test.rb:164`).
- [ ] [tests] `test_init_new_workflow_fresh_rolls_back_when_scaffold_commit_fails` stubs `scaffold_files!` to raise *before* the commit lock, so the unwind goes through `rollback_partial_init` (whole-worktree teardown), not `scaffold_descriptor_commit!`'s own inline `rollback_scaffold` (init.rb:174) — a regression in that specific rescue would be masked.
- [ ] [type-design] `Workflow.scaffold_files!` returns an undocumented symbol-keyed bag `{ id:, paths: { descriptor:, instruction_dir:, instruction:, workflows_dir: } }` (`lib/hive/commands/workflow.rb:435-453`); the shape is asserted nowhere and both call sites re-`.fetch` it by convention — a small value object would make the contract a construction invariant.
- [ ] [type-design] `WorkflowChoice`'s documented "descriptor is a fully resolved `Hive::Workflow`" invariant is unenforced (`lib/hive/commands/init.rb:54`); `Data.define(:descriptor, :source)` accepts anything, and `schema_files_test.rb:1253` constructs one with a bare `Struct.new(:id)` stand-in, proving the type permits a non-Workflow descriptor.
- [ ] [type-design] `WorkflowChoice#source` is an unconstrained symbol with four implicit valid values (`:flag`/`:prompt`/`:implicit`); a typo'd symbol would silently fall through the `== :implicit`/`== :flag` branch logic (`init.rb:90-92,213`).
- [ ] [type-design/simplify] The two scaffold-commit entry points compute pathspecs divergently with duplicated relative-path mapping — `relative_to_hive_state` (`init.rb:351`, includes `config.yml`) vs `relative_to_workflows_root` (`workflow.rb:240`, byte-identical body, descriptor+dir only); since init now depends on workflow, one shared helper would prevent path-relativity drift.
- [ ] [type-design] `scaffold_payload` emits `descriptor_path`/`instruction_path` as an unenforced pair (`init.rb:306-311`); the schema marks both optional on each arm, so a future caller passing a half-populated `extra` (one key) would still validate.
- [ ] [code-review] Latent path divergence: `GitOps#hive_state_path` is hardcoded to `<root>/.hive-state` (`git_ops.rb:15-17`) while `scaffold_paths` derives its dir from `config.yml`'s `hive_state_path`; for a non-default `hive_state_path` the `relative_path_from` pathspec would be wrong — unreachable through `hive init` today, but the two must stay in sync.
- [ ] [silent-failure] `rollback_scaffold`/`warn_failed_scaffold_cleanup` (`workflow.rb`) swallow all `rm_f`/`rm_rf` faults and detect failure only by re-checking `File.exist?`, then warn with a generic "remove them manually" that omits the underlying errno — the operator gets no signal as to *why* cleanup failed.
- [ ] [silent-failure] `reset_hive_state_index`'s warning truncates git's stderr to the first line (`init.rb:954`, `err…lines.first`), potentially dropping the actionable detail (e.g. `index.lock` contention) that git emits on a later line.
- [ ] [silent-failure] `current_default_workflow_with_status` (`init.rb:329-345`, pre-existing, now on the new rebind read path at :188) rescues `SystemCallError` as broadly as `Psych::Exception`, so an unreadable-but-valid config is treated as corrupt and the in-flight-task rebind warning under-reports `from`.
- [ ] [simplify] `write_existing_summary` recomputes `Shellwords.escape(File.basename(@project_path))` inline (`init.rb:241`) where the basename is already derived a few lines up — hoist a local as `print_summary` does (`safe_name`, init.rb:336).
- [ ] [simplify] `scaffold_and_bind_fresh` calls `scaffold.fetch(:paths)` twice (`init.rb:149,151`); bind `paths` once above the `if @json` to mirror `scaffold_and_bind_existing`.
- [ ] [simplify] `scaffold_descriptor_commit!` returns `{ id:, paths: }` but every caller reads only `:paths` (`init.rb:149,151`); returning `paths` directly drops the dead `:id` round-trip (leave if the symmetric shape with `scaffold_files!` is preferred).

ivankuznetsov:

# ce-code-review — `hive init --new-workflow ID` (pass 3)

Reviewed `git diff origin/main..HEAD` against the plan. Verified by running the
affected suites (all green): `test/integration/init_test.rb` (69 runs),
`test/integration/new_test.rb` (33 runs), `test/unit/commands/workflow_new_test.rb`
(17 runs), `test/unit/schema_files_test.rb` (111 runs), `test/unit/cli_test.rb`,
`test/unit/commands/init_test.rb`; `rubocop` clean on the three changed lib files.

Every Requirements-Trace item is implemented and covered by a test (fresh, existing,
reserved-id, mutual-exclusivity, fresh+existing rollback, collision, dogfood routing,
keyword-id quoting, reset-survival, JSON validation). No plan-required-but-missing
gaps. Composition of the two rollbacks, the single-commit atomicity, and the
index-reset leak-guard are all correct and well-tested. Findings below are minor.

## High

(none)

## Medium

(none)

## Nit
- [ ] Templates modified despite the plan listing them as out-of-scope / "reused as-is": `templates/workflows/blank/descriptor.yml.erb` now emits `id: "<id>"` and `templates/project_config.yml.erb` now emits `default_workflow: "<id>"`. The quoting is a genuine, necessary correctness fix (YAML.safe_load coerces keyword-like SAFE_SLUG ids such as `no`/`on`/`null` to booleans/nil), but it is a deviation from the stated Scope Boundary and also changes `hive workflow new`'s scaffolded descriptor bytes (`id: x` → `id: "x"`) — a cosmetic but observable change to a command the plan said should have "no observable change". Semantically identical YAML; no functional impact.
- [ ] `hive workflow new`'s `rollback_scaffold` now warns on a cleanup fault (`warn_failed_scaffold_cleanup`, surfacing the errno) where the old `rm_f`/`rm_rf` path was silent. This is a strict improvement (a silent cleanup failure used to resurface as a confusing `refuse_overwrite!` "already exists" on retry) and is tested, but it is technically an observable behavior change to `workflow new`, which the plan's scope boundary said to leave unchanged.
- [ ] The `default_workflow` quoting invariant is duplicated across two sites — `templates/project_config.yml.erb` (fresh render) and `Init#default_workflow_line` (rebind) — kept aligned only by a "keep the two in sync" comment. Each path is independently tested for correctness (`no` / `content_fixture`), so a removal of quoting on either side would fail a test; but no single test pins that the two render-paths agree on quote *style*, so a future divergence in formatting could pass unnoticed.
- [ ] Asymmetric index-leak guard between the two paths: `scaffold_and_bind_existing` resets the `.hive-state` index after a failed commit (`reset_hive_state_index`), but the fresh `scaffold_descriptor_commit!` relies solely on `rollback_partial_init` tearing down the whole worktree. That teardown is skipped when `hive_state_init` returns `:existed` (`rollback_on_failure = false`), so a concurrent-init TOCTOU race on the same fresh project — where the scaffold commit then fails after `git add -A` staged config.yml + descriptor — could leave a staged rebind that rides a later commit. Well-justified in practice (a brand-new orphan worktree has no concurrent committer, and concurrent `hive init` on one project is unsupported), so very low risk; noting for completeness.
- [ ] `print_summary`'s next-step hint now shell-escapes the project basename (`Shellwords.escape(name)`) on the normal (non-`--new-workflow`) path too, not just the new flow. This is a uniform, benign improvement (correct shell for names with spaces/special chars; identical output for ordinary basenames), but it is a small behavior change to the pre-existing path that falls slightly outside the stated scope.

ivankuznetsov:

# pr-review-toolkit — `hive init --new-workflow ID` (pass 3)

Union of six persona passes (code-reviewer, pr-test-analyzer, silent-failure-hunter,
comment-analyzer, type-design-analyzer, code-simplifier) over `git diff origin/main..HEAD`.

Overall: a well-hardened change. **code-reviewer** found no bugs/logic/convention issues
at or above threshold (rubocop clean, suites green, fresh/existing/rollback/schema paths
all correct). **silent-failure-hunter** found no genuine silent-failure risk — every
rescue warns-and-re-raises or is best-effort cleanup that surfaces the fault.

Two persona observations were dropped as not-real after verification:
(1) silent-failure's "`init_with_new_workflow` bypasses `call`'s rescue chain" is a **false
positive** — the `rescue` clauses are method-level on `def call`, so they cover the
`init_with_new_workflow(ops)` invocation; the `return` only runs on success. (2) the
`restore_config_snapshot`/`write_default_workflow!` rescue paths are by-design and were
self-verified safe (atomic_write keeps config.yml intact on a write fault).

## High

(none)

## Medium
- [ ] [tests] `warn_on_fieldless_tasks_rebinding` on the `--new-workflow` existing path (`lib/hive/commands/init.rb:189`) is untested: every fieldless-rebind warning test (`test/integration/init_test.rb:594-675`) drives the `--workflow` path via `update_existing_default_workflow!`, so a regression that drops/mis-warns the in-flight-task rebind on `hive init --new-workflow X` over an existing project would go uncaught — and this warning is the one user-facing guard against silent task-load failures after a default change.

## Nit
- [ ] [tests] No standalone keyword-like-id test at the `hive workflow new` unit level (e.g. `no`/`yes`/`off`): the descriptor `id:` quoting regression is only transitively covered through the init path and the shared template (`test/integration/init_test.rb:476`); a direct `workflow_new_test.rb` case would localize a descriptor-template regression to the command that owns it.
- [ ] [tests] The `current == id && !corrupt` suppression guard at `lib/hive/commands/init.rb:189` is effectively unreachable on the existing `--new-workflow` path — re-binding to the already-current id requires the descriptor to already exist, which `refuse_overwrite!` (`workflow.rb:143`) rejects first; dead/defensive branch, noted so a future reviewer does not add a redundant test for an unreachable condition.
- [ ] [comments] The `rollback_scaffold` NOTE (`lib/hive/commands/workflow.rb:~76-82`) frames the staged-index leak as unique to "hive init --new-workflow on an existing project," but `hive workflow new`'s own `commit_scaffold!` (`workflow.rb:253-261`) commits into the same long-lived `.hive-state` worktree with only filesystem rollback and no index reset — the same exposure. The comment could mislead a maintainer into thinking `workflow new` is index-safe (a real pre-existing asymmetry the init path now guards but `workflow new` does not; `workflow new` is out of scope for this task).
- [ ] [comments] Stale line reference (pre-existing, in a touched file): `wiki/commands/init.md:55` cites `Hive::GitOps#hive_state_init` at `lib/hive/git_ops.rb:34`, but the method is at line 65 (line 34 is `current_branch`).
- [ ] [type-design] Dead key `:workflows_dir` in the `scaffold_paths` Hash (`lib/hive/commands/workflow.rb:127`) is produced but never consumed by any caller (all use `:descriptor`/`:instruction_dir`/`:instruction`); pre-existing, but silently widens the producer/consumer contract for no reason.
- [ ] [type-design] `scaffold_descriptor_commit!`'s rescue re-derives `scaffold.fetch(:paths)` (`lib/hive/commands/init.rb:175`) while a local `paths` is already bound at line 161; deliberate (the local isn't set if `scaffold_files!` raised before assignment) and correct via the `if scaffold` guard, but reads as duplication — a one-line comment would settle intent.

## Notes (no action — persona positives)
- [type-design] The Hash-based `paths` contract is idiomatic here (`.fetch` fail-fast, small stable key set, matches `new.rb:122`); promoting to Data/Struct would add ceremony without strengthening an invariant. The class-method conversion exposed only the four intended cross-command entry points; all helpers stayed `private_class_method`. The `WorkflowChoice` "descriptor is a fully resolved `Hive::Workflow`" invariant now genuinely holds on the existing path (real `WorkflowSelection.fetch!`, not a stand-in).
- [simplify] No safe simplifications worth taking: the snapshot/restore + index-reset duplication vs `update_existing_default_workflow!`, and `relative_to_hive_state` vs `relative_to_workflows_root`, both encode load-bearing differences across two classes; unifying would obscure the index-leak defense or add cross-class coupling.

PR #573, opened 2026-06-23
feat(cli): advertise project-authored workflows in --workflow help

ivankuznetsov:
## Summary

The `--workflow` help on `hive init` and `hive new` listed only the built-in workflow names (`coding, content`), so an agent reading `--help` had no way to learn that projects can author their own workflows under `.hive-state/workflows/`. The help under-claimed relative to the runtime: the `UnknownWorkflow` error already accepts and lists project-authored workflows.

Both commands now share one `Hive::CLI.workflow_option_desc` helper. It renders the built-in names (still sourced from `Hive::Workflows::Registry.ids`, so there is no second hardcoded list) followed by a note that project-authored workflows are valid and how to make one:

```
coding, content, or any project-authored workflow (author one with `hive workflow new ID`)
```

The shared helper keeps `init` and `new` symmetric and leaves a single seam for a future dynamic, project-aware list (deliberately deferred, since Thor bakes option `desc:` at class-load time, before any project workflows register).

This is a help-text and discoverability change only: no behavior, no new command, and no `--json` output changes.

## Test plan

- `ruby -c lib/hive/cli.rb` → Syntax OK.
- `bundle exec rubocop lib/hive/cli.rb test/unit/cli_test.rb` → no offenses.
- `ruby -Ilib -Itest test/unit/cli_test.rb` → green (23 runs, 115 assertions, 0 failures, 0 errors); the 3 `--workflow` help tests pass (28 assertions).
- New tests pin: the wording (`or any project-authored workflow` + `hive workflow new`), `init`/`new` cross-command symmetry, that built-in names keep coming from the frozen `WORKFLOW_VOCABULARY` (no re-hardcoded literal), and the negative contract that help does **not** enumerate active-project workflows.
- Manual smoke: `bin/hive help init` and `bin/hive help new` both render the identical `--workflow` row shown above.

## Review summary

Three compound-engineering review passes plus three pr-review-toolkit (multi-persona) passes — all converged on **Ready to merge**, with the full Requirements Trace satisfied and scope boundaries respected (no `--json` change, `writing`/project descriptors not enumerated, `STAGE_VOCABULARY`/`UnknownWorkflow` untouched). No High or Medium findings remain open. Actionable findings were all test-side robustness fixes and were applied:

- Re-anchored the DRY assertion on the frozen `Hive::CLI::WORKFLOW_VOCABULARY` (immune to registry leaks under Minitest's randomized order) instead of a live `Registry.ids` recomputation.
- Tightened built-in advertisement to assert the joined `WORKFLOW_VOCABULARY` substring with custom failure messages, replacing loose `/coding/` `/content/` matches.
- Added a negative test that registers a project workflow and asserts its id never leaks into rendered help, plus a helper-level project-independence assertion.
- Broadened the source-scan DRY guard so it also catches the pre-PR `valid: coding, content` prefixed form.

Remaining items were triaged NO-FIX as intentional design choices (the two-test split; the order-specific smoke-guard; substring-not-equality symmetry, per the plan's help-brittleness mitigation). No escalations required a user decision.

## Linked task

Hive task: `ux-the-workflow-option-help-260622-8013`


ivankuznetsov:



## High

_None._

## Medium

_None._

## Nit

- [ ] `test_workflow_option_desc_is_symmetric_and_dry` anchors its DRY assertion on a *live* recomputation (`Registry.ids.join(", ")` at test time, cli_test.rb:146) instead of `Hive::CLI::WORKFLOW_VOCABULARY` — the constant the helper actually uses, frozen at class load to built-ins only: since Minitest randomizes order, a future test that registers a *project*-tier workflow without resetting would make `built_ins` diverge from the frozen `desc` and flake `assert_includes desc, built_ins`; asserting against the constant pins the real DRY source and removes the coupling.
- [ ] DRY regression guard `refute_match(/["']coding, content/, cli_source)` (cli_test.rb:158) only catches the exact `coding, content` literal in that order/spacing — a re-hardcode like `"content, coding"` or `"coding,content"` would slip through; adequate for the precise regression of concern, so optional.

ivankuznetsov:



## High

_None._

## Medium

- [ ] [pr-test-analyzer] Brittle live-vs-frozen DRY anchor (`test/unit/cli_test.rb:146,148`): `assert_includes desc, Hive::Workflows::Registry.ids.join(", ")` recomputes the built-in list *live* at test time, but `desc` is frozen to built-ins-only at class load (`WORKFLOW_VOCABULARY`, cli.rb:24); under Minitest's randomized order, any sibling test that registers a project/runtime workflow without cleanup would make `built_ins` diverge from `desc` and fail this assertion — anchoring on `Hive::CLI::WORKFLOW_VOCABULARY` (the exact frozen source the helper uses) tests the same DRY property and is immune to registry leaks (passes today only because every registrar cleans up via `ensure`).

## Nit

- [ ] [pr-test-analyzer] Loose unanchored matches `assert_match(/coding/, out)` / `/content/` (`test/unit/cli_test.rb:139-140`): `content` is a substring of words like "context"/"contents", so a future `long_desc` edit could satisfy these without the built-in actually being advertised — asserting the joined `coding, content` substring is tighter (each id currently appears only once on the `--workflow` line, so no false positive today).
- [ ] [code-simplifier] Hardcoded built-in names at `test/unit/cli_test.rb:139-140` lack custom failure messages (unlike the assertions above them) and don't track the registry; deriving them by iterating `Hive::Workflows::Registry.ids` would make the test self-documenting on failure and rot-proof if a built-in is renamed.
- [ ] [pr-test-analyzer] No negative assertion pinning the one behavioral guarantee (`test/unit/cli_test.rb`): the documented contract is that help renders built-ins-only and does NOT enumerate active-project descriptors — currently asserted only in wiki prose; registering a workflow via `with_registered_workflow` and asserting its id is absent from rendered `help new` would lock it (low value for a help-text change).
- [ ] [code-simplifier, code-reviewer] Source-text guard `refute_match(/["']coding, content/, cli_source)` (`test/unit/cli_test.rb:157-159`) re-reads `cli.rb` from disk and only catches a hardcode whose first two ids are literally `coding, content` in that order/spacing; it adds marginal value over the in-memory `assert_includes desc, built_ins` already present — adequate as an intentional smoke-guard against the specific regression, so optional.
- [ ] [code-simplifier] The two new tests overlap substantially (`test/unit/cli_test.rb:130-160`): `test_workflow_option_desc_is_symmetric_and_dry` largely supersets the help-text checks in `test_workflow_option_help_advertises_project_authored_workflows`, so they could collapse — though keeping them split ("help advertises the right thing" vs "the helper is the single DRY source") is a defensible readability choice.

ivankuznetsov:



## High

_None._

## Medium

- [ ] [pr-test-analyzer] DRY source-scan guard has a realistic blind spot (`test/unit/cli_test.rb:175`): `refute_match(/["']coding, content/, cli_source)` only catches a quote-hugging literal, so it would NOT catch the most-likely regression — re-introducing the pre-PR prefixed form `"valid: coding, content"` (quote, then `valid: `, then the names) — verified by direct regex probe; broaden the pattern to match the built-in names regardless of a leading prefix. (Same guard was raised in pass 1 and triaged NO-FIX/optional as an intentional smoke-guard; recorded here as the persona surfaced a sharper miss.)
- [ ] [pr-test-analyzer] `test_workflow_option_help_does_not_enumerate_project_workflows` guards the rendered boundary but not the load-bearing class-load freeze: because `option :workflow, desc: workflow_option_desc` is evaluated once at class-load (built-ins-only), a future rewrite of `self.workflow_option_desc` to dynamically enumerate `Registry.ids` would still leave the frozen option desc unchanged, so `content_fixture` never leaks and this test stays green while the helper silently becomes project-dependent — the project-independence contract rests on the eager freeze, which no test asserts directly. Consider a direct helper-level assertion (e.g. helper output is unchanged with a project workflow registered).

## Nit

- [ ] [pr-test-analyzer] Symmetry is asserted as `assert_includes new_out/init_out, desc` (containment) rather than equality of the two rendered `--workflow` rows (`test/unit/cli_test.rb`, test_workflow_option_desc_is_symmetric_and_dry): sufficient given both sites demonstrably call the shared helper, but it would not catch one command wrapping the shared string with extra prose while the other does not. Low value to tighten; noted for completeness.

ivankuznetsov:



## High

_None._

## Medium

_None._

## Nit

- [ ] [pr-test-analyzer] Symmetry asserted via substring, not equality (`test/unit/cli_test.rb:179-181`): both `help new`/`help init` are checked to *contain* `desc` via `assert_includes`, so a future divergence of `init` to a superset string would still pass; low risk since both options literally pass `desc: workflow_option_desc`.
- [ ] [pr-test-analyzer] No assertion that both options route through the single helper (`test/unit/cli_test.rb`): the DRY source-scan `refute_match(/coding, content/, ...)` catches a re-hardcoded name *list* but not a duplicated *helper call*; refactoring-resilience nuance only — the user-facing contract stays correct, not worth a brittle source-structure assertion.
- [ ] [comment-analyzer] `lib/hive/cli.rb:22-23`: the shorthand "a dynamic list changes that one method" is slightly cryptic for a cold reader; accurate but could spell out the seam (e.g. "making the list project-aware later means editing only that one method").
- [ ] [comment-analyzer] `wiki/cli.md:99`: "Thor still renders this at class-load time" is slightly loose wording — the desc is *captured/baked* at class-load and *displayed* later by `help`; the conclusion ("help does not dynamically enumerate active-project descriptors") is fully accurate, so this is optional phrasing only.
- [ ] [code-simplifier] `lib/hive/cli.rb:24`: `WORKFLOW_VOCABULARY` is now a bare comma-joined id list (the `"valid: "` caption was dropped), a semantic mismatch with sibling `STAGE_VOCABULARY` which remains a self-captioned phrase; `WORKFLOW_IDS`/`BUILTIN_WORKFLOW_IDS` would describe the bare list more truthfully. Leaving it is defensible (the new tests reference the constant by name; PR is deliberately surgical) — flagged so the divergence is a conscious choice, not drift.

PR #580, opened 2026-06-23
feat: make workflow selection a first-class setup step (CLI + web)

ivankuznetsov:
# feat: make workflow selection a first-class setup step (CLI + web)

## Summary

Choosing a project's `default_workflow` is now an explicit setup step on both surfaces that bootstrap a project. On a TTY, `hive init` renders a labeled `Workflow:` step (previously an easy-to-miss pre-prompt) and adds an inline `author a new workflow` entry. The hivebox `/repos/new` form, which had no workflow control at all and always wrote the implicit `coding` default, now carries a workflow `<select>` alongside the planning/development agent selects. Both write the same `default_workflow` into the same real `config.yml` — the choice the CLI already exposed is now equally reachable from the web (agent-native parity).

### CLI: explicit step plus inline authoring

The TTY prompt now prints a `Workflow:` header over the numbered list and appends one extra entry, `N) author a new workflow`. Selecting it asks for an id, re-prompts on invalid, reserved, or colliding ids, and then routes the whole init through the **existing** `--new-workflow` scaffold/bind path rather than duplicating any scaffold logic:

- `prompt_workflow` sets `@new_workflow` and returns `nil`; `call` checks `@new_workflow` immediately after `resolve_workflow_choice` and hands off to `init_with_new_workflow`. Because that path forks to `scaffold_and_bind_fresh` (which runs the full questionnaire) on a fresh init, the rest of setup still runs — authoring does not short-circuit it.
- The descriptor is never stuffed into `WorkflowChoice` (whose invariant is a fully resolved `Hive::Workflow`); the not-yet-scaffolded id flows through `@new_workflow` instead. `WorkflowChoice` gained a construction guard so the invariant is enforced rather than caller-discipline.
- Defaults, the `size > 1` guard, raw-id labels, and the EOF→`Prompts::Aborted` zero-footprint abort (ADR-023) are all unchanged. Non-TTY init still silently selects `coding`.

New `Hive::Commands::Workflow.scaffold_collision?` lets the prompt pre-check an id without reaching into scaffold internals. It and the scaffolder's hard `refuse_overwrite!` backstop share one private `scaffold_collisions` helper, so a TOCTOU collision after the pre-check still surfaces as the existing scaffold error. `hive init --help` now documents the `Workflow:` step and the inline-author entry.

### Web: a select-only, project-aware control

`/repos/new` gets a `settings[workflow]` select that mirrors the agent selects (raw ids, `coding` first, native form submit — no JS, no inline styles). It is intentionally select-only; inline authoring stays CLI-only for now (a documented parity gap).

| Form mode | Options listed | Preselected |
|---|---|---|
| Fresh clone | built-ins only (`coding`, `content`) | `coding` |
| Re-run setup | built-ins + that project's authored workflows | project's current `default_workflow` |

`InitSetup.workflows(project_root = nil)` drives the list. The `nil` (fresh) branch reads the frozen `Registry::WORKFLOWS.keys` directly rather than the mutable overlay, so a previously loaded project can't leak into a fresh-clone form; the registered-root branch wraps the load and `Registry.ids` read in `Project.synchronize` so a concurrent StatusFeed overlay swap can't split them. The controller threads the choice through to `Init.new(workflow:)` as a separate kwarg, leaving the `prompts:` answers hash untouched.

Hardening landed during review: the web-supplied id is validated against `SAFE_SLUG` before any path use; a corrupt project `config.yml` degrades the re-run form to `coding` instead of 500-ing; the view's `@workflows | [@current_workflow]` union keeps a persisted non-`coding` default selected so a bare submit can't silently downgrade it; and `WorkflowSelection.fetch!` now wraps its `load!`+`fetch` in `Project.synchronize`. An unknown id is rejected by Init's existing validation as a readable 422.

## Test plan

- `test/integration/init_test.rb`: non-TTY skip → implicit `coding`; the inline-author happy path (scaffold + bind, questionnaire still runs); re-prompt cases for reserved, invalid-format, and scaffold-collision ids (including a two-bad-entries loop boundary); EOF at the inline `New workflow id:` sub-prompt → exit 64, zero `.hive-state`; selection-by-name and out-of-range index; a pick-2-vs-3 discriminator.
- `test/unit/commands/init_test.rb` / `test/unit/commands/workflow_new_test.rb`: the `WorkflowChoice` construction-guard invariant; `scaffold_collision?` reports free vs. colliding ids.
- `web/test/integration/repos_test.rb`: fresh built-in list renders; a posted `settings[workflow]` writes the same real `default_workflow` CLI init writes; the re-run POST rebinds the on-disk default; a re-run that changes non-workflow answers asserts the deliberate discard; a bare POST (no `settings[workflow]`) → `AlreadyInitialized`/422; a broken-config `GET` degrades without 500.
- `web/test/system/repo_setup_flow_test.rb` (Capybara Playwright): selecting `content` on a fresh setup writes real `config.yml`; re-run lists a project-authored workflow and preselects the current default.

Verified across review passes: workflow-tagged init integration tests and the `workflow_new` unit tests pass; `bin/rails test test/integration/repos_test.rb` green; `rubocop` and `web/bin/rubocop` clean on every touched `.rb` file (`.erb` excluded per CLAUDE.md).

## Review summary

Three review passes ran (Claude `ce-code-review`, Codex `ce-code-review`, and `pr-review-toolkit` multi-persona), pass 3 reaching **no High or Medium correctness/security/scope findings**.

- **Stale-base High (all reviewers, pass 1):** the branch was 2 commits behind `origin/main`, so the raw `origin/main..HEAD` diff appeared to revert shipped PRs #578/#579 (`workflow new --template`). Resolved by merging `origin/main` and re-applying the `scaffold_collision?`/`scaffold_collisions` factoring on top of the template-based `scaffold_paths` (the inline author scaffolds from `blank`, so the pre-check uses `blank`); #578/#579 behavior verified intact.
- **Auto-fixed:** corrupt-config 500 → readable degrade; silent `coding` downgrade on re-run (union-select guard, matching the CLI's bare-Enter guard); `SAFE_SLUG` path-traversal validation on the web id; `fetch!` overlay synchronization; the inline-author routing made data-driven (nil-return seam, no flag re-read); plus comment-accuracy, test-robustness (derive author index / soften exact-count pins), and namespacing nits.
- **Escalation (web re-run honesty):** the re-run form renders the full questionnaire but `Init`'s existing-project path rebinds only `default_workflow`. Changing that touches `Init`'s re-init semantics and the answers hash — explicitly **out of this plan's scope** — so re-init semantics were left untouched and a regression test now documents the workflow-only rebind as deliberate. Reworking re-run to apply the full questionnaire is a separate follow-up.

## Linked task

- Hive task: `make-workflow-selection-a-first-260623-5a6a`
- Branch: `make-workflow-selection-a-first-260623-5a6a` → `main`
- PR: https://github.com/ivankuznetsov/hive/pull/580

---


ivankuznetsov:

## High
- [ ] `hive workflow new --template` was removed: `lib/hive/cli.rb:327` no longer declares or forwards `--template`, `Hive::Commands::Workflow.new!` / `scaffold_files!` no longer accept `template:`, and the `writing` / `research` templates plus their tests were deleted, breaking an existing public workflow-authoring path outside this plan's scope.

## Medium

## Nit

ivankuznetsov:

# Code Review — make workflow selection a first-class init step

**Scope:** `git diff origin/main..HEAD` (CLI `hive init` workflow prompt + inline author; web `/repos/new` workflow select).
**Mode:** report-only (reviewer spawn, read-only).
**Team:** correctness, adversarial, testing, maintainability, cli-readiness, kieran-rails, agent-native (+ scope/integration analysis).

**Verification:** branch is self-consistent — workflow-tagged init integration tests (42 runs) and the `workflow_new` unit tests (22 runs) pass green; `rubocop` clean on the touched `.rb` files. The code implements the plan's U1–U8 faithfully; the web view is native (`select_tag`/`options_for_select`, no inline styles, no JS) per CLAUDE.md.

**Critical context — the reviewed diff conflates two things.** The branch diverged at merge-base `34c161ec` and is **2 commits behind `origin/main`** (#578 `fix(workflow): point workflow new at the stub instruction`, #579 `feat(workflow): workflow new --template`). The task's real work is `34c161ec..HEAD`; everything the `origin/main..HEAD` diff shows about *deleting* `templates/workflows/{research,writing}`, removing `--template`, reverting `schemas/hive-workflow-new.v1.json`, and dropping the two `--template` tests is a **stale-base artifact**, not this task's intent. Findings below are scoped to the task's real changes, except the High finding which is precisely about that stale base.

## High
- [ ] Merging this branch as-is reverts shipped PRs #578/#579: the branch predates them, so the `origin/main..HEAD` diff deletes `templates/workflows/{research,writing}/*`, removes the `workflow new --template` CLI option (`lib/hive/cli.rb`), reverts `schemas/hive-workflow-new.v1.json`, and deletes `test_scaffolds_from_a_named_template` / `test_unknown_template_is_a_usage_error_listing_available` + the `wiki/commands/workflow.md` Templates section. Land this without first merging/rebasing `origin/main` and the `--template` feature silently disappears; `lib/hive/commands/workflow.rb` will also conflict because this branch's `scaffold_collision?`/`scaffold_collisions` were built on the pre-template `scaffold_paths` (single `instruction`) while main's `scaffold_paths` now returns a template-driven `instructions` array. Action: merge `origin/main` into the branch and re-apply the collision factoring on top of the template-based scaffolder, verifying #578/#579 behavior survives.

## Medium
- [ ] A corrupt project `config.yml` makes the re-run setup form 500 instead of a readable 422: `web/app/controllers/repos_controller.rb:36` calls `Hive::Config.load(project["path"])["default_workflow"]` unguarded, but `Config.load` (`lib/hive/config.rb:575`) does `YAML.safe_load` with no `Psych` rescue and `ApplicationController` only rescues `Hive::Error`/`ParameterMissing` — so a malformed config raises an unrescued `Psych::Exception`. The sibling `InitSetup.workflows(path)` survives a broken config via its fallback, so the workflow list renders but `@current_workflow` crashes — making the very UI an operator would use to repair the project un-openable. New call site (the old `new` never read project config). Fix: rescue and fall back to `coding`, matching the sibling.
- [ ] Web re-run can silently downgrade a non-`coding` `default_workflow` back to `coding`: `web/app/views/repos/new.html.erb:39` `options_for_select(@workflows, @current_workflow)` emits no `selected` attribute when the persisted default is absent from `@workflows` (e.g. its authored descriptor was deleted/renamed, so `valid_names` drops it; or an unquoted keyword value coerced to a non-string), so the browser auto-selects the first option (`coding`) and a bare "Apply" rebinds the project default to `coding` and re-resolves in-flight field-less tasks — with no warning. The CLI prompt explicitly guards this exact downgrade (bare Enter keeps the current default); the web select has no equivalent guard. Flagged independently by correctness, adversarial, testing, and kieran-rails.
- [ ] The web re-run rebind path has no end-to-end coverage: `web/test/system/repo_setup_flow_test.rb`'s rerun test asserts only the GET (list + preselect) and never submits, and the request test exercises only the fresh-clone branch — so `reinit!(project["path"], …, workflow: selected_workflow)` on the re-run branch (`repos_controller.rb:47`), which routes through `handle_existing_project` (a different bind path than fresh init), is never asserted to write `default_workflow` to disk. Add a re-run POST that reads back `config.yml`.

## Nit
- [ ] No test feeds EOF at the inline `New workflow id:` sub-prompt (`lib/hive/commands/init.rb:1011`): it is a distinct `Prompts::Aborted` raise site from the top-level workflow prompt, the behavior is verified correct (USAGE 64, zero footprint), but every author-branch fixture eventually supplies a valid id. Add a `"4\n"`-then-EOF case to lock the abort contract at that site.
- [ ] Implicit `@new_workflow` routing seam is the cheapest-to-misread spot in the change: `prompt_workflow` sets `@new_workflow` as a side effect and returns `nil`, `resolve_workflow_choice` wraps that in `WorkflowChoice.new(descriptor: nil, …)`, and `call` builds then immediately discards it at the second `if @new_workflow` guard (`init.rb:95`). Safe today (verified never dereferenced) but a latent `NoMethodError` trap for any future edit that reads `workflow_choice.descriptor` before line 95. Plan Risk #2 acknowledges it; consider an explicit author `WorkflowChoice` variant/sentinel so routing is data-driven rather than a re-read of a mutated flag.
- [ ] `hive init --help` long_desc (`lib/hive/cli.rb`) still lists only the questionnaire and never mentions the new `Workflow:` step or the inline "author a new workflow" entry; the wiki was updated but the in-CLI help surface drifted.
- [ ] On the fresh-clone web branch the workflow id is validated only after `clone!`/`normalize_origin!` (`repos_controller.rb:51-78`): a crafted POST with an unknown id clones first, then 422s — wasted work versus the CLI's validate-before-write ordering. The reject contract is identical, so this is robustness only; optionally pre-validate before `clone!`.
- [ ] `new` vs `create` project-lookup asymmetry: `new` degrades via `registered_projects.find` (an unknown `?project=` renders a misleading "Re-run setup" form that only fails on submit) while `create` raises via `find_project!`. Edge-only and fails closed, but consider rendering a not-found notice on `new` for symmetry.

ivankuznetsov:

# PR Review Toolkit — make workflow selection a first-class init step

**Scope reviewed:** the task's real changes (`git diff 34c161ec..HEAD`). Personas were steered off the stale-base artifact (see the High `[scope]` item) so their signal targets the actual work.
**Personas run:** code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer, code-simplifier.
**Verification:** new unit + integration tests pass (22 + workflow-tagged runs, 0 failures); RuboCop clean on every changed Ruby file (`.erb` correctly excluded). The web view is native `select_tag`/`options_for_select` — CLAUDE.md-compliant (no inline styles, no JS).

## High
- [ ] [scope] Merging the branch as-is reverts shipped PRs #578/#579: the branch is 2 commits behind `origin/main`, so the `origin/main..HEAD` diff deletes `templates/workflows/{research,writing}`, removes `workflow new --template` (`lib/hive/cli.rb`), reverts `schemas/hive-workflow-new.v1.json`, and drops the two `--template` tests — none of it this task's intent. `lib/hive/commands/workflow.rb` will also conflict (this branch's `scaffold_collision?`/`scaffold_collisions` were built on the pre-template single-`instruction` `scaffold_paths`; main's now returns a template-driven `instructions` array). Action: merge `origin/main` and re-apply the collision factoring on the template-based scaffolder, verifying #578/#579 survive. (Personas were scoped away from this artifact; surfaced via scope analysis.)

## Medium
- [ ] [code-reviewer, silent-failure-hunter] Corrupt project `config.yml` → unhandled `Psych::Exception` → raw 500 on the "Re-run setup" form: `web/app/controllers/repos_controller.rb:36` calls `Hive::Config.load(project["path"])["default_workflow"]` with no rescue; `Config.load` (config.rb:575) doesn't catch `Psych::Exception` and `ApplicationController` only rescues `Hive::Error`/`ParameterMissing`. Diverges from the house pattern — the sibling `tasks_controller.rb:337` wraps the same call in `rescue StandardError`, and the line directly above (`InitSetup.workflows`) survives the same broken file via its fallback. The repair UI 500s on the one project most likely to need repair. (silent-failure-hunter rated Critical; code-reviewer Important.) Fix: rescue → log → fall back to `coding`.
- [ ] [silent-failure-hunter, code-reviewer] Web "Apply" can silently rebind a non-`coding` default to `coding`: `web/app/views/repos/new.html.erb:39` `options_for_select(@workflows, @current_workflow)` marks nothing selected when the persisted default isn't in `@workflows` (its descriptor was deleted/renamed/corrupted, so `valid_names` drops it), so the browser auto-selects the first option (`coding`); a bare submit then rewrites `default_workflow` and re-resolves in-flight field-less tasks. The protective `warn_on_fieldless_tasks_rebinding` message goes only to the Puma log, not the operator. The CLI prompt explicitly guards this exact downgrade (bare Enter keeps current); the web select has no equivalent. (silent-failure-hunter rated High.)
- [ ] [pr-test-analyzer] The web re-run rebind path is never exercised end-to-end: the system rerun test stops at the GET (list + preselect, no submit, no disk read) and the request test covers only the clone branch — so `reinit!(project["path"], …, workflow: selected_workflow)` → `handle_existing_project` (repos_controller.rb:44-48) is untested. A regression dropping `workflow:` from the re-run `reinit!` call would rebind nothing yet keep every test green. Add a `post "/repos", params: {project:, settings:{workflow:}}` request test asserting the rebound `default_workflow` on disk.
- [ ] [pr-test-analyzer] EOF at the inline `New workflow id:` sub-prompt (`lib/hive/commands/init.rb:1011`) is untested: it's a distinct `Prompts::Aborted` raise site from the top-level prompt, reached only after selecting the author entry, and the abort→USAGE(64)→zero-disk-state contract there is unverified. Add a `"4\n"`-then-EOF case asserting USAGE exit and no `.hive-state`.
- [ ] [pr-test-analyzer] No test for web re-run when the persisted `default_workflow` is no longer loadable (codifies the silent-rebind finding above) — current system tests only cover loadable defaults.

## Nit
- [ ] [type-design-analyzer, code-simplifier, comment-analyzer] `WorkflowChoice.new(descriptor: nil, source: :prompt)` at `init.rb:972` violates the type's documented "descriptor is a fully resolved `Hive::Workflow`" invariant; it's safe only by temporal coupling to the duplicate `if @new_workflow` re-check at `init.rb:95`. The flag path (short-circuit at line 89) and `scaffold_and_bind_existing` (re-`fetch!`) deliberately uphold the invariant — the prompt path is the lone exception. Recommend returning a sentinel/`AuthorRequest` from `prompt_workflow`/`resolve_workflow_choice` so the invalid value is never built (also removes the twin guard and the mid-prompt `@new_workflow` mutation). Broad consensus across both review passes.
- [ ] [code-simplifier, type-design-analyzer] The new id is normalized/validated 3–4 times in a row (`init.rb:1013` → `scaffold_collision?` re-normalizes → `init_with_new_workflow` → `scaffold_files!`) — idempotent but redundant; validate once and pass through.
- [ ] [code-simplifier] Controller `new` duplicates the `Hive::Workflows::CODING_ID.to_s` fallback and nests `if @project_name`/`if project`; flatten to a single `project = @project_name && registered_projects.find{…}` and reuse the pre-assigned `@current_workflow` as the `.presence ||` fallback (behavior-preserving).
- [ ] [type-design-analyzer] `scaffold_collision?` is a `?`-predicate that can raise `UsageError` (via `normalize_and_validate_id!`) — unconventional for a boolean; the caller rescues it so it's intentional, but worth a doc note.
- [ ] [silent-failure-hunter] On a corrupt config, `scaffold_collision?` → `Loader.workflow_dir` → `Config.load` raises `Psych::SyntaxError` that surfaces (in the CLI prompt path) as a generic `InternalError "init failed: …"` rather than a `ConfigError`-shaped, actionable message — non-silent but less helpful.
- [ ] [pr-test-analyzer] Re-prompt tests assert the error message + final bound id but not the loop boundary; feed `bad1\nbad2\ngood\n` to prove a second invalid entry re-prompts again rather than aborting after one retry.
- [ ] [pr-test-analyzer] The author-entry tests hardcode `"4"`/`"4)"`, coupling them to the fixture's built-in count; if the built-in count changes, the input would silently select a real workflow at index 4 instead of the author entry. Derive the expected index from `Hive::WorkflowSelection.valid_names(...).size + 1`.
- [ ] [pr-test-analyzer] The system rerun test name ("…preselects current default") overpromises — it asserts only GET-side rendering, never the rebind; the disk-rebind assertion belongs in the request test (Gap above).
- [ ] [comment-analyzer] `init.rb:96-98` is accurate but could name the non-obvious mechanism — that `@new_workflow` is mutated as a side effect of `resolve_workflow_choice` (deep in `prompt_workflow`) — so a future "DRY" cleanup doesn't delete the line-95 guard thinking it duplicates line 89.
- [ ] [comment-analyzer] `repos_controller.rb:25-26` adds `File.basename(raw_project)` in `new` mirroring `create`'s traversal-hardening, but without the one-line note `create` carries; add a brief comment so the two lookups stay consistent.

ivankuznetsov:

# Code Review — make workflow selection a first-class init step (pass 2)

Scope: `git diff origin/main..HEAD` (7 commits). Reviewers: correctness, adversarial,
testing, maintainability, kieran-rails, agent-native, project-standards (ce-code-review
persona team). Mode: report-only (no files mutated).

Intent (from plan): relabel the CLI TTY workflow prompt as an explicit "Workflow:" step
and add an inline "author a new workflow" entry routing through the existing
`--new-workflow` scaffolder; add a select-only workflow `<select>` to the hivebox
`/repos/new` form, passing the chosen id through `Hive::Commands::Init.new(..., workflow:)`.
ADR-023 zero-footprint abort, the `size > 1` prompt guard, the non-TTY coding default, and
the fresh-vs-rerun control flow must be preserved.

Overall: the CLI prompt/inline-author work and the web `<select>` wiring are correct,
well-tested in isolation, and faithfully match the codebase's heavy-comment style.
Agent-native parity is fully preserved (PASS). The one substantive defect is on the web
re-run path, where three independent reviewers converged.

## High

- [ ] Web "Re-run setup" silently drops every non-workflow answer while reporting "settings applied" (`web/app/controllers/repos_controller.rb:44-48`, `web/app/views/repos/new.html.erb:4,155`, `lib/hive/commands/init.rb:250-262`): the re-run form renders the full questionnaire (planning/dev agent, claude mode, reviewers, patrol, triage, budgets, daemon enrollment), the header says "the same questions hive init asks, applied to the existing checkout", the button says "Apply settings", and the flash says "settings applied" — but because the always-populated `settings[workflow]` makes `resolve_workflow_choice` return source `:flag`/`:prompt`, `call` routes to `handle_existing_project`, which persists ONLY `default_workflow` and never consumes the `prompts:` answers; flagged independently by adversarial, kieran-rails, and testing.
- [ ] This PR newly makes that misleading-success path reachable, turning a visible error into silent partial-apply (`lib/hive/commands/init.rb:106-110`): on `origin/main` the web re-run passed no workflow, so it resolved `:implicit` and raised `Hive::AlreadyInitialized` → 422 every time; passing `workflow:` now flips routing to `handle_existing_project`, and in the everyday case (coding default unchanged) `update_existing_default_workflow!` early-returns a complete no-op (`init.rb:311`) yet still redirects "settings applied" — pick one fix: apply the questionnaire on the existing-project path, narrow the re-run form/notice to workflow-only, or change the copy+flash to "workflow updated".

## Medium

- [ ] Inline "author a new workflow" on a re-init crashes with `InternalError` (exit 70) when `config.yml` is unparseable YAML (`lib/hive/commands/init.rb:1019-1036`, `lib/hive/commands/workflow.rb:76-80`, `lib/hive/workflows/loader.rb:15-19`): `prompt_new_workflow_id` → `scaffold_collision?` → `Loader.workflow_dir` → `Config.load` → `YAML.safe_load` raises `Psych::Exception`, which the `rescue Hive::Commands::Workflow::UsageError` does not catch, so it escapes to `call`'s catch-all and surfaces as a raw "init failed: Psych::…" while every sibling config read in init degrades gracefully (`current_default_workflow_with_status`, web `persisted_default_workflow`); resolve scaffold paths via `Project#workflow_dir_for` (already falls back) or widen the rescue. (Correctness reviewer judged this acceptable since zero disk state is preserved; adversarial flagged the inconsistent error contract — surfaced for the user to decide.)
- [ ] The web-supplied `workflow` value reaches path resolution without slug validation (`web/app/controllers/repos_controller.rb:43`, `lib/hive/workflow_selection.rb:8-19`, `lib/hive/workflows/project.rb:95-109`): `params.dig(:settings, :workflow)` is passed verbatim to `Init.new(workflow:)` → `WorkflowSelection.fetch!`, which (unlike the scaffold path) never runs `normalize_and_validate_id!`, so a crafted `../../…` traverses `File.join(workflow_dir, "#{id}.yml")` into `File.file?`/`parse_file` and the resulting `ConfigError` message discloses the resolved path; impact is bounded (owner-authenticated, `.yml`-only, no contents returned) but this PR newly routes web input here — validate against `valid_names`/`SAFE_SLUG` before handing it to `Init`.
- [ ] `WorkflowSelection.fetch!`'s `load!`-then-`fetch` is unsynchronized while `InitSetup.workflows` wraps its `load!`+read in `Project.synchronize` (`lib/hive/workflow_selection.rb:8-19,32-35`, `web/app/models/init_setup.rb:63-69`): a concurrent Puma request (another setup POST, a `/repos/new` GET, or a StatusFeed tick) can swap the process-global project overlay between `load!(A)` and `Registry.fetch(:writing)`, making a genuinely valid project-authored workflow miss and 422 as "unknown workflow"; the plan's own Risk-1 mitigation applied `synchronize` to the GET-list path but not to the POST-resolve path this PR newly exercises from the web tier (low-confidence timing window, pre-existing helper).
- [ ] No test proves a re-run applies (or deliberately discards) its non-workflow answers (`web/test/integration/repos_test.rb:89-104`): the new "rerun setup rebinds the project default workflow on disk" asserts only `default_workflow`, so it encodes the High finding's misleading behavior rather than catching it; add a re-run test posting a changed reviewer/budget/claude_mode alongside the workflow and assert what actually lands, plus a case for a re-run POST that omits `settings[workflow]` (the new `.presence` branch → `workflow: nil` → `AlreadyInitialized`/422).

## Nit

- [ ] `Workflow.scaffold_collision?` raises `UsageError` despite the `?`-predicate naming convention (`lib/hive/commands/workflow.rb:76-80`): it runs `normalize_and_validate_id!` before the existence check, so a future caller reading it as a pure boolean (`if scaffold_collision?(id)` with no rescue) gets an uncaught raise; the sole caller rescues it and the comment documents it, but a name like `scaffold_collides!` or splitting validate-from-check would make the raise discoverable.
- [ ] The inline-author route signals intent by mutating `@new_workflow` four frames deep inside `prompt_workflow` and returning `nil` (`lib/hive/commands/init.rb:1007`, `955-985`, `94-105`): the nil-return seam is load-bearing and one stray `&.`/`|| default` away from silently swallowing the author branch; well-commented at every hop and called out as plan Risk-2, but a typed result (a third `WorkflowChoice` source or an explicit author-result object carrying the id) would make the branch impossible to drop accidentally.
- [ ] The `scaffold_collision?` comment over-claims template-independence (`lib/hive/commands/workflow.rb:69-75`): `scaffold_collisions` also iterates `paths.fetch(:instructions)`, which ARE template-dependent — the pre-check is correct only because the inline author always scaffolds from the `blank` default; reword to "the inline author always scaffolds from blank, so the pre-check uses blank too".
- [ ] `wiki/commands/web.md` body was edited (14 insertions) but its `updated:` frontmatter is stale at `2026-06-18` while the sibling `wiki/commands/init.md` was bumped to `2026-06-23` for the same change — bump it for consistency.
- [ ] The web built-in-count test assertions hardcode exactly two built-ins (`web/test/integration/repos_test.rb:47`, `web/test/system/repo_setup_flow_test.rb:14`): `assert_select "...option", 2` and `assert_equal %w[coding content]` break if a third built-in workflow is ever added, with no corresponding behavior regression; prefer asserting "includes coding/content" + "coding first" unless the exact-set pin is intentional.
- [ ] The web setup POST returns only a redirect + flash, with no machine-readable echo of the bound `default_workflow` (`web/app/controllers/repos_controller.rb:78`): an agent driving the web surface must re-read `config.yml` to confirm the binding, whereas the CLI `--json` path echoes it inline — a minor observability asymmetry, not a parity gap (agent-native verdict is PASS).

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — workflow selection as a first-class init step (pass 2)

Scope: `git diff origin/main..HEAD`. Personas run: code-reviewer, pr-test-analyzer,
silent-failure-hunter, comment-analyzer, type-design-analyzer. Read-only; no files modified.

Overall: the CLI prompt + inline-author flow and the web `<select>` wiring are correct and
well-tested in isolation; rubocop is clean and the CLI inline-author tests pass. The one
substantive defect (web re-run silently dropping non-workflow answers behind a "settings
applied" notice) was surfaced independently by three personas. The remaining items are
test-coverage gaps, an inconsistent error path, and type/comment polish.

## High

- [ ] [code-reviewer/silent-failure-hunter] Web "Re-run setup" reports `"<project> settings applied"` but persists ONLY `default_workflow`, silently discarding every other submitted answer (`web/app/controllers/repos_controller.rb:42-48`, `lib/hive/commands/init.rb:106-112,250-262`): the re-run form renders the full questionnaire (planning/dev agent, claude mode, reviewers, patrol, triage, budgets, daemon toggles) with an "Apply settings" button, but the always-present `settings[workflow]` routes `Init#call` to `handle_existing_project`, which never invokes `collect_prompt_answers`; this PR makes the false-success path reachable (on origin/main the no-workflow re-run raised `AlreadyInitialized`→422, a loud failure) — fix by applying the full answers on the existing-project path, or narrow the re-run form to the Workflow control and reword the button/notice to "workflow updated".

## Medium

- [ ] [pr-test-analyzer] No test proves the web re-run path applies (or deliberately discards) NON-workflow settings — the new "rerun setup rebinds the project default workflow on disk" test (`web/test/integration/repos_test.rb:89-104`) asserts only `default_workflow`, so it cannot catch the High finding and instead reinforces the impression the form works; add a re-run POST with a changed `claude_mode`/`triage_bias` and assert the real on-disk outcome.
- [ ] [silent-failure-hunter] Inline "author a new workflow" on a re-init crashes with `InternalError` (exit 70, raw `init failed: Psych::…`) when `config.yml` is corrupt (`lib/hive/commands/init.rb:1019-1036` rescues only `Workflow::UsageError`, but `scaffold_collision?`→`Loader.workflow_dir`→`Config.load`→`YAML.safe_load` raises `Psych::Exception`), while sibling reads in the same flow (`current_default_workflow_with_status`, `Project#workflow_dir_for`) degrade gracefully — widen the prompt-boundary rescue to a typed `Hive::Error`, or resolve scaffold paths through the fallback-wrapping `workflow_dir_for`.
- [ ] [type-design-analyzer] `Workflow.scaffold_collision?` is a `?`-suffixed predicate that RAISES `UsageError` (it calls `normalize_and_validate_id!` first) instead of returning a pure boolean (`lib/hive/commands/workflow.rb:76-80`): safe only because the sole caller's loop rescues `UsageError`; a future `if scaffold_collision?(id)` caller hits an uncaught raise — accept an already-validated id (also removing the redundant double-normalize) or rename to a bang/verb form.
- [ ] [type-design-analyzer] `resolve_workflow_choice`/`prompt_workflow` overload `nil` as an in-band sentinel for the "author new" outcome, with the id split into the `@new_workflow` ivar (`lib/hive/commands/init.rb:95,983,1007`): the third resolution outcome is invisible from the `WorkflowChoice` type and safe only by single-caller discipline; a typed `AuthorNewWorkflow = Data.define(:id)` variant + `case/in` in `call` would make a dropped branch a loud `NoMatchingPatternError` rather than a `nil.descriptor` crash deep in `render_project_config`.
- [ ] [pr-test-analyzer] A re-run POST that omits `settings[workflow]` is untested and 422s (`web/app/controllers/repos_controller.rb:43` → `workflow: nil` → `:implicit` → `AlreadyInitialized`): add `post "/repos", params: { project: name }` pinning the intended contract for scripted/curl callers, since the browser always submits the select.
- [ ] [pr-test-analyzer] The web `persisted_default_workflow` corrupt-config degrade has no test (`web/app/controllers/repos_controller.rb:93-98`) and `web/test` is outside the gem's coverage gate, so the rescue that keeps the repair form from 500-ing is unenforced — add a `GET /repos/new?project=...` with a broken `config.yml` asserting `:success` with the select still rendered.
- [ ] [pr-test-analyzer] Inline-author on a re-init happy path is covered only incidentally via the collision case (`test/integration/init_test.rb:332`); add a dedicated re-init inline-author happy-path test asserting the single-commit rebind + preserved project registration + fieldless-task rebind warning, matching the `--new-workflow` flag coverage.
- [ ] [pr-test-analyzer] Brittle hardcoded built-in count/list assertions (`web/test/integration/repos_test.rb:47` `option, 2`; `web/test/system/repo_setup_flow_test.rb:14` `%w[coding content]`): a third built-in breaks them with no behavior regression — derive from `InitSetup.workflows`/assert "coding first + content present", as the CLI tests already do by deriving `author_index`.

## Nit

- [ ] [silent-failure-hunter] `persisted_default_workflow` rescues bare `StandardError`, wider than its documented intent (`web/app/controllers/repos_controller.rb:95`): it also swallows `Hive::ConfigError` and `NoMethodError`-class bugs as a benign "fell back to coding" warn — narrow to `Psych::Exception, Hive::ConfigError, SystemCallError, IOError` to match the precedent in `Project#workflow_dir_for`.
- [ ] [comment-analyzer] `lib/hive/commands/init.rb:98` comment names the wrong method — "prompt_new_workflow_id set @new_workflow as a side effect"; the assignment is actually in `prompt_workflow:1007` (`@new_workflow = prompt_new_workflow_id`), and the sibling comment at `:978` already states it correctly — misleads a maintainer tracing `@new_workflow` writes.
- [ ] [comment-analyzer/type-design-analyzer] `lib/hive/commands/workflow.rb:72-75` over-claims collisions are "template-independent": `scaffold_collisions` also checks the template-dependent per-stage `instructions`; the result is template-independent only because each instruction lives under the already-checked `<id>/` dir (redundant for detection) — reword to state that real invariant.
- [ ] [type-design-analyzer] `WorkflowChoice = Data.define(:descriptor, :source)` has no construction guard (`lib/hive/commands/init.rb:56`): `descriptor: nil`/bad `source` are constructible, so the "descriptor is a resolved Hive::Workflow" invariant rests on caller discipline; if the typed-variant refactor is declined, add an `initialize` guard validating `source` and non-nil `descriptor`.
- [ ] [type-design-analyzer] Redundant double `normalize_and_validate_id!` on the inline-author happy path (`lib/hive/commands/init.rb:1026` then `lib/hive/commands/workflow.rb:77`) — harmless (idempotent) but a symptom of `scaffold_collision?` conflating validation with the collision check.
- [ ] [type-design-analyzer] `InitSetup.workflows` reads different sources per branch — `Registry::WORKFLOWS.keys` (nil arm) vs `Registry.ids` (project arm) (`web/app/models/init_setup.rb:24-31`): coherent in the web tier today but could skew if a process-wide overlay were ever active; unify on `Registry.ids` and add a one-line note that the nil arm is deliberately lock-free (reads only the frozen constant).
- [ ] [pr-test-analyzer] `web/test/system/repo_setup_flow_test.rb:18` uses `wait: 0` on a negative `assert_no_selector` — safe only because the prior `assert_selector "label", text: "Workflow", wait: 5` synced the render; it is the classic flake pattern CLAUDE.md E2E rule 6 warns against if that sync ever weakens.
- [ ] [pr-test-analyzer] Positional prompt-substring assertions (`test/integration/init_test.rb:174-176`: `"1) coding"`/`"2) content"`/`"3) content_fixture"`) hardcode both numbering and built-in order — refactor-fragile for cosmetic changes, though currently stable.

ivankuznetsov:

## High
- [ ] Web re-run setup still promises a full settings apply while only `default_workflow` is persisted: `new.html.erb` says the same `hive init` questions are applied and the submit/flash say settings were applied, but the existing-project `Init` path ignores `prompts:` and the new regression test confirms non-workflow answers are discarded.

## Medium

## Nit

ivankuznetsov:

# ce-code-review — make workflow selection a first-class init step (pass 3)

Reviewed `git diff origin/main..HEAD` against the plan. Scope: CLI relabel + inline-author
(U1/U2), `Workflow.scaffold_collision?` predicate, `WorkflowSelection.fetch!` synchronize,
web `InitSetup.workflows` (U3), controller wiring (U4), view select (U5), CLI tests (U6),
web request/system tests (U7), wiki updates.

**Verification performed (no code edited):**
- `ruby -Itest test/integration/init_test.rb` → 83 runs, 708 assertions, 0 failures.
- `ruby -Itest test/unit/commands/workflow_new_test.rb` → 24 runs, 0 failures.
- `bin/rails test test/integration/repos_test.rb` (web) → 15 runs, 0 failures.
- `rubocop` (CLI 5 files) and `web/bin/rubocop` (web 4 files) → 0 offenses.
- System test (`web/test/system/repo_setup_flow_test.rb`) not executed here (needs Chromium);
  its on-disk assertion is mirrored by the green request-level integration test.

Implementation matches the plan's intent: every Requirements-Trace row is implemented, the
ADR-023 zero-footprint abort contract is preserved (prompt runs before any write; EOF at the
inline new-id sub-prompt → `Prompts::Aborted` → exit 64 with zero `.hive-state`), the
`size > 1` guard and bare-Enter default logic are unchanged, the inline-author path reuses the
existing `--new-workflow` scaffolder rather than duplicating it, and the web path is select-only
with the documented path-traversal hardening (`SAFE_SLUG`) and corrupt-config degrade. No High
or Medium correctness, security, or scope findings.

## High

_None._

## Nit
- [ ] `WorkflowSelection.fetch!` rescue path reads `valid_names` outside the `Project.synchronize` block it just exited (`lib/hive/workflow_selection.rb:34`): `valid_names` re-runs `load!` then reads `Registry.ids` non-atomically, so under concurrent StatusFeed/web overlay churn the "valid workflows: …" list in the `UnknownWorkflow` 422 can reflect a different project's overlay — a cosmetic, error-message-only staleness (the happy resolve path is correctly synchronized; pre-existing `valid_names` non-atomicity, not introduced here). Wrapping the rescue's `valid_names` in the same reentrant `synchronize` would close it cheaply.
- [ ] `ReposController#new` assigns `@workflows`/`@current_workflow` twice — fresh built-in defaults (`repos_controller.rb:32-33`) then re-run overrides (`repos_controller.rb:38-39`) — and the re-run branch issues two independent disk reads (`InitSetup.workflows(path)` → `Project.load!` and `persisted_default_workflow` → `Config.load`). Idiomatic Rails default-then-override and negligible cost; noting only for completeness, no change needed.

ivankuznetsov:

# pr-review-toolkit (multi-persona) — make workflow selection a first-class init step (pass 3)

Ran the pr-review-toolkit personas in parallel over `git diff origin/main..HEAD`:
**code-reviewer**, **pr-test-analyzer**, **comment-analyzer**, **silent-failure-hunter**,
**type-design-analyzer**, **code-simplifier**. Findings below are the union, deduped, with the
out-of-scope items (web inline authoring, raw-id labels, re-run non-workflow answer discard,
`hive workflow new` index-safety) dropped per the plan's scope boundaries.

Consensus: the change meets standards. No High findings from any persona. code-reviewer found
nothing at threshold (security/concurrency/Rails/CLAUDE.md all clean); silent-failure-hunter
confirmed every rescue path is surfaced (logged/raised/422), not swallowed; comment-analyzer
found the dense comments unusually accurate (no rot); type-design rated the `WorkflowChoice`
guard 8–9/10. The two Mediums are interactive-selection test gaps; the rest are Nits.

## High

_None._

## Medium
- [ ] [tests] Top-level `prompt_workflow` unknown-answer re-prompt branch (`lib/hive/commands/init.rb:1025-1029`) is untested: all three re-prompt tests exercise the inline-author *new-id* sub-prompt (`prompt_new_workflow_id`), not the workflow-*selection* loop — a regression that broke the selection retry or dropped the `"unknown workflow … 1..N"` hint would go uncaught (`--workflow bogus` at init_test.rb:432 is a different `fetch!` path).
- [ ] [tests] Workflow selection by name and by out-of-range index (`resolve_workflow_answer`, `lib/hive/commands/init.rb:1052-1062`) is untested: every prompt-pick test feeds an in-range numeric (`"3\n"`); nothing types a workflow name (e.g. `content_fixture`) or an out-of-range number, so the case-insensitive name-match and bounds branches are pure assertion gaps.

## Nit
- [ ] [tests] TTY-pick test (`test/integration/init_test.rb:152`) proves `3` → `content_fixture` but never pins that a *different* pick yields a *different* on-disk default, so it wouldn't catch a "selection ignored, always writes the one registered non-coding default" regression — a pick-2-vs-3 discriminator would harden it.
- [ ] [tests] Fresh-clone "chosen settings reach init" (`web/test/integration/repos_test.rb:60`) asserts `workflow: content` lands on disk but doesn't assert the prompt path is skipped (web supplies the workflow as a flag); only implicitly covered by the system test.
- [ ] [type-design] `SOURCES` constant declared inside the `Data.define do…end` block (`lib/hive/commands/init.rb:60`) binds to the lexical enclosing `Hive::Commands::Init`, not the value object — it is `Init::SOURCES`, not `WorkflowChoice::SOURCES`. Harmless (the bare `SOURCES` reference resolves lexically) but reads as if it belongs to the type; define it as `WorkflowChoice::SOURCES = …` after the block for honest namespacing.
- [ ] [type-design] No direct unit test pins the new `WorkflowChoice` invariant: nothing asserts `WorkflowChoice.new(descriptor: nil, …)` or `source: :bogus` raises `ArgumentError` (existing tests only hit the happy `:flag` path) — a two-line `assert_raises(ArgumentError)` pair would lock the guard against a future `initialize`/`super` refactor silently dropping it.
- [ ] [comments] The `new`-action comment (`web/app/controllers/repos_controller.rb:26-27`) over-states the mechanism: it says `File.basename` stops a crafted `project` from "walk[ing] out of the registered-projects lookup," but that lookup is a string-equality `find` on `entry["name"]` where a `../`-laden name simply wouldn't match — the real value is defense-in-depth / parity with `create`. Reframe to avoid implying it blocks a traversal in the lookup itself.
- [ ] [comments] `scaffold_collision?` "so it never raises" (`lib/hive/commands/workflow.rb:73`) is scoped to the id argument, not absolute — the predicate can still raise on a corrupt project config (`workflow_dir` → `Config.load` → `ConfigError`); consider "never raises *on a pre-validated id*" if tightening.
- [ ] [errors] Inline-authored id is validated twice — `prompt_new_workflow_id` calls `normalize_and_validate_id!`, then `call` → `init_with_new_workflow` re-validates `@new_workflow` (`lib/hive/commands/init.rb:1023-1024`/160). Idempotent and harmless, but the second call can now only fail on a programming error, which would surface via the generic `StandardError → InternalError` arm rather than the clean USAGE path the first call uses. Awareness only.
- [ ] [errors] `persisted_default_workflow` logs the corrupt-config degrade at `Rails.logger.warn` with no error-ID/Sentry hook (`web/app/controllers/repos_controller.rb:114`). Reasonable for an expected operator-recoverable state (matches `normalize_origin!`'s `warn` breadcrumbs at 179/190); if the repo wires a Sentry/error-ID convention for web warnings, match it.
- [ ] [simplify] The `/\A\d+\z/` numeric-parse idiom now appears 3× in `init.rb` (author-entry check at `:1020`, `resolve_workflow_answer` at `:1053`). `Integer(value, exception: false) == author_index` collapses the guard+convert at `:1020` and a tiny private `numeric_choice(value)` helper would centralize the rest — behavior-preserving readability only.

---

**Persona notes (not separate findings):** code-reviewer reported zero issues. silent-failure-hunter
and comment-analyzer's substantive checks were all confirmations (rescue sets match the comments'
claims exactly; the view's `@workflows | [@current_workflow]` union genuinely converts a silent
coding-downgrade into a loud 422, not a hidden one). code-simplifier explicitly flagged the
nil-return routing seam, the extracted `scaffold_collisions`, the `fetch!` synchronize wrap, and
the `InitSetup.workflows` nil/project asymmetry as deliberate and **not** to be "simplified" away.

PR #587, opened 2026-06-24
feat(bot): add log severity and tame bot.log noise

ivankuznetsov:
## Summary

`bot.log` had become unreadable: benign Telegram long-poll timeouts and a per-tick dedupe line buried every meaningful event. In one current log file, `notification_skipped_dedupe` accounted for 31,828 lines against roughly 148 `notification_sent` events, and benign poll timeouts added ~20k more across rotations. This PR gives every log line a severity (`level`) and an optional `category`, then routes the two dominant noise sources to `debug`/`noise` while keeping genuine failures loud. Event names are unchanged, so existing greps-by-event keep matching; only the volume drops.

**What changed.** The bot-log schema bumps to `hive-bot-log.v3` (`SCHEMA_VERSION = 3`): `level` is now required on every line, `category` is an optional cross-cutting tag, and a new `poll_unhealthy` event is added. `v1` and `v2` stay checked in unchanged so historical lines remain valid.

| Event | Level / category | When it logs now |
|-------|------------------|------------------|
| benign `poll_failure` (transport timeout) | `debug` / `noise` | every benign `getUpdates` timeout — line kept, confirms the loop is alive |
| non-benign `poll_failure` | `warn` (map default) | per occurrence, unchanged |
| `poll_unhealthy` | `warn` | once per outage, after N consecutive failures or M seconds without a successful poll |
| `notification_skipped_dedupe` / `_backoff` | `debug` / `noise` | only on skip-state entry, fingerprint change, or after a row leaves and returns |
| `notification_sent`, `_active_conversation`, `_daemon_plan_pause` | `info` | unchanged |

**Design decisions.**

- *Severity lives in the Logger, not at every call site.* A frozen `LEVELS` map (event → level, default `info`) is the single source of truth, mirroring how the codebase already centralizes contracts. Call sites override only the two contextual downgrades (benign poll failures, transition-only skips); everything else inherits its level from the map. Note for reviewers: stub loggers capture raw kwargs and do not apply the map, so map-default levels are only observable through the real `Logger` (asserted in `logger_test`), while contextual overrides are passed explicitly and visible to stubs.
- *Benign timeouts stay visible, outages stay loud.* An error-class allowlist (`Faraday::TimeoutError`, `Faraday::ConnectionFailed`, `Net::ReadTimeout`) downgrades long-poll noise without deleting the line. A small injectable `PollHealth` collaborator tracks consecutive failures and time-since-success, firing `poll_unhealthy` once per unhealthy episode (latched until the next success re-arms it) so a real Telegram outage surfaces even though individual timeouts are quiet. Thresholds default to 5 consecutive / 60s.
- *Dedupe/backoff skips log on transition only.* The dispatcher keeps two in-memory sets keyed by fingerprint; each tick reconciles them against the current row set so a row that changes or leaves-and-returns logs again on genuine re-entry. Delivery dedupe itself is unchanged — only the log line is gated. A SIGHUP reload or process restart resets these sets, so one skip line per active fingerprint may reappear after a restart; this is accepted as within the spirit of "log on state transition."

## Test plan

- `bundle exec rake test` — full suite green: 6339 runs, 0 failures, 0 errors (6 pre-existing env skips, e.g. the real-API `telegram_smoke_test` gated on `HIVE_TELEGRAM_BOT_TOKEN`).
- `bundle exec rake coverage` — coverage gate passed at **100.00% lines** (27328/27328).
- `bundle exec rubocop` — clean on all 8 changed `.rb` files.
- Schema parity verified: `Logger::EVENTS` ↔ `hive-bot-log.v3.json` `event` enum exact (33/33); `LEVELS` keys ⊆ `EVENTS`; `v1`/`v2` schema files byte-unchanged vs `origin/main`.
- New / updated coverage:
  - `poll_health_test.rb` — consecutive vs silence escalation, once-per-episode latch, re-arm after success, backwards-clock clamp.
  - `telegram_test.rb` — benign vs real poll-failure levels, `poll_unhealthy` via both rescue arms, `:consecutive` and `:silence` escalation paths, success→re-arm seam, and the `build_update` parse-failure must-not-escalate invariant.
  - `notification_dispatcher_test.rb` — transition-only dedupe/backoff logging (entry, fingerprint change, leave/return) at `debug`/`noise`.
  - `logger_test.rb` — map defaults, explicit `level`/`category` overrides, invalid-level rejection, v2 back-compat, schema-enum bijection, and an info-stream noise-ratio assertion driven by the production `LEVELS` map.

## Review summary

Three independent review passes (ce-code-review, pr-review-toolkit's 6-persona union, and a Codex pass) ran against the branch. Every pass reported **no outstanding High findings and no plan-required-but-missing item**: the severity model, `poll_unhealthy` escalation latch/re-arm, two-arm rescue, and transition-only skip logging were each independently confirmed correct.

Findings raised across passes were either auto-fixed or resolved no-fix:

- **Auto-fixed:** de-circularized the noise-ratio acceptance test (now relies on the production `LEVELS` map, not hand-tagged levels); added backoff fingerprint-change + leave/return re-log tests; added the `build_update` must-not-escalate guard, `:silence` and StandardError-arm escalation tests, and the success→re-arm seam test; stamped the validated `level` *after* the attrs merge so a string-keyed `level`/`category` attr can't poison the v3 line; added `LEVELS ⊆ EVENTS` and schema-enum bijection guards; rounded `seconds_since_success` to an Integer; extracted `Logger::CATEGORY_NOISE`; and corrected/expanded wiki + schema docs.
- **Resolved no-fix (by-design, per plan/brainstorm):** in-memory dedupe sets that reset on reload (accepted one-line-per-fingerprint reappearance); consecutive-OR-silence outage detection (`any success = alive`); the once-per-episode latch with no recovery line; the `::Telegram::Bot::Exceptions::ResponseError` → `warn` reclassification; and the spec-locked explicit `level:` at noise call sites (stub loggers don't apply the map).

The only integration note is branch hygiene: the base is a couple of commits behind `origin/main`, so a raw `origin/main..HEAD` diff appears to touch `release.yml`/`CHANGELOG.md` — a stale-base artifact the branch's own commits never modify, so a normal merge preserves main's versions. Git-verified across all passes.

## Linked task

- Hive task slug: `telegram-bot-bot-log-is-260624-7d9b`
- Branch: `telegram-bot-bot-log-is-260624-7d9b`
- PR: https://github.com/ivankuznetsov/hive/pull/587

---


ivankuznetsov:

## High
- [ ] `.github/workflows/release.yml:280` runs the published-image smoke on `macos-15` with `colima start`: GitHub-hosted arm64 macOS runners do not support nested virtualization, so this release job will fail before `packaging/docker/smoke.sh` can run; keep the native `ubuntu-24.04-arm` Docker smoke or use a runner with supported Docker.

## Medium
- [ ] `lib/hive/bot/logger.rb:95` merges arbitrary attrs after stamping the validated `level`: a caller can pass string-keyed `level`/`category` attrs (for example `**{"level"=>"verbose"}`) and write a v3 line that fails `hive-bot-log.v3.json`, breaking the plan guarantee that every emitted v3 event has a validated severity.

## Nit

ivankuznetsov:

# ce-code-review — telegram-bot-bot-log-is-260624-7d9b (pass 1)

Reviewed `git diff origin/main..HEAD` (4 task commits: U1–U4). Verified locally:
full suite green (6339 runs, 0 failures/0 errors, 6 pre-existing env skips),
**coverage gate passed — 100.00% lines (27328/27328)**, rubocop clean on all 8
changed `.rb` files, EVENTS↔v3-schema-enum parity (33/33), LEVELS map keys/values
all valid, v2 schema byte-unchanged.

The implementation fully satisfies the plan's Goals and every row of the
Requirements Trace (U1 severity model + v3 schema; U2 benign classification +
`poll_unhealthy`; U3 transition-only dedupe/backoff; U4 kept-skip levels +
acceptance tests). No plan-required item is missing → **no High findings**.

## High


## Medium
- [ ] Branch is 2 commits behind `origin/main`; the reviewed diff shows reverts of `.github/workflows/release.yml` (#584) and `CHANGELOG.md` (#583): these are a stale-base artifact, not task work — `git merge-base origin/main HEAD` = `9efbca2a`, and main has since gained #583/#584. A normal merge won't actually revert them (the branch's own commits don't touch those files), but the branch should be rebased onto current main before merge so CI validates against current main and the diff stops appearing to undo the arm64 release-smoke fix (re-breaking that fix would re-break release CI on hosted macОS runners).

## Nit
- [ ] The criterion-(a)/(b) acceptance test `test_info_stream_excludes_noise_events_under_repeated_poll_and_dedupe_chatter` (`test/unit/bot/logger_test.rb`) is partly circular: it hand-tags the 100 `poll_failure` / 100 `notification_skipped_dedupe` lines with `level: :debug, category: :noise` itself, so on its own it only proves the Logger keeps explicitly-`debug` lines out of the `info` view — not that the production paths actually classify them as noise. (Low impact: the real classification is independently pinned by `telegram_test` benign-branch cases and the `notification_dispatcher_test` transition-only cases, so a production regression would still be caught there.)
- [ ] Backoff transition-logging has only the "logged once while unchanged" case (`test_backoff_skip_is_logged_once_while_fingerprint_is_unchanged`); unlike dedupe, there is no test pinning a backoff re-log after the fingerprint leaves `current` and returns (or changes). The reconcile mechanism (`reconcile_skip_log_state` / `keep_if`) is shared with dedupe and is covered by the dedupe leave/return + fingerprint-change tests, so this is redundancy rather than a gap — but the backoff path's re-entry behavior isn't independently asserted.

ivankuznetsov:

# pr-review-toolkit — telegram-bot-bot-log-is-260624-7d9b (pass 1)

Union of findings from 6 persona agents (code-reviewer, pr-test-analyzer,
silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier)
on `git diff origin/main..HEAD`. Personas were told to ignore the
`.github/workflows/release.yml` / `CHANGELOG.md` portions of the diff (a
stale-base artifact: branch base is 2 commits behind origin/main, which gained
#583/#584). Verified independently: full suite green (6339 runs, 0 failures),
coverage gate **passed at 100.00% lines**, rubocop clean.

No persona found a Critical/correctness defect or a plan-required-but-missing
item → **no High findings**. silent-failure-hunter explicitly confirmed the
benign-timeout downgrade cannot hide an outage (the dual consecutive-OR-silence
`poll_unhealthy` warn line is guaranteed; both rescue arms feed the health
machine; `record_success` only fires on a non-raising transport round-trip).

## High


## Medium
- [ ] [tests] Noise-ratio acceptance test is circular (`test/unit/bot/logger_test.rb:88-110`): it hand-passes `level: :debug, category: :noise` to `logger.event`, so it only proves the logger echoes the tags the test supplied — not that production classifies these events as noise. A regression of `LEVELS[:poll_failure]` to `:info` would still pass it. The `< 0.05` assertion (line 108) is also tautological since `noise_in_info.size == 0` is asserted on line 107 (0/20 is always < 0.05). This is the headline test for plan acceptance criterion (a); the real classification is covered elsewhere (`test_default_levels_come_from_event_map`, telegram/dispatcher tests), so the criterion holds in aggregate, but this test doesn't carry its weight.
- [ ] [tests] Backoff transition-logging is missing the re-log-on-fingerprint-change and leave-and-return cases (`test/unit/bot/notification_dispatcher_test.rb`): dedupe has all three (logged-once `:292`, change `:307`, leave/return `:323`) but backoff has only logged-once (`:822`). The `@backoff_logged.keep_if` eviction (`notification_dispatcher.rb:223-226`) and the distinct `backoff_active?` branch (`:199`) have no test asserting re-entry behavior; a refactor sharing one Set or dropping backoff from reconcile would go uncaught.
- [ ] [type-design] `PollHealth::Result` permits contradictory states (`lib/hive/bot/poll_health.rb:7-11`): `escalate` (bool) and `reason` (nullable) are independent fields with the "reason non-nil iff escalate" invariant enforced only by the single producer (`record_failure`). `Result.new(escalate: false, reason: :silence, ...)` is freely constructible. No live bug (producer is correct), but it's a public return type consumed by `telegram.rb:205-211`; deriving `escalate? = !reason.nil?` and dropping the `escalate` field would make the illegal states unrepresentable.
- [ ] [scope/branch-hygiene] Branch base is 2 commits behind `origin/main`; the reviewed diff therefore appears to revert `release.yml` (#584 arm64 release-smoke fix) and `CHANGELOG.md` (#583). It's a stale-base artifact, not task work, and a normal merge won't revert them — but the branch should be rebased onto current main before merge so CI validates against current main and the diff stops looking like it undoes the release-CI fix.

## Nit
- [ ] [type-design] `seconds_since_success` is emitted as a Float (`lib/hive/bot/poll_health.rb:32`; `Time - Time` → Float), so `poll_unhealthy` log lines carry e.g. `"seconds_since_success":61.0003` rather than an Integer. The unit test `assert_equal 11, ...` (`poll_health_test.rb:46`) passes only because `11 == 11.0`, masking the type; `.to_i`/`.round` would match the `consecutive_failures` Integer sibling and keep log lines tidy.
- [ ] [tests] `poll_unhealthy` escalation is only driven through the benign `Faraday::TimeoutError` rescue arm (`telegram_test.rb:227`); the `rescue StandardError` arm also calls `emit_poll_unhealthy_if_needed` (`telegram.rb:89`) but no test sustains a generic error to escalation — a copy-paste regression dropping that call from only the StandardError arm would be unnoticed.
- [ ] [tests] No Telegram-layer test exercises the `:silence` escalation reason end-to-end (only `:consecutive`); silence is unit-tested in `poll_health_test.rb` but the integration path emitting `reason: "silence"` is unverified.
- [ ] [tests] The `seconds_since_success` clamp `[..., 0].max` (`poll_health.rb:32`, guards a non-monotonic clock) is never exercised with a backwards clock.
- [ ] [simplify] Redundant `level: :warn` at the `poll_unhealthy` call site (`telegram.rb:208`): `LEVELS[:poll_unhealthy]` is already `:warn` (`logger.rb:52`), so the explicit arg duplicates the table (two places must agree). The benign `poll_failure` `level: :debug` (`:82`) and the dispatcher `level: :debug` calls genuinely override the table default and should stay.
- [ ] [simplify] `require "set"` (`notification_dispatcher.rb:2`) is unnecessary — the gemspec floor is Ruby `>= 3.4.0` and `Set` is autoloaded since 3.2. Harmless; remove or keep as explicit-dependency style.
- [ ] [type-design] `LEVEL_NAMES` (`logger.rb:10`) is an Array used purely for membership (`.include?`); a `Set` would express unordered-membership intent. Also the ascending severity order is currently incidental — worth a comment if a future "≥ warn" comparison ever relies on it.
- [ ] [comments] `wiki/modules/bot.md:23` labels all three benign classes as "transport timeouts," but `Faraday::ConnectionFailed` is a connection failure, not a timeout; "transport errors" (as the log.d fragment already says) is accurate.
- [ ] [comments] `schemas/hive-bot-log.v3.json:5` description says v3 "adds a required `level` field and optional `category` tag to every log line" — true for `level` (always stamped) but `category` is only conditionally present (`logger.rb:102` omits it when nil); the "to every line" phrasing over-implies category is always present.
- [ ] [comments] (pre-existing, outside diff scope) `wiki/commands/bot.md:285-286` still says the forward-skew path "logs a `poll_failure` skew warning," but the code logs `:poll_schema_skew` (`status_watcher.rb`); this diff corrected the sibling `modules/bot.md`, so the two pages now contradict — worth a follow-up though not introduced here.
- [ ] [code-review/silent-failure] Behavior change (intentional, tested): `::Telegram::Bot::Exceptions::ResponseError` was dropped from the benign list and now falls to `rescue StandardError`, logging `poll_failure` at the `warn` default instead of the prior quiet branch (`telegram.rb:81-89`; pinned by `test_poll_updates_logs_telegram_response_error_as_warning_default`). Defensible (a Telegram 401/409/429/5xx is signal, not noise) and it now also counts toward poll-health — flagged only so a reviewer confirms transient Telegram 429/5xx during long-poll surfacing as warnings is the intended outcome.

ivankuznetsov:

## High

## Medium
- [ ] `lib/hive/bot/notification_dispatcher.rb:223`: Backoff transition logging is latched for the whole active fingerprint, so a retry that leaves backoff and then fails again re-enters backoff without a new `notification_skipped_backoff` transition log.
- [ ] `.github/workflows/release.yml:280`: The branch reintroduces the macOS Colima smoke path (`macos-15` + `colima start`) that the in-tree wiki documents as currently failing before Docker starts, so tag release workflows can go red after image publish.

## Nit
- [ ] `CHANGELOG.md:21`: The diff rolls back unrelated 0.3.1 release-note entries for security, pipeline, reviewer, integration, and dependency fixes, leaving the published changelog incomplete.

ivankuznetsov:

# ce-code-review — telegram-bot-bot-log-is-260624-7d9b (pass 2)

Reviewed `git diff origin/main..HEAD` (6 task commits: U1–U4 plus two follow-up
fix/doc commits). Independently verified locally:

- Focused bot suites green: `logger_test` (17), `telegram_test` (32),
  `notification_dispatcher_test` (51), `poll_health_test` (5),
  `status_watcher_test` (16), `supervisor_test` (141), and all
  `test/integration/bot/*` (the one `telegram_smoke_test` skip is a pre-existing
  real-API gate on env tokens, not introduced here) — 0 failures / 0 errors.
- `rubocop` clean on all 8 changed `.rb` files.
- `schemas/hive-bot-log.v3.json` parses as valid JSON; `EVENTS` ↔ v3 `event`
  enum parity is exact (33/33); `LEVELS` keys/values all valid; `v1`/`v2` schema
  files byte-unchanged vs `origin/main`.
- Production reachability confirmed: `@telegram` (and thus the `PollHealth`
  instance) is created once in `Supervisor#initialize` and reused across the
  `poll_loop` (`supervisor.rb:72,231`), so cross-iteration state persists and
  `poll_unhealthy` (acceptance c) actually fires; SIGHUP rebuilds only the
  dispatcher (`supervisor.rb:1604`), not `@telegram`, so poll-health survives a
  reload as the plan intends.
- Wiki doc corrections match code: success-path forward-skew is emitted under the
  distinct `:poll_schema_skew` event (`status_watcher.rb:176,212`), not
  `:poll_failure`, matching the corrected `wiki/commands/bot.md` wording.

The implementation fully satisfies the plan's Goals and every row of the
Requirements Trace (U1 severity model + v3 schema; U2 benign classification +
`poll_unhealthy` consecutive/silence escalation with once-per-episode latch and
re-arm; U3 transition-only dedupe/backoff via reconciled in-memory sets; U4
kept-skip `info` levels + acceptance/outage tests). No plan-required item is
missing and no scope boundary is crossed → **no High findings**.

Pass-1 auto-fixes are present in the current diff: the acceptance test
(`test_info_stream_excludes_noise_events_under_repeated_poll_and_dedupe_chatter`)
now emits the high-frequency events with **no** explicit level so severity comes
from the production `LEVELS` map (non-circular), and the backoff path now has
independent fingerprint-change and leave/return re-log tests.

## High


## Medium


## Nit
- [ ] `Telegram#initialize`'s `now:` keyword (`telegram.rb:65`) is used only to seed the default `poll_health: PollHealth.new(now: now)` — it is never stored or referenced by `Telegram` itself, and every test injects a fully-built `poll_health:` rather than `now:`. The seam is harmless and defensible (lets a future caller override the clock without constructing a `PollHealth`), but it could be dropped to simplify the signature, or carry a one-line comment clarifying it governs only the default collaborator's clock.

ivankuznetsov:

# pr-review-toolkit — telegram-bot-bot-log-is-260624-7d9b (pass 2)

Ran the multi-persona review (`code`, `tests`, `silent-failure`, `type-design`,
`comments`, `simplify`) over the real task changes in `git diff origin/main..HEAD`
(scoping out the stale-base `release.yml`/`CHANGELOG.md` artifact, which the
branch's own commits never touch). Union of findings below.

Consensus across all six personas: **no current code defect and no
plan-required-but-missing item** — the implementation matches its stated intent,
is rubocop-clean, and carries thorough, largely non-circular tests. Findings are
a test-coverage seam, one type-design asymmetry, and documentation/clarity nits.

## High


## Medium
- [ ] [tests] No telegram-layer test for the success→re-arm seam: deleting `@poll_health.record_success` (`telegram.rb:79`) leaves all 32 `telegram_test` cases green, so acceptance (c) is proven only for the *first* outage — a regression dropping that call would latch `@unhealthy_latched` true and silently suppress every subsequent `poll_unhealthy`. Add a test: escalate → successful `get_updates` → escalate again, asserting two `poll_unhealthy` events. (persona rated High; downgraded to Medium as the current code is correct and the re-arm is covered at the `PollHealth` unit level — only the cross-layer wiring is unpinned.)
- [ ] [type-design] `category` is stamped via `category.to_s` with no validation or closed set (`logger.rb:85,105`), unlike the guarded `level` enum, yet the noise-suppression filter keys on the exact string `"noise"` — a typo (`category: :noies`) would silently emit a line that escapes the filter. Consider a `CATEGORY_NAMES = %i[noise].freeze` membership guard mirroring `level`, or document the open-set decision near `LEVELS`. (Caveat: the v3 schema deliberately specifies `category` as free-form `type: string` per the plan, so this is a hardening/consistency suggestion, not a schema violation.)

## Nit
- [ ] [tests] `poll_unhealthy: :warn` in the `LEVELS` map (`logger.rb:52`) is dead-tested — removing it keeps both suites green because the only call site (`telegram.rb:208`) passes `level: :warn` explicitly. Add `poll_unhealthy` to `test_default_levels_come_from_event_map` to pin the map default. (persona rated Medium.)
- [ ] [tests] `test_dedupe_skip_logs_again_when_fingerprint_changes` (`notification_dispatcher_test.rb`) overstates what it isolates: `first`/`changed` are never simultaneously present, so it exercises a brand-new fingerprint's own once-cycle, not the `reconcile_skip_log_state` keep_if path (which is already covered by the leave/return test). Rename to reflect "a new fingerprint gets its own once-logged skip"; same applies to the backoff twin.
- [ ] [type-design] `category` lacks the symmetric string-keyed-attr protection `level` has: a caller passing `**{"category" => "x"}` with no explicit `category:` survives the merge (`logger.rb:105` only assigns `unless category.nil?`). Harmless today (category is open-string, no enum to violate) but inconsistent with the deliberately-hardened `level` path.
- [ ] [type-design] `PollHealth::Result.new` is public and unvalidated (`poll_health.rb:11`), so `reason` is only conventionally constrained to `{nil, :consecutive, :silence}`; fine for a single-producer internal value object, noted only so it isn't later exposed without a guard.
- [ ] [comments] `build_update`'s rescue (`telegram.rb:195-201`) deliberately does NOT call `emit_poll_unhealthy_if_needed` (per-update parse failures must not trip the outage escalator) — this load-bearing asymmetry vs. the two `poll_updates` rescues is uncommented; a one-line note would prevent a future "fix the inconsistency" regression. (persona rated Medium.)
- [ ] [comments] `poll_health.rb:22` (`@last_success_at = @now.call`) seeds the silence window to construction time so the first `record_failure` can't instantly read as a `:silence` outage — a subtle correctness decision worth a short comment. (persona rated Medium.)
- [ ] [comments] `wiki/commands/bot.md` "Events include `bot_started`, `poll_failure`, …" list omits `poll_unhealthy` (the event this change centers on); the list is explicitly non-exhaustive so not wrong, but adding it would make the v3 doc self-consistent.
- [ ] [comments] Wiki `updated:` frontmatter not bumped despite substantive v3 edits — `wiki/modules/bot.md:6` still `2026-06-15`, `wiki/commands/bot.md:6` still `2026-06-18`; bump to the change date (2026-06-24).
- [ ] [simplify] `PollHealth#record_failure` (`poll_health.rb:38-43`) uses `should_escalate` both to latch and to re-derive the `reason:` field via a second ternary; nulling `reason` directly (`reason = nil unless should_escalate`) before building `Result` removes the inline conditional. Behavior-identical readability tweak.
- [ ] [simplify] (do-NOT-apply) Explicit `level:` at the three noise call sites (`notification_dispatcher.rb:232,242`, `telegram.rb:208`) duplicates the `LEVELS` map default, but is spec-locked: the stub loggers capture raw kwargs and never apply the map, and tests assert `attrs[:level] == :debug`/`:warn`. Removing it (as the pass-1 suppressed "redundant level" nit suggests) would fail those assertions — flagging the conflict so triage does not auto-apply the removal. (`category: :noise` is genuinely non-redundant and must stay.)
- [ ] [simplify] `emit_poll_unhealthy_if_needed` (`telegram.rb:204`) reads as a query (`_if_needed`) but unconditionally mutates poll-health via `record_failure` before conditionally emitting; minor naming smell, no change recommended (renaming churns the diff for negligible gain).
- [ ] [code] By-design observability edge (consistent with the plan's transition rule): `@dedupe_logged`/`@backoff_logged` membership persists while a fingerprint stays in `current`, so a continuously-present row that cycles skip-states (e.g. backoff → delivered → backoff) won't re-log the second debug/noise skip line. Matches "re-log only on fingerprint change or leave/return"; genuine `send_failure` lines remain loud. Noted, not a defect.

ivankuznetsov:

# Code Review — telegram-bot-bot-log-is-260624-7d9b (pass 3)

**Scope:** branch changes only, diffed against the merge-base
(`9efbca2a..HEAD` over `lib/ schemas/ test/`), not `origin/main`.
`git diff origin/main..HEAD` additionally shows `.github/workflows/release.yml`
and `CHANGELOG.md`, but those are **not** this branch's work — the branch is
behind `origin/main` and those files moved on main after it was cut. They are
excluded from review (a normal merge will keep main's versions).

**Intent:** structured severity model (`level`/`category`) for the bot JSONL
log, schema v2→v3, benign long-poll timeouts → `debug`/`noise` with a loud
`poll_unhealthy` escalation (`PollHealth`), and transition-only dedupe/backoff
skip logging.

**Reviewers:** correctness, adversarial, testing, maintainability, reliability,
api-contract. **Tests:** the four touched unit suites pass locally
(logger 17 / telegram 33 / poll_health 5 / dispatcher 51 runs, 0 failures).
The diff faithfully implements U1–U4; no requirement in the plan's Requirements
Trace is missing.

## High

_None. No P0/P1 defects found; the core severity model, escalation latch/re-arm,
and transition-only skip logging are implemented correctly and match the plan._

## Medium
- [ ] `LEVELS` map has no drift guard against `EVENTS` (`lib/hive/bot/logger.rb:52-72`): `event` validates the name against `EVENTS`, but `LEVELS.fetch(name, :info)` silently ignores a stale or misspelled `LEVELS` key, so a wrong/orphaned entry never surfaces — add `assert_empty(Hive::Bot::Logger::LEVELS.keys - Hive::Bot::Logger::EVENTS)` to pin the subset relationship. (maintainability)
- [ ] The "build_update parse failure must not escalate poll-health" invariant is untested (`test/unit/bot/telegram_test.rb`, `test_poll_updates_logs_build_update_backtrace_once_per_error_class`): the deliberate comment at `lib/hive/bot/telegram.rb:198-202` is unguarded, so a copy-paste adding `emit_poll_unhealthy_if_needed` to the `build_update` rescue would pass — assert `poll_unhealthy` is absent (ideally with an injected `PollHealth(max_consecutive: 1)`). (testing)
- [ ] dedupe↔backoff state transition is silently dropped despite the "log on state transition" contract (`lib/hive/bot/notification_dispatcher.rb:223-246`): `reconcile_skip_log_state` prunes the two `Set`s only when a fingerprint leaves `current`, never cross-clearing, so a still-present fingerprint that goes dedupe→backoff (or backoff→dedupe) emits no skip line; impact is limited to a missing `debug`/`noise` audit line — fix by deleting the fingerprint from the other state's `Set` when logging a skip. (correctness + adversarial — cross-corroborated)
- [ ] Flapping/degraded outage may never trip `poll_unhealthy` (`lib/hive/bot/poll_health.rb:29` `record_success`): if one `get_updates` succeeds inside every `max_silence_sec` window, each success resets both the consecutive counter and the silence clock, so neither threshold is reached and acceptance criterion (c) "outage produces a loud line" is evadable by a partial outage; may be an accepted trade-off of the plan's consecutive-OR-silence design, but worth confirming (a sliding-window failure ratio would close it). (adversarial)

## Nit
- [ ] Boot-time false `poll_unhealthy` if startup exceeds `max_silence_sec` (`lib/hive/bot/poll_health.rb:24`): `@last_success_at` is seeded at construction (during `Supervisor#initialize`), so a slow startup followed by a first failure reads as a 60s "silence" outage with no real outage — rare, but defeats the signal; lazy-seed on the first `record_*` call instead. (adversarial + reliability — cross-corroborated)
- [ ] `Net::OpenTimeout` is absent from `BENIGN_POLL_ERRORS` (`lib/hive/bot/telegram.rb:11-15`): it is a sibling of `Net::ReadTimeout` (not a subclass), so a TCP connect timeout lands in the `warn` arm and counts toward escalation instead of `debug`/`noise` — louder, not swallowed; add it if connect timeouts should be treated as benign. (reliability)
- [ ] `now:` keyword on `Telegram#initialize` is inert when `poll_health:` is injected (`lib/hive/bot/telegram.rb:64-68`): documented, but a misleading affordance with zero test coverage in its intended use — consider dropping it and having callers wire the clock into `PollHealth` directly. (maintainability)
- [ ] `emit_poll_unhealthy_if_needed` runs inside both `rescue` arms (`lib/hive/bot/telegram.rb:84-92`): a raise there (e.g. injected clock or unexpected `Logger#event` error) escapes `poll_updates` to the poll-loop's fatal handler rather than being contained; theoretical under current valid call values, but a defensive `begin/rescue` would harden the hot loop. (reliability)
- [ ] v3 schema `$id` points to the mutable `/blob/main/` ref (`schemas/hive-bot-log.v3.json:3`): any consumer dereferencing it resolves the latest schema, not the v3 snapshot — pre-existing pattern (v1/v2 identical), not introduced here; pin to a tag/SHA if external consumers ever resolve `$id`. (api-contract — pre-existing)
- [ ] No reverse bijection test for the schema `event` enum (`schemas/hive-bot-log.v3.json` vs `EVENTS`): the dangerous direction (an `EVENTS` entry missing from the enum) is already caught by `test_every_documented_event_is_accepted_and_schema_valid`; only an extra enum entry not in `EVENTS` is unguarded — a strict `enum == EVENTS.map(&:to_s).sort` assertion would close the remaining gap. (maintainability)
- [ ] Pre-existing: `Logger#event` has no mutex around `@file.puts`/`@file.flush` while `poll_loop`, `status_loop`, and `reaper_loop` all log concurrently (`lib/hive/bot/logger.rb:111-119`); this branch adds new high-frequency call sites (`poll_unhealthy`, skip lines) but does not introduce the race — noted for awareness, out of scope. (adversarial + reliability)

ivankuznetsov:

# PR Review (pr-review-toolkit) — telegram-bot-bot-log-is-260624-7d9b (pass 3)

**Scope:** branch changes only, diffed against the merge-base
(`9efbca2a..HEAD` over `lib/ schemas/ test/`). `git diff origin/main..HEAD`
also shows `.github/workflows/release.yml` and `CHANGELOG.md`, but those are
**not** this branch's work (the branch is behind `origin/main`); excluded.

**Personas run:** code-reviewer, pr-test-analyzer (`tests`), comment-analyzer
(`comments`), silent-failure-hunter (`errors`), type-design-analyzer
(`type-design`), code-simplifier (`simplify`). Findings below are the union.

**Overall:** the runtime implementation is sound — every persona independently
confirmed the severity model, escalation latch/re-arm, two-arm rescue, and
transition-only skip logging behave as intended; affected unit tests pass. No
runtime-correctness blocker was found. The one High item is a test-hardening
gap on an explicitly-documented invariant.

## High
- [ ] [tests] `build_update` parse-failure must-not-escalate invariant is unguarded (`lib/hive/bot/telegram.rb:198-202`; exercised by `test/unit/bot/telegram_test.rb:133,517`): the runtime code is correct and commented, but both parse-path tests use default thresholds (5/60s) and only 1–2 failures, so a future refactor unifying the three rescues into one that calls `emit_poll_unhealthy_if_needed` would silently page on every malformed update and no test would fail — add a test injecting `PollHealth(max_consecutive: 2)` + frozen clock, feeding 3+ malformed updates, asserting `poll_unhealthy` is never emitted while `poll_failure` is.

## Medium
- [ ] [tests, code-reviewer] dedupe/backoff skip re-arm is only tested for "row leaves `current`"; the within-`current` state cycle (backoff → recover/deliver → backoff, fingerprint never leaves) and the cross-state dedupe→backoff path are untested (`lib/hive/bot/notification_dispatcher.rb:223-246`): because the two `Set`s are pruned only on disappearance and never cross-cleared, a second backoff episode for a persistent fingerprint is silently suppressed, contradicting the "log on state transition" intent — add tests pinning whether re-entry re-logs (`== 2`) or stays suppressed (`== 1`) with a comment; code-reviewer judges it a deliberate tradeoff (optional fix: clear both Sets on successful delivery).
- [ ] [errors] flapping/partial outage never escalates (`lib/hive/bot/poll_health.rb:29-34` `record_success`, `:53-57` `unhealthy_reason`): every success resets the consecutive count, the silence window, AND the latch, so a link failing ~80% of polls but succeeding within each 60s window emits only `debug`/`noise` and never a `warn` — acceptance criterion (c) "outage produces a loud line" is evadable by a degraded-but-not-dead connection; add a windowed failure-rate signal (or don't fully reset the silence anchor on a single success), or at minimum document the deliberate "any success = alive" gap.
- [ ] [type-design] the schema `event` enum has no bidirectional parity guard with `Logger::EVENTS` (`test/unit/bot/logger_test.rb:43` vs `schemas/hive-bot-log.v3.json`): the existing test only proves `EVENTS ⊆ schema-enum`; schema-side drift (a stale/typo'd/removed enum value the producer can never emit) passes green and misleads non-Ruby consumers, whereas every other closed enum in this repo is pinned via `schema_files_test.rb` `assert_equal` — add `assert_equal Logger::EVENTS.map(&:to_s).sort, schema_doc.dig("properties","event","enum").sort`.
- [ ] [comments, errors] `lib/hive/bot/logger.rb:51` comment states "The noise-suppression filter keys on the literal 'noise'" as present-tense fact, but no such filter exists in the gem — only the three producers set `category: :noise`; suppression is convention-only (corroborated by `errors`): reword the comment to describe `noise` as a downstream convention nothing in-repo currently consumes, and document the consumer-side filter contract in the wiki/schema description.

## Nit
- [ ] [type-design] the load-bearing `"noise"` category is a magic string duplicated across `lib/hive/bot/telegram.rb:85`, `lib/hive/bot/notification_dispatcher.rb:232,242`, and `test/unit/bot/logger_test.rb:107`: extract a `Logger::CATEGORY_NOISE` constant to make the one significant value greppable while keeping the category set open.
- [ ] [type-design] `PollHealth::Result#seconds_since_success` / `#consecutive_failures` are only meaningful when `escalate?` is true, but the type doesn't express it (`lib/hive/bot/poll_health.rb:11-15`): a future caller could read them on the no-escalation path and trust a meaningless number — add a doc comment scoping them to escalations (mirror the existing `escalate?`/`reason` comment); do not introduce ADT variants.
- [ ] [type-design] `PollHealth::Result.new` bypasses every producer invariant (negative seconds, `consecutive_failures: 0`, arbitrary reason all constructible) (`lib/hive/bot/poll_health.rb:11`): low practical risk given the single internal producer — note only, no custom constructor needed.
- [ ] [type-design] `LEVELS` lists only non-`:info` events, so a newly added `:warn`/`:error` event silently defaults to `:info` unless the author remembers to add it (`lib/hive/bot/logger.rb:52-72`): nothing forces the map entry; acceptable terseness, partially pinned by `logger_test.rb:74` — awareness only.
- [ ] [type-design, simplify] `record_failure` reuses the local `reason` as both computed-reason and report-flag via `reason = nil unless should_escalate` (`lib/hive/bot/poll_health.rb:39-42`): correct and documented, but the variable changes meaning mid-method — optionally compute escalation into a separate var and set `reason:` via a ternary at `Result.new` so the detected reason stays immutable (taste-level, fine to leave).
- [ ] [errors] the once-per-episode latch emits exactly one `warn` for an arbitrarily long outage and no recovery line (`lib/hive/bot/poll_health.rb:36-49`, `lib/hive/bot/telegram.rb:211-219`): an operator who misses the onset line (rotation/restart/dedup) has no ongoing `warn` signal — consider periodic re-emission while unhealthy and/or a `warn` recovery line when `record_success` clears the latch.
- [ ] [comments] the once-per-skip-state throttling (`reconcile_skip_log_state` + `log_*_skip_once` + the two `Set`s) has no inline comment despite being central to this branch's noise-reduction goal (`lib/hive/bot/notification_dispatcher.rb:223-246`): only the wiki documents it — add a one-line comment noting the in-memory Sets throttle the debug/noise skip lines to state transitions (entry / fingerprint change / leave-and-return), distinct from `AlertStore` delivery dedup.
- [ ] [tests] `test/unit/bot/telegram_test.rb:243` asserts `seconds_since_success == 0`, which only holds because the clock is frozen across the three calls (pins clock-freezing, not health behavior): add a clarifying comment, or rely on the meaningful `== 11` assertion already present in the silence test.
- [ ] [tests] `test/unit/bot/poll_health_test.rb:5-22` bundles escalate-once + sustained-no-repeat + re-arm-after-success in one method: consider splitting the re-arm tail for sharper failure localization (DAMP-compliant as-is; optional).
- [ ] [simplify] the explicit `level:` at `lib/hive/bot/telegram.rb:215` and `lib/hive/bot/notification_dispatcher.rb:232,242` duplicates the `LEVELS` map default and looks removable, but is NOT safely removable — the test `StubLogger` doesn't apply the map and asserts on the raw `level:` attr; flagged as a minor coupling smell, **no change recommended**.

PR #588, opened 2026-06-24
fix(bot): suppress stuck alerts for live-agent retries

ivankuznetsov:
## Summary

The Telegram bot announced "⚠ <stage> stuck" for tasks whose agent was actively running. When a live task lock made `hive status` report `action=agent_running` while the state file still carried a stale recovery marker (`error`, `review_error`, `review_stale`, `review_ci_stale`, `execute_stale`) from a previous run, `NotificationBuilders.build` never consulted its own `SKIP_ACTIONS` set and let `recovery?` match the stale marker — paging the operator about a task that was already being retried.

`build` now treats the live `action` as the authoritative gate: `agent_running` and `archived` rows are dropped before recovery matching, logging `notification_skipped_live_agent` (project/slug/stage/marker/action) so the stale-marker rate stays observable instead of going silently invisible.

Suppressing the push at `build` exposed a second, subtler bug. Once the stuck row vanishes from the current snapshot, the dispatcher's recovery path would, after the grace window, fire a premature "✅ Recovered" for a task whose agent was still running. `NotificationDispatcher` now reads the live `agent_running` identities from the raw rows and **holds** any stored recovery alert whose project/slug/stage is currently live, re-arming the grace clock while the hold is active so a transient single-tick row drop (e.g. a per-project status degrade that keeps the envelope `ok:true`) can't release it early. The result is silent throughout the retry: if it fails the original stuck alert dedupes (no duplicate page), and if it succeeds the normal "✅ Recovered" confirmation fires exactly once after the lock disappears.

```mermaid
stateDiagram-v2
    [*] --> Alerted: action=error → "⚠ stuck" sent
    Alerted --> Held: action=agent_running (live retry)
    Held --> Held: still running (no "Recovered", entry retained)
    Held --> Deduped: error reappears → dedupe, no new page
    Deduped --> Held: agent_running again
    Held --> Recovered: row gone, no live identity → "✅ Recovered" once
    Recovered --> [*]
```

**Design notes**

- `recovery?` is deliberately left untouched — it stays a pure marker/action predicate. The gate lives in `build`, before fingerprinting, so dedup logic is unaffected.
- The skip-log is gated on `recovery?(row)` so only the live-vs-stale-marker contradiction is recorded — healthy `agent_working` and lingering `archived` rows stay quiet.
- The new event is additive to `Hive::Bot::Logger::EVENTS` and the `hive-bot-log.v2` schema enum; `logger_test.rb`'s "emitted symbols ⊆ EVENTS" scan enforces the registration.
- Out of scope (tracked separately): upstream marker-clear on run start, and the `/status` button, which routes a live `agent_running+error` row through the same `recovery?` branch.

## Test plan

All green (verified during review):

- `ruby -Itest test/unit/bot/notification_builders_test.rb` — 46 runs
- `ruby -Itest test/unit/bot/notification_dispatcher_test.rb` — 52 runs
- `ruby -Itest test/unit/bot/logger_test.rb` — 11 runs
- `ruby -Itest test/integration/bot/scenarios/alert_lifecycle_test.rb`
- `ruby -Itest test/integration/bot/scenarios/notification_exactly_once_test.rb`
- `ruby -Itest test/integration/bot/scenarios/s2_review_error_test.rb`
- `bundle exec rubocop` on the three changed `lib/` files — clean
- `rake coverage` — 100% line coverage gate; new branches in `build` and `process_recoveries` exercised

Coverage includes the #9281 `agent_running` + `error` regression fixture, every stale recovery-marker variant under `agent_running`, the archived suppression + skip-log assertion, the unchanged `error` / `recover_execute` / `recover_review` notifications, the clean-retry (`agent_running` + `agent_working`) marker-independent hold, project/slug/stage identity scoping, the mid-hold per-project-degrade regression, and the dispatcher retry-hold / retry-failure-dedupe / retry-success-recovery lifecycle.

## Review summary

Three review passes (claude-ce, codex-ce, and the multi-persona pr-review-toolkit each ran 3 times). Pass 3 surfaced one genuine **High**: a premature/false "✅ Recovered" during a held retry, because the grace clock was consumed *during* the hold, so a single-tick `agent_running` drop (a per-project status degrade keeps the envelope `ok:true` while emptying that project's rows) released the hold and fired "Recovered" though nothing had recovered. Fixed by re-arming the grace/settling window while the hold is active, correcting the overclaiming hold-invariant comment, and adding a mid-hold-degrade regression test (commit `cb090949`).

All remaining findings were test-strength additions, comment/doc-accuracy fixes, an observability narrowing (skip-log gated on `recovery?`), and behavior-preserving simplifications — all auto-fixed or resolved from the plan, brainstorm, and prior-pass decisions. No finding required a user decision; there were no open escalations.

## Linked task

- Hive task: `telegram-bot-contradictory-stage-stuck-260624-c552`
- Regression fixture: real-bot.log line #9281 (`project=hive`, `slug=bug-agentlimit-false-positives-on-260623-a4df`, `action=agent_running`, `marker=error`)

---


ivankuznetsov:

# Code Review — telegram-bot-contradictory-stage-stuck-260624-c552 (Pass 1)

Scope reviewed: `git diff origin/main..HEAD`. Real task changes are the 9 bot
files (`lib/hive/bot/{notification_builders,notification_dispatcher,logger}.rb`,
`schemas/hive-bot-log.v2.json`, two test files, three wiki/log files). The
`.github/workflows/release.yml` and `CHANGELOG.md` entries in the diff are NOT
from this branch — see the first Nit.

Verification performed:
- `notification_builders_test.rb` (46 runs), `notification_dispatcher_test.rb` (47 runs), `logger_test.rb` (11 runs) — all green.
- Integration: `alert_lifecycle_test.rb`, `s2_review_error_test.rb`, `notification_exactly_once_test.rb` — all green.
- `rubocop` on all 5 changed Ruby files — clean.
- Confirmed `recovery?` is byte-for-byte unchanged; guard placement is after the legacy short-circuit and before READY/needs_input/recovery dispatch (plan Unit 1).
- Confirmed `agent_running`/`archived` are real action keys (`lib/hive.rb:163-164`, `TaskAction#universal_action` returns `:agent_running` under a live lock) — neither branch is dead code.
- Confirmed Unit 3 traces correctly: live hold suppresses premature "✅ Recovered", retry-failure dedupes, retry-success fires Recovered once; `absent_since` is intentionally not reset so Recovered fires promptly post-completion.

The implementation faithfully realizes the plan (Units 1–4; Q1=include Unit 3; Q2=skip-event observability; Q3=`/status` button left out of scope). No High or Medium defects found.

## High
- [ ] None.

## Medium
- [ ] None.

## Nit
- [ ] Branch is stale relative to `origin/main` (merge-base is #582; `origin/main` already has #583/#584): the `.github/workflows/release.yml` and `CHANGELOG.md` deletions shown in `git diff origin/main..HEAD` are drift artifacts, not changes from this task (`git log merge-base..HEAD -- release.yml CHANGELOG.md` is empty). Rebase/merge `origin/main` before merging so the PR doesn't appear to revert the arm64 smoke fix (#584); no code change in this worktree is needed.
- [ ] `test/unit/bot/notification_dispatcher_test.rb:784` relaxed `assert_empty logger.events` to `assert_equal :notification_skipped_live_agent, logger.events.first.first`, which only checks the first event and no longer guarantees the skip event is the *only* event emitted; `assert_equal [:notification_skipped_live_agent], logger.events.map(&:first)` would preserve the original "single event" strength.
- [ ] Cosmetic action-type asymmetry: `notification_builders.rb:46` uses `SKIP_ACTIONS.include?(row.action)` (no `.to_s`) while `notification_dispatcher.rb:280` uses `row.action.to_s == "agent_running"`. Both are correct because `row.action` is always a String parsed from `hive status --json` (`status_watcher.rb:292`), and each matches its own file's convention (builder mirrors `recovery?`/`READY_ACTIONS`; dispatcher mirrors `recovery_identity`'s defensive `.to_s`). No behavioral impact.

ivankuznetsov:

# PR Review Toolkit — telegram-bot-contradictory-stage-stuck-260624-c552 (Pass 1)

Multi-persona review of `git diff origin/main..HEAD`, scoped to this task's bot
changes (the `release.yml`/`CHANGELOG.md` diff entries are origin/main drift past
the merge-base and were excluded). Personas run: **code-reviewer**,
**pr-test-analyzer**, **comment-analyzer**, **silent-failure-hunter**,
**type-design-analyzer**.

**Overall verdict:** no High/Critical/Important issues from any persona. The
`SKIP_ACTIONS` gate is correctly placed before fingerprinting and `recovery?`,
returns `nil` for live/archived rows with an auditable skip event, and the
dispatcher's live-agent hold is correctly positioned and uses a consistent
`[project,slug,stage]` identity. silent-failure-hunter confirmed the suppression
is auditable in production (logger never nil at the dispatcher call site), bounded
by the daemon's `StaleAgentHealer` (no forever-hold), and not reachable into a
false "✅ Recovered". The findings below are test-strength, comment/doc-accuracy,
and maintainability improvements.

## High
- [ ] None.

## Medium
- [ ] [silent-failure-hunter] `notification_dispatcher.rb:150,155` — `absence_passed_grace?` runs before the live-identity hold, so the grace clock starts/elapses during a long `agent_running` hold; the first non-held tick then fires "✅ Recovered" with no fresh post-recovery settling. Intentional per the plan ("already-elapsed grace lets Recovered fire promptly") and NOT a false-recovered risk (the intermediate "neither agent_running nor error" state is unreachable from `TaskAction#universal_action`), but the grace-bypass-on-long-hold deserves an inline comment so a maintainer doesn't read it as a bug.
- [ ] [pr-test-analyzer] `notification_dispatcher_test.rb:780-785` — `test_rows_without_notification_are_ignored` weakened `assert_empty logger.events` to `assert_equal :notification_skipped_live_agent, logger.events.first.first`; it no longer pins event count, so a future spurious/duplicate event after the skip passes silently. Tighten to `assert_equal [:notification_skipped_live_agent], logger.events.map(&:first)`.
- [ ] [pr-test-analyzer] No dispatcher test asserts `:notification_skipped_live_agent` is logged during the *held* retry lifecycle (only the single-tick `test_rows_without_notification_are_ignored` covers the log); a regression that stopped logging the skip mid-hold would leave the "why no Recovered yet" audit trail gone with no failing test.
- [ ] [pr-test-analyzer] The `absent_since` non-reset-during-hold invariant is exercised only implicitly — the success test (`:499-520`) relies on the grace timestamp set at step 2 never being reset, but nothing names it; add an explicit `absent_since` assertion on the retained entry during the hold so the grace-continuity intent is refactor-proof.
- [ ] [comment-analyzer] `notification_builders.rb:43-45` — the guard comment explains only the `agent_running` race, but the guard keys off `SKIP_ACTIONS = %w[agent_running archived]`; `archived` (new in this change) gets no rationale and "those rows must stay silent" reads as agent_running-only. Add a clause covering why `archived` is suppressed (terminal — nothing left to announce).
- [ ] [comment-analyzer] `wiki/log.d/20260624T184142Z-telegram-live-agent-suppression.md:21` — "archived stale markers" overstates coverage: the only archived test (`notification_builders_test.rb`) uses a single `error` marker, and `error` is not a *stale* (`*_stale`) marker. Reword to "an archived row with a stale recovery marker" / "an archived error row".

## Nit
- [ ] [type-design-analyzer, silent-failure-hunter] `notification_dispatcher.rb:280` hardcodes the literal `"agent_running"` while the builder uses the `SKIP_ACTIONS` constant; reference the canonical `Hive::Schemas::TaskActionKind::AGENT_RUNNING` (`lib/hive.rb:163`) so a future action-string rename can't silently desync the dispatcher hold from the builder suppression. (Pre-existing convention — `SKIP_ACTIONS` at `notification_builders.rb:35-38` also hardcodes the literals — so this is consistency polish, not a regression.)
- [ ] [pr-test-analyzer] No test feeds one `agent_running` row through both layers to lock the cross-module contract (builder returns `nil` AND dispatcher holds the matching fingerprint); the builder-vs-dispatcher action-matching divergence is currently untested.
- [ ] [pr-test-analyzer] No negative identity-scoping test: a live `agent_running` row for a *different* slug must NOT hold an unrelated stored recovery — a bug that compared only `project` or dropped `stage` from `recovery_identity` would pass every current test.
- [ ] [pr-test-analyzer, code-reviewer] `archived` has thinner coverage than `agent_running`: `test_archived_error_marker_is_suppressed` asserts only the `nil` return (no logger passed → no skip-event assertion), and no lifecycle test confirms the intended asymmetry (archived suppresses the build but, unlike agent_running, does NOT hold a stored recovery → can fire "Recovered"). Both personas judged the asymmetry intentional/correct; a one-line test would pin it.
- [ ] [comment-analyzer] `wiki/log.d/20260624T184142Z-telegram-live-agent-suppression.md:20-21` — "all stale recovery-marker variants under agent_running" is slightly loose: the parameterized test covers 4 markers; the 5th (`error`) is covered separately and isn't a "stale" variant.

ivankuznetsov:

# Code Review — telegram-bot-contradictory-stage-stuck (pass 2)

Diff reviewed: `git diff origin/main..HEAD` (lib/hive/bot/{notification_builders,notification_dispatcher,logger}.rb, schemas/hive-bot-log.v2.json, tests, wiki).

Verification run: `notification_builders_test` (46), `notification_dispatcher_test` (52), `logger_test` (11), `alert_lifecycle_test`, `notification_exactly_once_test`, `s2_review_error_test` all green; `rubocop` clean on the three changed lib files. All plan units (1 build-gate, 2 logger/schema registration, 3 dispatcher hold, 4 acceptance tests incl. the #9281 fixture) are implemented and exercised — no plan-required-but-missing gap.

## High
- [ ] (none) — the core fix is correct: `build` short-circuits `SKIP_ACTIONS` before fingerprinting, `recovery?` is untouched, and the dispatcher holds a stored recovery while the same project/slug/stage is live `agent_running`. The grace-before-hold ordering, identity scoping, and archived/agent_running asymmetry all behave as the plan specifies and are well tested.

## Medium
- [ ] Skip-log fires for healthy `agent_working` and lingering `archived` rows every tick, not just the stale-marker case A6 targeted: the log call sits unconditionally inside the `SKIP_ACTIONS` guard (`notification_builders.rb:49-54`), so it emits `:notification_skipped_live_agent` for *every* `agent_running` row (every active agent, marker `agent_working`) and *every* `archived` (9-done) row on each 30s poll tick — both were previously silent (they fell through to `else nil`). `hive status --json` includes all 9-done rows with no age filter (the >3d hide in `commands/status.rb:439` is render-only; the JSON path at `:194-197` does not apply it), and the bot forwards every row unfiltered (`supervisor.rb:222`), so archived skip-logs grow unboundedly with the done-task backlog and dilute the stale-marker rate A6 wanted observable. The plan's risk #3 accepted per-tick cadence only for the `agent_running+error` case; the archived/healthy amplification is unanalyzed. Consider gating the log to recovery/stale markers (or excluding `agent_working`/`archived`) so the diagnostic isolates the actual suppression A6 cares about.

## Nit
- [ ] Dispatcher comment overstates a load-bearing safety claim (`notification_dispatcher.rb:150-158`): it says the intermediate "neither agent_running nor error" state is "unreachable from `TaskAction#universal_action`", but a held retry can resolve to `needs_input` / `ready_to_*` (via `kind_action`) or `manual_steering` (which `universal_action` itself returns), releasing the hold and firing "✅ Recovered" even though no `error` re-occurred. The behavior is defensible (the error did clear), but the blanket "unreachable" wording is inaccurate and could mislead a future maintainer reasoning about the grace-before-hold ordering; the `agent_running → needs_input` transition is also not covered by the dispatcher tests.
- [ ] Event name is a slight misnomer for the `archived` branch (`notification_builders.rb:50`): `notification_skipped_live_agent` reads as "a live agent suppressed this", but an `archived` task is terminal, not live. The logged `action: archived` makes each line self-describing, so this is cosmetic.
- [ ] Minor defensive-style asymmetry between the two new action checks: the builder uses `SKIP_ACTIONS.include?(row.action)` (no `.to_s`, `notification_builders.rb:49`) while the dispatcher uses `row.action.to_s == ...AGENT_RUNNING` (`notification_dispatcher.rb:289`). Both are correct because `row.action` is always a String here, but the adjacent inconsistency is avoidable.

ivankuznetsov:

# PR Review Toolkit — telegram-bot-contradictory-stage-stuck (pass 2)

Multi-persona review of `git diff origin/main..HEAD`. Personas run: code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, code-simplifier. Union of findings below (deduped; severity is the aggregator's call, with dissenting persona ratings noted inline).

Consensus: the core fix is correct — no swallowed-alert / data-loss path was found. The operator always gets the original "stuck" alert before any retry; the new logic only governs the follow-up "✅ Recovered". Findings are observability quality, one inaccurate load-bearing comment + the edge it mis-describes, and test-coverage gaps.

## High
- [ ] (none) — no blocking correctness defect and no plan-required-but-missing item. Two personas rated facets of the comment/transition cluster "High" (see first two Medium items); the aggregator judges realistic operator impact as Medium and surfaces the dissent there.

## Medium
- [ ] [comments] Dispatcher comment makes a false load-bearing claim (`notification_dispatcher.rb:156-158`): "the intermediate 'neither agent_running nor error' state is unreachable from `TaskAction#universal_action`" is contradicted by `task_action.rb:282`, which returns `:manual_steering` right there. Consequence: a held `agent_running` retry that resolves to `manual_steering` (or, via `kind_action`, `needs_input`/`ready_*`) is absent from both `current` and `live_identities`, so `process_recoveries` fires a spurious/contradictory "✅ Recovered" — the exact failure class this task targets, on a different transition. Fix the wording to state the real invariant. (comment-analyzer rated **High**.)
- [ ] [tests] The held-retry-resolves-to-a-live-non-error-state transition class is entirely untested (`notification_dispatcher_test.rb:470-644`): hold tests only resolve via same-fingerprint re-error (dedupe) or the row vanishing (Recovered). No test pins `agent_running → needs_input/ready_*/manual_steering`, the realistic production resolution and the one that yields a premature/contradictory "Recovered". (pr-test-analyzer rated **High**.)
- [ ] [code][errors] Skip-log volume regression (`notification_builders.rb:49-54`): `:notification_skipped_live_agent` fires for every healthy `agent_running` (marker `agent_working`/`review_working`) and every lingering `archived` row on every 30s poll tick — rows that never would have alerted. Previously silent; now dilutes the stale-marker audit signal A6 wanted and churns the rotating JSONL. Consider gating the log on `recovery?(row)` so only the live-vs-stale-marker contradiction is recorded. (code-reviewer + silent-failure-hunter, both Medium.)
- [ ] [tests] Identity-scoping negative test only varies slug, not stage (`notification_dispatcher_test.rb:598-619`): its comment claims it guards a "dropped stage" regression, but a `recovery_identity` that dropped `stage` still differs by slug and passes. Same-project/same-slug/different-stage hold-masking (a stage-B live row wrongly holding a stage-A stuck alert) is unguarded. (pr-test-analyzer)
- [ ] [errors] The load-bearing hold decision itself is unlogged (`notification_dispatcher.rb:162-164`): `next if live_identities.include?(recovery_identity(row))` withholds a real recovery alert silently — observability depends on correlating against the separately-logged builder skip event by (project,slug,stage). The most important event (a stuck alert is being held) is silent while the routine one is loud. Consider a dedicated `:notification_recovery_held` event (which could also let the noisy builder log be narrowed). (silent-failure-hunter)

## Nit
- [ ] [comments][simplify] Event name `:notification_skipped_live_agent` under-describes the `archived` (terminal, not live) row it also logs; the logged `action:` makes each line self-describing, but the name is narrower than the behavior. Renaming is an observable log-contract change (also in `schemas/hive-bot-log.v2.json`) — note for awareness, not necessarily this PR. (comment-analyzer + code-simplifier)
- [ ] [code] `.to_s` asymmetry: builder uses `SKIP_ACTIONS.include?(row.action)` (`notification_builders.rb:49`) while dispatcher uses `row.action.to_s == ...AGENT_RUNNING` (`notification_dispatcher.rb:289`); both correct for today's string-typed actions, but they would diverge on a symbol and nothing asserts they agree. (code-reviewer + code-simplifier + pr-test-analyzer)
- [ ] [simplify] `live_identities = live_recovery_identities(rows)` (`notification_dispatcher.rb:40`) is computed at the top of `process_rows` but unused on the `fresh_install?` early-return path; moving it down to just above its single use co-locates it and skips the pass on fresh installs. Cosmetic. (code-simplifier)
- [ ] [tests] `test_absent_since_is_not_reset_while_live_retry_holds_the_row` (`:549-569`) is implementation-coupled — it reads the private `AlertStore#absent_since` rather than an observable outcome; the same grace-not-reset property is already behaviorally guarded by the retry-success test. Would break on a grace-representation refactor with no behavior change. (pr-test-analyzer)
- [ ] [tests] Hard-coded `"✅ Recovered: \"#42 Readable Recovery\" — Review"` literal duplicated across three/four test sites (`notification_dispatcher_test.rb:313, 495, 516`); a shared expectation constant would reduce brittleness to a harmless reword. (pr-test-analyzer)
- [ ] [tests] Held-retry-fails-into-a-DIFFERENT-marker (supersede during hold → silent, no Recovered) is uncovered; the supersede path is tested only in isolation, not combined with the new hold. (pr-test-analyzer)
- [ ] [code] `StubLogger` (events-collecting stub) is duplicated byte-for-byte between `notification_builders_test.rb:564` and `notification_dispatcher_test.rb`; self-contained test files are a valid choice, so low value. (code-simplifier)
- [ ] [errors] Flicker-after-grace: a single partial-but-`ok` status snapshot that momentarily drops a still-live held row would satisfy the already-elapsed grace and fire a premature "Recovered"; mitigated by the grace window + `result.ok`-only dispatch, and largely pre-existing to the recovery mechanism. (silent-failure-hunter)

---
Scope note (dropped per plan): silent-failure-hunter flagged that archiving a stuck task fires "✅ Recovered" (a task that was abandoned, not healed). The plan explicitly accepts this — "archived is terminal; clearing a stale alert for an archived task … is acceptable and not part of A5's 'silent during retry' guarantee" — and pins it with `test_archived_row_suppresses_build_but_does_not_hold_recovery`. Not raised as a finding.

ivankuznetsov:

## High
- [ ] `notification_dispatcher.rb` ~164–169 (`process_recoveries`): the grace-before-hold ordering fires a **premature/false "✅ Recovered"** (and deletes the alert entry) when an in-flight `agent_running` row transiently drops from a *single* status snapshot while the retry is still genuinely running. Because grace (`absent_since`) is stamped and fully elapses *during* the legitimate hold, the new `next if live_identities.include?(recovery_identity(row))` becomes a binary gate: any one tick where the identity is absent from `live_identities` immediately releases it. A per-project degraded status (`project_payload_or_degraded` / `missing_project_path` / `not_initialised`, `status.rb:155-181`) keeps top-level `ok:true` (`status.rb:141`) while dropping that project's rows (`extract_rows` `next if project_doc["error"]`, `status_watcher.rb:273`); `status_tick` still calls `process_rows` because `result.ok` is true (`supervisor.rb:167-170`). Failure: P/S/stage is stuck, a retry holds the lock (`agent_running`), grace elapses while held; one tick later P momentarily fails its dir check (mount/worktree blip) → P's row absent, `ok:true` → hold releases → "✅ Recovered" sent + entry removed, though nothing recovered. The in-code comment ("not a false-recovered risk … a held row leaves the hold *only* by ceasing to be agent_running") is provably wrong, and this defeats A5. Secondary trigger: the markers-clear→re-dispatch gap between retry attempts. Pre-diff the grace window absorbed a one-tick drop; post-diff the hold consumes grace, so a single-tick drop is no longer buffered.

## Medium
- [ ] `notification_dispatcher_test.rb` ~54 (`live_retry_row` default `marker: "review_error"`): the hold's **marker-independence is unverified** — every hold test uses an `agent_running` holder whose marker is a recovery marker, but the realistic production hold (a retry that has already *cleared* its stale marker yet still holds the lock, i.e. `agent_running` + `agent_working`) is never exercised. A refactor that "tidies" `live_recovery_identities` to also gate on `NotificationBuilders.recovery?(row)` would silently drop the hold in exactly that window and fire a premature Recovered — and the whole suite would still pass.

## Nit
- [ ] `notification_builders_test.rb` ~290/327 (and dispatcher archived test): the **`archived` + recovery-marker log path is never behaviorally asserted** — `test_archived_error_marker_is_suppressed` and the dispatcher asymmetry test pass *no* logger, so nothing pins that `:notification_skipped_live_agent` fires for an archived row with `action: "archived"`. The builder comment intends the archived contradiction to "stay diagnosable," but a regression narrowing the log gate to `agent_running` only would drop it undetected (line coverage comes only from the shared `agent_running` gate).
- [ ] `notification_dispatcher_test.rb` ~607/718: the **`project` component of `recovery_identity` is never independently exercised** — all fixtures use project `"hive"`. The negative-scoping tests vary slug and stage, but a refactor dropping `project` from `recovery_identity` would let a live `agent_running` retry in project A hold/supersede a same-slug+stage recovery in project B, and every existing test would still pass.
- [ ] `notification_dispatcher_test.rb` ~656 (`test_live_retry_resolving_to_manual_steering_fires_recovered_once`): asserts only `recovered.size == 1`, never total `telegram.messages.size`, so the test comment's claim that "manual_steering produces no notification of its own, so Recovered is the only message" is **not actually pinned** — if a resolution action later emits its own push, the test stays green.

ivankuznetsov:

## High
- [ ] [comments] `notification_dispatcher.rb:156-159` comment ("This is not a false-recovered risk: a held row leaves the hold **only** by ceasing to be agent_running, which **means** the retry finished") overclaims a safety invariant the code does not hold — and the underlying logic fires a false "✅ Recovered": the hold (`next if live_identities.include?(...)`, line 169) lifts whenever the `agent_running` row is merely *absent* from a snapshot, not only when the retry finishes. A transient per-project status degrade (`status.rb` `project_payload_or_degraded`/`missing_project_path`/`not_initialised` keeps top-level `ok:true` while emptying that project's tasks; `status_watcher.rb:273` `next if project_doc["error"]`; `supervisor.rb:167-170` still calls `process_rows` when `result.ok`) drops the still-running row for one tick; since grace already elapsed *during* the hold, line 169 no longer skips → `recovered_message` is sent and the entry removed though nothing recovered. Violates A5; the existing tests cover only clean action transitions, never a mid-hold degrade.

## Medium
- [ ] [tests] `notification_dispatcher_test.rb:54-56` (`live_retry_row` marker defaults to `review_error`, never overridden): the hold's **marker-independence is unpinned**. The dominant production retry is `action=agent_running, marker=agent_working` (clean retry, no stale marker), but every hold test uses a recovery marker. A plausible "tighten it" refactor adding `&& NotificationBuilders.recovery?(row)` to `live_recovery_identities` would pass 100% of the suite yet stop holding the common clean-retry case → premature "✅ Recovered" while it runs — the exact regression this feature prevents.
- [ ] [errors] `notification_dispatcher.rb:169` + `notification_builders.rb:57`: in the common clean-retry case (`agent_running`+`agent_working`) the hold is **invisible** — `build`'s skip-log is gated on `recovery?(row)` (false for a cleared marker) so it emits nothing, and the dispatcher hold logs nothing either, so a withheld "Recovered" leaves no audit breadcrumb. Only the stale-marker variant produces a log; the held-clean-retry path is untested and unlogged.
- [ ] [errors] `notification_dispatcher.rb:171-176` (pre-existing; unchanged line, surfaced via the touched function): partial multi-chat delivery silently drops "Recovered" — `if send_notification(recovered_message(row)).any?` removes the fingerprint after a *single* chat succeeds, so a chat that raised never receives "Recovered" and (entry gone) gets no retry, unlike the initial-alert path which tracks `delivered_to`/`pending_chat_ids`. Multi-operator deployments leave some chats with a dangling "⚠ stuck".

## Nit
- [ ] [tests] `notification_dispatcher_test.rb`: the `project` component of `recovery_identity` (`[project, slug, stage]`) is never independently exercised — all fixtures use project `"hive"` (slug and stage have negative-scoping tests; project does not). Dropping `project` from `recovery_identity` would pass the whole suite while letting a live retry in project A hold/supersede a same-slug+stage recovery in project B.
- [ ] [tests] `notification_builders_test.rb:327-331` / dispatcher archived test: the **archived skip-log is never asserted with a logger** — `assert_live_agent_skip_logged` is only called with `action: "agent_running"`. Hardcoding `action: "agent_running"` instead of `action: row.action` (notification_builders.rb:58-60), or dropping the log for archived rows, would pass everything (line coverage met, value/branch unguarded).
- [ ] [tests] `notification_dispatcher_test.rb:645-647` (and `:519-521`): count/substring-only assertions — `recovered.size == 1` via `include?("Recovered")` with no total `telegram.messages.size` guard and (archived test) no exact `RECOVERED_MESSAGE` body check, so a malformed Recovered body or a spurious extra message would slip.
- [ ] [comments] `notification_builders.rb:54`: "on every 30s poll tick" hardcodes a configurable default — `bot.poll_interval_sec` defaults to 30s but is operator-tunable (5s floor). Say "every poll tick (default 30s)" so the figure can't silently drift.
- [ ] [comments] `notification_builders.rb:52`: the bare "A6" plan-tag reference is comment rot — "A6" appears nowhere in `lib/`, `wiki/`, or `docs/`, so a future reader can't resolve it. State the concrete intent (or link `wiki/log.d/20260624T184142Z-telegram-live-agent-suppression.md`).
- [ ] [comments] `notification_builders.rb:55-56`: "recovery? is the exact gate: true only for the recovery/error markers" omits `recovery?`'s action clause (`recover_execute`/`recover_review`/`error`); it's accurate only because `row.action` is `agent_running`/`archived` here. Qualify as "for these rows `recovery?` reduces to the marker check" so it isn't read as a general description.
- [ ] [simplify] `notification_dispatcher.rb:294`: redundant `.to_s` — `row.action.to_s == AGENT_RUNNING` is inconsistent with every other `row.action` comparison in both files (`:89`, `:108`, `:346`, builders `:49`) which treat it as a plain String. Dropping `.to_s` aligns the builder/dispatcher action-comparison contract (and removes a latent string-vs-symbol divergence where the builder would suppress but the dispatcher wouldn't, or vice-versa).

PR #590, opened 2026-06-24
fix(bot): every needs-input row gets a working button or is suppressed

ivankuznetsov:
# fix(bot): every needs-input row gets a working button or is suppressed

## Summary

A Telegram operator who got a "Needs input" push often had no way to act on it: many announced `needs_input` rows shipped with only a **Show details** button, that button dead-ended in "No diagnostic available", and the `/status` surface rendered no button at all for the same rows. The two surfaces computed their keyboards independently and had drifted.

This collapses both surfaces onto one resolver. `Hive::Bot::RowActions.resolve(row)` is now the single source of truth for which chat actions a status row offers; push notifications and `/status` both render from it, so they can no longer disagree. Every announced needs-input row now carries a button that actually dispatches the right `hive` verb, or is suppressed at the source. **Show details** renders the live row summary plus a next-step hint instead of dead-ending.

```mermaid
flowchart TB
    Row["StatusWatcher::Row"] --> RA["RowActions.resolve → Resolution{actions, kind}"]
    RA --> NB["NotificationBuilders (push keyboards)"]
    RA --> SUP["Supervisor#status_action_button (/status)"]
    NB --> CB["CallbackHandlers (rerun / approve_plan / details)"]
    SUP --> CB
    CB --> Q["daemon dispatch-request queue"]
```

### Button mapping

`RowActions` keys off `(action, marker, stage, workflow, attrs)`. Needs-input resolutions:

| Row | Actions (primary first) |
| --- | --- |
| coding `2-brainstorm` + `waiting` | Answer |
| coding `3-plan` + `waiting` | Approve plan, Show details |
| `review_waiting` | Accept all, Reject all, Show details |
| `review_waiting` + `reason=fix_guardrail` | Show details (terminal) |
| `execute_waiting` | Re-run develop, Show details |
| finalize `waiting` | Run finalize, Show details |
| generic / non-coding needs-input | Run |
| `needs_input` + `marker=none` / `complete` | suppressed (no push, no `/status` button) |

Manual-only and recovery rows keep today's behavior: `execute_stale` and `fix_guardrail` are terminal Details, retryable recovery is Autofix, `ready_to_*` is Approve.

### Design decisions

**New dispatch callbacks.** `rerun:<project>:<slug>:<stage>:<verb>` advances the next workflow verb (develop / finalize / generic run) through the daemon dispatch-request queue. The existing `approve:` callback asserts a terminal source-stage marker, which is wrong for a paused row, so Run/Re-run use `--from` (workflow verbs) or `--stage` (generic `run`) instead.

**Plan-approve from chat.** `approve_plan:` mirrors `Hive::Daemon::PlanApproval`: it validates the plan command, rewrites it to `develop`, and flips the on-disk `:waiting` marker to `:complete` **before** dispatch (the flip is gated behind validation, so a corrupt `plan.md` or malformed command leaves the marker untouched and replies with an actionable error). An already-advanced `:complete` / develop-shaped row dispatches a single well-formed `develop`; a non-approvable marker refuses cleanly rather than crashing.

**Working Details.** The Details button and `/details <slug>` now route through the cached `hive status` snapshot and `Supervisor#render_details` (the same path `/status <slug>` uses), appending a next-step hint derived from the resolved primary action. Recovery diagnostics still run via the unchanged `refresh_diagnose:` path.

**Root-cause reclassification.** The presentation-layer suppression is the belt; this is the suspenders. In `TaskAction`, markerless coding brainstorm/execute rows and markerless finalize rows now classify as `READY_TO_RUN`, not phantom `NEEDS_INPUT`. A stray `:complete` at execute-stage maps to `READY_TO_OPEN_PR`; other unexpected terminal markers there fail loud as `ERROR`.

**Type safety.** `Resolution`/`Action` are immutable `Data` values with a closed `ROLES`/`KINDS` vocabulary and an exactly-one-primary invariant enforced at construction. Role→presentation tables and the `needs_input` dispatch fail loud (`.fetch` / `raise`) on an unmapped role or kind rather than silently degrading. Per-row notification builds are isolated behind a widened `rescue StandardError` so one malformed row can no longer abort a whole notification tick or `/status` keyboard.

## Test plan

All targeted suites green and `rubocop` clean on the 8 changed lib files; the 100% line-coverage gate passes on changed files.

- `bin/rake test TEST=test/unit/bot/button_coverage_test.rb` — enumerates the `(action, marker, stage, workflow)` product from the closed `TaskActionKind` enum and the stage/marker registries (~42k assertions). For every row it asserts the resolver either suppresses it or yields a tap-resolvable primary callback, that `/status` agrees with the resolver on actionability, and that no non-terminal needs-input row ships with only Details. A guard test fails if a new `TaskActionKind` is added without a `RowActions` mapping.
- `bin/rake test TEST=test/unit/bot/row_actions_test.rb` — pins the button mapping table, the `none`/`complete` suppression, and `coding_row?` gating.
- `bin/rake test TEST=test/unit/bot/callback_handlers_test.rb` — `rerun:` argv shapes (`--from` for develop/finalize, `--stage` for generic run), `approve_plan` marker flip / validate-before-flip / already-advanced / corrupt-state paths, and unknown-callback fallback.
- `bin/rake test TEST=test/unit/task_action_test.rb` and the dispatch-parity test — the `none`/`complete` reclassification and the fail-loud foreign-marker branch.
- `bin/rake test TEST=test/unit/bot/notification_builders_test.rb test/unit/bot/supervisor_test.rb test/unit/bot/slash_handlers_test.rb test/unit/bot/router_test.rb test/unit/daemon/plan_approval_test.rb` — push keyboards, `/status` keyboard + working Details, slash routing, callback intents, and the plan-approval helper.

Not covered by automated tests: live Telegram callback redelivery and an end-to-end daemon-off dispatch are not exercisable headlessly (the surface is Telegram inline keyboards reachable only via a live bot token; the review stage also recorded `browser=skipped`), so there is no demo capture for this PR.

## Review summary

Three Compound Engineering review passes (`ce-correctness`, `ce-adversarial`, `pr-review-toolkit` personas, plus codex passes). Every auto-fixable Medium/Nit was resolved and verified regression-free in the later passes: kind-tag dispatch, fail-loud `needs_input`/`next_step_hint`, the widened per-row `rescue`, the declared `row_actions` require, the collapsed `:run`→`:rerun` role, a realistic already-advanced `approve_plan` fixture, corrupt-state rescues, fail-closed `WORKFLOWS`/`MARKERS` derivation, the `findings_reject` role-table parity, and routing unknown non-terminal execute markers to `:error`.

One High item remains an operator decision rather than a code defect: the new `approve_plan:`/`rerun:` buttons route the same queue-routable verbs (`develop`/`finalize`/`run`) the **pre-existing** ready-row Approve button already uses, which the daemon's `project_disabled` gate (added in PR #241, not this branch) drops for daemon-disabled projects. Whether that strands a tap depends on an operational fact about the daemon-off deployment (does a daemon drain the queue) that can't be determined from code. In-tree mitigations bound the blast radius: on daemon-off the flipped `:complete` plan row reclassifies to `ready_to_develop` and re-pushes Approve (recoverable, not stranded), and the already-advanced row now dispatches a single well-formed `develop`. The escalation is left open for operator confirmation.

The daemon-safety concern from pass 1 (U6 reclassification vs. the `Policy` coding-immunity invariant) was resolved by keeping the `TaskAction` reclassification consistent with the documented intent and pinning it with the parity and `task_action` tests; gates `stage_literal_guard` and `daemon_stale_agent_healing` are green.

## Linked task

- Hive task: `telegram-bot-every-announced-needs-260624-ce0e`
- This umbrella stacks after the `none`/`complete` (9398) and dead-end-Details (9399) sub-tasks, folding their root causes (U6 reclassification, U5 working Details) where not already landed.

---


ivankuznetsov:

## High
- [ ] `.github/workflows/release.yml:280` switches the post-publish smoke to `macos-15` + `colima start`: GitHub-hosted arm64 macOS runners do not support nested virtualization, so the release job will fail before `packaging/docker/smoke.sh` runs.

## Medium
- [ ] `lib/hive/bot/row_actions.rb:155` assumes `Hive::Bot::NotificationBuilders` is already loaded: requiring the new canonical resolver directly and resolving any ready/recovery/details action raises `NameError`, so the resolver is not safely reusable as its own source of truth.

## Nit

ivankuznetsov:

# Code Review — telegram-bot-every-announced-needs-input (pass 1)

Scope: `git diff origin/main..HEAD` (commits U1–U7). Reviewed the canonical
`RowActions` resolver, the push (`notification_builders.rb`) and `/status`
(`supervisor.rb`) surfaces, the new `rerun`/`approve_plan` callbacks, the
Show-details rewiring, and the `task_action.rb` reclassification.

Gates verified locally: `rubocop` clean on all 8 changed lib files; all
changed test suites green (`row_actions` 11, `button_coverage` 3/18376
assertions, `task_action` 81, `notification_builders` 44, `supervisor` 144,
`callback_handlers` 41, `slash_handlers` 33, `router` 40, parity 1). The
verb/flag math (`develop`/`finalize` use `--from`, `run` uses `--stage`) is
correct against `cli.rb`/`stage_action.rb`; the `status_command?` exact-match
still routes Show-details in-process; the `/\Aapprove:/` vs `/\Aapprove_plan:/`
regex order does not shadow (literal colon). No must-fix defects found; the
items below are robustness, anti-drift, and test-quality improvements.

Note: `.github/workflows/release.yml` and `CHANGELOG.md` appear in the diff
only because the branch is 2 commits behind `origin/main` (stale base) — they
are NOT this branch's changes and are excluded from review (see Nit).

## High

_No must-fix-before-merge defects found._

## Medium
- [ ] `approve_plan` flips the plan marker to `:complete` on disk (callback_handlers.rb:106, via `PlanApproval.prepare` → `Markers.set`) during Result construction, *before* the `develop` dispatch is durably enqueued and with no rollback: a rescued `enqueue_dispatch_request` failure (read-only state dir / ENOSPC → nil) strands the task advanced to `:complete` with no `develop` queued and nothing driving it. Defer the flip until after the request is confirmed written, or restore `:waiting` on a nil enqueue result.
- [ ] `approve_plan` does the marker flip via a blocking `flock(LOCK_EX)` (Markers.set) on the single poll thread (callback_handlers.rb:106); contention with another markers-lock holder (e.g. a daemon child for the same slug) blocks *all* bot update processing — a new pattern vs. the other handlers which dispatch async. Consider `LOCK_NB` + short retry, or route the flip through the daemon.
- [ ] Anti-drift goal only partially met: `RowActions.resolve` delegates back into `NotificationBuilders` (`verb_for_action`, `details_callback`, `recovery?`/`retryable_recovery?`/`autofix_callback`) and the role-keyed presentation tables stay split and duplicated — `READY_ROLES` (row_actions.rb:21) vs `READY_ACTIONS` + `label_for_action` (notification_builders.rb) vs `status_action_emoji` (supervisor.rb:1459) — with no shared source or parity test, which is the exact drift the task set out to eliminate.
- [ ] `notification_builders#needs_input` dispatches on exact role-array literals (notification_builders.rb:137, e.g. `when [:approve_plan, :details]`); any reorder or extra action added in `RowActions` silently falls through to `default_needs_input` instead of the intended surface. Add a `kind`/tag field to `Resolution` and dispatch on that.
- [ ] Long-slug `approve_plan:`/`rerun:` callbacks routinely exceed Telegram's 64-byte cap (this branch's own slug → `approve_plan:hive:<slug>:3-plan` ≈ 71 bytes), so they depend on the in-memory compaction registry (process-lifetime, TTL/size-capped); after a bot process restart or registry eviction a tapped button degrades to "I did not understand that." (router.rb:301) rather than acting. Reply with an actionable "button expired — reopen /queue" when an unresolved `#`-token reaches `:unknown`.
- [ ] The "new TaskActionKind must be mapped" guard test passes for an unsound reason: inert kinds (`agent_running`/`archived`/`manual_steering`) only resolve via the incidental `recovery?` path when paired with `marker=error`, not because they have real `RowActions` decisions, so a future inert kind could slip through the guard (button_coverage_test.rb).
- [ ] `approve_plan`'s two early-exit branches are untested — row-not-in-snapshot ("Task status changed - reopen /queue.") and empty `state_file` ("Plan state is unavailable - reopen /queue.") — and the stale-button-tap (cold/advanced snapshot) is the common real case (callback_handlers_test.rb).

## Nit
- [ ] `coding_stage?(row, "7-finalize")` is dead/unreachable — the coding finalize stage dir is `8-finalize` (already checked on the same line); index 7 is `7-artifacts`. Drop the `|| coding_stage?(row, "7-finalize")` or fix to the intended stage (row_actions.rb:71).
- [ ] `waiting_input` is dead code after the `needs_input` rewrite — no remaining callers; its logic now lives in `RowActions`. Remove it (notification_builders.rb:161).
- [ ] `finalize_action` is indented at 6 spaces while every sibling method uses 4 (task_action.rb:437-443); rubocop doesn't flag it, but it reads as wrongly nested. Re-indent to 4 spaces.
- [ ] `RowActions.terminal_details?` is public production API invoked only by tests (row_actions.rb:54; row_actions_test.rb, button_coverage_test.rb) — a test-oracle smell. Either give it a production call site or fold the assertion into a test helper.
- [ ] `next_step_hint` derives the verb via `action.callback.split(':').last` (supervisor.rb:1507-1509) — brittle if the callback format ever gains a trailing token; the generic case also renders "Next: tap Run to run run." Prefer an explicit `verb` field on the `Action` struct.
- [ ] `approve_plan` does not rescue `SystemCallError`/`IOError` from the marker write (read-only/full state dir); only `ArgumentError` (handler-level) and `NotApprovable` are caught, so an I/O failure propagates to the poll loop's generic rescue and the operator gets no reply (callback_handlers.rb:96-117).
- [ ] `plan_action` still classifies marker `:none` as `plan_waiting` (needs_input) while `brainstorm_action`/`execute_action`/`finalize_action` now reclassify `:none` → `generic_ready_to_run` (task_action.rb:430-435); the bot suppresses it, but TUI/`hive status`/daemon consumers see the asymmetry. Low reachability (plan `:none` with a complete artifact is unusual), but inconsistent with the U6 "fix at the source" intent.
- [ ] Branch is 2 commits behind `origin/main`, so the diff shows `release.yml`/`CHANGELOG.md` as reverted (stale-base artifact — reverting the #584 arm64-native smoke fix and #583 changelog, not actual branch changes). Rebase onto `origin/main` before merge so the review/merge diff is clean.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — pass 1

Scope: `git diff origin/main..HEAD`. Personas run: code-reviewer,
pr-test-analyzer, comment-analyzer, silent-failure-hunter,
type-design-analyzer, code-simplifier. Findings below are the union across
personas, deduplicated (overlapping findings carry multiple persona tags).

`.github/workflows/release.yml` / `CHANGELOG.md` are excluded — they appear in
the diff only because the branch is 2 commits behind `origin/main` (stale
base), not as real changes.

**Three High items are gate/test failures I reproduced directly** (commands
below), so the branch is not mergeable as-is:
- `bundle exec rake test TEST=test/unit/stage_literal_guard_test.rb` → 1 failure
- `bundle exec rake test TEST=test/integration/daemon_stale_agent_healing_test.rb` → 1 failure
- `waiting_input` confirmed unreachable (no callers in `lib/` or `test/`) → 100%-line coverage gate fails.

## High
- [ ] [code-reviewer] Stage-literal guard test FAILS: `lib/hive/bot/row_actions.rb:63,64,71` hardcode coding stage literals (`"2-brainstorm"`, `"3-plan"`, `"8-finalize"`, `"7-finalize"`) without the mandatory `# coding-scoped: <reason>` annotation every other lib coding-stage literal carries — `test/unit/stage_literal_guard_test.rb` is red (reproduced). Add the inline `# coding-scoped:` annotation (or a block annotation) to each line.
- [ ] [code-reviewer][comment-analyzer][code-simplifier] 100%-line coverage gate FAILS (~99.95%): `waiting_input` (`lib/hive/bot/notification_builders.rb:161-166`) is now dead — the rewritten `needs_input` routes everything through `RowActions.resolve`, so no caller remains in `lib/` or `test/` and its lines are uncovered. Delete `waiting_input` (and the now-unreachable `actions: nil` fallbacks it was the sole user of).
- [ ] [code-reviewer] Daemon-side regression from the U6 `:none`→`generic_ready_to_run` reclassification: `test/integration/daemon_stale_agent_healing_test.rb:296` FAILS (expects `needs_input`, gets `ready_to_run`) after a finalize `unpushed_commits` heal leaves `marker=none`. `TaskAction` feeds `Hive::Daemon::Policy`, whose "coding is immune — markerless post-run coding stage classifies as debounced `needs_input`" invariant (policy.rb:81-82) is now broken, risking cap-draining `hive run` re-dispatch every tick; neither Policy nor the integration test was updated. Either scope the reclassification to the bot's `RowActions` (keep `TaskAction` coding `:none`→`needs_input`) or deliberately update Policy + the test and prove the run debounce protects markerless coding rows. (lib/hive/task_action.rb:402,419,440; lib/hive/workflows/coding.rb)

## Medium
- [ ] [silent-failure-hunter] `approve_plan` marker write can raise `SystemCallError`/`IOError` (read-only/full state dir, via `PlanApproval.prepare`→`Markers.set`) that nothing rescues — only `NotApprovable` (handler) and `ArgumentError` (handler-level) are caught; it propagates to the poll-loop rescue which logs `:fatal` and sends no reply, so the operator sees the acked spinner clear then nothing. Add `rescue SystemCallError, IOError` returning an actionable reply (lib/hive/bot/handlers/callback_handlers.rb:96-117).
- [ ] [silent-failure-hunter] `approve_plan` flips the marker to `:complete` (callback_handlers.rb:106) before the develop dispatch is enqueued; a rescued `enqueue_dispatch_request` failure leaves the marker advanced with no rollback. Recoverable (the row resurfaces as ready-to-develop and the failure is surfaced), but the deliberate "advance-then-resurface" contract is undocumented — defer the flip until after enqueue or add a comment noting the window.
- [ ] [silent-failure-hunter] An unmapped role raises `KeyError` in `label_for_action` (`.fetch`) with no per-row rescue in `NotificationDispatcher#current_notifications` (notification_dispatcher.rb:66) — one malformed row would abort ALL push notifications that tick (not just the offending row). No live gap today, but defensive: wrap the per-row build in a rescue, or have both role tables fall back + log instead of raising. (lib/hive/bot/notification_builders.rb label_for_action)
- [ ] [pr-test-analyzer] `approve_plan` early-exit branches untested: row-not-in-snapshot ("Task status changed - reopen /queue.") and empty `state_file` ("Plan state is unavailable - reopen /queue.") — the stale-button-tap is the common real case (lib/hive/bot/handlers/callback_handlers.rb:99-102). Add `handlers_with_rows([])` and a row-without-state_file case.
- [ ] [pr-test-analyzer] The new `execute_action` `else → :error` branch (a stray terminal marker like `:review_complete` at the execute stage) has no coverage; `button_coverage_test` injects `action` directly and never exercises the classifier, so a future collapse back to `:generic_ready_to_run` would offer a runnable `hive run` against a structurally-wrong marker with no test failing (lib/hive/task_action.rb:422-426). Add a `task_action_test.rb` case asserting `:error`.
- [ ] [comment-analyzer] Stale comment: "...diagnose replies still fire for Show details" (lib/hive/bot/supervisor.rb:1016) — Show details now routes through the synchronous in-process `render_details` path, not the child reaper's diagnose reply (only `refresh_diagnose` reaches that now). Update/drop the parenthetical.
- [ ] [comment-analyzer] Stale rationale: the `plan_waiting` comment "the operator reviews the draft and advances it from the CLI — so it points at the plan, not the /answer flow" (lib/hive/bot/notification_builders.rb:184-189) predates this PR's in-chat `approve_plan` button (keyboard is now Approve + Details). Update to reflect the operator can now tap Approve in chat.
- [ ] [comment-analyzer] Misleading user-facing string: "Tap Show details to dump it here" (lib/hive/bot/supervisor.rb:1282-1283, in `diagnose_reply_for_child`) over-promises — Show details now renders the row summary (title/action/marker/attrs/hint), not the diagnostic body. Reword to point at the Refresh/`refresh_diagnose` path.
- [ ] [code-simplifier] Anti-drift gap: `READY_ROLES` (row_actions.rb:21-32) has every value `:approve` (so `.fetch` at :103 is always `:approve`) and its key set duplicates both `NotificationBuilders.verb_for_action`'s keys and `NotificationBuilders::READY_ACTIONS` — three parallel role tables with no shared source. Demote `READY_ROLES` to a membership Set, or define `ready_action?` via `verb_for_action` to make it the single source of truth.
- [ ] [type-design-analyzer] `Action.role` is a free symbol validated nowhere at construction; the closed vocabulary lives only in the consumers' `Hash#fetch` tables (status_action_emoji / label_for_action), so a typo'd or newly-added role becomes a deep-path `KeyError` instead of a clear boundary error. Add a frozen `ROLES` set and validate in `Action#initialize` (lib/hive/bot/row_actions.rb:15-19).

## Nit
- [ ] [code-reviewer][code-simplifier] Dead branch `coding_stage?(row, "7-finalize")` (lib/hive/bot/row_actions.rb:71) — the coding finalize stage dir is `8-finalize` (index 7 is `7-artifacts`); legacy `7-finalize` is migrated before status reports it. Drop the disjunct.
- [ ] [code-reviewer][comment-analyzer][code-simplifier] `finalize_action` is indented 6 spaces vs the 4-space siblings (lib/hive/task_action.rb:437-443); rubocop passes but it reads as wrongly nested. Re-indent to 4 spaces.
- [ ] [code-reviewer] Awkward copy "Next: tap Run to run run." for the generic-run hint (lib/hive/bot/supervisor.rb:1509, `action.callback.split(':').last` on the `rerun:…:run` callback). Special-case the verb or use "...to run this stage."
- [ ] [code-reviewer] Now-unreachable `--diagnose` lookup in `inferred_success_slug`'s `when "status"` (lib/hive/bot/supervisor.rb:1389) — Show details/`/details` send `["hive","status","--json"]` and return early via `status_command?` before this runs. Simplify/drop.
- [ ] [code-simplifier] Once `waiting_input` is removed, the `actions: nil` keyword defaults and `|| [ RowActions.action(...) ]` fallbacks in `default_needs_input`/`brainstorm_waiting`/`plan_waiting`/`review_waiting` are unreachable (always called with `actions:` from the live path); make `actions:` required and drop the fallbacks so keyboards come solely from the resolver (lib/hive/bot/notification_builders.rb).
- [ ] [code-simplifier] `approve` and `rerun` duplicate the argv builder (`stage_flag = verb == "run" ? "--stage" : "--from"` + the same `dispatch_then_reply` argv); extract a shared private helper (lib/hive/bot/handlers/callback_handlers.rb:65-94).
- [ ] [pr-test-analyzer] `approve_plan` empty-`suggested_command` fallback (synthesize `hive plan <slug> --from <stage>`) is untested — a typo would ship silently (lib/hive/bot/handlers/callback_handlers.rb:104-105). Add a row-with-nil-suggested_command case.
- [ ] [comment-analyzer] The core `:none`/`:complete` reclassification branches carry no inline "why" comment while neighboring branches do (lib/hive/task_action.rb:398-440); rationale lives only in `wiki/modules/task_action.md`. Add a one-line "markerless = nothing ran yet → runnable, not an input gate" note.
- [ ] [type-design-analyzer] `Resolution` can represent contradictory states (`suppress:true` with non-empty actions; `actions:[]`+`suppress:false`); and `notification_builders#needs_input` reverse-engineers the surface intent by matching exact role arrays (`when [:approve_plan, :details]`, …) — fragile if `RowActions` reorders/adds an action (silently falls to `default_needs_input`). Add a `kind` tag to `Resolution` and dispatch on it (row_actions.rb:9-13; notification_builders.rb:137-151).
- [ ] [type-design-analyzer] `Action.primary` is not constrained to exactly one per resolution; both surfaces rely on `actions.find(&:primary) || actions.first`. Correct by discipline today; consider a `Resolution#primary` accessor and/or a one-primary assertion (lib/hive/bot/row_actions.rb:15-19).

ivankuznetsov:

## High

## Medium
- [ ] `approve_plan` is not race-safe once the status row is already `ready_to_develop`: `CallbackHandlers#approve_plan` feeds the current row's `hive develop ...` suggested command into `PlanApproval.prepare`, which only accepts `hive plan ...`, so a stale Approve tap on an already-complete plan replies "Bot got confused" instead of the planned no-op flip + develop dispatch (`lib/hive/bot/handlers/callback_handlers.rb:107`).

## Nit
- [ ] `RowActions` is not self-contained when required directly: `row_actions.rb` references `Hive::Bot::NotificationBuilders` without loading it, so `require "hive/bot/row_actions"; RowActions.resolve(...)` raises `NameError` unless another file happened to require notification builders first (`lib/hive/bot/row_actions.rb:193`).

ivankuznetsov:

# Code Review — telegram-bot: every announced needs-input row carries a working button or is suppressed (Pass 2)

Scope: `git diff origin/main..HEAD` (merge-base `54fd3455`). 12 commits; `lib/hive/bot/*` (new `row_actions.rb`), `lib/hive/task_action.rb`, `lib/hive/workflows/coding.rb`, schema/logger, wiki, and tests.

Intent: one canonical `RowActions.resolve(row)` feeding both the push keyboard and the `/status` keyboard so they can't drift; new `approve_plan:`/`rerun:` chat-dispatch callbacks; in-process "Show details"; reclassify `none`/`complete` markers away from phantom `needs_input`.

Review team: ce-correctness (opus), ce-adversarial (opus), ce-reliability, ce-maintainability, ce-testing, ce-project-standards. Findings verified against the daemon dispatch path (`dispatcher.rb`), `plan_approval.rb`, `dispatch_request_queue.rb`, and the coding stage table.

## High
- [ ] New `approve_plan:`/`rerun:` buttons dispatch queue-routable verbs (`develop`/`finalize`/`run`) that the daemon **blocks for daemon-disabled projects** — the exact daemon-OFF target these buttons serve (`callback_handlers.rb:91-96,114-122` → `supervisor.rb:1001-1003` enqueue → `dispatcher.rb:1279` `project_disabled`): the push only appears when `daemon.enabled != true` (`notification_dispatcher.rb:63,117-126`), `project_enabled?` reads the same `daemon.enabled` key (`dispatcher.rb:1702`), and `ALLOWED_VERBS` (`dispatch_request_queue.rb:18`) includes develop/finalize/run but not `approve`; so the tap silently no-ops (no ack — `supervisor.rb:1014`; request expires at 600s) while `approve_plan` has already flipped the on-disk marker `:waiting→:complete`. The headline U2 capability is non-functional for the deployment it targets.
- [ ] `approve_plan` mishandles the realistic already-approved / advanced plan row, and the test masks it with an impossible fixture: a `:complete` plan row carries `suggested_command: "hive develop …"` (`plan_complete` → command `develop`, `task_action.rb:41-45`), which `PlanApproval.rewrite_to_develop` rejects with `ArgumentError` (requires `argv[1]=="plan"`, `plan_approval.rb:69`); `approve_plan` rescues only `NotApprovable`/`SystemCallError`/`IOError`, so it falls to the generic `rescue ArgumentError` in `handle` (`callback_handlers.rb:51-54`) → "Bot got confused - please retry from /queue." + `callback_malformed`, instead of dispatching the valid develop or replying "Plan is no longer waiting." `PlanApproval`'s race-safe `:complete` no-op branch (`plan_approval.rb:56-58`) is unreachable from the bot. `test_approve_plan_callback_allows_already_complete_marker` (`callback_handlers_test.rb:373-384`) uses a fixture that can't occur (`:complete` row with `action:"needs_input"` and a `hive plan` command) and asserts develop IS dispatched — certifying behavior production never produces.

## Medium
- [ ] `/status` + `/queue` render the Approve-plan button with **no daemon-enabled gate** (`supervisor.rb:1454` `status_action_button` → `RowActions.resolve`), while the push is gated by `suppress_daemon_plan_pause?` (`notification_dispatcher.rb:63`) — so on a daemon-ENABLED project the operator can tap it, becoming a second writer of the plan marker racing the daemon's `PlanApproval` auto-dispatch (`dispatcher.rb` plan_approval trigger); the redundant `hive develop` can WrongStage and surface a spurious failure alert. Mitigated (not eliminated) by the daemon's per-slug in-flight gate; contradicts the plan's stated "buttons target the daemon-OFF operator" safety invariant. Gate the `:approve_plan` action the same way the push is gated.
- [ ] Per-row notification isolation is incomplete: `current_notifications` calls the three `suppress_*` predicates and `fingerprint(row)` OUTSIDE the rescue, and `build_notification` catches only `ArgumentError, KeyError` (`notification_dispatcher.rb:60-86`); a malformed row raising any other class (`NoMethodError`/`TypeError`) aborts the whole tick via `status_loop`'s generic `:fatal` rescue → ALL notifications dropped that tick, recurring every tick, with no row attribution. Same narrow-rescue pattern on `status_action_button` (`supervisor.rb:1473`) can abort the whole `/status` keyboard. Widen both to `StandardError` and isolate the full per-row body.
- [ ] Role→presentation is enumerated in three independent tables — `label_for_action` (`notification_builders.rb:472`), `status_action_emoji` (`supervisor.rb:1476`), and the `case action.role` in `next_step_hint` (`supervisor.rb:1521`) — and only the first two `.fetch` (loud `KeyError`); `next_step_hint`'s catch-all `else` (`supervisor.rb:1530`) silently degrades a new primary-capable role to the generic "open on a laptop" hint, and no test ever invokes `next_step_hint`, so `button_coverage_test` won't catch it. Consolidate into one role-keyed table or make the hint fetch-backed/raise.
- [ ] `RowActions`↔`NotificationBuilders` is a bidirectional runtime dependency with an undeclared require: `row_actions.rb` requires only `hive`/`hive/workflows` (`:1-2`) yet calls `NotificationBuilders.{recovery?,manual_only?,verb_for_action,retryable_recovery?,details_callback,autofix_callback}` (`:192-194`) — so requiring `row_actions` alone and resolving a recovery row raises `NameError`. The "single source of truth" is two mutually-dependent modules held together by a deliberately-omitted require. Extract the shared predicates into a lower-level module both require one-directionally.

## Nit
- [ ] `["hive","status","--json"]` is a magic sentinel argv built by `show_details`/`details` (`callback_handlers.rb:166-174`, `slash_handlers.rb:234-242`) only so `status_command?` (`supervisor.rb:1549`) routes them to the in-process `render_details` (which renders plain text, never JSON): the `--json` token is misleading and the literal is duplicated across two producers + the matcher with no shared constant, so drift silently regresses to shelling out. Give the Result a dedicated `:render_row_details` action or hoist a named constant.
- [ ] `:run` vs `:rerun` roles split semantically-identical paused-stage re-run buttons that share one `rerun:` callback (execute→`:rerun` label "Re-run"; finalize/generic→`:run` label "Run") (`row_actions.rb:92-103`); pick one role for "paused stage, re-run the agent" and let the `verb` field carry develop/finalize/run.
- [ ] `needs_input` dispatch still ends in `else default_needs_input` (`notification_builders.rb`), re-introducing (milder) the silent-misroute the `kind` tag was added to prevent — a new `needs_input` KIND renders generic text rather than failing loud. Replace the `else` with an explicit case that raises on an unexpected kind.
- [ ] `approve_plan` does not distinguish a corrupt `plan.md` from a malformed callback: `Hive::Markers.current` scans a UTF-8 string (`markers.rb:54`), so invalid bytes raise `ArgumentError` → generic "Bot got confused" + `callback_malformed` instead of an actionable corrupt-state message. Add a local `rescue ArgumentError` distinguishing parse/marker failures (the validate-before-flip invariant is also untested — add a test with a non-`hive plan` command asserting the marker stays `:waiting`).
- [ ] `button_coverage_test.rb:14-23` hand-lists `WORKFLOWS`/`MARKERS` while deriving actions from the closed enum and stages from the registry; derive `WORKFLOWS = Registry.all.map { … }` so the workflow axis is also fail-closed (a 3rd workflow currently wouldn't extend the guarantee).
- [ ] The parity test's brainstorm cell is now tautological — the legacy arm calls the production `brainstorm_action` (`task_action_dispatch_parity_test.rb:28`), the same method the kind-dispatch path resolves, so the brainstorm `:none→generic_ready_to_run` change is no longer cross-checked here (it is independently pinned in `task_action_test.rb`). Add a clarifying comment like the existing command-column note.
- [ ] Review-triage keyboard's 2-wide layout (Accept all / Reject all on one row, Show details on the next — `notification_builders.rb` `keyboard_for_actions`) is asserted only via a flattened label check (`notification_builders_test.rb:345-355`); pin the un-flattened row structure so a regression to one-per-row or all-on-one-row fails.

## Coverage
- Cross-reviewer corroboration promoted the two High findings (correctness+testing on the already-complete path; adversarial+reliability on the daemon-gating mismatch).
- project-standards returned clean: wiki discipline (RowActions documented as a row in `wiki/modules/bot.md`, two `log.d` fragments, `wiki/log.md` untouched), stage-literal guard annotations correct, schema/logger parity intact, rubocop clean.
- Verified independently: `8-finalize` is the correct coding finalize stage dir; `render_details` is reached (not the `format == :json` branch); daemon `project_enabled?` and bot `daemon_enabled_for?` read the identical `daemon.enabled` key; `approve` is absent from `ALLOWED_VERBS` so the pre-existing ready-row Approve button (in-process) is unaffected by the High #1 daemon-block.
- Not independently runtime-tested: actual Telegram callback redelivery behavior; live daemon-off end-to-end stall.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — telegram-bot row-actions (Pass 2)

Scope: `git diff origin/main..HEAD` (merge-base `54fd3455`). Personas run: code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer, code-simplifier. Findings below are the deduplicated union; persona prefix shows the surfacing lens (corroborating personas noted inline).

Verification note: code-reviewer and pr-test-analyzer claimed the `approve_plan` `ArgumentError` escapes to the poll loop as a silent `:fatal`. I verified against `callback_handlers.rb` that `CallbackHandlers#handle` wraps the dispatch in `rescue ArgumentError` → reply "Bot got confused - please retry from /queue." + `callback_malformed` log (matching silent-failure-hunter F3 and comment-analyzer F2). The High item below reflects that verified outcome (misleading message, not silence).

## High
- [ ] [code-reviewer] (corr: silent-failure, comment-analyzer) `approve_plan`/`rerun` buttons silently no-op on daemon-disabled projects — the verbs are queue-routable (`develop`/`finalize`/`run` ∈ `DispatchRequestQueue::ALLOWED_VERBS`) and the daemon blocks them for projects where `daemon.enabled != true` (`dispatcher.rb:1279` `project_disabled`, same key as the bot's `daemon_enabled_for?` and `project_enabled?` at `dispatcher.rb:1702`); the `plan_waiting`/ready pushes only appear when daemon-disabled (`suppress_daemon_plan_pause?`/`suppress_ready_action?`), so on the only surface that shows them the tap enqueues, gets no ack (`supervisor.rb:1014`), and is dropped — while `approve_plan` has already flipped the marker `:waiting→:complete`, then the `:complete` row re-pushes Approve → infinite silent loop; recoverable only from the CLI.
- [ ] [pr-test-analyzer] (corr: code-reviewer, silent-failure, comment-analyzer) `approve_plan` mishandles the realistic stale/already-advanced plan row and the test masks it: a `:complete` plan row carries `action: ready_to_develop` + `suggested_command: "hive develop …"` (`plan_complete` command is `develop`), which `PlanApproval.rewrite_to_develop` rejects with `ArgumentError` (requires `argv[1]=="plan"`); `approve_plan` rescues only `NotApprovable`/`SystemCallError`/`IOError`, so it falls to `handle`'s generic `rescue ArgumentError` → "Bot got confused - please retry from /queue." (retry re-fails), and `PlanApproval`'s race-safe `:complete` no-op branch is unreachable from the bot. `test_approve_plan_callback_allows_already_complete_marker` (`callback_handlers_test.rb:373-384`) hardcodes an impossible fixture (`:complete` row with `action:"needs_input"` + `"hive plan …"`) and asserts develop IS dispatched — certifying behavior production never produces.

## Medium
- [ ] [silent-failure] (corr: code-reviewer, reliability) Per-row isolation rescues catch only `ArgumentError, KeyError`, and `current_notifications` runs the three `suppress_*` predicates + `fingerprint(row)` OUTSIDE the rescue (`notification_dispatcher.rb:60-86`); a forward-skew/legacy row raising `NoMethodError`/`TypeError` aborts the whole notification tick into an unattributed `:fatal source=status_loop` (all pushes dropped, recurring every tick). `status_action_button` (`supervisor.rb:1454`) has the same narrow rescue and can drop the entire `/status` reply. Widen to also catch `NoMethodError, TypeError` and move the suppress/fingerprint calls inside the per-row rescue with row attribution.
- [ ] [silent-failure] (corr: comment-analyzer) `approve_plan`'s unrescued `ArgumentError` (develop-shaped command, or `Markers.current` hitting invalid UTF-8 in a corrupt `plan.md`) is misattributed: it lands in `handle`'s `:callback_malformed` log keyed on the (well-formed) callback string with no project/slug/stage, and replies "retry from /queue" — unactionable for a corrupt-state condition. Add a local `rescue ArgumentError` in `approve_plan` logging row identity + a "open it on a laptop; retry won't fix it" reply.
- [ ] [type-design] `Resolution#primary` papers over a missing invariant — `actions.find(&:primary) || actions.first` (`row_actions.rb:34-36`) tolerates zero or many primaries despite the "exactly one" comment; a future resolver path forgetting `primary: true` silently makes the first action primary. Enforce `count(&:primary) == 1` (for non-empty actions) in `initialize` and drop the fallback.
- [ ] [type-design] (corr: silent-failure F4, comment-analyzer F3, code-simplifier) The closed `ROLES` validation protects only the construction site, not the three external role-keyed tables (`label_for_action` notification_builders.rb:472, `status_action_emoji` supervisor.rb:1476, `next_step_hint` supervisor.rb:1521); adding a role touches 4 sites, and `next_step_hint`'s catch-all `else` silently degrades a new primary role to the generic "open on a laptop" hint while the other two `.fetch` loudly — and no test invokes `next_step_hint`. Consolidate into one role-keyed presentation table next to `ROLES`, and make the hint fetch-backed/raise.
- [ ] [pr-test-analyzer] `button_coverage_test.rb:14-23` hand-lists 2 of its 4 axes, narrowing the headline "every (action, marker, stage, workflow)" guarantee: `WORKFLOWS = %w[coding content]` isn't derived from `Registry.all` (a new workflow escapes the sweep), and `MARKERS` omits `manual_steering` (a `Markers::KNOWN_NAMES` entry), so every `action × manual_steering` combination is never tested. Derive both from the closed registries.

## Nit
- [ ] [silent-failure] (corr: comment-analyzer F4) `needs_input` still ends in `else default_needs_input` (`notification_builders.rb`), narrowing but not closing the silent-fallthrough the `kind` tag was meant to eliminate — a new `KINDS` entry renders generic text with no failing test; raise on an unhandled needs-input kind or add a test binding each kind to a non-default branch.
- [ ] [pr-test-analyzer] Review-triage 2D keyboard layout (Accept all/Reject all on one row, Show details on the next — `keyboard_for_actions`) is asserted only via flattened label checks (`notification_builders_test.rb:345-355`); a regression to one-per-row/all-on-one-row passes. Assert `keyboard[0]`/`keyboard[1]` row shapes directly.
- [ ] [pr-test-analyzer] (corr: code-reviewer) Parity test's brainstorm cell is now tautological — the legacy arm calls production `brainstorm_action` (`task_action_dispatch_parity_test.rb:28`), the same method the kind-dispatch resolves, so brainstorm `:none→generic_ready_to_run` is no longer cross-checked here (it is independently pinned in `task_action_test.rb`). Restore an inline expectation or add a clarifying comment like the existing command-column note.
- [ ] [type-design] `suppress` boolean is redundant with `kind == :suppressed` (`row_actions.rb:23,60`) — only `suppress` is read, `:suppressed` is write-only, and contradictory states (`suppress:true, kind::brainstorm_waiting`) are representable. Drop one, or document `suppress` as canonical.
- [ ] [type-design] `Action` enforces `role` membership but not `verb`/`callback` coherence — a nil `verb` on `:rerun`/`:run` would render "Next: tap Re-run to re-run ." (`supervisor.rb:1523`) and `callback` accepts nil; low live risk (single construction site) but a make-illegal-states-unrepresentable gap.
- [ ] [code-simplifier] (corr: type-design) `:run` vs `:rerun` roles duplicate the same `rerun:` callback + handler for the three paused-stage re-run buttons, differing only in label/hint already derivable from `Action.verb`; collapse to one role (preserving the existing Run/Re-run wording) to remove a row from each presentation table.
- [ ] [code-simplifier] `keyboard_for_actions` repeats `button(label_for_action(a), a.callback)` across both branches (`notification_builders.rb:460-470`); map actions→buttons once, then arrange rows.
- [ ] [code-simplifier] (corr: code-reviewer) `RowActions`↔`NotificationBuilders` is bidirectional — `RowActions` delegates classification predicates (`recovery?`/`manual_only?`/`verb_for_action`/`retryable_recovery?`) back to `NotificationBuilders` via a lazy accessor that exists only to dodge a require cycle (`row_actions.rb:192-194`); move the pure classification predicates into `RowActions` to break the cycle and make "single source of truth" literal.
- [ ] [code-simplifier] `execute_action`'s explicit `when :none → generic_ready_to_run` (`task_action.rb`) is redundant with its non-terminal `else` (`:none ∉ TERMINAL_MARKER_NAMES`); fold it, or keep for cross-method symmetry with `brainstorm_action`/`finalize_action` (which genuinely need it).
- [ ] [code-simplifier] `result_stage` guards `respond_to?(:stage)` (`supervisor.rb`) though `stage` is now a declared `Result` member, so the guard is always true; droppable. `status_action_emoji`'s `findings_reject: "✅"` entry is unreachable (never primary) and exists only to mirror `label_for_action` — dissolved by the table consolidation above.

## Coverage / agreement
- Both High items reproduce independently across three pr-review-toolkit personas each, and match the prior ce-code-review pass.
- comment-analyzer flagged the `approve_plan` "resurfaces driveable" / `plan_waiting` "tap Approve in chat" comments as overclaiming for the daemon-OFF target (the same root as High #1) — fix the code or qualify the comments.
- Strong parts confirmed by the personas: the cross-surface `status_action_button` vs resolver consistency assertion and the fail-closed new-`TaskActionKind` guard in `button_coverage_test.rb`; the `--from` vs `--stage` argv pins; `PlanApproval`'s validate-before-flip invariant pinned (at the unit layer in `plan_approval_test.rb`); the `Data`-based immutable value objects; `# coding-scoped:` annotations and `8-finalize` correct.
- code-simplifier explicitly recommends NOT changing the `["hive","status","--json"]` sentinel argv (pre-existing convention shared by `/status` and `/queue`), so it is omitted here despite being a readability wart.

ivankuznetsov:

# Code Review — telegram-bot: every announced needs-input row carries a working button or is suppressed (Pass 3)

Scope: `git diff origin/main..HEAD` (merge-base `54fd3455`). 14 commits; `lib/hive/bot/*` (new `row_actions.rb`), `lib/hive/task_action.rb`, `lib/hive/workflows/coding.rb`, `lib/hive/daemon/plan_approval.rb`, schema/logger, wiki, and tests.

Intent: one canonical `RowActions.resolve(row)` feeding both the push keyboard and the `/status` keyboard so they can't drift; new `approve_plan:`/`rerun:` chat-dispatch callbacks; in-process "Show details"; reclassify `none`/`complete` markers away from phantom `needs_input` (U6).

Pass-3 context: the latest 5 commits (`ba62b464`…`24e0c1be`) are review-fix passes that resolved every auto-fixable Medium/Nit from pass 2 (kind-tag dispatch, fail-loud `needs_input`/`next_step_hint`, widened per-row `rescue StandardError`, declared `row_actions` require, collapsed `:run`→`:rerun`, realistic already-advanced `approve_plan` fixture, corrupt-state rescues, fail-closed `WORKFLOWS`/`MARKERS` in `button_coverage_test`). This pass verifies those fixes are regression-free and surfaces residual items. Verified independently: all 608 `bot/` unit tests, 83 `task_action`, the dispatch-parity test, and 12 `plan_approval` tests pass; rubocop clean on all 8 changed lib files. Review team: ce-correctness (opus), ce-adversarial (opus), plus manual integration tracing (`build_router` snapshot wiring, `coding.rb` `handler:` dispatch, `TERMINAL_MARKER_NAMES`, `ACTIONS` keys).

## High
- [ ] CARRYOVER (still open, pending operator decision — escalations-02 A1 unanswered): `approve_plan:`/`rerun:` dispatch queue-routable verbs (`develop`/`finalize`/`run`) that the daemon **blocks** for daemon-disabled projects (`dispatcher.rb` `project_disabled` gate, `daemon.enabled != true`) — the exact daemon-OFF surface these pushes appear on (`suppress_daemon_plan_pause?`/`suppress_ready_action?` only push when daemon OFF). No commit in this pass addresses it; the diff adds no bot-side gate or in-process fallback. `approve_plan` flips the on-disk marker `:waiting→:complete` before that possibly-dropped dispatch. Re-surfaced so it stays in the pass-3 gating loop; resolution depends on the operational fact (does the daemon-OFF deployment drain the dispatch queue) that drove the original escalation. Mitigation already in tree: on daemon-OFF the flipped `:complete` row reclassifies to `ready_to_develop` and re-pushes Approve (recoverable, not stranded — confirmed via `plan_action`/`verb_for_action`).

## Medium
- (none — the pass-2 Mediums were all auto-fixed and verified regression-free this pass.)

## Nit
- [ ] `status_action_emoji` (`supervisor.rb:1492-1501`) and the `next_step_hint` `case` (`supervisor.rb:1505-1535`) omit the `:findings_reject` role, while `label_for_action` (`notification_builders.rb:392-404`) and `RowActions::ROLES` (`row_actions.rb:67`) include it: the closed-vocabulary parity the pass-2 fail-loud `.fetch`/`raise KeyError` was meant to guarantee is incomplete in two of the four role tables. Safe only because `review_waiting` makes `findings_accept` (never `findings_reject`) the primary; if that invariant ever changes, the loud `KeyError` is swallowed by the surrounding `rescue StandardError` and silently drops a `/status` button. Add `findings_reject` to both tables for true parity.
- [ ] `execute_action`'s broadened `else` (`task_action.rb:426-431`) routes any unexpected **non-terminal** marker to `:generic_ready_to_run` (a runnable `READY_TO_RUN`). No real execute-stage marker reaches it today (the runner only stamps `execute_waiting`/`execute_complete`/`execute_stale`/`:none`; universal markers are intercepted earlier), but it trades the phantom-`needs_input` U6 removed for a future phantom-*runnable*: a later execute-stage input-gate marker added without an explicit `when` silently classifies as ready_to_run instead of failing loud. Consider routing unknown non-terminal execute markers to `:error` (loud) like the terminal branch already does.
- [ ] Pre-existing (flagging for awareness, not introduced here): `NotificationBuilders` compaction `@registry` is a plain Hash mutated by the status_loop thread (`compact_callback`/`prune_registry!`) and read by the poll_loop thread (`resolve_callback`) with no mutex. The new `approve_plan:`/`rerun:` callbacks always exceed Telegram's 64-byte cap, so they always depend on this registry. Worst case under heavy callback churn is a benign FIFO-eviction-as-expired ("That button expired — reopen /queue") rather than a crash, so no action is required for this task; noted because this change increases reliance on the registry.

## Coverage
- No new High/Medium defects in the changed code. The single High is the unresolved pass-2 escalation, carried forward unchanged; the fix commits introduced no regressions.
- Adversarially probed and confirmed SAFE: callback payload abuse / `split_callback` arity (garbled → `ArgumentError` → "Bot got confused"; expired `#`-token → `:callback_expired` clean reply); `approve_plan` validate-before-flip ordering (corrupt `plan.md`/bad verb raise BEFORE the marker flip — pinned by `test_approve_plan_keeps_waiting_marker_and_replies_corrupt_on_malformed_command`); already-advanced/`:complete` develop-shaped row dispatches a single well-formed develop with no duplicate `--project` (`test_approve_plan_callback_dispatches_develop_for_already_advanced_row`); `status_row` matches the full `(project, slug, stage)` triple and `prepare` re-reads the live marker, so a stale snapshot can't mis-dispatch; the per-row `rescue StandardError` bodies only call `Data.define` accessors and can't re-raise; fail-loud `needs_input`/`next_step_hint` raises are unreachable for production rows (every `needs_input` kind has an explicit arm; `findings_reject` is never primary).
- The testing gaps the sub-agents speculated about are already covered: `test_rerun_callback_dispatches_{develop,finalize,generic_run}_*` pin the argv shapes (`--from` for develop/finalize, `--stage` for generic run), and `button_coverage_test` (3 runs, 42k assertions) enforces the cross-surface actionability guarantee.
- The `/status` Approve-plan button appearing on a daemon-ENABLED 3-plan pause (cross-surface suppression divergence) was raised by ce-correctness at P3 but is DROPPED: the plan explicitly scopes daemon-ON suppression as unchanged/out-of-scope, and pass-2 already adjudicated it RESOLVED/NO-FIX (race bounded by the daemon per-slug in-flight gate + idempotent `:complete` no-op + validate-before-flip).

ivankuznetsov:

# PR Review Toolkit — telegram-bot: every announced needs-input row carries a working button or is suppressed (Pass 3)

Scope: `git diff origin/main..HEAD` (merge-base `54fd3455`), 14 commits. Personas run: code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier. This is pass 3 (the latest 5 commits are review-fix passes); all targeted suites are green and rubocop is clean, so no lint/test-failure findings. Already-known items excluded per task: the daemon-disabled `approve_plan`/`rerun` dispatch viability (escalated, pending), the `hive status --json` in-process sentinel argv, and the `/status` Approve-on-daemon-ON divergence (plan scopes daemon-ON out of scope).

## High
- (none — no persona surfaced a High-severity defect in the changed code.)

## Medium
- [ ] [silent-failure] False "✅ Recovered" self-heal: if a previously-alerted recovery/error row later **deterministically** fails to build (a persistently-malformed `attrs`/workflow value raising in `build_notification`), the per-row `rescue` drops it from `current_notifications`, so `process_recoveries` sees its stored fingerprint as absent, waits out the grace window, sends `recovered_message` and clears the alert — a still-broken task gets a false "Recovered" ping and goes silent (`notification_dispatcher.rb` `current_notifications`↔`process_recoveries`); only the `:notification_build_failed` log remains. Narrow trigger (builds once, then malforms) but it is exactly the "silently lost" outcome on the recovery path.
- [ ] [type-design] `Resolution`/`Action` construction-time validation is bypassable: `Data.define`'s positional constructor (`Action.new(:rerun,"cb",true,nil)`) and the `.with` copy method both skip the keyword-form `initialize`, so role/kind/callback/verb/one-primary checks are silently skipped — the "illegal states unrepresentable" contract holds by caller discipline, not by the type (`row_actions.rb:30`, `:71`). Latent today (all construction goes through the keyword helpers; nothing calls `.with`); hardening = override `self.new`/`with` to re-run validation.
- [ ] [comment-analyzer] `rerun_label`/`rerun_hint` comments justify the verb→label split semantically — "Re-run once an agent has already run the stage … Run for stages that have not yet completed (finalize, generic run)" — but a `finalize` rerun is only produced for `finalize_waiting`, which means the finalize agent **did** run and paused; the real rule is purely lexical (`develop`→"Re-run", everything else→"Run") (`notification_builders.rb` `rerun_label` ~:406; `supervisor.rb:1538-1540`). Risk: a maintainer "fixes" finalize to "Re-run" on the false premise that the label tracks execution history.

## Nit
- [ ] [code-reviewer][code-simplifier] `result_stage(result)` is a pure pass-through (`result.stage`) with a single caller and no guard/transform since the fix pass stripped its `respond_to?` check; inline `result.stage` at the call site (`supervisor.rb:1549-1551`, called at `:976`).
- [ ] [pr-test-analyzer][type-design] `status_action_emoji` (`supervisor.rb:1473`) omits `:findings_reject` even though it is in `RowActions::ROLES` (`row_actions.rb:67`); safe only because `findings_reject` is structurally never the primary action, and no test pins the gap — if a future refactor promoted it to primary, `.fetch(:findings_reject)` would raise and be swallowed by `status_action_button`'s `rescue StandardError` (silently dropped `/status` button). `label_for_action` does not share the gap (maps all 8 roles).
- [ ] [type-design] `Resolution.new(actions: [], kind: :plan_waiting)` — empty actions with a needs-input kind — is representable: the exactly-one-primary invariant only fires for non-empty actions, so an empty needs-input resolution slips construction and would make `needs_input` silently drop the push via its `actions.empty?` guard (`row_actions.rb:35-39`; `notification_builders.rb:135`). No current resolver produces it; the "a needs-input kind must carry ≥1 action" rule is real business logic left unencoded.
- [ ] [code-simplifier] `next_step_hint`'s terminal-`:details` early return (`supervisor.rb:1512`) and the `when :details` arm (`:1525-1529`) emit the identical "open on a laptop" string, so the early return is redundant for this method's output (author kept it deliberately so the hint and `button_coverage_test`'s `terminal_details?` guard agree — defensible to leave).
- [ ] [code-simplifier] `execute_action`'s explicit `when :none` arm (`task_action.rb:422-431`) duplicates the `else` non-terminal fallthrough — `:none` is not in `TERMINAL_MARKER_NAMES`, so the `else` already returns `generic_ready_to_run`; redundant (explicit-over-compact), and asymmetric with `brainstorm_action`/`finalize_action` where the `:none` arm IS load-bearing.
- [ ] [silent-failure] U6 `execute_action` `else`: an impossible-but-present non-terminal foreign marker (e.g. a hand-edited `:review_waiting` on an execute row) now classifies as `generic_ready_to_run` (a runnable command) rather than surfacing as an anomaly; the dangerous terminal half is correctly routed to `:error`. Defensive-classification drift, not a regression — the `else` no longer has any "unrecognized" surface (every unknown non-terminal is presumed runnable).
- [ ] [silent-failure] The new per-row `rescue StandardError` blocks log with `row.project/slug/marker/action` unguarded (`notification_dispatcher.rb:69-71,96-98`; `supervisor.rb:1467-1469`); if the **row object** (not its attrs) were malformed so a top-level reader raised, the log line would re-raise inside the rescue and escape to the tick's `:fatal` rescue — the very abort the rescue claims to prevent. Narrow (real `StatusWatcher::Row`/`LegacyStageDirs` always respond); capturing the attrs into locals before the `begin` would close it.
- [ ] [pr-test-analyzer] `supervisor_test.rb:10` exercises `render_details`/`next_step_hint`/`status_action_button` against a local `Row = Struct.new(...)` rather than the production `StatusWatcher::Row`; works today (code only calls `respond_to?`/readers the Struct satisfies) and the real Row is covered by `row_actions_test`/`button_coverage_test`, but a future Row reader-with-logic change wouldn't surface in these specific tests.
- [ ] [comment-analyzer] `inferred_success_slug` comment (`supervisor.rb:1390-1394`) enumerates the two callers (Show details / `/details`) of the in-process `hive status --json` path by name; if a third caller ever emits that argv through the child-spawn path the comment becomes a false guarantee (the `""` return would drop a real slug). Phrase it as "any plain `hive status --json` is rendered in-process" without binding to the named callers.

## Coverage / strengths (no action)
- **No High-severity defect from any of the six personas.** The fix commits genuinely close the headline silent-failure risks (corrupt `plan.md`, marker-write failure, per-row build crash — each now logs a typed event AND surfaces an actionable reply instead of dying in the poll loop's `:fatal` rescue).
- pr-test-analyzer verified the U7 guard is **non-vacuous** via two mutations: adding a `TaskActionKind` to the schema alone fails the `ALL`-equality check; adding it to `expected` without a `RowActions` surface fails the actionable-resolution check (the `MARKERS - RECOVERY_MARKERS` exclusion is what stops every kind resolving vacuously via the error-marker recovery path). The `(action × marker × stage × workflow)` sweep is real (42,316 assertions) and `MARKERS`/`WORKFLOWS` derive from the closed registries.
- type-design rated `Resolution` and `Action` strong (8/8/9/7 and 8/8/8/7); deriving `suppress` from `kind == :suppressed` and the exactly-one-primary enforcement both genuinely remove contradictory-state classes.
- comment-analyzer confirmed the large structural comments are accurate: the `approve_plan` advance-then-resurface / validate-before-flip / rescue-scope block, the `row_actions.rb` require-cycle note, the `:callback_expired` 64-byte note, the `rerun` `--from` "bypasses the source-stage terminal-marker gate" claim (verified `develop`/`finalize` expose only `--from` and the target-stage path skips `validate_marker!`), and all four `# coding-scoped:` stage-literal annotations.
- code-simplifier confirmed the per-surface label/emoji/hint split is intentional and not reducible, and `dispatch_stage_verb`/`ensure_flag!`/the `RowActions` helpers are appropriately factored.

PR #589, opened 2026-06-24
feat(bot): accept task ids for slash commands

ivankuznetsov:
## Summary

Every Telegram push notification and `/status` row identifies a task by its `#NNNN` id, but the four slash commands that act on a task (`/answer`, `/approve`, `/autofix`, `/details`) only accepted the date+hex slug — a string the operator never sees. Targeting a task from your phone meant deriving a slug you were never shown. These commands now accept the `#NNNN` id directly (with or without the `#`), so what you read in a notification is what you can type back. Slug arguments keep working unchanged.

A numeric target (`#?\d+`) is resolved against the bot's current `StatusWatcher` snapshot — the same in-memory view the operator is looking at — and rewritten to the task's slug before each command's existing action path runs. Anything else is treated as a slug exactly as before (real slugs always carry letters and dashes, so there is no ambiguity).

Resolving against the live snapshot rather than letting the CLI resolve an id on disk is the deliberate choice here. It keeps three properties the on-disk path could not:

- Archived or scrolled-out ids are untargetable — they are not in the snapshot, so they cannot be acted on by accident.
- An id miss echoes the requested id (`No active task #9281 — was it archived?`) instead of a generic CLI error.
- No wrong-project dispatch: the id maps to one globally-unique slug before anything runs, so there is no guessed project. The slug path keeps its existing cross-project ambiguity refusal.

| Target typed | Resolution |
|---|---|
| `9281` / `#9281` | Snapshot row with `id == 9281` → its slug |
| `stuck-260525-abcd` (slug) | Matched verbatim, unchanged behavior |
| `9281` not in snapshot | `No active task #9281 — was it archived?`, no dispatch |
| id before snapshot loads | `Status is still loading — try again in a moment.` |

The shared `resolve_status_row` resolver gained the id branch, so `/details` and `/autofix` needed no call-site change; `/approve` and `/answer` gained a one-line pre-resolution step. Usage hints, `/help`, the Telegram command descriptions, and the voice-answer hint now read `<id|slug>`.

## Test plan

- `bundle exec rake test TEST=test/unit/bot/slash_handlers_test.rb` — green
- `bundle exec rubocop` on all changed Ruby files — clean
- `bundle exec rake coverage` (100%-line gate) — passed, Lines 100.00%

[...]

PR #591, opened 2026-06-24
fix: suppress incoherent needs_input notifications

ivankuznetsov:
## Summary

Operators were getting nonsensical Telegram alerts like `#9281 Brainstorm / Needs input: none` and `#5 Execute / Needs input: complete`. Internal marker names were leaking into push copy for rows that aren't actually waiting on anyone. A row can wear the `needs_input` action while carrying a marker that is not a real input state (`none`, `complete`), and every operator surface trusted the action blindly: the bot dumped the raw marker, `hive status` grouped the row under "Needs your input", and the TUI offered an answer editor.

This adds one shared allowlist — `Hive::Markers.input_marker?`, true only for `waiting`, `execute_waiting`, and `review_waiting` — and applies it at each operator boundary. Incoherent rows are dropped or hidden; genuine ones get operator copy with no raw marker text.

The bug was reported on the bot, but `hive status` and the TUI read the same rows, so fixing only the bot would leave them inconsistent. All three consume the one predicate, so a future stray marker on a `needs_input` row is handled the same way everywhere.

### What changed

| Surface | Behavior |
|---|---|
| `Hive::Markers` | New `INPUT_MARKER_NAMES` allowlist plus `input_marker?` / `incoherent_needs_input?` predicates (String/Symbol tolerant): the single source of truth. |
| Telegram dispatch | Incoherent `needs_input` rows are dropped before the alert is built, logged as `notification_skipped_incoherent` (project/slug/stage/marker/action), and never enter the alert dedupe store. |
| Telegram copy | Genuine states reworded without raw markers: `execute_waiting` shows "Execute paused — needs your input.", the generic fallback shows "This task needs your input."; both keep the existing Show details button. |
| Telegram `/status` & `/queue` | Pull replies filter incoherent rows through the same predicate, matching proactive suppression. |
| `hive status` (text) | Incoherent rows are excluded from the "Needs your input" group and the `unanswered_questions` count, and collapse into a muted "N tasks hidden" breadcrumb pointing at `--json`. |
| `hive tui` | The visible projection drops incoherent rows (with a hidden-count breadcrumb), and `KeyMap` defensively treats them as non-actionable so Enter won't open an answer editor. |

### Why the JSON contract stays put

`hive status --json` still emits `action: "needs_input"` for these rows. The daemon, web, and the bot's own `StatusWatcher` consume that field, so rewriting it would ripple far beyond this bug. The predicate is applied at each human render boundary instead. Re-classifying *why* a `none`/`complete` row gets the `needs_input` action upstream is left to the umbrella task that owns the marker-to-action map (#9512).

## Test plan

Regression coverage is seeded from the real `bot.log` evidence (22 `none` rows, 1 `complete`, 4 `execute_waiting`). All suites pass and `rubocop` is clean on every touched lib file.

- `ruby -Itest test/unit/markers_test.rb` — predicate accepts the 3 real markers, rejects `none`/`complete`/terminal/unknown; normalization (nil, uppercase) pinned; `INPUT_MARKER_NAMES ⊆ KNOWN_NAMES` invariant.
- `ruby -Itest test/unit/bot/notification_dispatcher_test.rb` — `none`/`complete` suppressed and logged, `execute_waiting` passes (asserted by text, not count), suppressed rows kept out of the alert store, gate-ordering vs active-conversation pinned.
- `ruby -Itest test/unit/bot/notification_builders_test.rb` — reworded copy plus an invariant that no produced notification text contains `none`/`complete`/`execute_waiting`.
- `ruby -Itest test/unit/bot/logger_test.rb` — `notification_skipped_incoherent` registered in the event allowlist.
- `ruby -Itest test/unit/bot/supervisor_test.rb` — `/status` & `/queue` pull replies suppress incoherent rows.
- `ruby -Itest test/unit/commands/status_test.rb` / `test/integration/status_test.rb` — text omits incoherent rows (and "no active tasks" when they are the only rows), JSON keeps the action and reports `unanswered_questions: 0`, no raw marker text leaks to the CLI surface.
- `ruby -Itest test/unit/tui/key_map_test.rb` / `test/unit/tui/snapshot_test.rb` — projection and keymap honor the predicate (`none` and `complete` cases); the raw snapshot still preserves every row for `--json` consumers.

Manual: captured a demo of `hive status` hiding the two incoherent rows behind a muted breadcrumb while `hive status --json` still emits all four `needs_input` rows (GIF kept local; Screenote not connected for hosted stills).

## Review summary

Three compound-engineering review passes (claude-ce + codex-ce + pr-review-toolkit personas) plus orchestrator grounding. **No High findings**; every plan Requirements-Trace item was verified present and tested, and the deferred root-cause tasks (#9512/#9400/#9399/#9401) were correctly left untouched.

Auto-fixed across passes: a standalone-load `NameError` in `incoherent_needs_input?` (now loads `Hive::Schemas`), the silent-removal "no active tasks" all-clear (added a hidden-incoherent count breadcrumb mirroring the archived-hidden summary), supervisor `/status`/`/queue` gap, shared-predicate extraction to remove triplication, comment-accuracy corrections, and added coverage (normalization, gate-ordering, evidence-matrix text, drift-guard, CLI no-leak).

Two open questions were escalated (plan's own Q3/Q4) and resolved under courageous mode via the plan's recommended defaults: genuine `finalize_waiting`/`plan_waiting` recovery rows that surface as `marker=none` are suppressed like any other `none` (proper fix needs a dedicated marker → #9512), paired with the discoverability breadcrumb so a stuck task is never a silent all-clear. The `release.yml` / `CHANGELOG.md` churn in the two-dot diff is a stale-merge-base artifact (origin/main #584/#583 not yet picked up); the branch's own commits never touch those files and a merge-base-aware merge keeps main's versions.

## Linked task

- Task **#9398 — Fix Telegram Needs Input Alerts** (`telegram-bot-needs-input-none-260624-2c96`)
- Deferred root-cause / follow-up tasks referenced but intentionally out of scope: **#9512** (re-classify why `none`/`complete` reach the `needs_input` action), **#9400** (contradictory marker-vs-action root cause), **#9399** (enrich the Show-details payload), **#9401** (debug/noise severity-log model for the new suppression event).

---


ivankuznetsov:

## High
- [ ] `.github/workflows/release.yml:280` reintroduces a non-starting macOS/Colima smoke job: `macos-15` is a macOS arm64 hosted runner and `colima start` needs nested virtualization, so this release job will fail before smoking the published image.

## Medium
- [ ] `lib/hive/bot/supervisor.rb:1482` leaves `/status` and `/queue` outside the incoherent `needs_input` boundary: `actionable_queue_rows` only filters `archived`/`agent_running`, so `needs_input` rows with `marker=none` or `marker=complete` still render in Telegram pull replies instead of being suppressed like proactive notifications.

## Nit

ivankuznetsov:

# Code Review — telegram-bot-needs-input-none (pass 1)

**Scope:** `git diff origin/main..HEAD` — feature commits `cb6bf444..9008d6c8` (U1–U5 + wiki).
**Intent:** Stop the Telegram bot leaking nonsensical operator alerts (`Needs input: none` / `Needs input: complete`). Add a shared `Hive::Markers.input_marker?` allowlist (`waiting`/`execute_waiting`/`review_waiting`); the bot suppresses+logs incoherent `needs_input` rows and rewords genuine ones without raw marker text; `hive status` and the TUI stop presenting incoherent rows as pending questions. JSON `--json` `action` contract must stay stable.
**Reviewers:** correctness, maintainability, testing, api-contract (+ orchestrator verification: tests run, rubocop clean).

**Grounding performed by orchestrator:** all 6 changed test files run green (markers 34 / dispatcher 49 / builders 42 / status 67 / key_map 124 / snapshot 26, 0 failures); `rubocop` clean on all 7 touched lib files; JSON `action` contract verified preserved (status.rb:287 emits all rows; only text-render + TUI projection filter); `finalize_action` (task_action.rb:420-424) verified to emit `needs_input`+`marker=none` for a real recovery state.

The core fix is correct, well-tested, and contract-safe. No High (must-fix) findings: plan-required scope (U1–U5) is implemented, deferred items (9512/9400/9399/9401) are correctly untouched. Findings below are a plan-acknowledged tradeoff that deserves the user's open-question confirmation (Q3/Q4/Q5), plus minor maintainability/coverage items.

## High

_(none)_

## Medium
- [ ] **`finalize_waiting` (marker=none) genuine recovery state is suppressed at every human surface** (`lib/hive/markers.rb:39`, `lib/hive/task_action.rb:420-424`): `finalize_action` returns `needs_input`+`marker=none`+`command=finalize` when `pr.md` exists but the marker was cleared (e.g. after a `hive markers clear` / StaleAgentHealer on `unpushed_commits`); the allowlist treats this as incoherent, so the bot suppresses the push, `hive status` text drops the row, and the TUI removes it. The plan **explicitly accepts this** (R2 / open-question Q3, deferring the proper fix to 9512) — but its premise ("no finalize case" in the evidence) is empirically contradicted by existing tests (`test_finalize_waiting_when_pr_md_exists`, `daemon_stale_agent_healing_test.rb:241`), and because `suppress_incoherent_needs_input?` is **not** daemon-gated, daemon-*disabled* projects lose the only signal to re-run `hive finalize`. Surface for the user's Q3 confirmation.
- [ ] **TUI `key_map` `input_marker?` Enter-guard is unreachable under real wiring** (`lib/hive/tui/key_map.rb:209-210, 306-308`): two reviewers agree the `enter_message` `needs_input` branch only receives rows from `Snapshot#visible_projection`, which now strips incoherent rows via `without_incoherent_needs_input` (snapshot.rb:272) before key handling; the ternary's false branch is dead in production and only the unit test (which bypasses the projection) exercises it. Either drop the guard+helper and rely on the snapshot strip, or add a comment marking it explicit defense-in-depth so a future reader doesn't treat it as load-bearing.
- [ ] **TUI fully removes incoherent rows, diverging from plan U5's "keep visible, neutral handling"** (`lib/hive/tui/snapshot.rb:238-244, 269-274`): U5 specified guarding the key handler so an incoherent row "falls through to the same neutral handling other non-actionable rows get" (still visible); the implementation instead removes the row from the visible projection entirely. Combined with the status text removal, a genuinely-stuck task (e.g. the finalize-none case above) vanishes from both human surfaces with **no log/audit trail** there (only the bot logs `notification_skipped_incoherent`; status/TUI drop silently). Relates to open-questions Q4/Q5 — confirm full removal (vs a muted diagnostic line) is intended for the live TUI dashboard.
- [ ] **Untested branch: `hive status` prints "no active tasks" when a project's only rows are incoherent** (`lib/hive/commands/status.rb:445`): the guard was changed from `visible_rows.empty?` to `render_rows.empty?`, so a project whose sole task is an incoherent `needs_input` row now reports "no active tasks"; the new test always keeps a visible genuine `waiting` row, so this new behavioral branch is never exercised. Add a `render_project` test with a single `marker=none`/`needs_input` row asserting the "no active tasks" output (and slug absent).

## Nit
- [ ] **Composite "incoherent needs_input" rule re-coded in 3 surfaces** (`lib/hive/commands/status.rb:465-467`, `lib/hive/tui/snapshot.rb:294-296`, `lib/hive/bot/notification_dispatcher.rb:95-97`): only the leaf `input_marker?` was shared; the `action == NEEDS_INPUT && !input_marker?` composite is independently re-implemented at each boundary. Plan R3 explicitly accepts applying the predicate per-boundary, but a single `Hive::Markers.incoherent_needs_input?(action:, marker:)` would remove the triplication and the drift risk the file's own `TERMINAL_MARKER_NAMES` comment warns about.
- [ ] **`Markers.input_marker?` nil/uppercase normalization paths untested** (`lib/hive/markers.rb:172-174`, `test/unit/markers_test.rb`): the predicate does `name.to_s.downcase.to_sym` but tests only feed lowercase symbols/strings; add `assert input_marker?("WAITING")` and `refute input_marker?(nil)` to pin the advertised normalization contract.
- [ ] **Gate-ordering interaction untested (incoherent row + active conversation)** (`lib/hive/bot/notification_dispatcher.rb:65-66`): `suppress_active_conversation?` runs before `suppress_incoherent_needs_input?` and short-circuits, so an incoherent row that also has an active conversation logs only the active-conversation skip; no test pins this (active-conversation tests all use `marker:"waiting"`). Add a `marker:"none"`+active-conversation case asserting exactly one skip event, or document the scenario as unreachable (incoherent rows never offer "Answer in chat").
- [ ] **Schema prose drift for `unanswered_questions`** (`schemas/hive-status.v4.json` field description): `unanswered_question_count` now force-returns 0 for incoherent `needs_input` brainstorm rows, but the field's description still describes it purely as a brainstorm count "0 for every other row"; not a contract break (type/range unchanged, no machine consumer reads the prose — daemon/web re-parse `brainstorm.md` directly), but the prose under-describes when 0 is emitted.
- [ ] **`needs_input` `else → default_needs_input` arm is now defensive-only** (`lib/hive/bot/notification_builders.rb:139-140`): the dispatcher suppresses incoherent markers before `build`, so the case's `else` arm no longer fires for them (still reachable via `waiting_input` fallthrough). Harmless layered defense; flagging only because the allowlist knowledge now lives in two layers (suppression gate + builder case) and a future reader may be unsure which is authoritative.
- [ ] **Diff carries unrelated `release.yml`/`CHANGELOG.md` churn from a stale merge-base** (`.github/workflows/release.yml`, `CHANGELOG.md`): the branch forked before the arm64-smoke fix landed on `origin/main`, so the diff *appears* to revert it; the branch's own commits never touch these files, so a normal 3-way merge will keep main's version (no actual reversion). Not a defect — rebasing on `origin/main` would clean the review scope and avoid the false alarm.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — telegram-bot-needs-input-none (pass 1)

**Scope:** `git diff origin/main..HEAD` — feature commits `cb6bf444..9008d6c8` (U1–U5 + wiki). The `.github/workflows/release.yml` / `CHANGELOG.md` portions of the diff are a stale merge-base artifact (the branch never touches them; a 3-way merge keeps main's version) and were excluded from review.
**Personas run:** code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier (all in parallel).
**Intent:** Stop the Telegram bot leaking nonsensical alerts (`Needs input: none`/`complete`); shared `Hive::Markers.input_marker?` allowlist (`waiting`/`execute_waiting`/`review_waiting`); bot suppresses+logs incoherent `needs_input` rows and rewords genuine ones without raw marker text; `hive status` + TUI hide incoherent rows; JSON `action` contract stays stable. Root causes deliberately deferred to tasks 9512/9400/9399/9401.

**Consensus:** No Critical issues. The allowlist is correct (all `*_waiting` actions are driven by allowlisted markers in the canonical path), the JSON `action` contract is preserved, and `input_marker?` is robust to String/Symbol/case/nil. Findings cluster on (a) the diagnosability asymmetry between the bot (logs suppressions) and status/TUI (drop silently), (b) a plan-acknowledged tradeoff worth Q3 confirmation, and (c) test-coverage + maintainability polish.

## High

_(none)_

## Medium
- [ ] [pr-test-analyzer] Status "only-incoherent rows → `no active tasks`" branch is untested (`lib/hive/commands/status.rb:441-450`): the guard changed from `visible_rows.empty?` to `render_rows.empty?`, creating a new state where a project with only `marker=none`/`complete` rows prints "no active tasks"; the sole status-text test always keeps a coherent row, so a regression to a blank "Needs your input" header or wrong emptiness guard would pass green.
- [ ] [silent-failure-hunter] `hive status` text and the TUI projection remove incoherent rows with **no log/audit trail** (`lib/hive/commands/status.rb:441,465`; `lib/hive/tui/snapshot.rb:238,294`): unlike the bot (which emits `notification_skipped_incoherent`), the two human surfaces drop rows silently and, worst case, print "no active tasks" — a misleading all-clear; recommend a debug-level log or an on-surface "N incoherent rows hidden — see `hive status --json`" hint to match the bot's audit standard.
- [ ] [silent-failure-hunter] `finalize_waiting` (`marker=none`, `command=finalize`) is a genuine recovery state suppressed at all three surfaces (`lib/hive/task_action.rb:420-424` ↔ `lib/hive/markers.rb:39`): the gate is **not** daemon-gated, so daemon-*disabled* projects lose the only re-run prompt, and status/TUI drop it with no audit trail — plan-acknowledged tradeoff (R2 / open-question Q3, root cause → 9512); surface for the user's Q3 confirmation.
- [ ] [pr-test-analyzer, code-reviewer] Gate-ordering (incoherent row + active conversation) is untested (`lib/hive/bot/notification_dispatcher.rb:65-66`): `suppress_active_conversation?` precedes `suppress_incoherent_needs_input?`, so such a row logs `notification_skipped_active_conversation` and never `_incoherent`; both still suppress (no leak), but a reorder would silently flip the audit event with no failing test — add a `marker:"none"`+active-conversation case asserting exactly one skip event.

## Nit
- [ ] [code-reviewer] `test_unanswered_question_count_degrades_to_zero_on_parse_error` is now de-fanged (`test/unit/commands/status_test.rb:726-736`): the new guard `return 0 unless input_marker?(row[:marker_name])` (`status.rb:309`) returns 0 *before* the `BrainstormParser.parse` stub runs, so the `rescue StandardError` path it was written to pin is no longer exercised; add `marker_name:"waiting"` (+ `workflow`/`stage`) to restore coverage.
- [ ] [pr-test-analyzer] `input_marker?` nil/uppercase normalization is unpinned (`lib/hive/markers.rb:172-174`): tests only feed lowercase symbols/strings, but on-disk markers are uppercase and the `nil`→false guard is load-bearing; add `assert input_marker?("WAITING")` and `refute input_marker?(nil)`.
- [ ] [pr-test-analyzer] `without_incoherent_needs_input` has no direct test and its kept-set fixture omits `review_waiting` (`test/unit/tui/snapshot_test.rb:527-531`): coverage runs only through `visible_projection`; add `review_waiting` to the kept fixtures and/or a direct method test to harden against a future caller.
- [ ] [pr-test-analyzer] `unanswered_questions` JSON value is asserted only for the `none` row, not the `complete` row (`test/unit/commands/status_test.rb:267-272`): low-value, but pinning the `complete` incoherent row at 0 would fully lock the gate.
- [ ] [type-design, code-simplifier, code-reviewer] The composite `action == NEEDS_INPUT && !input_marker?(marker)` invariant is re-coded at 3 boundaries (`status.rb:465`, `snapshot.rb:294`, `notification_dispatcher.rb:96`): plan R3 explicitly accepts applying the predicate per-boundary, but a shared `Hive::Markers.incoherent_needs_input?(action:, marker:)` would collapse the triplication and the drift risk; reviewers split on whether it is worth forcing given the differing row shapes.
- [ ] [type-design] No asserted `INPUT_MARKER_NAMES ⊆ KNOWN_NAMES` relationship (`lib/hive/markers.rb:39`): a boot-time subset assertion (also for `TERMINAL_MARKER_NAMES`) would make an illegal/non-parseable allowlist entry unrepresentable at load rather than discoverable only via a silently-suppressed row.
- [ ] [type-design] The daemon's `brainstorm_answers_pending?` (`lib/hive/daemon/dispatcher.rb:~856`) is a 4th `NEEDS_INPUT` boundary that does not consult `input_marker?`: this is intentional per the plan (Q2/R3 keep the daemon/web on the stable JSON `action` contract) — flag only to confirm the divergence is the intended scope boundary.
- [ ] [comment-analyzer] `lib/hive/markers.rb:37` cites "(see umbrella task 9512)" which is unresolvable from within the repo: it is consistent with the plan's umbrella-task id, so likely valid — consider citing a repo-resolvable issue/PR or keeping the self-contained explanation so a future reader isn't blocked.
- [ ] [comment-analyzer] `lib/hive/markers.rb:35-38` says "legacy classification paths" but `none`/`complete` reaching `needs_input` is the current, live config-driven mapping in `task_action.rb`, not deprecated code — consider "config-driven classification paths" to avoid implying dead code.
- [ ] [comment-analyzer] Wiki completeness: `wiki/log.d/20260624T224650Z-input-marker-boundary.md` omits `review_waiting` from the "still notify" set, and `wiki/modules/bot.md` ("logged as `notification_skipped_incoherent`") doesn't note the active-conversation precedence edge — both minor accuracy touch-ups.
- [ ] [code-simplifier] `lib/hive/bot/notification_dispatcher.rb:119` uses the bare string `"needs_input"` while the new/adjacent gates use `Hive::Schemas::TaskActionKind::NEEDS_INPUT` (`:87`,`:96`): pre-existing inconsistency, trivial consistency win if the file is touched.
- [ ] [code-simplifier] `execute_waiting_input` / `default_needs_input` / `plan_waiting` are now near-identical (`header + sentence + Show-details keyboard`, `notification_builders.rb:159-201`): an optional `details_only(row, sentence)` helper would dedupe the three-times-repeated keyboard construction.
- [ ] [silent-failure-hunter] `unanswered_question_count` rescues bare `StandardError` → 0 (`lib/hive/commands/status.rb:319`, pre-existing): the new `input_marker?` guard changes which rows reach the disk read; consider narrowing the rescue to `IOError`/parse classes so unrelated bugs aren't masked as "0 unanswered" when 9512 is addressed.
- [ ] [code-simplifier, silent-failure-hunter] The TUI key_map Enter guard (`lib/hive/tui/key_map.rb:209-210`) is unreachable in the live path because `Snapshot#visible_projection` pre-strips incoherent rows before the cursor can land on one: keep as defense-in-depth, but add a one-line comment noting it is redundant with `without_incoherent_needs_input` so a future reader doesn't treat it as load-bearing.

ivankuznetsov:

## High
- [ ] `.github/workflows/release.yml:289` reintroduces Colima on hosted `macos-15`: `macos-15` is arm64 and GitHub documents arm64 macOS runners as lacking nested virtualization, so the post-publish smoke job is expected to fail before `packaging/docker/smoke.sh` runs.

## Medium
- [ ] `lib/hive/markers.rb:186` new shared predicate depends on `Hive::Schemas` without loading it: `require "hive/markers"; Hive::Markers.incoherent_needs_input?(...)` raises `NameError`, making the exported marker-boundary helper unsafe for standalone marker consumers.

## Nit

ivankuznetsov:

# Code Review — telegram-bot-needs-input-none (pass 2)

**Scope:** `git diff origin/main..HEAD` — feature commits `cb6bf444..8273f5de` (U1–U5 + the pass-1 auto-fix commits `7b5d3f6c` refactor, `6c839fe6` tests, `a054436a` `details_only` dedup, `8273f5de` docs).
**Intent:** Stop the Telegram bot leaking nonsensical operator alerts (`Needs input: none` / `complete`). Shared `Hive::Markers` allowlist (`waiting`/`execute_waiting`/`review_waiting`); bot suppresses+logs incoherent `needs_input` rows and rewords genuine ones without raw markers; `hive status`, TUI, and bot `/status`/`/queue` inherit the rule. JSON `--json` `action` contract stays stable.
**Reviewers:** correctness, maintainability, testing, api-contract (+ orchestrator grounding below).

**Grounding performed:** all 7 changed test files run green (markers 38 / builders 42 / dispatcher 50 / supervisor 142 / status 68 / key_map 124 / snapshot 27, 0 failures); `bot/logger` 11 and `daemon/dispatcher` 132 regression suites green; `rubocop` clean on all 8 touched lib files. JSON `action` contract verified preserved (incoherent rows stay in `--json` with `action:"needs_input"`; only text-render, TUI projection, and bot pull-replies filter). `unanswered_questions` semantic change verified safe (daemon/web compute counts from `BrainstormParser` directly, not the JSON field; sole JSON consumer — the TUI — strips incoherent rows). All TUI surfaces (`key_map`, `bubble_model` auto-continue at `bubble_model.rb:387`) resolve rows via `visible_projection`, so they inherit the filter.

**Verdict:** The pass-1 auto-fixes are correctly applied and behavior-preserving; plan-required scope (U1–U5 + supervisor pull-replies) is fully implemented and the deferred items (9512/9400/9399/9401) remain correctly untouched. No High findings. One **new** Medium was introduced by the pass-1 shared-predicate extraction (corroborated by codex pass-2). The two pass-1 escalations (Q1/Q2) remain open — code unchanged, `escalations-01.md` `A1`/`A2` still blank — and are re-surfaced below. The stale-base `release.yml`/`CHANGELOG.md` churn is already in `suppressed.md` (no-fix) and is intentionally NOT re-raised.

## High

_(none)_

## Medium
- [ ] **`Hive::Markers.incoherent_needs_input?` raises `NameError: uninitialized constant Hive::Schemas` when loaded standalone** (`lib/hive/markers.rb:186`): introduced by the pass-1 refactor (`7b5d3f6c`) that centralized the composite into `markers.rb`, which `require`s only `securerandom`. Empirically confirmed: `ruby -Ilib -e 'require "hive/markers"; Hive::Markers.incoherent_needs_input?(action:"needs_input", marker:"none")'` → `NameError`, whereas the sibling `input_marker?` works standalone. In-app it never fires (every caller loads `require "hive"`, which defines `Hive::Schemas` in `lib/hive.rb`), but the method is explicitly advertised as the single cross-surface source of truth for the marker boundary, so a future consumer requiring just `hive/markers` hits a latent load-order trap that the sibling predicate doesn't have. Fix by ensuring `Hive::Schemas` is loaded before the constant reference (or compare the action without the `Hive::Schemas::TaskActionKind` constant). Corroborated independently by codex pass-2.
- [ ] **[carry-over from pass 1 — still open, `escalations-01.md` A1 blank] `finalize_waiting` (marker=none) genuine recovery state is suppressed at every human surface** (`lib/hive/markers.rb:40`, `lib/hive/task_action.rb:419-424`, `:123-127`): re-verified — `finalize_action` returns `ACTIONS[:finalize_waiting]` (key `NEEDS_INPUT`, command `finalize`) when `pr.md` exists but the marker was cleared (`hive markers clear` / StaleAgentHealer on `unpushed_commits`), producing `action=needs_input` + `marker=none`. The allowlist classes `none` as incoherent, so the bot suppresses the push, `hive status` text drops the row, and the TUI removes it. Plan R2/Q3 explicitly accepts this (defer root cause to 9512), but `suppress_incoherent_needs_input?` is not daemon-gated, so daemon-*disabled* projects lose the only signal to re-run `hive finalize`. Pending the user's Q3/A1 decision.
- [ ] **[carry-over from pass 1 — still open, `escalations-01.md` A2 blank] Incoherent rows are fully removed from `hive status` text and the TUI with no audit trail** (`lib/hive/commands/status.rb:441-450`, `lib/hive/tui/snapshot.rb:238-244,269-274`): unlike the bot (which logs `notification_skipped_incoherent`), the two human surfaces drop rows silently; a project whose only task is incoherent now prints "no active tasks" — and `test_text_status_reports_no_active_tasks_when_only_incoherent_rows` pins that misleading all-clear. This also diverges from plan U5's "keep visible, neutral non-actionable handling." Plan Q4 poses full-removal vs muted-diagnostic-line as a genuine open question; A6 wants suppressions diagnosable. Pending the user's Q4/A2 decision (a hidden-incoherent-count hint mirroring `hidden_old_archived_count`, pointing at `--json`, would honor A6 cheaply).

## Nit
- [ ] **Diagnosability asymmetry between bot and CLI/TUI suppression** (`lib/hive/bot/notification_dispatcher.rb:95-104` vs `lib/hive/commands/status.rb:465-467`, `lib/hive/tui/snapshot.rb:294-296`): the bot path emits a full `notification_skipped_incoherent` audit event per suppression, but the status/TUI filters carry no equivalent breadcrumb (the row survives only in `--json`). Lower-severity facet of the open Q2/A2 decision above; folded here so it isn't lost if Q2 is resolved as "keep full removal." (Pre-existing suppressed Nits — the bare-`StandardError` rescue in `unanswered_question_count`, the unresolvable 9512 citation, and the daemon `brainstorm_answers_pending?` 4th boundary — are intentionally not re-raised; they are in `suppressed.md`.)

ivankuznetsov:

# PR Review (pr-review-toolkit, pass 2) — telegram-bot-needs-input-none

**Scope:** `git diff origin/main..HEAD` (feature files only; `release.yml`/`CHANGELOG.md` are stale-merge-base artifacts already triaged no-fix in pass 1 and were excluded by all personas).
**Personas run:** code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier. All 7 changed test files run green; rubocop clean on all 8 lib files (verified by multiple personas).
**Union of findings below.** Strong cross-persona convergence on one High theme (over-suppression of genuine `needs_input`+`marker=none` recovery rows). This theme overlaps the still-open pass-1 escalations Q1/Q2 (`escalations-01.md` `A1`/`A2` still blank), now re-surfaced at higher severity with concrete `plan`/`finalize` grounding. Severity is reported as the persona(s) rated it; where personas disagreed, the split is noted.

## High
- [ ] [code-reviewer + silent-failure-hunter] **Genuine, operator-actionable `needs_input`+`marker=none` recovery rows are now suppressed at every human surface** (`lib/hive/task_action.rb:36-39` `plan_waiting` key=NEEDS_INPUT cmd `plan`, `:123-127` `finalize_waiting` key=NEEDS_INPUT cmd `finalize`, `:413-424`; suppressed via `lib/hive/markers.rb:185-188` at `commands/status.rb:441`, `tui/snapshot.rb:240`, `bot/supervisor.rb:1489`, `bot/notification_dispatcher.rb:96`): a plan/finalize task whose marker was cleared/interrupted but whose `plan.md`/`pr.md` exists resolves to `action=needs_input`+`marker=none` (confirmed: `finalize_action` falls through to `:finalize_waiting`; `plan_action` to `:plan_waiting` when artifact present); the allowlist treats `none` as incoherent and drops the row from `hive status`, the TUI, and bot `/status`/`/queue`, so the operator loses the only `hive finalize`/`hive plan` re-run signal. On `origin/main` these render. Overlaps still-open pass-1 Q1/A1; the plan's R2 only reasoned about finalize as "not a pending question" and never anticipated the `plan` case or the CLI/TUI *visibility* loss for an actionable command.
- [ ] [silent-failure-hunter] **Suppression is not daemon-gated, so daemon-disabled projects have no fallback path** (`lib/hive/bot/notification_dispatcher.rb:95-106`, and the ungated `commands/status.rb:441` / `tui/snapshot.rb:240` / `supervisor.rb:1483`): unlike `suppress_daemon_plan_pause?` (`:117-122`) which gates on `daemon_enabled_for?`, the incoherent gate suppresses unconditionally; a daemon-enabled project can still auto-resume via `EDIT_RESUME_ACTIONS`, but a daemon-OFF project (a documented mode) has no daemon and no operator surface, leaving the H1 finalize/plan row fully invisible and un-resumable.
- [ ] [silent-failure-hunter] **Misleading "no active tasks" / "No active Hive tasks." all-clear with no hidden-count breadcrumb** (`lib/hive/commands/status.rb:445-447`, `lib/hive/bot/supervisor.rb:1404`, `lib/hive/tui/snapshot.rb:238-244`): incoherent rows are excluded from `render_rows` but are NOT counted in `hidden_rows`, so a project whose sole task is incoherent prints a clean empty state with no "N hidden" footnote (unlike `render_archived_hidden_summary`) — trading a noisy-but-true alert for a quiet-but-false all-clear. Overlaps still-open pass-1 Q2/A2.

## Medium
- [ ] [type-design-analyzer + code-reviewer] **`Hive::Markers.incoherent_needs_input?` raises `NameError` when loaded standalone** (`lib/hive/markers.rb:186` references `Hive::Schemas::TaskActionKind::NEEDS_INPUT` but the file only `require`s `securerandom`): empirically `require "hive/markers"; Hive::Markers.incoherent_needs_input?(...)` → `NameError: uninitialized constant Hive::Schemas`, while the sibling `input_marker?` works — a module-boundary defect in a helper whose own doc-comment advertises broad cross-surface reuse. **Severity split:** type-design rated High (the doc invites a standalone `require` then crashes); code-reviewer rated Nit (matches a pervasive pre-existing pattern; never fires in shipped paths since every caller loads `hive`). Reported here as Medium. One-line fix: `require "hive"` (or the file defining `Hive::Schemas`) at the top of `markers.rb`.
- [ ] [type-design-analyzer] **Stringly-typed `action:`/`marker:` boundary fails silently on the status Hash shape** (`lib/hive/commands/status.rb:466` passes `action: row[:action_key]`): the two `Data.define` row shapes raise `NoMethodError` on a wrong accessor (fail-fast), but the status plain Hash returns `nil` for a mistyped key (`row[:action]` → `nil`), which the predicate treats as "not needs_input → coherent," silently disabling suppression — the weakest shape is the unprotected one.
- [ ] [pr-test-analyzer] **The 4 surviving `execute_waiting` pushes are asserted by count only, not text** (`test/unit/bot/notification_dispatcher_test.rb:217-237`): the evidence-matrix test asserts `telegram.messages.size == 4` but never inspects message text, so a regression that suppressed correctly but re-leaked `execute_waiting` or dropped the rewording on the integrated dispatch path would still pass; the "rephrased, not suppressed" acceptance criterion is only half-verified end-to-end. Add `assert m[:text].include?("Execute paused")` / `refute m[:text].include?("execute_waiting")`.
- [ ] [comment-analyzer] **`INPUT_MARKER_NAMES` comment misattributes/understates where incoherent rows originate** (`lib/hive/markers.rb:35-39`): the comment says other markers reach `needs_input` "through the config-driven classification paths (see umbrella task 9512)," but live in-repo producers include the *hardcoded* `plan_action`/`finalize_action` (`task_action.rb`) emitting `marker=none`+NEEDS_INPUT — i.e. exactly the genuine actionable states in the High finding — which the comment does not mention; the project's own `wiki/modules/markers.md` describes this more honestly as "Historical/upstream classification."
- [ ] [comment-analyzer] **`else`-arm comment cites an impossible fall-through example** (`lib/hive/bot/notification_builders.rb:144-146`): the comment calls the `case row.marker` `else` "layered defense for the remaining genuine fall-through (a `waiting` marker outside coding's 2-brainstorm/3-plan stages)," but a `waiting` marker is caught by `when "waiting"` (`:133`) → `waiting_input` and can never reach the `else`; the parenthetical example describes a path that cannot occur here (the surrounding "incoherent path no longer reaches here" claim is accurate — only the example is wrong).
- [ ] [type-design-analyzer] **Three row-shape field conventions converge on one predicate with no shared adapter** (`markers.rb:185-188` consumed at `notification_dispatcher.rb:66`, `supervisor.rb:1489`, `status.rb:466`, `snapshot.rb:294`): `.action`/`.action_key`/`[:action_key]` and `.marker`/`.marker`/`[:marker_name]` are each re-spelled per surface; absence of a thin `MarkerRow` adapter reintroduces the drift risk the file's own `TERMINAL_MARKER_NAMES` comment warns about. **Counter-view:** code-simplifier judged the per-surface wrapper methods acceptable (they mirror adjacent established helpers like `hide_archived_row?`/`without_old_archived`).

## Nit
- [ ] [pr-test-analyzer] **`execute_waiting` arm of the no-leak test is trivially green** (`test/unit/bot/notification_builders_test.rb:284-308`): nothing on that path interpolates the marker, so `refute_includes "execute_waiting"` could never fail regardless of the lib; the `none`/`complete` arms are meaningful, the `execute_waiting` arm adds no discriminating power.
- [ ] [pr-test-analyzer] **`refute_match(/\bwaiting\b/, …)` is an over-broad copy guard** (`test/unit/bot/notification_builders_test.rb:230,255`): asserting the word "waiting" never appears in operator copy is correct today but would fail a legitimate future sentence ("…is waiting on you"); brittleness watch-point, not a defect.
- [ ] [pr-test-analyzer] **Gate-ordering test is implementation-coupled by design** (`test/unit/bot/notification_dispatcher_test.rb:331`): pins `suppress_active_conversation?` running before the incoherent gate by asserting the exact skip-event list; legitimate (both gates suppress; only the audit trail differs, and the inline comment says so) but will fail on a harmless reorder — noted so it isn't mistaken for a behavioral assertion.
- [ ] [silent-failure-hunter] **`unanswered_question_count` bare `rescue StandardError → 0`** (`lib/hive/commands/status.rb:319-320`): the new `input_marker?` guard (`:309`) only narrows which rows reach the disk read; the broad rescue (pre-existing, not worsened by this diff) still masks permission/parse/encoding errors as "0 unanswered" with no log. Optional: log the swallowed exception. (Matches a pass-1 suppressed Nit.)
- [ ] [comment-analyzer + code-reviewer] **Orphaned tracker reference "(see umbrella task 9512)"** (`lib/hive/markers.rb:36`): "9512" is unresolvable from the repo (no commit/doc/wiki/PR, no system prefix), yet it is the only pointer to the real cause of incoherent rows; qualify it or inline the explanation. (Matches a pass-1 suppressed Nit.)
- [ ] [type-design-analyzer] **`input_marker?` silently absorbs arbitrary unknown input** (`lib/hive/markers.rb:174`): `INPUT_MARKER_NAMES.include?(name.to_s.downcase.to_sym)` maps any garbage/typo'd constant to `false` with no signal; correct & safe default for suppression, flagged only because a typo'd marker name elsewhere degrades silently to "not input."
- [ ] [type-design-analyzer] **`INPUT_MARKER_NAMES ⊆ KNOWN_NAMES` enforced by test, not by the type** (`lib/hive/markers.rb:40` vs `test/unit/markers_test.rb:198-207`): the lowercase-symbol vs uppercase-string split means the subset can only be checked at test time; the test is good defense — Nit only given the repo's low-ceremony preference.
- [ ] [code-simplifier] **Comment proportionality: defensive/explanatory blocks larger than the code they guard** (`lib/hive/bot/notification_builders.rb:140-146` 7-line `else` comment over a 1-line call; `:166-169` `details_only` docstring; `lib/hive/tui/key_map.rb:210-215` 6-line defense-in-depth comment over a ternary): each could compress without losing intent; `details_only` extraction itself is a good dedup (`plan_waiting` 6→1 lines) — keep it.
- [ ] [code-simplifier] **`key_map.rb` `input_marker?(row)` is a single-use wrapper inconsistent with its neighbors** (`lib/hive/tui/key_map.rb:312-314`, called once at `:216`): could read `Hive::Markers.input_marker?(row.marker)` inline; low-confidence — leave it if the team prefers uniform per-surface predicates (the snapshot/status wrappers are precedent for keeping it).

ivankuznetsov:

## High
- [ ] `.github/workflows/release.yml:286` switches the post-publish arm smoke to `macos-15` + `colima start`: GitHub-hosted arm64 macOS runners do not support nested virtualization, so this replaces the native Linux arm64 smoke with a job that fails tagged releases.

## Medium
- [ ] `CHANGELOG.md:21` removes existing 0.3.1 release-note details: the release workflow publishes this section as GitHub release notes, so this drops documented security/integration/fix items from the next release artifact.

## Nit

ivankuznetsov:

# Code Review — telegram-bot-needs-input-none (pass 3)

**Scope:** `git diff origin/main..HEAD`. Branch's own work (three-dot `origin/main...HEAD`,
merge-base `9efbca2`) touches `lib/hive/{markers.rb, bot/*, commands/status.rb, tui/*}`,
two versioned schemas, tests, and wiki. **Note:** the two-dot diff also shows
`.github/workflows/release.yml` and `CHANGELOG.md` being *reverted* — those are
`origin/main` commits #584 and #583 the branch has not picked up (it is behind), not
changes this branch authored (see first Medium finding).

**Mode:** report-only (no files mutated). Reviewers: correctness, adversarial, testing,
maintainability, project-standards, api-contract, cli-readiness, agent-native, learnings.

**Verdict:** Ready with minor fixes. The change is correct, complete against the plan
(every Requirements-Trace item verified present and tested), and well-tested — all 8 touched
unit files pass (517 assertions, 0 failures) and `rubocop` is clean on the 8 changed lib files.
Load is cycle-free (verified `require "hive/markers"` standalone). No High-severity defects.
The implementation sensibly extends the plan's U1–U5 to the bot pull-reply path (supervisor)
and TUI snapshot projection to honor A5's "all surfaces" mandate.

## High

_None. No plan-required item is missing; no correctness/security/data defect found._

## Medium
- [ ] Branch is stale vs `origin/main` — rebase before merge: the literal review target `git diff origin/main..HEAD` reverts `release.yml` (#584, the arm64-native release-smoke fix that replaced the broken macOS/colima job) and `CHANGELOG.md` (#583); a squash/naive apply from this diff re-breaks the release smoke job. A normal merge-base-aware merge is safe, but rebasing onto origin/main removes the risk and makes the review diff accurate.
- [ ] Incoherent-row suppression is silent on every surface except the bot proactive-push path — only `NotificationDispatcher` logs `notification_skipped_incoherent`; `supervisor.rb:1483` (`/status` & `/queue`), `status.rb:441` (text render), and `snapshot.rb:238` (TUI) drop rows with no log event and no summary line, so a genuinely stuck incoherent task becomes undiscoverable from any human surface (unlike archived rows, which get a count line). Plan left this as Open Q4; consider a muted "N rows in an incoherent state (see `--json`/bot.log)" line and/or extending A6's "log every suppression" to the pull/render paths.
- [ ] `lib/hive/markers.rb:2` adds `require "hive"` (the 640-line top-level entrypoint) solely to read the `NEEDS_INPUT` string constant — wrong dependency direction for a leaf vocabulary module that 25+ files require independently. Safe today only because `hive.rb` has no requires of its own; inline the literal `"needs_input"` (closed frozen enum, comment the canonical constant) or require a narrower schemas path.
- [ ] Future-marker drift is unguarded: a new genuine input marker routed to `action=needs_input` but omitted from `INPUT_MARKER_NAMES` would be silently suppressed at every surface — the hardest failure to notice. The existing `test_input_and_terminal_marker_names_are_subsets_of_known_names` only checks the subset direction, not that every `needs_input`-producing `TaskAction` marker is in the allowlist; add an invariant test tying the two.

## Nit
- [ ] `lib/hive/bot/supervisor.rb:7` uses `Hive::Markers.incoherent_needs_input?` but pulls `Hive::Markers` only transitively (via `require "hive/bot/notification_dispatcher"` at line 16) — add an explicit `require "hive/markers"` so a future require reorder can't `NameError`.
- [ ] U3 invariant ("no raw marker text leaks") is asserted for the bot builder but not the `hive status` text surface — add `refute_includes out, "none"`/`"complete"` to `test_text_status_excludes_incoherent_needs_input_rows_but_json_contract_stays_stable` (`status_test.rb`).
- [ ] TUI `key_map` incoherent guard is exercised only for `marker:"none"`; add a `marker:"complete"` case for the other incoherent variant (`key_map_test.rb`).
- [ ] Per-surface wrapper names drift — `incoherent_needs_input?` (status.rb) vs `incoherent_needs_input_row?` (snapshot.rb) vs `input_marker?` (key_map.rb) for the same concept; align for grep-ability (the underlying shared predicate is correctly single-sourced).
- [ ] For agents reading `hive status --json`, `unanswered_questions: 0` cannot distinguish "no pending questions" from "incoherent/suppressed"; the schema description names the three markers but consider a one-line machine-readable hint (or an additive `incoherent` boolean) so consumers needn't re-derive the predicate. JSON `action` field is correctly kept stable.
- [ ] `wiki/index.md` "Updated" date is ~2 days stale relative to the new `wiki/log.d` fragment (cosmetic — no new catalog page, so the new-page rule isn't triggered).
- [ ] `key_map.rb:213` defense-in-depth comment ("Kept as a redundant guard exercised by the unit test") reads as circular justification; reword to state it protects a future caller that bypasses `visible_projection`.

---

_Suppressed below confidence gate / demoted to coverage: lowercase-`execute_waiting` not asserted
in the composite-predicate test (conf 50); nested-loop structure in the no-leak invariant test makes
some cross-row assertions trivially pass (conf 50); no isolated-`complete` case in
`without_incoherent_needs_input` (covered transitively). Pre-existing, untouched: `review_waiting`'s
`fix_guardrail` branch still interpolates `marker_with_attrs` (a recognized input marker, so not the
leak class this task targets)._

ivankuznetsov:

# PR Review (pr-review-toolkit) — telegram-bot-needs-input-none (pass 3)

**Scope:** `git diff origin/main..HEAD`; personas reviewed the branch's own work
(three-dot `origin/main...HEAD`, merge-base `9efbca2`). Six personas run: code-reviewer,
pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer, code-simplifier.

**Overall:** No correctness/security defect; every plan Requirements-Trace item is implemented
and tested; the touched suite passes and rubocop is clean. code-reviewer and code-simplifier found
the change well-factored. The findings below are observability, drift-guard, comment-accuracy, and
test-completeness items (union of all personas). Disagreements are noted inline.

## High

_None. No plan-required item is missing; no behavioral defect found._

## Medium
- [ ] [comment-analyzer] `lib/hive/markers.rb:37-40` mis-attributes the `complete`→`needs_input` path to `TaskAction#plan_action`/`#finalize_action`: those route `marker==:complete` to `:plan_complete`/completion actions (verified, `task_action.rb:413`), so they never emit `complete`+needs_input. The real path is `execute_action`'s `else` branch (`task_action.rb:407-410`, matching the plan's `#5` evidence). Comment is self-contradictory ("…classify any non-complete plan/finalize row as NEEDS_INPUT") and duplicated in `wiki/modules/markers.md` + the `log.d` fragment — fix all three; will mislead the 9512 follow-up.
- [ ] [type-design/pr-test-analyzer/silent-failure] No producer-side tripwire ties `TaskAction`'s genuine `needs_input`-producing markers to `INPUT_MARKER_NAMES`; the existing `test_input_and_terminal_marker_names_are_subsets_of_known_names` only checks `INPUT_MARKER_NAMES ⊆ KNOWN_NAMES` (wrong direction). A future genuine input marker wired to `needs_input` but forgotten in the allowlist would be silently suppressed at every surface — the exact bug class inverted. Add a test enumerating `needs_input`-producing markers and asserting each is `input_marker?` (or in an explicit incoherent set).
- [ ] [silent-failure-hunter] `lib/hive/commands/status.rb:441-448`: when a project's only task is incoherent, `hive status` prints "no active tasks", masking a present-but-stuck task and breaking the file's own convention (archived/legacy hidden rows get a summary count line; incoherent rows get nothing). Mirror the archived-hidden pattern with a muted "N task(s) hidden — needs_input with no answerable marker; see `--json`" line. (Plan Q4 left presentation open; code-reviewer judged the suppression acceptable — surfaced for the user to decide.)
- [ ] [silent-failure-hunter] Suppression is un-logged on three of four surfaces — `supervisor.rb:1482` (`/status` & `/queue`), `status.rb` render, `snapshot.rb` projection drop rows with no event, while only the dispatcher logs `notification_skipped_incoherent`. Plan A6 ("log every suppression") is only half-honored; an operator pulling `/status` sees a task vanish with no `bot.log` trail. `@logger` and the event are already available on the supervisor. (code-reviewer dissents: per-poll pull logging would be noise — note the trade-off; guard against double-logging since `actionable_queue_rows` runs for both `render_queue` and `status_keyboard`.)
- [ ] [type-design-analyzer] `lib/hive/markers.rb:2` `require "hive"` is a layering inversion: the low-level marker-vocabulary module now depends on the top-level entrypoint solely to read `Hive::Schemas::TaskActionKind::NEEDS_INPUT`. Safe today only because `hive.rb` has no requires; inline the `"needs_input"` literal (closed frozen enum) or relocate the composite predicate to a small classification helper, keeping `input_marker?`/`INPUT_MARKER_NAMES` self-contained.

## Nit
- [ ] [type-design-analyzer] `incoherent_needs_input?` normalizes `marker` (`to_s.downcase.to_sym`) but not `action` (`action.to_s == NEEDS_INPUT`, no `.downcase`) — latent asymmetry; safe today since all callers pass lowercase action keys. Normalize both or document the lowercase-string contract.
- [ ] [silent-failure-hunter] `input_marker?` does not strip whitespace (`"  waiting  "` → `:"  waiting  "` → treated incoherent); latent only, since markers are normalized at source via `Markers.current`.
- [ ] [silent-failure-hunter] `notification_dispatcher.rb` orders `suppress_active_conversation?` before `suppress_incoherent_needs_input?`, so an incoherent row that also has an active conversation logs the active-conversation event, not the incoherent one — currently intentional and pinned by a test, but the incoherent reason is the more accurate attribution.
- [ ] [type-design-analyzer] `/details` (`supervisor.rb render_details`) bypasses the coherence gate, so an incoherent row's content can still be pulled up. `render_details` content is explicitly deferred to task 9399, so document the exemption in the `actionable_queue_rows` comment rather than leaving the reader to infer it.
- [ ] [comment-analyzer] `key_map.rb:210-213` "Kept as a redundant guard exercised by the unit test" is circular justification; reword to the defense-in-depth rationale and cross-reference `Snapshot#without_incoherent_needs_input` by name so a future removal of the upstream filter lands here.
- [ ] [comment-analyzer] `notification_builders.rb:140` and `supervisor.rb:1485` "e.g. none/complete" examples present `complete` as an equal incoherent input; lead with `marker=none` and label `complete` as the `execute_action`-path/defensive case to match the corrected `markers.rb` attribution.
- [ ] [type-design-analyzer] Per-surface wrapper names drift — `incoherent_needs_input?` (status.rb) vs `incoherent_needs_input_row?` (snapshot.rb) vs `input_marker?` (key_map.rb) for the same gate; standardize for grep-ability (underlying shared predicate is correctly single-sourced).
- [ ] [pr-test-analyzer] TUI `key_map` U5 guard (`key_map_test.rb:971`) pins only `marker:"none"`; add a `marker:"complete"` case to match the evidence matrix and the predicate's own two-value coverage.
- [ ] [pr-test-analyzer] No test characterizes the deliberate bot-vs-CLI divergence: the bot rewords `execute_waiting` ("Execute paused…") while `hive status` text renders the raw `execute_waiting` label. Add a status-text test asserting a genuine `execute_waiting` row renders, pinning the raw label so a future "reword/suppress everywhere" change is conscious.
- [ ] [pr-test-analyzer] The U3 no-leak invariant is asserted only on the bot builder; add `refute_includes out, "none"`/`"complete"` to the `hive status` text test (`status_test.rb`) so the CLI surface is pinned too.
- [ ] [code-simplifier] `notification_builders.rb:168` `details_only` uses `header(row) + "\n" + sentence`; interpolation `"#{header(row)}\n#{sentence}"` matches the sibling builders in the same file (behavior-identical).
- [ ] [merge-hygiene] Branch is behind `origin/main`: the two-dot `git diff origin/main..HEAD` shows `release.yml`/`CHANGELOG.md` reverted (origin/main commits #584/#583 not yet picked up). A merge-base-aware merge is safe, but rebase before merge to avoid any squash/naive-apply reverting the #584 arm64 release-smoke fix.

---

_Positives (personas): evidence-matrix test pins the exact `{none=>22, complete=>1}` skip distribution
with `.tally` + content assertions (not count-only); gate ordering and alert-dedupe-state are explicitly
tested; the raw TUI snapshot preserves the full JSON contract while only the projection filters;
the logger test structurally scans `lib/hive/bot/**` for unregistered events; `hive status --json`
correctly keeps incoherent rows + `action` field stable (the machine-readable audit trail). The
`details_only` extraction and the `NEEDS_INPUT` constant swap are clean. Deferred scope boundaries
(9512/9400 upstream re-classification, `--json` action stability, 9399 details content) were respected
and not flagged as defects._

ivankuznetsov:
Closing as superseded by #590 (`fix(bot): every needs-input row gets a working button or is suppressed`), now merged.

Both PRs targeted the same "needs input: none" bug, with different strategies:

- **This PR (#591)** suppressed the incoherent rows at the *surface* — a shared `Hive::Markers.incoherent_needs_input?` predicate applied across the bot dispatcher, TUI, and `hive status`.
- **#590** fixed the *root cause* in `Hive::TaskAction`: a markerless (`:none`) or terminal (`:complete`) row no longer classifies as `needs_input` at all (`:none` → `ready_to_run`/`done`, `:complete` → `done`/`ready_to_advance`); only a real `:waiting` input marker maps to `needs_input`. So the incoherent rows this PR was suppressing can no longer be produced.

With #590 merged, this PR's suppression is dead code for the `none`/`complete` cases, and its blanket "`complete` is incoherent" rule was actually too aggressive — it tripped `StatusLegacyLayoutTest` by over-suppressing a coherent row, which #590's nuanced per-stage classification handles correctly.

Net of the bot-notification work also landed via #587, #588, #589, #592. No action needed; the root cause is fixed upstream.

PR #592, opened 2026-06-25
fix(bot): make Show details always render row content

ivankuznetsov:
## Summary

Tapping **Show details** on a Telegram notification was a dead end for the rows
that need it most. For `needs_input`, `plan_waiting`, and `review_waiting` tasks
(none of which are in a red recovery state) the button dispatched
`hive status --diagnose <slug>`, which came back with an empty diagnostic and
replied **"No diagnostic available."** Show details and the `/details` slash
command now render from the cached status row instead, so every tap returns
content: a task summary, a next-step hint tailored to the row's surface, and the
diagnostic appended only when one actually exists.

`NotificationBuilders.details_reply(row)` is now the single source of truth for
the Show-details body, composing three parts:

- **Summary** (always present): title, project/slug, stage, action, marker, and
  attrs — the same baseline `Supervisor#render_details` used to build inline.
- **Next-step hint** (per surface): `plan_waiting` points at the plan file and
  `/approve`; a manual-only recovery row gives the cause sentence, the
  manual-repair hint, and a log path; a `fix_guardrail` review row says to
  inspect on a laptop; a coding brainstorm row says `/answer`; everything else
  falls back to the row's own next-action or suggested command, then to "Open on
  a laptop to advance."
- **Diagnostic** (only when `row.diagnostic` is a non-empty Hash): the cached
  `summary` + `detail`, truncated to stay under Telegram's 4096-character cap.

Three surfaces share this renderer: the inline `details:` callback, the
`/details` slash command, and `Supervisor#render_details` (the `/status`
🔍 Details button). The callback resolves the row from the in-memory status
snapshot by project + slug (with stage to disambiguate duplicate slugs). A stale
button whose task is gone replies "no longer active" rather than dispatching
anything, and a render-time fault soft-degrades to a logged "Status lookup
failed" reply — so the button is never a dead end.

**Design decisions**

- **The escalation signal is in memory, not on disk.** The resolved snapshot row
  already carries the computed `diagnostic` Hash, populated by the same
  `hive status --json` poll (run `with_diagnostic: true`) the bot already
  performs. So "does a diagnostic exist for this row?" is decided at tap time
  with zero filesystem I/O and no subprocess. Trade-off: the appended diagnostic
  is as fresh as the last poll, which matches everything else the bot renders.
- **`refresh_diagnose` is untouched.** Regenerating a fresh agent verdict still
  runs `hive status --diagnose --write --force`. `Supervisor#diagnose_reply_for_child`
  keeps its empty-envelope branch as defense for that path; a comment notes Show
  details no longer routes through it.
- **The callback wire format is byte-identical.** `details:project:slug[:stage]`
  parses exactly as before, so buttons on already-delivered notifications keep
  resolving.

One operator-visible shift: Show details no longer spawns a diagnose subprocess,
so the old two-step ("Diagnosis is available… tap Show details to dump it here")
is gone for these rows. One tap returns content directly. `refresh_diagnose`
remains the way to generate a fresh agent verdict.

## Test plan

- `test/unit/bot/notification_builders_test.rb` covers each surface branch
  (plan-approval pause, brainstorm Q&A, fix-guardrail review, manual-only and
  auto-recoverable recovery, dirty-worktree), the diagnostic append on/off, the
  `next_step_hint` structured-vs-flat precedence, and the truncation cap
  (including the summary-exhaustion guards).
- The callback and slash-handler tests assert `:reply` with no `--diagnose`
  argv, exercise the stale-button / still-loading / render-fault soft-degrade
  paths, and confirm no in-scope path can emit "No diagnostic available."
- `bundle exec rake test TEST='test/unit/bot/**/*_test.rb'` — 600 tests pass.
- `bundle exec rake test TEST='test/integration/bot/**/*_test.rb'` — 29 tests
  pass (1 skip).
- `bundle exec rubocop` — clean on every changed Ruby file.
- `bundle exec rake coverage` — 100.00% lines (6340 runs), gate holds.

## Review summary

Three CE review passes (claude-ce, codex, and the pr-review-toolkit multi-persona
suite) all returned **ready with minor fixes** and found no High/Critical,
security, or contract defect. The "never replies 'No diagnostic available' on any
in-scope Show-details surface" guarantee was independently verified, the callback
wire shape is byte-identical, and `refresh_diagnose` is left intact. Findings were
quality/robustness/coverage items, all auto-fixed across the passes:

- Wrapped the `details_reply(row)` render in a logged soft-degrade rescue on every
  surface, and added `:details_render_failed` / `:details_lookup_failed` log
  events so a snapshot-provider fault is visible in bot.log instead of silently
  degrading.
- Added branch-exercising tests for the truncation guards, the `next_step_hint`
  precedence and `suggested_command`/`-` fallbacks, the plan-path fallback, the
  diagnostic summary-only/detail-only/blank cases, and the actionable
  manual-recovery hints (dirty-worktree, fix-status-check) through the
  Show-details surface.
- Consolidated the shared stage/marker predicates and `sorted_attr_pairs` helper,
  promoted the soft-degrade reply copy to shared frozen constants, introduced a
  small `Diagnostic` Data type with coercing initializer, and reconciled stale
  comments/copy (the `diagnose_reply_for_child` "dump" wording and routing notes).

Recurring stale-base notes (`.github/workflows/release.yml`, `CHANGELOG.md`
appearing reverted under the two-dot review diff) were confirmed phantom: this
branch's `merge-base..HEAD` touches neither file, `origin/main` already runs the
arm64-native smoke (#584), and a standard merge will not regress them. The
Telegram 4096-char cap measured in codepoints vs UTF-16 units was resolved as a
pre-existing transport convention (shared with `Telegram#split_message`), bounded
and logged, not a regression. No reviewer findings were escalated to the user.

## Linked task

- Hive task **9399** — "Fix Dead Show Details Button"
  (`telegram-bot-show-details-button-260624-96cc`).
- Origin: operators tapping Show details on `needs_input` / `plan` / `review`
  notifications hit `hive status --diagnose`, which replied "No diagnostic
  available for <slug>." because the button was attached unconditionally in
  `default_needs_input` / `plan_waiting` / `review_waiting`.

---

Claude_Code-D97757?logo=claude&logoColor=white)


ivankuznetsov:

## High
- [ ] `.github/workflows/release.yml:280` runs the Colima smoke on `macos-15`: GitHub-hosted macOS arm64 runners do not support nested virtualization, so `colima start` can fail before the release image smoke runs.

## Medium

## Nit

ivankuznetsov:

# Code Review — telegram-bot-show-details-button (pass 1)

Scope: `git diff origin/main..HEAD`. Intent: route Telegram "Show details" to an
in-process renderer (`NotificationBuilders.details_reply(row)`) built from the cached
StatusWatcher snapshot row — always returns content (summary + per-surface hint +
optional cached `row.diagnostic`, truncated to 4096 chars) — replacing the
`hive status --diagnose` dispatch that dead-ended non-red rows with "No diagnostic
available". Reviewers: correctness, testing, maintainability, reliability, adversarial,
project-standards.

Verdict: **Ready with minor fixes.** All plan Requirements-Trace items are implemented
and verified (no in-scope path can emit "No diagnostic available"; callback-data wire
shape preserved; `refresh_diagnose` left intact; hot push loop does no FS work). The
adversarial pass found no breakable defect. Findings below are quality/robustness/
coverage improvements, not blockers.

## High
- [ ] _None._ All plan goals are met; adversarial review found no exploitable/breaking defect, and the truncation arithmetic is provably bounded to ≤4096 chars in every branch.

## Medium
- [ ] Truncation guard branches are never exercised (`lib/hive/bot/notification_builders.rb:466` detail-empty early-return, `:469` available-not-positive early-return): the 100% **line**-coverage gate passes because the lines run, but neither branch is ever true in tests, so a regression in these guards (the exact 4096-char overflow protection the plan flags as a risk) would ship green.
- [ ] `details_hint` (`lib/hive/bot/notification_builders.rb:414`) re-implements the same coding stage/marker classification already encoded in `waiting_input`/`needs_input`/`review_waiting` (`:134`–`:216`): the plan_waiting / brainstorm / fix_guardrail tuples now live in two method trees that must stay in sync, or the push notification and the Show-details reply silently disagree after a future stage/marker change.
- [ ] `resolve_details_row` rescue (`lib/hive/bot/handlers/callback_handlers.rb:277`) swallows every `StandardError` without logging, even though `@logger` is injected and already used for `:callback_malformed` (`:56`): a recurring snapshot-provider fault degrades the reply to "Status lookup failed" while leaving nothing in bot.log — observability lost exactly where it is cheap.

## Nit
- [ ] `truncate_diagnostic_reply` (`lib/hive/bot/notification_builders.rb:464`) re-splits the string `diagnostic_detail` just joined with `\n` (`:460`) via `split("\n", 2)`: a lossy join-then-resplit that is only safe today because the canonical producer (`lib/hive/task_action/diagnostic.rb`, `SUMMARY_MAX=120`) guarantees a single-line summary; pass summary/detail through structurally instead.
- [ ] Branch-coverage gaps the line gate can't catch — `next_step_hint` `suggested_command` fallback (`:446`) and `== "-"` sentinel (`:448`), plan-waiting `File.join(folder,'plan.md')` fallback when `state_file` is nil (`:416`), and the manual-only hint with no folder / no "Logs/artifacts:" line (`:426`): each executes but never decides the outcome; add targeted `details_reply` tests.
- [ ] Telegram's 4096 cap is measured in Ruby characters, not UTF-16 code units (`lib/hive/bot/notification_builders.rb:353` and `telegram.rb` `split_message`): emoji/non-BMP-heavy diagnostic detail can be ≤4096 chars but >4096 units → HTTP 400; pre-existing in the transport, bounded (`safe_send_message` rescues, never crashes the poll thread), and low-probability for ASCII log text, but this change newly routes raw agent-authored diagnostic text into that path.
- [ ] `diagnose_reply_for_child` success message "…Tap Show details to dump it here." (`lib/hive/bot/supervisor.rb:1281`) is now slightly stale: after `refresh_diagnose` writes a fresh verdict, Show details renders the *cached* `row.diagnostic` (refreshed only on the next poll) and summarizes rather than "dumps"; `refresh_diagnose` is out of scope so this is wording-only.
- [ ] Minor cleanup: near-duplicate `resolve_details_row` (`callback_handlers.rb:263`) vs `resolve_status_row` (`slash_handlers.rb:351`) share nil-guard/0-1-many/rescue scaffolding; the attr `key=value` sort/format pipeline is written three times (`notification_builders.rb:86`, `:336`, `:409`); and `next_step_hint` opens with a dead `command = nil` (`:443`) overwritten on the next line.
- [ ] Plan-waiting hint yields a bare relative `plan.md` when both `state_file` and `folder` are absent (`lib/hive/bot/notification_builders.rb:416`): `File.join("".., "plan.md")` → non-actionable path; cosmetic, and `folder` is present on production rows, but worth a guard.
- [ ] Branch is behind `origin/main`: the two-dot review diff shows phantom reverts of `CHANGELOG.md` and `.github/workflows/release.yml` (from main-only commits `54fd3455`, `fc6b884c` that landed after the merge-base). The branch's own commits (merge-base...HEAD) do **not** touch these files and a standard 3-way merge/rebase will **not** regress them — rebase onto `origin/main` for a clean diff.

ivankuznetsov:

# PR Review Toolkit — telegram-bot-show-details-button (pass 1)

Scope: `git diff origin/main..HEAD`. Multi-persona review (code-reviewer, pr-test-analyzer,
comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier). Union of all
persona findings below.

Verdict: **Ready with minor fixes.** No persona raised a blocking (Critical) issue. The
code-reviewer independently ran the full gate — `rake coverage` passed (100.00% lines,
6340 runs, 0 failures) and `rubocop` reported no offenses on all 5 changed files — and the
plan's core goal is verified (no in-scope Show-details path can emit "No diagnostic
available"; callback wire shape byte-identical; `refresh_diagnose` left intact). Findings are
observability, test-branch-coverage, comment/copy accuracy, and quality items.

## High
- [ ] _None._ No persona surfaced a Critical/merge-blocking defect; the full test suite, 100% line-coverage gate, and rubocop all pass.

## Medium
- [ ] [silent-failure-hunter] `resolve_details_row`'s `rescue StandardError` (`lib/hive/bot/handlers/callback_handlers.rb:277`) swallows the exception with zero logging, though `@logger` is injected and already used for `:callback_malformed` (`:56`): a future I/O-backed `status_snapshot_provider` failing would be invisible in bot.log — log before degrading (requires adding a `:details_lookup_failed`-style constant to `Logger::EVENTS`).
- [ ] [comment-analyzer] `lib/hive/bot/supervisor.rb:1280-1281` success-path string "Diagnosis is available… Tap Show details to dump it here." is now misleading — Show details renders the *cached* snapshot row, so a diagnostic `refresh_diagnose` just wrote is not visible until the next poll, and the reply summarizes rather than "dumps"; the new comment guards the branch without flagging this contradiction.
- [ ] [code-simplifier, code-reviewer] `truncate_diagnostic_reply` (`lib/hive/bot/notification_builders.rb:464`) joins summary+detail with `\n` in `diagnostic_detail` then re-splits via `split("\n", 2)`; a multi-line summary mis-assigns its tail into "detail" and skews the truncation boundary — thread the structured summary/detail through instead (output stays ≤4096 today, so currently cosmetic).
- [ ] [type-design-analyzer] `test/unit/bot/slash_handlers_test.rb:106` uses a hand-maintained `Row = Struct.new(...)` duplicate of `StatusWatcher::Row` (widened from 6→14 fields in this diff); the renderer's `respond_to?` guards convert would-be `NoMethodError` drift into silent divergence — use the real `Hive::Bot::StatusWatcher::Row` (as `callback_handlers_test.rb` already does).
- [ ] [pr-test-analyzer] Two `truncate_diagnostic_reply` early-return branches are unasserted (`lib/hive/bot/notification_builders.rb:466` detail-empty, `:469` available-not-positive), and the truncation test only checks `length`/marker, not that the summary prefix survives — the 100% **line** gate masks these branch gaps.
- [ ] [pr-test-analyzer] `next_step_hint` `suggested_command` fallback (`lib/hive/bot/notification_builders.rb:446`) only ever runs against `nil`, and the `== "-"` sentinel (`:448`) is never hit — no test asserts a `suggested_command`-sourced "Next step:" or the `"-"` fallthrough.
- [ ] [pr-test-analyzer] `diagnostic_detail` (`lib/hive/bot/notification_builders.rb:453-460`) is never tested with a summary-only, detail-only, or blank/`{}` diagnostic — add a `{"summary"=>"","detail"=>""}` row asserting no diagnostic section is appended (the `reject(&:empty?)`/`!diagnostic.empty?` guards).
- [ ] [pr-test-analyzer] Plan-waiting `File.join(folder,'plan.md')` fallback (`lib/hive/bot/notification_builders.rb:416`, `state_file` nil) is untested — every plan-waiting test supplies a non-nil `state_file`, so the realistic folder-derived path is unguarded.

## Nit
- [ ] [code-reviewer] Plan-hint renders a misleading absolute `"/plan.md"` when both `state_file` and `folder` are blank (`lib/hive/bot/notification_builders.rb:416`, `File.join("","plan.md")`) — guard the fallback and only emit "Plan draft:" when a path resolves.
- [ ] [comment-analyzer] `lib/hive/bot/supervisor.rb:1276` comment says the path "remains defensive for refresh_diagnose and manual diagnose child paths" (plural), but no "manual diagnose" child path exists — `refresh_diagnose` is the sole producer of the diagnose envelope.
- [ ] [comment-analyzer] Pre-existing comment `lib/hive/bot/supervisor.rb:993` lists `status/--diagnose` among read-only in-process verbs; this PR removed the read-only `/details` diagnose, so the only surviving diagnose spawn is the write-mode `--diagnose --write --force` refresh — update so the "read-only" rationale isn't contradicted.
- [ ] [silent-failure-hunter] Telegram's 4096 cap is measured in Ruby `String#length` (code points) not UTF-16 units (`lib/hive/bot/notification_builders.rb:353` and `telegram.rb` `split_message`): an emoji/astral-heavy diagnostic can pass the cap and still 400 → operator sees no reply (the spinner is cleared); rare and logged via `:send_failure`, so observable but not surfaced to the user.
- [ ] [type-design-analyzer] `details_reply`'s contract over `Row` (`lib/hive/bot/notification_builders.rb:336-345`) mixes `respond_to?`/`present_value` guards with hard reads of the same always-present `Data` members — dead defensiveness; pick one stance so required vs optional fields are legible.
- [ ] [type-design-analyzer] `diagnostic` is an untyped, stringly-keyed Hash (`"summary"`/`"detail"`/`"suggested_next_action"`) accessed at 3+ sites with no shared accessor — a thin `Data`/reader would centralize the key contract so a producer-side rename is a one-line change.
- [ ] [type-design-analyzer] `Row` carries both `next_action["command"]` and `suggested_command`, with the "structured wins over flat" precedence existing only inside `next_step_hint` (`lib/hive/bot/notification_builders.rb:442-451`) — add a one-line comment documenting the precedence.
- [ ] [code-simplifier] The sorted `attrs … map { "#{k}=#{v}" }` pipeline is spelled three times (`lib/hive/bot/notification_builders.rb:86`, `:336`, `:409`) — extract one `sorted_attr_pairs(row)` helper (byte-identical output).
- [ ] [code-simplifier] `details_hint` (`lib/hive/bot/notification_builders.rb:414-440`) re-derives the coding stage/marker classification already encoded in `waiting_input`/`review_waiting` — extract shared predicates (`plan_pause?`/`brainstorm_qna?`/`fix_guardrail_review?`), preserving branch order, so the two trees can't drift.
- [ ] [code-simplifier] Dead `command = nil` init in `next_step_hint` (`lib/hive/bot/notification_builders.rb:443`) is immediately overwritten — make the assignment unconditional (`next_action.is_a?(Hash) ? next_action["command"] : nil`).
- [ ] [code-simplifier] The `.compact` in `details_reply` (`lib/hive/bot/notification_builders.rb:350`) is dead — both `details_summary` and `details_hint` always return non-nil strings.
- [ ] [pr-test-analyzer] Minor coverage gaps: manual-only hint "no folder" branch (`lib/hive/bot/notification_builders.rb:426`) untested; the `resolve_details_row` non-conforming-row skip guard (`callback_handlers.rb:268`) untested; the many `refute_includes "No diagnostic available"` assertions are near-tautological regression guards (keep, but they are not content coverage).

ivankuznetsov:

## High

## Medium
- [ ] `.github/workflows/release.yml:289` restores the macOS/Colima hivebox smoke: repo docs and `origin/main` record this hosted-runner path failing at `colima start`, so tag release verification will go red after publishing instead of using the native arm64 Docker runner/fallback.

## Nit
- [ ] `CHANGELOG.md:21` drops existing 0.3.1 release-note details from `origin/main`: unrelated deletions remove shipped patrol, reviewer, integration, and dependency items from the published changelog.
- [ ] `lib/hive/bot/supervisor.rb:1014` still says diagnose replies fire for Show details: Show details now renders in-process, so the comment points future maintainers at the old child-diagnose path.

ivankuznetsov:




## High

- [ ] _None._ The shipped code is correct: all plan goals met, truncation arithmetic is provably bounded to ≤4096 chars on every branch, no in-scope Show-details surface can emit "No diagnostic available", and tests + rubocop + the 100% line-coverage gate all pass.

## Medium

- [ ] Router→CallbackHandlers `status_snapshot_provider` wiring (`lib/hive/bot/router.rb:103`) is unguarded by any test: deleting that line leaves all 590 bot unit tests AND the 100% line-coverage gate green (the local is still referenced on `router.rb:94`), yet every live inline "Show details" tap would silently degrade to the "no longer active" reply, since production injects a populated provider at `supervisor.rb:218`. The plan explicitly called this out as the "Callback wiring regression" risk and noted the U2 test guards it "only if the test constructs the handler with a populated provider" — that guard was never added at the Router level (the slash path has `test_details_slash_resolves_against_default_empty_snapshot_provider`, but an empty-provider test cannot distinguish wired-but-empty from unwired-default). Fix: add a `router_test` case that routes a `details:<project>:<slug>` callback through `@router.handle` with a non-empty `status_snapshot_provider` and asserts the rendered row summary (e.g. `"Action:"`) appears.

## Nit

- [ ] Stale comment at `lib/hive/bot/supervisor.rb:1014` — "(and diagnose replies still fire for Show details)" is now inaccurate: Show details renders in-process via `details_reply`; only `refresh_diagnose` (`--diagnose --write --force`) still produces a diagnose reply. The pass-1 comment-reconciliation commit (8f259b3d) updated the :993/:1276/:1280 sites but missed this one. (Corroborates codex-ce-code-review-02.) Fix: reword to reference the refresh-diagnostic path, not Show details.
- [ ] `next_step_hint` precedence is documented (`lib/hive/bot/notification_builders.rb:470-472`: structured `next_action["command"]` wins over flat `suggested_command`) but never tested with BOTH fields present — `test_details_reply_prefers_structured_next_step_when_available` supplies only `next_action`, and `..._uses_flat_suggested_command...` supplies only `suggested_command`. Inverting the precedence would keep both tests green. Fix: add a row carrying both `next_action {command: "A"}` and `suggested_command: "B"` and assert the reply contains `"Next step: A"`.
- [ ] Inconsistent defensiveness in `resolve_details_row` (`lib/hive/bot/handlers/callback_handlers.rb:227-230`): the match guards `row.respond_to?(:project)` and `:slug` (to skip non-Row snapshot junk — see `test_show_details_skips_non_conforming_snapshot_rows`) but then reads `row.stage.to_s` ungated. Harmless today (the only objects passing the project+slug guards are real Rows, which have `stage`, and the surrounding `rescue StandardError` would catch a stray `NoMethodError`), but the partial guard is misleading. Fix: either add `&& row.respond_to?(:stage)` for symmetry, or drop the project/slug guards if non-Row snapshot entries are not a real concern.

ivankuznetsov:

# PR Review (pr-review-toolkit) — telegram "Show details" in-process renderer (pass 2)

Scope: `git diff origin/main..HEAD`. Personas run: code-reviewer, pr-test-analyzer, silent-failure-hunter, comment-analyzer, type-design-analyzer, code-simplifier. Union of findings below; severity mapped Critical→High, Important→Medium, Suggestion→Nit.

`code-reviewer` returned **no issues** — rubocop clean on the 4 changed lib files, bot test suite green, wiring (`status_snapshot_provider` threaded through constructor + SIGHUP reload path) and truncation math verified, `:reply` action sent at supervisor.rb:411. Findings are concentrated in test-strength, error-handling, the new type, and wording/cleanup.

## High

- [ ] [silent-failure-hunter] `details_reply(row)` renders OUTSIDE any rescue (`callback_handlers.rb:114`, `slash_handlers.rb:322`): the degrade rescue guards only the snapshot lookup, so a render-time raise escapes (`handle` rescues only `ArgumentError`) to the `:fatal` poll handler *after* the callback was already ack'd → no reply is ever sent, contradicting the plan's central "never a dead end" goal and producing silence where the old subprocess path produced a reply. Latent today (the provider only ever yields full `StatusWatcher::Row`s, and `details_summary`'s comment leans on that by dropping the old `respond_to?` guards), but cheaply closed: wrap the render in a logged rescue (`:details_render_failed` + soft reply). Corroborated by the prior review pass (reliability + adversarial).

## Medium

- [ ] [pr-test-analyzer] The plan-traced "plan_waiting must NOT mention a diagnostic" guard is vacuous (`notification_builders_test.rb:310,425`): `refute_match(/diagnostic/i)` runs only against `diagnostic: nil` rows, yet `details_reply` appends `diagnostic_detail(row)` unconditionally — a `3-plan`+`waiting` row carrying a diagnostic hash *does* get it appended (analyst reproduced this). The test gives false confidence in a guard the code doesn't enforce. Add a plan_waiting row WITH a populated diagnostic and assert the intended behavior (or drop the misleading `refute_match` if appending is acceptable for plan rows).
- [ ] [pr-test-analyzer] `next_step_hint` precedence is documented as a contract but never behaviorally tested (`notification_builders_test.rb:280,455`): the structured-vs-flat tests each set only one of `next_action["command"]` / `suggested_command`, so a refactor that flipped the precedence (comment at notification_builders.rb:470-472) would pass every test. Add one row with both fields set to different values and assert the structured command wins.
- [ ] [silent-failure-hunter] `resolve_details_row`'s broad `rescue StandardError` (`callback_handlers.rb:277`) degrades deterministic programming errors (NoMethodError/TypeError from the match block) to the transient-sounding "Status lookup failed — try again in a moment." and logs no backtrace, so a logic bug masquerades as a retryable I/O blip and the operator retries forever. Narrow to I/O classes (`SystemCallError`/`IOError`/`Timeout::Error`) or add `backtrace:` + non-retryable copy. (Sibling `resolve_status_row` at slash_handlers.rb:361 is the same pattern and doesn't even log.)
- [ ] [silent-failure-hunter] Truncation counts `String#length` (codepoints), Telegram counts UTF-16 code units (`notification_builders.rb:382,508,515`): an emoji-dense cached diagnostic ≤4096 chars can exceed the real cap → Telegram 400 → `safe_send_message` rescues + logs a generic `:send_failure` (no slug/project) and drops the reply silently. Pre-existing `String#length` convention (`Telegram#split_message` shares it), and the truncation tests use only BMP `'x'*6000`, so the path is unverified. Measure/slice in UTF-16 units and add an astral-plane test. (Also raised by pr-test-analyzer as a lower-severity item, who notes `split_message` chunking reduces — but does not eliminate — the 400 risk.)

## Nit

- [ ] [pr-test-analyzer] Tautological assertion (`supervisor_test.rb:1561`): `assert_equal NotificationBuilders.details_reply(rows.first), text` compares the delegated method to itself (`render_details` *is* `details_reply(row)`), so it can never fail; pin an exact expected string or delete the line — the adjacent `assert_includes` lines carry the real coverage.
- [ ] [pr-test-analyzer] `next_step_hint`'s "present `next_action` Hash but blank/missing `command` → fall back to flat `suggested_command`" branch (`notification_builders.rb:475`) is unexercised; only the nil-`next_action` fallback is tested. Add a row with `next_action: {"kind"=>"run"}` (no `"command"`) plus a `suggested_command`.
- [ ] [pr-test-analyzer · code-simplifier] `respond_to?` guards contradict the change's own stated contract: `details_summary` reads Row members directly (commenting that the renderer "only ever receives a real Row"), while `next_step_hint`/`diagnostic_detail`/`present_value` still `respond_to?`-guard the same always-present `StatusWatcher::Row` members. Pick one contract — drop the guards for consistency, or soften the comment.
- [ ] [silent-failure-hunter] `refresh_diagnose` pre-tick staleness (`supervisor.rb:1280`): the freshly written verdict isn't in the cached snapshot until the next status tick, so a Show-details tap before the tick renders the *stale* pre-refresh diagnostic with no staleness cue. The new copy hedges "updates on the next status refresh" but doesn't flag the shown content as stale. Low priority (no crash); consider surfacing snapshot age or nudging a tick.
- [ ] [type-design-analyzer] `Diagnostic = Data.define(:summary,:detail)` accepts nil members (`notification_builders.rb:487`): `Diagnostic.new(summary: nil, detail: nil)` is constructible and breaks `#text`'s `reject(&:empty?)` with NoMethodError — all four invariants live in the `diagnostic_detail` factory, not the type, unlike sibling `Notification` which validates in `initialize`. Currently unreachable, but a 3-line `initialize` coercion (`super(summary: summary.to_s, detail: detail.to_s)`) closes the gap at the type boundary. (Type shape itself is justified, not over-engineering.)
- [ ] [code-simplifier] `truncate_diagnostic_reply` rebuilds the identical `(sections + [diagnostic.text]).join("\n\n")` expression in both early-return guards (`notification_builders.rb:504-509`) — the same string `details_reply` already composed as `text`. Thread the computed `text` through and compute `prefix` only on the path that uses it. Behavior-preserving (covered by the two summary-exhaustion truncation tests).
- [ ] [code-simplifier] `reject(&:empty?)` on `sections` is partly dead (`notification_builders.rb:505,512`): `details_summary`/`details_hint` always return non-empty strings, so only `diagnostic.summary` can ever be dropped. Keep just that filter (`[ *sections, diagnostic.summary ].reject(&:empty?)`) or leave a note so a reader doesn't assume `sections` can contain blanks. Cosmetic.
- [ ] [comment-analyzer] The "can't misfire" claim on `plan_pause?`/`brainstorm_qna?` (`notification_builders.rb:170-172`) asserts an invariant the code doesn't structurally enforce (it's a property of the current predicate bodies), and the "runs against arbitrary rows" justification overclaims — all three call sites pass live Rows. Soften to describe the rule ("each predicate re-checks the marker itself; keep that check when editing").
- [ ] [comment-analyzer] The "always-present Data members … instead of guarding them with `respond_to?`" comment (`notification_builders.rb:362-364`) can mislead a maintainer into stripping the `respond_to?` guards the sibling helpers still use; scope it to the core members (action/marker/stage/project/slug) and note optional fields stay guarded.
- [ ] [comment-analyzer] "mirroring the `:callback_malformed` rescue above" (`callback_handlers.rb:278-280`) is a positional cross-reference that dangles if `resolve_details_row` is ever moved; name the method (`handle`'s `:callback_malformed` logging) instead. Optional.

ivankuznetsov:

## High
- [ ] `.github/workflows/release.yml:280` macOS smoke job starts Colima on an unsupported runner: `macos-15` is arm64 and GitHub documents nested virtualization as unsupported there, so `colima start` will fail before the smoke and leave releases red.

## Medium
- [ ] `lib/hive/bot/supervisor.rb:957` `/status` details buttons resolve against a cold or stale cache: `execute_dispatch` renders buttons from `fetch_result.rows` but never refreshes `@latest_status_rows`, so an immediate tap can get "Status is still loading" or old details instead of the row summary.
- [ ] `lib/hive/bot/notification_builders.rb:478` retryable diagnostics can get a laptop fallback instead of the retry command: `next_step_hint` ignores `row.diagnostic["suggested_next_action"]["command"]`, so `review_stale`/`error` rows with retry diagnostics but no `row.next_action`/`suggested_command` hide the actionable command.

## Nit
- [ ] `lib/hive/bot/handlers/slash_handlers.rb:338` `/details` render-error logging can re-raise on malformed matched rows: `resolve_status_row` accepts slug-only rows, but the rescue logs `row.project`, defeating the soft-degrade path when a bad provider row lacks that method.

ivankuznetsov:

# Code Review — claude-ce-code-review (pass 03)

Scope: branch `telegram-bot-show-details-button-260624-96cc` vs merge-base
(`git diff $(git merge-base origin/main HEAD)..HEAD`). Reviewed the bot
Show-details renderer change (U1–U4): `lib/hive/bot/notification_builders.rb`,
`handlers/callback_handlers.rb`, `handlers/slash_handlers.rb`, `router.rb`,
`logger.rb`, `supervisor.rb`, `schemas/hive-bot-log.v2.json`, and the four bot
unit test files.

Reviewers: correctness, adversarial, reliability, maintainability, testing
(plus requirements-trace check against the plan).

Scope note (NOT a finding): `.github/workflows/release.yml` and `CHANGELOG.md`
appear under `git diff origin/main..HEAD` but are a stale-base artifact —
`origin/main` advanced with #583/#584 after this branch was cut, and the branch
authors no commit touching either file (`git diff merge-base..HEAD` excludes
them). A real merge resolves against the merge-base and will not revert them, so
they are out of review scope.

Requirements trace: U1–U4 all satisfied; "No diagnostic available" is confirmed
unreachable from every in-scope Show-details surface (only the out-of-scope
`refresh_diagnose` envelope path can still emit it). No plan-required item is
missing. Verdict: ready with minor fixes.

## High

(none)

## Medium
- [ ] Triplicated render-fault soft-degrade rescue — `lib/hive/bot/handlers/callback_handlers.rb:125`, `lib/hive/bot/handlers/slash_handlers.rb:335`, `lib/hive/bot/supervisor.rb:1481`: the same `details_reply(row)` call + `rescue StandardError` + `:details_render_failed` log (`backtrace: Array(e.backtrace).first(3)`) + identical `"Status lookup failed — try again in a moment."` reply is copied near-verbatim at three sites; a future change to the degrade copy, event name, or backtrace depth must touch all three or the surfaces silently drift — the exact failure mode this PR consolidates away for the render body itself. Hoist into e.g. `NotificationBuilders.details_reply_safe(row, logger:, **log_context)` and collapse each call site to one delegating line (bodies differ only in log context: supervisor adds `source:`, the callback derives project/slug from tokens).

## Nit
- [ ] Truncation bounds output by Ruby `String#length` (codepoints) but Telegram enforces the 4096-char cap in UTF-16 code units — `lib/hive/bot/notification_builders.rb:525,533`: an emoji/astral-plane-heavy `diagnostic.detail` (or `display_name`) can pass the `<= TELEGRAM_MESSAGE_MAX_CHARS` check at ≤4096 codepoints yet exceed 4096 UTF-16 units, so `send_message` returns HTTP 400 and the operator gets no reply (surfaced only as a logged `send_failure`; `write_last_seen` still runs, so no redelivery loop). Flagged independently by correctness and adversarial; both note it is a pre-existing codebase convention shared with `Telegram#split_message` (so the new code is internally consistent, not a regression) and recommend not blocking. If hardening: measure UTF-16 units (`str.encode(Encoding::UTF_16LE).bytesize / 2`) in both `truncate_*` and `split_message`, and add an emoji-boundary truncation test.
- [ ] `details_hint` recovery-but-auto-recoverable fall-through is line-covered but its hint content is never asserted — `test/unit/bot/notification_builders_test.rb` (branch at `lib/hive/bot/notification_builders.rb:458`): the `recovery?(row) && manual_only_recovery?(row)` predicate is only asserted on the manual-only TRUE path (`execute_stale`); a recovery row that is auto-recoverable (e.g. `recover_review`/`review_error` with non-manual attrs) passes the 100% line gate but its resulting hint line is unpinned. Add a test for such a row asserting the hint falls through to `"Open on a laptop to advance."` (or `next_step_hint` output) and `refute_includes "no automatic recovery"`, so a regression that wrongly emits the manual-only hint is caught.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — pass 03

Scope: branch `telegram-bot-show-details-button-260624-96cc` vs merge-base
(`git diff $(git merge-base origin/main HEAD)..HEAD`). Reviewed the bot
Show-details renderer change (U1–U4) across `lib/hive/bot/notification_builders.rb`,
`handlers/callback_handlers.rb`, `handlers/slash_handlers.rb`, `router.rb`,
`logger.rb`, `supervisor.rb`, `schemas/hive-bot-log.v2.json`, and the four bot
unit test files.

Personas run: code-reviewer, pr-test-analyzer, comment-analyzer,
silent-failure-hunter, type-design-analyzer, code-simplifier.

Scope note (NOT a finding): `.github/workflows/release.yml` and `CHANGELOG.md`
appear under `git diff origin/main..HEAD` but are a stale-base artifact
(origin/main advanced with #583/#584 after this branch was cut; the branch
authors no commit touching them, and a real merge resolves against the
merge-base, so they will not be reverted). Excluded from review.

Verdict: **No Critical or Important findings.** code-reviewer reported nothing
actionable (598 bot tests pass, rubocop clean, schema valid, production wiring
confirmed). silent-failure-hunter confirmed the five new `rescue StandardError`
blocks are a deliberate, correctly-logged, honest soft-degrade — not silent
failures. type-design-analyzer rated the new `Diagnostic` Data type sound
(encapsulation 8/10, enforcement 8/10). The Suggestion-level items below map to
Nit.

## High

(none)

## Medium

(none)

## Nit
- [ ] [pr-test-analyzer] No `details_reply` test renders the *actionable* manual-recovery hints through the Show-details surface — `test/unit/bot/notification_builders_test.rb:373`: the only manual-only `details_reply` test uses `execute_stale` (the generic "open it on a laptop" arm of `manual_only_reply`); the actionable variants (`error`+`reason=dirty_worktree` → "…commit or discard the changes, then tap Autofix.", `review_error`+`reason=fix_status_check_failed` → "…repair Git state, then rerun review.") are only pinned via `RecoverySequence` (a different caller), so a regression that forwards the wrong `attrs` and degrades Show-details to the generic line would stay green. Add a focused test rendering a `dirty_worktree` error row through `details_reply` and asserting the actionable git hint.
- [ ] [code-simplifier] Soft-degrade reply literals are duplicated and meant to stay in lockstep — `"Status lookup failed — try again in a moment."` at 5 sites (`callback_handlers.rb:131,304`; `slash_handlers.rb:341,387`; `supervisor.rb:1491`) and `"Status is still loading — try again in a moment."` at 2 sites (`callback_handlers.rb:282`; `slash_handlers.rb:370`); promote each to a shared frozen constant in `NotificationBuilders` (already required into both handlers) so a future wording change can't drift across surfaces (no coverage-gate risk — each rescue/branch line stays distinct). Persona explicitly recommends NOT DRY'ing the three `render_details_reply` rescue wrappers or the two resolvers (intentional per-context variations: distinct log payloads/event names and match predicates).
- [ ] [silent-failure-hunter] `:details_render_failed` log payload omits `stage` — `callback_handlers.rb:128` and `supervisor.rb:1487`: the lookup path (`resolve_details_row`) carries `stage`, which disambiguates duplicate slugs, but the render-failure log does not; when a render fault recurs for only one stage of a duplicated slug the log can't identify the row. Thread `stage:` into the `render_details_reply` log payload (it's already in scope in `show_details`). Diagnosability nicety, not a correctness issue.
- [ ] [comment-analyzer] `details_hint` plan-path comment is imprecise — `notification_builders.rb:452-454`: it claims `File.join` "would emit a bogus bare/absolute 'plan.md'" when neither `state_file` nor `folder` resolves, but the `File.join` on line 451 is guarded by `if folder`, and `present_value` returns `nil` or a non-empty stripped string (never `""`), so that hypothetical can't trigger (a missing folder would leave `plan` nil, not produce `/plan.md`). The implemented behavior is correct; only the justifying comment is misleading. Tighten the wording.
- [ ] [type-design-analyzer] `Diagnostic` comment overstates the parallel to the sibling `Notification` type — `notification_builders.rb:493-495`: it frames the `.to_s` member coercion as "like sibling Notification's validate-in-initialize", but `Notification` *raises* on a bad `parse_mode` while `Diagnostic` *coerces* (lenient — correct for a render-path type that must not crash a reply). Soften the comment so a future reader doesn't expect a raise. The type design itself is sound; no code change needed.

PR #604, opened 2026-06-27
feat(setup): persist selected agent backends globally

ivankuznetsov:
## Summary

Hive had nowhere to record which agent backends an operator actually uses. `hive doctor` and `hive init` probe every configured skill across Claude, Codex, and Pi, so an operator who never touches Pi still sees Pi assets reported as missing — and there was no path to *install* the assets that genuinely are missing. This PR lands the foundation for fixing both: a global, persisted record of selected backends under `agents.selected` in `~/.config/hive/config.yml`, plus the prompt that collects it.

This is the first slice (U1) of the larger `hive setup` plan. It ships on its own because it is self-contained and unblocks the rest — later units (the `setup` command, the skill/runtime installers, and making doctor/init backend-aware) read the selection written here.

**What changed**

- `Hive::Config.load_global_agents` / `write_global_agents!` round-trip the selection through the existing `update_global_config!` machinery. The list is normalized to the canonical `GLOBAL_AGENT_BACKENDS` order (`claude, codex, pi`) and de-duped, so persisted output is stable regardless of input order.
- `Hive::Commands::Setup::BackendPrompt` is a TTY multi-select modelled on the existing `Init::Prompts` idioms: numbered and named tokens, case-insensitive, blank accepts the default, with valid names/indices re-listed on every reprompt. Claude + Codex are pre-checked; Pi is opt-in.
- Choices are filtered to the agents actually registered in `AgentProfiles` via the new shared `Config.registered_agent_names` helper (which `Init::Prompts` now also routes through), so a machine without Codex is never offered or defaulted to a Codex backend.

**Design decisions**

- **Safe defaults, no surprises off-TTY.** A non-interactive invocation short-circuits to `[claude, codex]` and emits a one-line `hive:` summary instead of blocking, so daemon and CI callers never hang on a prompt. Absent config returns that same default, and a closed stdin degrades to defaults rather than raising.
- **One validating boundary.** `normalize_global_agents` guards both the read and write paths: it enforces ≥1 backend, canonical order, and case-insensitive resolution, and returns a frozen array of frozen strings on every path so callers can treat the selection as read-only. A load-time assertion pins `DEFAULT_GLOBAL_AGENTS ⊆ GLOBAL_AGENT_BACKENDS`, matching the boot-time parity check in `Init::Prompts`.
- **Preserve operator-owned config.** `write_global_agents!` merges `selected` into any existing `agents` hash rather than overwriting it, and attributes a bad *caller* argument to the argument rather than to the config file.
- **Loud on drift.** A malformed `agents` block, a non-array `agents.selected`, or an unknown backend each raise `ConfigError`; the message distinguishes a genuine typo from a known-but-not-installed-here backend.

**Scope and follow-up**

Execution stopped after U1 on a deliberate plan/source mismatch rather than silently redesigning. The plan assumes a bare Claude `/plan` skill, but the real marketplace exposes it as `/llm-wiki:wiki-plan`, and the placeholder `ivankuznetsov/hive-skills` source the install path was to use is still unpublished (`Repository not found`). The `hive setup` command, the real plugin-CLI skill installer, the runtime-prereq installer (e.g. the missing Arch `erb` gem that motivated this task), and the doctor/init verify-only changes are deferred until the canonical marketplace identifiers are confirmed.

## Test plan

- `ruby -Itest test/unit/commands/setup/backend_prompt_test.rb` — non-TTY default, blank-accepts-default, numeric / case-insensitive / mixed name+number selection with de-dup, whitespace-only → default, out-of-range and `0` index reprompt, unknown-token reprompt, EOF → `Aborted`, empty-registry `ArgumentError`, registered-backend filtering (interactive and non-interactive menus), frozen-return contract, and `summary_io` silence on interactive paths.
- `ruby -Itest test/unit/config_global_agents_test.rb` — default-when-absent (including no-config-file cold start and present-but-null `selected`), write/load round-trip in canonical order, realistic re-run overwrite, preservation of sibling `agents` keys, the malformed-config rejection matrix (non-Hash `agents`; non-array / empty / non-string `selected`; unknown backend), `registered_agent_names`, and the zero-default `ConfigError`.
- `ruby -Itest test/unit/commands/init/prompts_test.rb` — confirms the `Init::Prompts` refactor onto the shared `Config.registered_agent_names` helper is behavior-preserving.
- Repo gates: `rubocop` clean on all changed files; `rake coverage` at 100% line coverage (the new error/rescue branches each have a covering test); `brakeman` clean.

## Review summary

Three review passes ran (`ce-code-review` + `pr-review-toolkit`, multi-persona, plus `codex`).

- **Pass 01** found one High: the required 100% line-coverage gate would have gone red because two new `ConfigError` raise lines — and later the `default_global_agents` zero-default raise — were never exercised. Fixed by adding the covering tests.
- **Passes 02–03** surfaced no live defects, but a cluster of contract, robustness, and convention items: enforce the ≥1-backend invariant and frozen-return contract on every path, add the `DEFAULT ⊆ BACKENDS` boot guard, split the "unknown vs not-installed-here" error message, guard `interactive?` against a closed stream, make resolution case-insensitive on load, derive prompt copy from the defaults, align the `hive:` summary prefix, consolidate the intersection idiom, and fill test gaps (cold start, restricted-registry menu, out-of-range index, mixed input). All were applied.
- **Codex** repeatedly escalated the absence of U2–U7. This is the deliberate scope boundary described above: the install path is blocked on confirming the canonical marketplace/plugin identifiers, an open question the plan itself flags (Risk #1 / Open Question #2). Its remaining High/Medium items were diff artifacts of the branch trailing `origin/main` by five patrol commits, resolved by rebasing.

No High or Medium findings remain open against the delivered U1 code.

## Linked task

- Hive task `can-we-install-skills-if-260626-481a` (id 12294, "Install Missing Hive Skills").
- Origin: operator request — *"Can we install skills if they are not available?"* — raised after `hive doctor` reported missing Claude commands (`/ce-brainstorm`, `/plan`, `/ce-code-review`), a missing Codex `ce-code-review` skill, and a missing Arch `erb` runtime gem across freshly-initialized checkouts.

---


ivankuznetsov:

## High
- [ ] Plan-required `hive setup` install path is absent: `lib/hive/cli.rb:260` still goes from `doctor` directly to `update`, and the only setup command file is a backend prompt, so R1/R2/R4/R6-R11 are not delivered.
- [ ] Doctor/init still ignore the persisted backend selection: `lib/hive/commands/doctor.rb:318` and `lib/hive/commands/doctor.rb:350` verify every configured stage/reviewer while `lib/hive/commands/init.rb:711` warns on every missing row, so unselected Pi backends still count against R5/R12.
- [ ] Babysitter dry-run can execute caller-controlled Ruby startup code before its guard: `lib/hive/babysitter/dry_run_env.rb:84` now writes a Ruby shim without unsetting `RUBYOPT`/`RUBYLIB`, so injected startup code runs before lines 86-87 pin the real binaries.

## Medium
- [ ] `hive-e2e replay` follows symlinked scenario directories again: `bin/hive-e2e:161` only checks the final `repro.sh` with `File.lstat`, so a symlinked parent can point replay at an executable outside the run artifact.
- [ ] `bin/hive-eval` no longer converts invalid-byte argument failures into usage exits or stale-report cleanup: `bin/hive-eval:96` rescues only `OptionParser::ParseError`, dropping the prior `ArgumentError` handling for invalid byte sequences.
- [ ] Missing real `gh` passthrough now leaks a raw exec failure instead of the documented 127 diagnostic: `bin/hive-babysitter-stub-gh:460` calls `exec(real, *argv)` without rescuing `SystemCallError`.

## Nit

ivankuznetsov:

# Code Review — `/ce-code-review` (pass 1)

**Scope.** This branch's only commit is `52fbd681 feat(setup): U1 persist
selected agent backends` (4 new files: `lib/hive/commands/setup/backend_prompt.rb`,
`lib/hive/config.rb`, `test/unit/commands/setup/backend_prompt_test.rb`,
`test/unit/config_global_agents_test.rb`). The remaining files shown by
`git diff origin/main..HEAD` (babysitter `dry_run_env`, `bin/hive-e2e`,
`bin/hive-eval`, several `wiki/` fragments) are **origin/main commits that
landed after this branch was cut** — the merge-base (`16fff941`) is behind
`origin/main`, so they surface as "deletions" but are not this branch's work.
They are out of review scope. Per the plan, only U1 is delivered here; U2–U7
are separate downstream units and are correctly not implemented in this task.

Verified green: both new unit tests pass (`13 runs, 45 assertions, 0 failures`);
`rubocop` clean on all 4 files; `BackendPrompt` matches `Init::Prompts`
conventions (`input:$stdin, output:$stderr, summary_io:$stdout`, identical
`interactive?`); `require "hive/agent_profiles"` auto-registers claude/codex/pi
so the registry is always populated (no load-order bug in `default_global_agents`).

## High
- [ ] Coverage gate (`rake coverage`, CI-required with `HIVE_COVERAGE_MIN_LINE: "100"`) will fail — two new executable lines are never exercised: `lib/hive/config.rb:808` (the `raise ConfigError "agents ... must be a Hash"` in `write_global_agents!`, reached only when an existing config has a non-Hash `agents` value) and `lib/hive/config.rb:826` (the `raise ConfigError "... must contain non-empty strings"` in `normalize_global_agents`, reached only by a non-String / empty-String element in `selected`). Verified directly via the Coverage API; grep confirms the only callers of these methods are the two new test files, so aggregate line coverage drops below 100% and the required gate fails (the plan also lists "100% line coverage on new files" as a U1 acceptance criterion). Fix: add a `write_global_agents!` test where the existing `agents:` block is a non-Hash, and a `load_global_agents`/`normalize` test passing `selected: [123]` or `selected: [""]`.

## Medium
- [ ] (none)

## Nit
- [ ] `lib/hive/commands/setup/backend_prompt.rb:18` — the constructor's `raise ArgumentError, "registered_agents must include at least one setup backend" if @backends.empty?` guard has no covering test (no scenario passes `registered_agents: []`). The line is line-covered (the `if`-modifier always executes, so it won't break the 100% gate), but the actual raise behavior is unverified; a one-line test would pin it.

ivankuznetsov:

# PR Review — `/pr-review-toolkit:review-pr` (multi-persona, pass 1)

**Scope.** Reviewed only commit `52fbd681` (U1 — persist selected agent
backends): `lib/hive/commands/setup/backend_prompt.rb` (new),
`lib/hive/config.rb` (added `default_global_agents`, `load_global_agents`,
`write_global_agents!`, `normalize_global_agents`, constants
`GLOBAL_AGENT_BACKENDS` / `DEFAULT_GLOBAL_AGENTS`), and the two new unit test
files. All other files in `git diff origin/main..HEAD` are unrelated
origin/main commits that landed after this branch was cut (merge-base behind
origin/main) — out of scope. U2–U7 (the `hive setup` orchestrator, installers,
doctor/init wiring, docs) are separate downstream units and correctly not
implemented here; findings that would only flag that deferred wiring were
dropped per scope rules.

Personas run: code-reviewer, pr-test-analyzer, silent-failure-hunter,
type-design-analyzer, comment-analyzer, code-simplifier.

## High
- [ ] [tests] 100% line-coverage gate (`rake coverage`, CI sets `HIVE_COVERAGE_MIN_LINE=100`, aggregate) will fail — two new executable lines are never exercised: `config.rb:808` (the `raise ConfigError "agents … must be a Hash"` in `write_global_agents!`, reached only when an existing config already holds a non-Hash `agents:` value) and `config.rb:826` (the `raise ConfigError "… must contain non-empty strings"` in `normalize_global_agents`, reached only by a non-String/empty element in `selected`). Verified via Coverage API; the only callers are the two new test files. Fix: add a `write_global_agents!` test pre-seeding `agents: "claude"`, and `load`/`normalize` cases for `selected: ["claude",""]`, `["  "]`, `[123]`.

## Medium
- [ ] [code-review][silent-failure][type-design] `BackendPrompt#collect` can silently return an empty backend selection: when the registry contains only a non-default backend (e.g. `registered_agents: %w[pi]`), `default_backends` (`backend_prompt.rb:52`) is `[]`, and both the non-interactive path (`:55-58`, prints "using default backends — " with nothing after) and the interactive blank-Enter path (`:38-39`, `return selection if selection` — `[]` is truthy in Ruby) hand back `[]` for a prompt whose job is to yield ≥1 backend. Its model class `Init::Prompts` raises `ArgumentError` in the analogous "default not registered" case (`init/prompts.rb:114-119`); add an equivalent guard.
- [ ] [type-design] `load_global_agents` return contract is frozen on one path and not the other — `config.rb:571` freezes `default_global_agents`, so the absent-key path returns a frozen array, while the configured path returns the unfrozen `normalize_global_agents` result (`config.rb:840`). A consumer that mutates the result (`agents << x` / `agents.sort!`) succeeds with a stored selection but raises `FrozenError` when it falls through to defaults — a latent, path-dependent crash. Freeze both or neither.
- [ ] [type-design][code-review] The "≥1 backend" invariant is enforced at the prompt boundary but not at persistence/load — `normalize_global_agents` accepts `[]` (the `agents.each` loop over `[]` yields `{}` → returns `[]`), so `write_global_agents!([])` or a hand-edited `agents: {selected: []}` persists/loads zero backends, a state the prompt can never produce. Producers disagree on whether "zero selected backends" is legal; decide and enforce at the mutation boundary.
- [ ] [code-review] Global `agents.selected` overloads the top-level `agents` key that project-config `DEFAULTS["agents"]` (`config.rb:108`) already uses for per-CLI profile overrides; `load_global_digest_config`'s `merge_defaults` (`config.rb:1060`) yields a mixed hash `{claude=>{…}, codex=>{…}, selected=>[…]}`. Safe today only because no code enumerates the `agents` block (`AgentProfiles.lookup` does targeted `dig`), but fragile — any future `cfg["agents"].each_value` expecting profile hashes would choke on the array. Consider a distinct key (e.g. `agent_backends`).

## Nit
- [ ] [tests] `BackendPrompt` constructor's empty-backends `ArgumentError` (`backend_prompt.rb:18`) is line-covered (so the gate passes) but its raise branch is never triggered — no test passes `registered_agents: []` or a non-intersecting set (`%w[ghost]`). Add `assert_raises(ArgumentError)`.
- [ ] [tests] The reprompt branch where `resolve_selection` returns `nil` because every token strips away (e.g. user types just `","`, which is non-blank so bypasses the blank-default path) is untested (`backend_prompt.rb:62`); only the unknown-token reprompt is covered. Add an input like `",\nclaude\n"`.
- [ ] [tests] `default_global_agents` backend-filtering (`config.rb:570`) and `normalize_global_agents`'s registered-intersection `allowed` (`config.rb:822`) are never exercised with a missing/unregistered backend — all three profiles auto-register and tests never `reset_for_tests!`, so the filter always returns the full set; the filter-out / "valid-constant-but-unregistered → unknown backend" behavior is unverified.
- [ ] [comments] New `config.rb` methods (`load_global_agents`/`write_global_agents!`/`normalize_global_agents`) and constants (`GLOBAL_AGENT_BACKENDS`/`DEFAULT_GLOBAL_AGENTS`) carry zero explanatory comments though every sibling in this file is documented (`REVIEWER_KINDS`, `DEPENDENCY_GATE_STAGES`, `registered_projects`, `load_global_config`); CLAUDE.md says match surrounding comment density. In particular the canonical-reorder+dedupe intent of `normalize`'s trailing `.select` and the merge-preserve intent of `(existing || {}).merge(...)` are load-bearing and invisible. `BackendPrompt` also lacks the substantial class-level doc comment its sibling `Init::Prompts` (`init/prompts.rb:7-16`) has.
- [ ] [type-design] `config.rb` has no `# frozen_string_literal: true`, so `GLOBAL_AGENT_BACKENDS = %w[claude codex pi]` holds mutable strings and `normalize_global_agents` (`config.rb:840`) returns aliases of them — an in-place caller mutation (`result.first.upcase!`) would corrupt the shared constant process-wide. Dup/freeze the returned names or add the magic comment.
- [ ] [type-design][simplify] "GLOBAL_AGENT_BACKENDS restricted to registered names, canonical order" is written three different ways — `BACKEND_ORDER.select { registered.include? }` (`backend_prompt.rb:18`), `GLOBAL_AGENT_BACKENDS & registered` (`config.rb:822`), `DEFAULT_GLOBAL_AGENTS.select { registered.include? }` (`config.rb:571`). Settle on one idiom for consistency. (Do NOT merge the methods — `BackendPrompt`'s `registered_agents:` injection seam is an intentional test seam.)
- [ ] [simplify] `normalize_global_agents`'s `selected = {}` hash-as-set buys no real dedupe — the result comes from `GLOBAL_AGENT_BACKENDS.select { selected[name] }` iterating the already-unique constant, so a plain array (`selected << name` / `…select { selected.include? }`) is simpler and matches `BackendPrompt#resolve_selection`'s idiom. Output is byte-identical (`config.rb:823,837,840`).
- [ ] [simplify] `default_backends` is pure but recomputed once per listing-loop iteration and again on every prompt iteration in `BackendPrompt#collect`; hoist once (`defaults = default_backends`) at the top of `collect` (`backend_prompt.rb:30,35,38`).
- [ ] [test-quality] `test_interactive_accepts_numbers_and_persists_backend_order` (`backend_prompt_test.rb:42`) asserts only `collect`'s return ordering — it never writes or reads config, so "persists" overstates what it verifies; rename.
- [ ] [silent-failure][comments] `BackendPrompt::Aborted` is a bare `StandardError` with no in-commit catcher (wiring is U5, deferred). Forward-looking note: when the U5 orchestrator is built, ensure it rescues `Aborted` and renders a clean "setup aborted" rather than letting a generic `StandardError` rescue surface a Ctrl-D as an internal error (exit 70); the abort message currently never reaches `output`/`summary_io`.

ivankuznetsov:

## High
- [ ] Missing `hive setup` command/orchestrator: only `lib/hive/commands/setup/backend_prompt.rb` exists, with no `lib/hive/commands/setup.rb`, CLI entrypoint, sentinel, JSON/dry-run path, or summary, so R1/R2/R11/R13 cannot be exercised.
- [ ] Missing asset and runtime installers: no `lib/hive/setup/*` manifest, skill source registry, real plugin CLI installer, or `erb`/`qmd` runtime-prereq installer exists, so R4/R6/R7/R8/R9/R10/R14 remain unimplemented.
- [ ] `doctor` and `init` never consume the selected-backend config: `Hive::Config.load_global_agents` is added but not called from `lib/hive/commands/doctor.rb` or `lib/hive/commands/init.rb`, so unselected Pi can still be checked/warned and R5/R12 are not met.
- [ ] Babysitter dry-run wrappers again run Ruby before scrubbing startup env: `lib/hive/babysitter/dry_run_env.rb:84` writes Ruby shims, so caller-controlled `RUBYOPT`/`RUBYLIB` execute before the wrapper can pin handoff env or reach the stub.

## Medium
- [ ] `bin/hive-e2e replay` accepts symlinked parent artifact directories: `bin/hive-e2e:161` now only `lstat`s `repro.sh`, so a symlinked `scenarios/<name>` directory can point replay at an executable outside the run artifact.
- [ ] `bin/hive-eval` regresses invalid-byte usage handling: `bin/hive-eval:67` uses UTF-8 string matching/splitting on `--report=` after the invalid-byte rescue was removed, so malformed argv can crash instead of returning exit 64 and clearing stale reports.
- [ ] Setup docs are still the deferred stub: `install.md:141` still tells users not to run skill install commands and no `wiki/commands/setup.md` exists, so U7's documented `hive setup` flow and installer nudge are missing.
- [ ] Unrelated wiki changelog fragments are deleted: the diff removes three `wiki/log.d/20260626...` files, which violates the append-only fragment convention and risks stale `hive wiki compile-log --check` output.

## Nit

ivankuznetsov:

# Code Review — `hive setup` U1 (persist selected agent backends), pass 2

**Scope note (not a finding):** This branch's *authored* change set is **U1 only** —
`lib/hive/commands/setup/backend_prompt.rb` (new), the global-agents methods in
`lib/hive/config.rb`, and their two test files (`git diff 16fff941..HEAD` = 4 files,
+497 lines). The plan decomposes the feature into U1–U7 as separate units; U2–U7 are
deferred to downstream tasks by the plan's own structure, so their absence here is **not**
a missing-requirement gap. U1's stated deliverables and test scenarios are all present and
pass. The babysitter/e2e/eval changes that appear in `git diff origin/main..HEAD` are
**not** authored by this branch (see first Nit). Verified locally: `rubocop` clean on all
4 files; new tests green (`backend_prompt_test` 11 runs / `config_global_agents_test`
13 runs, 0 failures).

## High

_None._

## Medium

_None._

## Nit

- [ ] Branch is 5 commits behind `origin/main`; the babysitter/e2e/eval "deletions" in `git diff origin/main..HEAD` are not authored changes: `git diff 16fff941..HEAD` (the real change set) touches only the 4 U1 files. `origin/main` advanced by patrol fixes #598–#602 (RUBYOPT bypass, replay parent-symlink, invalid-UTF-8 cleanup, gh-stub backtrace, gh caller-Ruby) after this branch's merge-base. A standard merge/rebase preserves those commits (no real revert risk), but the branch should be rebased onto `origin/main` so the reviewed diff equals what lands and the security hardenings aren't misread as removed.

- [ ] `Config.load_global_agents` can return an empty array on its default path, contradicting the "never empty" invariant the comments tout (`config.rb:578-586`, `:797-820`): `default_global_agents` is `DEFAULT_GLOBAL_AGENTS.select { registered }.freeze`, which yields `[]` if neither `claude` nor `codex` is registered — whereas `normalize_global_agents`/`write_global_agents!` and `BackendPrompt` all enforce ≥1 backend. Unreachable today (both default backends auto-register on `require "hive/config"`), but the read-default path is the one asymmetry where the guarantee isn't actually enforced; worth a guard or a comment acknowledging it before U6's doctor/init consume it.

- [ ] `normalize_global_agents` couples persisted-config validity to the live `AgentProfiles` registry (`config.rb:861-875`): membership is checked against `allowed` (the registry-filtered subset of `GLOBAL_AGENT_BACKENDS`), so a *known* backend that is merely unregistered is rejected as `unknown backend "<x>"; expected one of [<registered>]`. Today all three backends register unconditionally so this can't fire, but if any backend's registration ever becomes conditional (e.g. register `pi` only when its CLI is present), a previously-valid persisted `selected:` would hard-raise `ConfigError` and brick `load_global_agents` — and therefore U6's doctor/init — instead of degrading gracefully. Consider validating membership against the static `GLOBAL_AGENT_BACKENDS` and treating backend *availability* as a separate, non-fatal concern.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — `hive setup` U1, pass 2

**Scope:** Authored change set is U1 only (`git diff 16fff941..HEAD`, 4 files):
`lib/hive/commands/setup/backend_prompt.rb` (new), the global-agents constants/methods in
`lib/hive/config.rb`, and `test/unit/commands/setup/backend_prompt_test.rb` +
`test/unit/config_global_agents_test.rb`. The babysitter/e2e/eval/wiki deletions in
`git diff origin/main..HEAD` are an artifact of origin/main being 5 patrol commits ahead of
the merge-base, **not** authored here, and were excluded from persona review. Personas run:
code-reviewer, comment-analyzer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer.
None found a Critical/live bug; the new code is well-structured, rubocop-clean, and its tests
pass. U2–U7 are plan-deferred separate units (not flagged as missing). Note: U1's code has no
`lib/` caller yet (infrastructure awaiting U5 wiring), so every finding below is latent/contract-level.

## High

_None._ (U1 is fully delivered against the plan; no plan-required-but-missing item, no live defect.)

## Medium

- [ ] [correctness/comments/errors/type-design] `Config.default_global_agents` can return frozen `[]`, and `load_global_agents` returns it **un-normalized** on the absent-`agents:` (`config.rb:810`) and absent-`selected` (`config.rb:817`) paths — bypassing the ≥1-backend invariant that `normalize_global_agents` (`config.rb:856-859`) and `BackendPrompt` (`backend_prompt.rb:38-41`) enforce; the docstring at `config.rb:580-582` calls this "the same return contract as `normalize_global_agents`," which is false on cardinality: unreachable today (claude/codex auto-register) but a real contract hole + inaccurate comment on a public read path U6 will consume. Converged across 4 personas.
- [ ] [errors] `agents: { selected: <null> }` (YAML key present but null) silently falls back to defaults at `config.rb:817`, while `selected: []` raises (`config.rb:856-859`) — same operator intent ("no/empty selection"), divergent outcomes, and the null path can silently mask a hand-edit these methods exist to surface (uses `agents["selected"]`-nil rather than `key?`). Either distinguish key-presence (raise on present-but-null) or document the asymmetry at `config.rb:797-801`. This null→defaults outcome is also behaviorally untested (the line is covered only via its false branch).

## Nit

- [ ] [tests] `BackendPrompt#resolve_token` numeric out-of-range / `"0"` / negative is untested (`backend_prompt.rb:103-106`): the `index >= 1` guard that stops `"0"` from selecting `@backends[-1]` (= `pi`) never has its false-branch outcome exercised, so a regression dropping it would silently mis-select. Add `make_prompt("0\nclaude\n")` and `make_prompt("9\nclaude\n")` → reprompt then `%w[claude]`.
- [ ] [type-design] Two sources of truth for "known backends" with no drift guard: `GLOBAL_AGENT_BACKENDS` (`config.rb:555`, strings) vs the `AgentProfiles` registry (symbols); registering a 4th profile won't make it selectable and nothing flags the divergence — unlike the load-time CHOICES/MODES guard at `init/prompts.rb:50`. Consider a load-time assertion or test that every member is registerable.
- [ ] [type-design] Registry symbol→string conversion `AgentProfiles.registered_names.map(&:to_s)` is duplicated 3× (`config.rb:584`, `config.rb:861`, `backend_prompt.rb:29`); centralize in one private helper so the registry-representation coupling is a one-line edit.
- [ ] [type-design] `BackendPrompt#collect` returns a **mutable** array on every path (`backend_prompt.rb:63`, `:77`, `:95`), inconsistent with the Config layer's stated "frozen array of frozen strings on every path" contract; elements are frozen so there's no mutation risk, but the two collaborators that both produce "a backend selection" disagree on mutability. Freeze the three return sites.
- [ ] [type-design] `BackendPrompt::Aborted` (`backend_prompt.rb:20`) duplicates `Init::Prompts::Aborted` semantics as a distinct class; the eventual `hive setup` wiring must rescue the right one — consider a shared `Hive::Commands::PromptAborted`. (Forward-looking: that call site must also rescue `Aborted`/`Interrupt` so stdin-EOF becomes a clean cancel, not an uncaught exit 70.)
- [ ] [errors/clarity] `write_global_agents!` passes `source_path: global_config_path` into `normalize_global_agents` (`config.rb:829`), so a bad **caller argument** (e.g. `[123]`) yields an error attributing fault to the config file — with a misleading "(defaults; no file present)" suffix from `describe_source` — even though nothing is wrong with the file. Use an argument-oriented source label on the write path.
- [ ] [comments] `config.rb:880-882` "in one pass" is loosely accurate: `GLOBAL_AGENT_BACKENDS.select { selected.include? }` is O(n·m), not a single linear pass (the dedupe/canonical-order result is correct); say "in a single `select` over the canonical constant."
- [ ] [tests] Mixed names+numbers in one answer (`"claude,3"`) is untested — existing tests are homogeneous (`"3,1"`, `"PI,claude,pi"`); add `make_prompt("claude,3\n")` → `%w[claude pi]`.
- [ ] [tests] Whitespace-only answer (`"   "`) → defaults is untested; it specifically pins `read_line`'s `.strip` (`backend_prompt.rb:115`). Add `make_prompt("   \n")` → `%w[claude codex]`.
- [ ] [tests] The frozen-return contract is asserted nowhere — add `assert written.frozen?` (or on `default_global_agents`) so removing a `.freeze` (`config.rb:585`, `:882`) is detectable.
- [ ] [tests] `write_global_agents!` overwriting a pre-existing `selected:` while keeping siblings (the realistic re-run-`hive setup` path) is not directly asserted; `test_write_global_agents_preserves_other_agent_settings` only preserves a sibling while writing a fresh selection.
- [ ] [tests] No integration test wires `collect → write_global_agents!` (expected until U5 lands); add one once the command exists, since the prompt's "≥1 / canonical-order" output is the normalize input contract and the two enforce overlapping-but-not-identical invariants.

### Credited strengths (persona consensus)
Single validating factory (`normalize_global_agents`) guards both read and write boundaries; `write_global_agents!` validates before locking and refuses to clobber a non-Hash `agents:` while preserving sibling keys; the `BACKEND_ORDER = Config::GLOBAL_AGENT_BACKENDS` alias makes prompt-listing order == on-disk order structural; EOF→`Aborted` is correct (no infinite reprompt); `with_restricted_registry` is sound (restores the global registry; serial execution); the round-trip test asserts return value + reloaded value + on-disk YAML shape + sibling preservation in one strong case; the malformed-input rejection matrix on both load and write is thorough.

ivankuznetsov:

## High
- [ ] `hive setup` is not wired into the CLI or command layer: `lib/hive/cli.rb` is unchanged and `lib/hive/commands/setup.rb` is absent, so R1/R2/R11 cannot be exercised despite the plan requiring a one-time setup command with summary/sentinel.
- [ ] Required asset source manifest and skill installer are absent: `lib/hive/setup/asset_manifest.rb`, `skill_sources.rb`, and `skill_installer.rb` are missing, so no selected backend can install the required Claude/Codex/Pi assets via real plugin CLIs (R4/R6-R8/R10/R13).
- [ ] Runtime prerequisite installation is absent: `lib/hive/setup/runtime_prereqs.rb` is missing, so `erb`, `GEM_PATH`, and `qmd` gaps remain warnings instead of being fixed by setup as required (R1/R6/R9/R11/R13).
- [ ] `doctor` and `init` are not backend-aware for setup selections: `lib/hive/commands/doctor.rb` and `lib/hive/commands/init.rb` are unchanged, so unselected Pi rows still warn/fail instead of being skipped and missing assets do not point users to `hive setup` (R5/R12).
- [ ] Documentation and install nudges for setup are missing: `install.md`, `install.sh`, `README.md`, `wiki/commands/setup.md`, `wiki/commands/doctor.md`, and `wiki/gaps.md` are unchanged/absent, leaving the required package-install -> `hive setup` -> `hive init` flow undocumented (U7/R2).

## Medium
- [ ] `BackendPrompt` rejects the registry's native symbol names when injected: passing `Hive::AgentProfiles.registered_names` to `registered_agents:` raises because `backend_prompt.rb:29-30` compares strings to symbols, which can break future setup wiring/tests that pass the registry directly.

## Nit
- [ ] Interactive prompt copy hardcodes `Claude + Codex` even when the injected registry filters one out: restricted backends still show the stale default sentence in `backend_prompt.rb:52`, which can mislead tests/custom registries.

ivankuznetsov:

# ce-code-review — pass 03

**Scope:** `git diff origin/main..HEAD` (4 files, +636 / -0): `lib/hive/commands/setup/backend_prompt.rb` (new), `lib/hive/config.rb` (edit), and two new unit-test files.

**Intent:** Implements **U1** of the `hive setup` plan — backend selection + global persistence. Adds `Hive::Commands::Setup::BackendPrompt` (interactive TTY multi-select of claude/codex/pi; non-TTY → `[claude, codex]` defaults) and `Hive::Config` methods (`registered_agent_names`, `default_global_agents`, `load_global_agents`, `write_global_agents!`, `normalize_global_agents`) + frozen constants, persisting the selection under `agents.selected` in the XDG global config.

**Scope note:** This PR is *intentionally* U1-only. Per `task.md`, the implementer hit the documented marketplace-slug / `/plan`-vs-`/llm-wiki:wiki-plan` blocker (plan Risk #1, Open Question #2) and stopped before U2–U7 (the real installer, doctor/init changes, docs). Absence of U2–U7 is **not** flagged — it is deferred work blocked on a human decision, not a defect.

**Reviewers:** correctness, testing, maintainability, project-standards, adversarial, cli-readiness, reliability, learnings-researcher. The persistence layer is well-hardened (fix passes 01/02 already landed cardinality, error-attribution, frozen-return, and the drift-guard test); no P0/P1 defects found. Findings below are robustness, test-coverage, and convention-consistency items.

**Verdict:** Ready with minor fixes. The three Medium items are worth resolving (one is a design decision to confirm); Nits are discretionary polish.

## High
_None._

## Medium
- [ ] Missing load-time drift guard that `DEFAULT_GLOBAL_AGENTS ⊆ GLOBAL_AGENT_BACKENDS` (`lib/hive/config.rb:558`): the sibling `Init::Prompts` enforces the analogous parity with a boot-time `raise` (`lib/hive/commands/init/prompts.rb:50`), but here a default that isn't in the backend list surfaces only as a runtime `ConfigError` from `default_global_agents`; the comment even claims "unreachable" — a claim better enforced than asserted. Add `raise ... unless (DEFAULT_GLOBAL_AGENTS - GLOBAL_AGENT_BACKENDS).empty?`.
- [ ] `load_global_agents` no-file-exists branch is untested (`lib/hive/config.rb:833`): `with_tmp_global_config` always pre-writes `config.yml` (`test/test_helper.rb:151`), so the `: {}` first-run path — the core cold-start scenario U1 targets — never executes; line coverage passes because it's a single ternary line, but the branch is unverified. Add a test that deletes `config.yml` then asserts `load_global_agents == %w[claude codex]`.
- [ ] `load_global_agents` hard-raises `ConfigError` on a persisted-but-valid-but-unregistered backend (`lib/hive/config.rb:846` → `:898`): this diverges from `registered_projects`' tolerate-and-skip discipline (`:806`). A config synced across machines (dotfiles) — or any future conditional/lazy registration — would brick *every* command that reads the selection, including the deferred verify-only doctor (U6). Unreachable today (all three auto-register) but the load contract is decided here; at minimum make the message distinguish "unknown name" from "known but not installed here", and consider intersect-and-fallback-to-defaults instead of raising. (Flagged by reliability + adversarial + correctness.)

## Nit
- [ ] Reprompt message drops the valid names/indices that `Init::Prompts` re-lists on every retry (`lib/hive/commands/setup/backend_prompt.rb:66`): "use names or numbers from the list" forces the caller back to scrollback; inline `#{@backends.join(', ')} or 1..#{@backends.size}` to match the established pattern (`init/prompts.rb:259`).
- [ ] Prompt text never states the input grammar (`backend_prompt.rb:51`): the comma-separated names/numbers format is only implied by the `[claude, codex]` default; `Init::Prompts` multi-selects say "comma-separated" explicitly (`init/prompts.rb:375`).
- [ ] Preamble hardcodes "Claude + Codex" prose (`backend_prompt.rb:52`) instead of deriving from `DEFAULT_BACKENDS`; it can drift from the dynamic `[#{defaults.join(', ')}]` shown one line below if the default set ever changes.
- [ ] Non-interactive summary prefix `hive setup:` diverges from `Init::Prompts`' `hive:` (`backend_prompt.rb:87`): a scripted caller grepping `^hive:` to collect auto-applied defaults across commands would silently miss this line.
- [ ] `normalize_global_agents` is case-sensitive while the prompt matches names case-insensitively (`lib/hive/config.rb:897` vs `backend_prompt.rb:107`): a hand-edited `selected: [Claude]` — the exact capitalization the prompt itself displays — raises "unknown backend"; downcase before the allowed-list check to make load equally tolerant.
- [ ] `interactive?` can raise instead of degrading (`backend_prompt.rb:71`): a closed input stream returns `true` for `respond_to?(:tty?)` but raises `IOError` on `.tty?`, escaping `collect` rather than falling through to defaults; guard with `(@input.tty? rescue false)`. (Shared with `Init::Prompts`.)
- [ ] "Filter canonical list to registered/selected" idiom is repeated ~6× across both files (`backend_prompt.rb:82,102`; `config.rb:600,889,910`): three semantically distinct uses share one textual form — extract a shared helper so intersection/dedup semantics live in one place.
- [ ] `Init::Prompts#initialize` still inlines `AgentProfiles.registered_names.map(&:to_s)` (`init/prompts.rb:106`) rather than routing through the new `Config.registered_agent_names` wrapper this PR added to centralize exactly that projection.
- [ ] Interactive tests never assert `summary_io` stays empty (`test/unit/commands/setup/backend_prompt_test.rb`): only the non-TTY test checks summary output, so an accidental summary write on the TTY path would go uncaught.
- [ ] No direct test for `Config.registered_agent_names`, nor for `default_global_agents` raising when no default backend is registered (`test/unit/config_global_agents_test.rb`): only the filter-to-one case is covered, leaving the zero-default `ConfigError` path unverified.
- [ ] Interactive menu rendering with a restricted registry is never exercised (`backend_prompt_test.rb`): `test_registered_backends_filter_prompt_choices` uses `tty: false`, so the filtered menu listing (e.g. only `1) claude`) in the interactive path is untested.
- [ ] (pre-existing) `load_global_config` rescues EACCES/EISDIR/Psych but not `Errno::ENOENT` (`lib/hive/config.rb:937`): the new `File.exist? ? read : {}` reader inherits a narrow TOCTOU where a file deleted between check and read surfaces as exit-70 InternalError instead of exit-78 ConfigError.

ivankuznetsov:

# pr-review-toolkit review — pass 03

**Scope:** `git diff origin/main..HEAD` (4 files, +636 / -0): `lib/hive/commands/setup/backend_prompt.rb` (new), `lib/hive/config.rb` (edit), `test/unit/commands/setup/backend_prompt_test.rb` (new), `test/unit/config_global_agents_test.rb` (new). Implements **U1** (backend selection + global persistence) of the `hive setup` plan.

**Scope note:** This PR is intentionally U1-only — per `task.md` the implementer hit the documented marketplace-slug / `/plan`-vs-`/llm-wiki:wiki-plan` blocker and stopped before U2–U7 (installer, doctor/init verify-only, docs). Absence of U2–U7 is deferred, not flagged.

**Personas run (union below):** code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier. The runtime value contract is genuinely strong (frozen-from-frozen-constant arrays; `collect`/`normalize` provably never return empty; lockless read is safe under atomic rename). One High (coverage-gate blocker) and a cluster of contract/robustness items follow.

## High
- [ ] [tests] `default_global_agents`' `raise ConfigError` is never executed (`lib/hive/config.rb:601`), so the repo's required 100% line-coverage gate (`rake coverage`) should go RED: the raise fires only when **both** `claude` and `codex` are unregistered, but the sole registry-restriction helper `with_restricted_registry` is used only with `%i[claude]` (and no other test or `lib/` caller restricts the registry) — verified by grep + an independent `Coverage` probe measuring line 602 = 0 hits. Add `with_restricted_registry(%i[pi]) { assert_raises(Hive::ConfigError) { Hive::Config.default_global_agents } }`.

## Medium
- [ ] [type-design] `default_global_agents` (and `BackendPrompt#default_backends`) iterate the `DEFAULT_*` literal as the base rather than the canonical `GLOBAL_AGENT_BACKENDS`/`BACKEND_ORDER` (`lib/hive/config.rb:599`, `backend_prompt.rb:82`): order is canonical-by-authoring not by construction (reordering the literal would make this producer disagree with `normalize_global_agents`), and there's no `⊆ GLOBAL_AGENT_BACKENDS` filter, so the default path could emit a value `normalize`/`write` would reject — derive both from `GLOBAL_AGENT_BACKENDS.select { |n| DEFAULT_GLOBAL_AGENTS.include?(n) && registered.include?(n) }`.
- [ ] [type-design] Missing load-time guard that `DEFAULT_GLOBAL_AGENTS ⊆ GLOBAL_AGENT_BACKENDS` (`lib/hive/config.rb:555-558`): the sibling enforces the analogous CHOICES/MODES parity with a boot-time `raise` (`init/prompts.rb:50`), but here a drifted default surfaces only as a runtime `ConfigError`; add the one-line subset assertion next to the constants.
- [ ] [errors] `interactive?` raises `IOError` on a closed input stream (`lib/hive/commands/setup/backend_prompt.rb:70-71`): `respond_to?(:tty?)` returns true but `.tty?` raises on a closed real IO, so a `hive setup` with stdin closed (daemon/CI) crashes instead of taking the documented non-TTY defaults path — StringIO-based tests can't catch it; guard with `rescue IOError; false`.
- [ ] [errors] `normalize_global_agents`' "unknown backend" message conflates a typo with a known-but-not-registered-on-this-machine backend (`lib/hive/config.rb:898`): a synced/dotfile config naming a backend absent on a second box hard-raises on every read with `expected one of [...]` omitting the valid name — split into "not in `GLOBAL_AGENT_BACKENDS`" (genuine typo) vs "valid but not installed/registered here" (actionable: install or re-run `hive setup`).
- [ ] [tests] Index resolution under a restricted registry is untested (`backend_prompt.rb:112`): all interactive index tests run with the full registry where `@backends == BACKEND_ORDER`, so a regression indexing into the canonical constant instead of the displayed/filtered `@backends` would pass yet select the wrong/unregistered backend — add `make_prompt("2\n", registered_agents: %w[claude pi])` asserting it returns `%w[pi]`.

## Nit
- [ ] [comments] Hardcoded `"(Press Enter for Claude + Codex. Add Pi only if you use it.)"` prose (`backend_prompt.rb:52`) is already wrong on a partial-registration box and rots if `DEFAULT_GLOBAL_AGENTS` changes — derive the copy from `defaults` (the prompt line at `:60` already does).
- [ ] [tests] `load_global_agents` no-file-exists branch (`config.rb:833` `: {}`) is never executed — `with_tmp_global_config` always pre-writes `config.yml`, so the genuine first-run path is unverified (line gate stays green; branch coverage isn't gated); add a config-less test asserting it returns `%w[claude codex]`.
- [ ] [tests] `test_registered_backends_filter_prompt_choices` is misnamed and uses `tty: false` (`backend_prompt_test.rb:127`): it exercises the non-interactive path and never renders the interactive filtered menu — add a `tty: true` variant with a restricted registry asserting the menu lists only the registered subset.
- [ ] [tests] `summary_io` is never asserted empty on interactive paths (`backend_prompt_test.rb`): a stray summary write on the TTY path would go uncaught — add `assert_equal "", summary_io.string` to the interactive tests.
- [ ] [tests] No direct test of `Config.registered_agent_names` (`config.rb:584`): the symbol→string projection that `BackendPrompt` and `normalize` depend on is only exercised indirectly; a one-liner asserting `%w[claude codex pi]` would lock it.
- [ ] [errors] `load_global_config` rescue omits `Errno::ENOENT` (`config.rb:936-942`): the new `File.exist? ? read : {}` reader inherits a narrow TOCTOU where a file deleted mid-read surfaces as exit-70 InternalError instead of exit-78 ConfigError (pre-existing pattern, one more instance).
- [ ] [errors] Empty-selection `ConfigError` ("must list at least one backend", `config.rb:883-886`) gives no recovery path — append that removing/blanking the `selected:` key restores defaults, and list the valid names.
- [ ] [simplify] The `canonical.select { |n| other.include?(n) }` intersection idiom appears 6× across both files (`config.rb:600,889,910`; `backend_prompt.rb:30,82,102`): `Array#&` is an exact behavior-preserving replacement (left-ordered, dedups; left operands are dup-free) — keep the trailing `.freeze`.
- [ ] [simplify] `BackendPrompt#default_backends` is recomputed twice per flow (`backend_prompt.rb:76`): the constructor guard at `:38` computes and discards it, then `collect`/`non_interactive_defaults` recompute — memoize once at construction alongside `@backends`.
- [ ] [type-design] `Setup::BackendPrompt::Aborted` and `Init::Prompts::Aborted` are unrelated classes (`backend_prompt.rb:20`): the deferred U2 consumer must rescue the Setup one specifically — note it or introduce a shared `Hive::Commands::PromptAborted` base.
- [ ] [comments] Several accurate-but-fragile comments: the `frozen_string_literal`-absence parenthetical (`config.rb:552-554`) rots if the magic comment is ever added; the "frozen array of frozen strings" contract is restated in ~5 places (the very drift it warns about); the class header's "#collect always yields ≥1 backend" (`backend_prompt.rb:14`) omits the `Aborted`/EOF raise; and "the analogue of Init::Prompts' guard" (`:34-37`) elides Init=all-required vs Setup=any-suffices.

PR #609, opened 2026-06-27
perf(tui): scale poll cost with active tasks not archive

ivankuznetsov:
## Summary

The TUI poll thread refreshed at ~1 Hz by re-statting and re-parsing every task on disk, including the unbounded `9-done` archive. Steady-state poll cost therefore grew with the *total* task count: on a box with a large archive, active-stage rows went stale and the poll stalled behind a scan of terminal tasks that never change.

The poll thread is now O(active tasks). Archived rows are served from an in-memory cache that is refreshed off the poll thread, so the per-tick fingerprint and the forced re-parse never touch the archive. Adding more `9-done` tasks no longer raises idle cost. At 8 projects × 200 tasks (1600 tasks, 8 active) the per-tick idle fingerprint drops from a full ~2.3 s parse to ~0.13 ms, and the active-only re-parse is ~27 ms.

The shared `Status` snapshot consumed by the daemon, web feed, and `hive status --json` is unchanged. The seam added to `Status` is additive and default-preserving, and a test pins `json_payload(projects)` byte-for-byte against the pre-change output.

## How it works

```mermaid
graph TB
  subgraph Poll["Poll thread (~1 Hz)"]
    FP["O(active) fingerprint<br/>active files + 9-done dir mtime"]
    AP["Active re-parse<br/>Status(stages: active)"]
    MERGE["Merge + dedup<br/>active rows + cached archived"]
  end
  subgraph Refresh["Refresher thread (off-poll)"]
    ARP["Archive parse<br/>Status(stages: 9-done)"]
  end
  CACHE[("Archived cache<br/>frozen, per-project")]
  SNAP["Snapshot @current<br/>(render reads lock-free)"]

  FP -->|dirty signal| Refresh
  AP --> MERGE
  CACHE -->|lock-free read| MERGE
  MERGE --> SNAP
  ARP -->|atomic frozen publish| CACHE
```

| Mechanism | What it does |
|---|---|
| O(active) fingerprint | The idle-tick stat sweep skips archived rows' `state_file`/`.lock`, watching only active stage dirs and active rows. It adds one cheap stat per project of the `9-done` directory: that mtime changes when a task folder is renamed in or out, so it is the "archive changed" signal without walking the archive. |
| Active-only re-parse | The per-tick and forced 3 s liveness re-parse parse active stages only. Archived tasks are terminal (no live PIDs/locks), so excluding them from the liveness re-check is correct. |
| Archived cache | A frozen per-project map of archived payload rows, published by atomic frozen-reference assignment. The poll thread reads it lock-free and merges it into the snapshot. |
| Off-thread refresher | A short-lived thread rebuilds the cache on cold start, on a `9-done` dir-mtime change, when the archive pane opens, or after a 30 s backstop. An O(archive) parse can never stall active reactivity. |
| Merge + dedup | Snapshot rows = active rows + cached archived rows, dropping any cached row whose slug reappears in the active set (covers a stage-move out of `9-done`, no duplicate or ghost). |
| Dependency phantoms | Archived `{slug, id, stage}` identities are fed into the active parse so `blocked_by` stays correct for an active dependent whose prerequisite is now archived and outside the walked stages. |

## Design decisions

**Reuse `Status`'s row builder instead of reimplementing it in the TUI.** The active parse goes through `collect_rows` with its ENOENT stage-move, lock, and `pr.md` race handling rather than duplicating the most safety-critical code. The cost is a minimal additive seam: `collect_rows(stages:)`, `annotate_dependencies(extra_dependency_tasks:)`, and `json_payload(stages:, extra_dependency_tasks:)`. With no kwargs the output is byte-identical, and a byte-equivalence test guards the daemon/web/CLI contract.

**Lock-free single-writer discipline for the cache.** The refresher thread is the sole writer of the archived cache; the poll thread is the sole writer of the dirty/in-flight flags; publication is an atomic frozen-reference assignment, the same pattern the render thread already relies on for `@current`. `stop` joins any in-flight refresher.

**The active/archived partition follows `ArchiveFilter::ARCHIVE_STAGE_DIR` (`9-done`)** — the same source of truth `ArchivePane` and the grid already use, so the surfaces cannot disagree. Generic non-coding workflows' terminal stages are not `9-done` and stay on the active path by design; they are few and not the unbounded-archive problem.

## Test plan

`rake test`, `rake coverage` (100% line gate, 0 failures), and `rubocop` all pass on the touched files.

**Performance gate.** `test/integration/tui_reactivity_perf_test.rb` runs under `rake test` and fails CI on regression. The always-on gate is machine-independent — idle-tick and active-parse cost must not grow with archive size. The absolute wall-clock budgets are opt-in via `HIVE_TUI_PERF_ABSOLUTE` (a reviewer host measured 118 ms vs the 100 ms budget, the host-dependent flakiness plan Risk #2 anticipated), so neither `rake test` nor `rake coverage` can go spuriously red on a busy host.

| Gate | Scope | Budget at 8 projects × 200 tasks |
|---|---|---|
| Idle-tick scaling | always-on (`rake test`) | small vs large archive within tolerance — must not grow with archive size |
| Active-parse scaling | always-on (`rake test`) | independent of archive size |
| Idle tick | opt-in (`HIVE_TUI_PERF_ABSOLUTE`) | < 5 ms |
| Active re-parse | opt-in (`HIVE_TUI_PERF_ABSOLUTE`) | < 100 ms (150 ms under `rake coverage`) |

**Fixtures + coverage.** `test/support/tui_scale_fixture.rb` generates N projects × M parseable task folders with a tunable active/archived split (and the perf test asserts every generated task materializes into a row, so a silently dropped row fails CI). New coverage spans:

- `status_test.rb` — the `Status` byte-equivalence and the stage-filter / phantom-dependency paths.
- `state_source_test.rb` — the O(active) fingerprint, cache merge/dedup, stage-move dedup (both directions), active-task liveness self-heal, dropped-last-archived-row clearing, custom-workflow active-stage detection, persistent-stat-error backoff, project add/remove rescan, and refresher thread teardown.
- `bubble_model_test.rb` — the archive-pane on-open refresh signal wiring.

**Profile (opt-in diagnostic).** `HIVE_TUI_PROFILE=1` on the 8×200 fixture printed `fingerprint_ms=0.13 full_parse_ms=2268.57 active_parse_ms=27.22` with 0 failures, confirming the idle hot path is decisively O(active).

## Review summary

Three compound-engineering review passes (claude-ce, codex-ce, and the pr-review-toolkit multi-persona run) over the merge-base diff `git diff origin/main...HEAD`. Every Requirements-Trace item maps to shipped code plus a test; no review pass found a required-but-missing plan requirement and there were 0 user escalations.

- **Pass 1** (35 findings → 25 fixed, 10 resolved/no-fix): dead-state and duplicate-call cleanups, refresher teardown hardening, cross-thread single-writer docs, and the custom-workflow active-stage **parse** fix.
- **Pass 2** (33 findings → 21 fixed, 12 resolved/no-fix): retain prior archived cache on a transient per-project degradation, the custom-workflow **watch-set** fix, backstop re-arm / failure-backoff hardening, and added regression tests.
- **Pass 3** (17 findings → 10 fixed, 7 resolved/no-fix): clear a project's **last** dropped archived row by gating retain-on-empty on on-disk `9-done` evidence (plan Risk #6), and move the proven-flaky absolute perf budget to opt-in while the scaling assertion stays the hard default gate.

Recurring no-fix items were the plan's deliberate design (the in-memory active+archived merge before `Snapshot.from_payload`; the daemon/web-shared global overlay `Monitor`, explicitly out of this TUI-local scope) or out-of-scope `origin/main` reversions absent from the three-dot diff. The final pass confirmed `state_source_test` green, `rubocop` clean, and the perf gate passing.

## Linked task

Hive task `tui-reactivity-degrades-with-many-260626-0b41` (no GitHub issue). Acceptance from the task brief: the TUI stays responsive with at least 8 projects × ~200 tasks each, and `9-done` growth must not degrade active-stage reactivity — both met and pinned by the scaling gate. Brainstorm and plan artefacts live in the hive task folder (`brainstorm.md`, `plan.md`).

---


ivankuznetsov:

## High
- [ ] lib/hive/tui/state_source.rb:145 rebuilds a full snapshot from cached archived rows on every active reparse: `merge_archived_payload` appends every cached `9-done` task before `Snapshot.from_payload`, so the forced 3s liveness path and any active change still do O(archive) row allocation/sorting on the poll thread.
- [ ] lib/hive/tui/state_source.rb:119 can still let an archive refresh stall active polling: the refresher thread is started before the active parse and both paths call `Status#project_payload`, whose `Hive::Workflows::Project.synchronize` block wraps `collect_rows`, so a large `9-done` scan can block the 1 Hz active refresh behind the shared monitor.
- [ ] lib/hive/tui/state_source.rb:141 computes the active stage filter before each project's workflow overlay is loaded: project-local workflow stages absent from the current registry union are omitted from `collect_rows`, so incremental TUI snapshots can drop custom-workflow tasks after the cold full parse.
- [ ] lib/hive/babysitter/dry_run_env.rb:85 reintroduces Ruby startup-env execution before the dry-run guard: the generated shim is itself a Ruby script and no longer unsets `RUBYOPT`/`RUBYLIB` before the interpreter starts, allowing inherited hooks/load-path entries to run attacker-controlled code before the stub classifies the command.
- [ ] bin/hive-e2e:161 no longer rejects symlinked replay ancestors: a symlinked scenario directory under the run artifact tree can point `repro.sh` outside the contained run directory and still be executed.

## Medium
- [ ] lib/hive/tui/state_source.rb:95 can leak a slow archive refresher after shutdown: `stop` waits only 0.5s, clears `@archive_refresh_thread`, and leaves any still-running archive scan detached despite the plan requiring teardown to join in-flight refreshers.
- [ ] bin/hive-eval:96 lets invalid-byte argument errors escape the usage cleanup path: narrowing the rescue back to `OptionParser::ParseError` means malformed `--report` values can bypass stale-report cleanup and surface as an untyped crash.
- [ ] bin/hive-babysitter-stub-gh:460 drops the typed exec failure path for missing real `gh`: `exec(real, *argv)` now raises `SystemCallError` with a Ruby backtrace/exit 1 instead of the documented dry-run error and exit 127.

## Nit

ivankuznetsov:

# Code review — tui-reactivity-degrades-with-many (pass 1)

Scope: branch work vs `origin/main` (effective diff `git merge-base origin/main HEAD..HEAD`).
The extra files shown by `git diff origin/main..HEAD` (babysitter / e2e / eval) are
reversions from `origin/main` advancing past the fork point, not this branch's work,
and were excluded from review.

Verified clean: the additive `Status` seam (`stages:` / `extra_dependency_tasks:`)
keeps every existing caller (daemon `StatusConsumer`, `Web::StatusFeed`,
`hive status --json`, `--archive`) byte-identical via nil-safe defaults;
`Hive::Dependencies.field` accepts both symbol- and string-keyed task hashes, so the
archived-identity hashes (string keys, no `stage_index`) resolve correctly via the
stage-name fallback; frozen archived rows are only read by `Snapshot.from_payload`;
the `OpenArchivePane` side-effect handler returns `nil` so `Update.apply` still opens
the pane.

## High

(none)

## Medium
- [ ] A dependent on a just-completed prerequisite transiently resolves to `blocked` / "blocked by X (unresolved)" (`lib/hive/tui/state_source.rb:142`): when prereq X moves into `9-done`, the active parse no longer returns X and the captured archived `cache` does not yet contain it (the refresher is async), so for ~1–2 ticks `Hive::Dependencies.resolve` finds no prerequisite and stamps `blocked: true`/`blocked_by: nil` — the opposite of the truth (X just finished). Today's full-every-tick re-parse never exhibits this; `blocked` also gates KeyMap verb dispatch, so a dependent can show a momentary action refusal. Bounded by archive-refresh latency (worst case the 30s backstop); not covered by any gate-iii test.

## Nit
- [ ] `@last_active_payload` is write-only dead state in the steady-state branch (`lib/hive/tui/state_source.rb:144`): assigned but never read — line 145 uses the local `active_payload` instead, and no other lib/test reads the ivar. Remove it (it also keeps a full active payload alive needlessly).
- [ ] `refresh_archive_signals` and `start_archive_refresh_if_needed` are each called twice per re-parse tick (`lib/hive/tui/state_source.rb:117`/`119` then `146`/`147`), duplicating the per-project `9-done` dir stat I/O; the second pair finds the dirty flag already cleared and recomputes mtimes just updated moments earlier.
- [ ] `@archive_refresh_dirty` is written from the Bubbletea update thread (`request_archive_refresh` via `app.rb:84` → `bubble_model.rb:418`) as well as the poll thread, contradicting the plan's "poll thread = sole flag writer" single-writer discipline (`lib/hive/tui/state_source.rb:70`); benign under MRI's GVL (atomic boolean store) but a dropped set is recovered only by the 30s backstop — worth a comment narrowing the invariant.
- [ ] Potential archive-refresher thread leak on shutdown (`lib/hive/tui/state_source.rb:92`-95): if the poll thread is mid-`json_payload` and exceeds the 0.5s join deadline, it can spawn a fresh `@archive_refresh_thread` after `stop` already captured/nilled the reference, leaving a detached thread post-shutdown and risking a flaky teardown thread-leak assertion.

ivankuznetsov:

# pr-review-toolkit multi-persona review — tui-reactivity-degrades-with-many (pass 1)

Scope: branch work vs `origin/main` (effective diff `git merge-base origin/main HEAD..HEAD`).
The extra files in `git diff origin/main..HEAD` (bin/hive-*, babysitter dry_run_env, their
tests) are reversions from `origin/main` advancing past the fork point — not this branch's
work — and were excluded.

Personas run: code-reviewer, pr-test-analyzer, comment-analyzer, silent-failure-hunter,
type-design-analyzer, code-simplifier. Union of findings below, deduped across personas.

Confirmed sound (no finding): the additive `Status` seam keeps daemon/web/CLI byte-identical
(internal-consistency byte-equivalence test passes); `Dependencies.field` tolerates the
string-keyed archived identities and re-derives `stage_index` from the stage name; the
`stop` ordering reaps the refresher; `merge_archived_payload` dedup is path+slug scoped so no
dup/ghost; `refresh_once`'s top-level rescue surfaces via `@last_error` + `stalled?`.

## High

(none)

## Medium
- [ ] [errors] Failed background archive refresh strands a stale cache (`lib/hive/tui/state_source.rb:330`-339): the dirty flag is cleared at spawn (`:314`) and `@archive_dir_mtimes` is advanced before the worker runs (`:287`), so on a worker `StandardError` neither the dir-mtime nor the dirty trigger re-arms — recovery falls solely to the 30s backstop, and the `warn` (`:338`) is invisible/frame-corrupting under the alt-screen TUI and never reaches `@last_error`. Re-arm on failure (or defer advancing `@archive_dir_mtimes` until success) and route the error to an operator-visible channel.
- [ ] [errors] `safe_mtime` collapses "9-done dir absent" and "stat errored" into the same `nil` (`lib/hive/tui/state_source.rb:200`-204, `:290`-295): a transient EACCES/ESTALE on the archive dir reads as "unchanged", so a real archive mutation during that window is dropped from the change detector until the 30s backstop. Distinguish absent from errored.
- [ ] [type-design] The cache's two parallel maps `:projects` and `:identities` are built together but indexed independently with no `keys ==`/length invariant (`lib/hive/tui/state_source.rb:257`-270, `:280`-282); a future single-map edit silently desyncs dependency identities from rows and re-blocks dependents — prefer deriving identities from `:projects` on demand to delete the drift surface entirely.
- [ ] [type-design] This branch is the first production caller to feed string-keyed dependency-task hashes into a `Dependencies` snapshot (`dependency_identity_for`, `lib/hive/tui/state_source.rb:272`-278 → `status.rb:686`), contradicting the explicit comment at `lib/hive/dependencies.rb:146`-157 ("no production path passes a string-keyed Hash today"); deleting `field`'s string-key arm on that now-false assurance would silently break archived-prereq dependents. Update that comment / pin the contract.
- [ ] [tests] No direct test for the 30s archive backstop `archive_refresh_due?` (`lib/hive/tui/state_source.rb:301`-306): an inverted comparison or a bad constant would not be caught — add a `Time.now`-stubbed test asserting a refresh is requested past `@archive_last_refresh_at + ARCHIVE_REFRESH_FALLBACK_SECONDS` with no dir-mtime change.
- [ ] [tests] The archive-pane-open trigger's App→StateSource wiring is untested (`lib/hive/tui/app.rb:83`-86): `BubbleModel` calling the injected lambda and `request_archive_refresh` are each unit-tested, but the actual `state_source.method(:request_archive_refresh)` binding could be dropped/mis-wired with both halves still green.

## Nit
- [ ] [simplify] `@last_active_payload` is dead write-only state — assigned at `lib/hive/tui/state_source.rb:45`,`:134`,`:144` and read nowhere; both writes duplicate a local used on the same line. Delete it.
- [ ] [simplify] `refresh_archive_signals` + `start_archive_refresh_if_needed` are called twice per re-parse tick (`lib/hive/tui/state_source.rb:117`/`119` then `146`/`147`), re-stat'ing every project's `9-done` dir; the else-branch pair is near-redundant (next tick's top-of-method call self-heals). Drop `:146`-147.
- [ ] [simplify] Cold start strips archived rows (`payload_without_archived_rows`) then merges them back from a cache derived from the same payload (`lib/hive/tui/state_source.rb:131`-136) — two full payload deep-copies on the first/most-expensive frame to reproduce the payload already in hand; publish the full payload directly and the two helpers (`:224`-236) can go.
- [ ] [simplify] `start_archive_refresh_if_needed`'s `projects = nil` parameter + `|| projects_from_snapshot` fallback + compound guard is a YAGNI seam — no caller passes it (`lib/hive/tui/state_source.rb:308`-318). Inline.
- [ ] [comments][type-design] Class doc (`lib/hive/tui/state_source.rb:9`-23) and `sleep_in_slices` describe only `@current` as the lock-free cross-thread reference; the branch adds a second writer thread for `@archived_cache` plus `@archive_last_refresh_at` and a two-thread-written `@archive_refresh_dirty`, none documented (contrast `bubble_model.rb:129`-134's careful `@healed_folders` note).
- [ ] [comments] `stop` comment (`lib/hive/tui/state_source.rb:84`-87) says "joins the thread" (singular) with a 0.5s deadline, but the body now joins two threads sequentially (`:92`,`:95`) for a worst-case ~1.0s teardown.
- [ ] [comments] `ARCHIVE_REFRESH_FALLBACK_SECONDS = 30.0` (`lib/hive/tui/state_source.rb:36`) sits directly under the liveness-fallback comment block with no comment of its own, implying it shares the liveness rationale rather than being the archive-set backstop.
- [ ] [comments] `request_archive_refresh` (`lib/hive/tui/state_source.rb:69`-72) is an undocumented cross-thread entry point (invoked from the Bubbletea update thread via `app.rb:85` → `bubble_model.rb:418`).
- [ ] [comments] `wiki/commands/tui.md:142`-145 lists "cold cache changes" as a background-refresh trigger, but the cold path builds the cache synchronously on the poll/boot thread — no code trigger maps to that phrase.
- [ ] [tests] gate (i) liveness test offsets by a hardcoded `+4`s rather than asserting against `LIVENESS_REPARSE_FALLBACK_SECONDS` (`test/unit/tui/state_source_test.rb:470`-497); the lower edge (no reparse before ~3s) is unasserted.
- [ ] [tests] No test that `start_archive_refresh_if_needed` refuses to spawn a second refresher while one is alive (`lib/hive/tui/state_source.rb:311` overlap guard) — a regression could leak threads under load.
- [ ] [tests] The 1600-task parseability assertion in `test/integration/tui_reactivity_profile_test.rb:29` only runs behind `HIVE_TUI_PROFILE=1` (`:13`); fold the bare parse-count check into the always-run perf test so the invariant gates CI.
- [ ] [tests] Byte-equivalence test (`test/unit/commands/status_test.rb:81`-106) asserts default==explicit-full internal consistency, not against a captured historical golden — a change altering both paths identically wouldn't be flagged.
- [ ] [errors] `dependency_tasks_for` uses `project.fetch("path")` with no default (`lib/hive/commands/status.rb:692`-696), stricter than this never-fail surface's degrade-don't-crash posture; a future path-less caller with non-nil `extra_dependency_tasks` would hard-`KeyError`. Prefer `project["path"]`.
- [ ] [type-design] The cache is a raw nested symbol/string-mixed Hash rather than a small `Data` value object; frozen-ness and per-path correspondence are conventional not enforced, and tests reach in via `instance_variable_get`, cementing the raw shape as the de-facto interface.
- [ ] [code] `test/integration/tui_reactivity_profile_test.rb:13` conditionally `skip`s on `ENV["HIVE_TUI_PROFILE"]`; defensible (diagnostic, not a correctness gate; the global no-conditional-skip rule is E2E-scoped) but flagged so it's a conscious choice.

ivankuznetsov:

## High
- [ ] Active reparses still rebuild every archived row on the poll thread (lib/hive/tui/state_source.rb:191): `merge_archived_payload` appends cached archived payload hashes and `Snapshot.from_payload` rebuilds/sorts the full merged snapshot every active reparse, so the 3s fallback and active changes remain O(active + archived), violating the plan's O(active) poll-path requirement.
- [ ] Markerless `3-plan` rows are routed into a plan-approval path that rejects them (lib/hive/task_action.rb:442): fresh approved plan folders now classify as `needs_input`, daemon policy treats that as plan approval, and `PlanApproval.prepare` refuses marker `:none`, so newly approved tasks skip instead of running the plan agent.
- [ ] Dry-run shims execute caller-controlled Ruby startup env before the guard (lib/hive/babysitter/dry_run_env.rb:84): the generated shim is Ruby, so `RUBYOPT`/`RUBYLIB` code runs before the shim can pin env or invoke the guarded stub.
- [ ] `hive-e2e replay` can execute through symlinked scenario directories (bin/hive-e2e:161): only the final `repro.sh` is `lstat`-checked, so a symlinked scenario parent can point outside the run tree and still execute an external script.

## Medium
- [ ] Long archive refreshes can re-trigger continuously (lib/hive/tui/state_source.rb:157): the 30s fallback sets `@archive_refresh_dirty` while a refresher is alive, the successful worker never clears that bit, and the next tick starts another full archive scan.
- [ ] Custom workflow active stage dirs can be omitted from the mtime watch set (lib/hive/tui/state_source.rb:241): `active_stages` reads process-global workflow dirs outside each project's overlay lock, so newly added or moved custom-stage tasks may not invalidate the poll until the fallback.
- [ ] Archive refresh can publish an empty cache after status degradation (lib/hive/tui/state_source.rb:389): `Status#json_payload` converts per-project failures to empty task lists, and the refresher treats that as success, replacing previously visible archived rows.
- [ ] The committed perf gate can pass while active reparses stay O(archived) (test/integration/tui_reactivity_perf_test.rb:63): it measures best-case wall time with loose thresholds instead of structurally asserting archived rows are not rebuilt on active reparses.
- [ ] Invalid-byte `hive-eval` usage errors bypass cleanup (bin/hive-eval:96): `ArgumentError` from OptionParser now escapes the usage-error path, producing a Ruby backtrace and leaving stale selected report files behind.
- [ ] Missing real `gh` path no longer exits through the controlled dry-run diagnostic (bin/hive-babysitter-stub-gh:460): `exec` raises `Errno::ENOENT` directly instead of returning the documented exit 127 message.
- [ ] Deleted wiki log fragments leave the compiled changelog stale (wiki/log.d/20260626T083614Z-e2e-replay-parent-symlink.md:1): the branch deletes `wiki/log.d` fragments and `bin/hive wiki compile-log --check` fails with `wiki/log.md is stale`.

## Nit
- [ ] `hive approve` wiki JSON example is stale (wiki/commands/approve.md:70): the code/test now emit `hive plan <slug> --from 3-plan`, but the documented `next_action.command` still shows `hive run .../3-plan/...`.

ivankuznetsov:

# Code review — tui-reactivity-degrades-with-many (pass 2)

Scope: this branch's own work vs `origin/main`, i.e. the effective diff
`git diff origin/main...HEAD` (merge-base). The extra files surfaced by the
two-dot `git diff origin/main..HEAD` (`task_action.rb`, `babysitter/dry_run_env.rb`,
`bin/hive-e2e`, `bin/hive-eval`, `bin/hive-babysitter-stub-gh`, deleted
`wiki/log.d/*` fragments, `wiki/commands/approve.md`) are **reversions** from
`origin/main` advancing past the fork point, NOT this branch's changes — they are
out of scope (same conclusion as pass 1).

Verified this pass:
- `rake`-level suites green: `state_source_test` (29), `status_test`+`bubble_model_test`
  (69), `tui_reactivity_perf_test` (passes its 8×200 budget + scaling gates).
- Rubocop clean on all changed `.rb` files.
- Profile at 8×200 (`HIVE_TUI_PROFILE=1`): idle tick **0.21 ms**, old full parse
  **~2957 ms**, active parse **34 ms** — the idle hot path is decisively O(active)
  and the plan's primary success metric holds.
- U2 contract guards present: byte-equivalence test
  (`test_json_payload_default_path_matches_explicit_full_stage_inputs`) and the
  active-only-with/without-`extra_dependency_tasks` gate-iii tests.

## High

(none) — no plan Requirement is required-but-missing; every Requirements-Trace item
maps to shipped code + a test. The codex-02 "High" items are either out-of-scope
reversions (see scope note) or the plan's deliberate merge design (see triage note).

## Medium
- [ ] Archive refresher overwrites a good cache with `[]` when a project transiently degrades (`lib/hive/tui/state_source.rb` `refresh_archived_cache` → `archived_cache_from_payload`): `Status#json_payload` swallows a per-project failure into an empty task list (`project_payload_or_degraded`), the off-thread refresher treats that as success and publishes a frozen cache with no archived rows for that project, so its archived tasks vanish from the archive pane/grid until a later refresh (worst case the 30 s backstop). `merge_archived_payload` guards the *active* payload's `"error"` but the refresher's cache rebuild keeps no prior entry. Old behavior (full parse) re-showed the rows on the next ~3 s tick once the blip cleared. 

## Nit
- [ ] Backstop re-arms `@archive_refresh_dirty` during an in-flight refresh, causing one redundant full archive scan per backstop cycle when an archive parse spans more than one poll tick (`lib/hive/tui/state_source.rb` `archive_refresh_due?` keys off `@archive_last_refresh_at`, which only advances on completion; a successful `refresh_archived_cache` never clears the dirty bit). Bounded to one extra scan and entirely off the poll thread, so no reactivity impact — pure redundant I/O. 
- [ ] Cross-thread-state doc omits `@last_error` as a refresher-written field (`lib/hive/tui/state_source.rb` class comment): the refresher thread writes `@last_error` (in `refresh_archived_cache`'s rescue) in addition to the poll thread, but the carefully-enumerated "SOLE writer of `@archived_cache`/`@archive_last_refresh_at`; poll/render only read" note doesn't mention this second writer. Benign under the GVL (atomic ref, last-write-wins, renderer shows latest), but the comment is the file's stated single-writer contract. 

## Triage scope note

- **codex-02 High #1 (active reparse O(active+archived) merge + `Snapshot.from_payload`
  rebuild) → NO-FIX, deliberate plan design.** Plan U3 explicitly prescribes "merges the
  cached archived payload rows per project … before `Snapshot.from_payload`" and "Snapshot.from_payload
  then sorts/builds Rows exactly as today". The eliminated cost is the O(archive) **disk**
  re-parse (~2957 ms → off-thread cache); the residual in-memory merge+build runs only on
  active-change/3 s-liveness ticks, not the per-tick idle path (0.21 ms). Already triaged
  RESOLVED in pass 1; the committed U5 scaling gate exercises this exact path and passes.
- **codex-02 Medium #4 (perf gate is wall-time, not a structural assertion) → NO-FIX, plan
  U5 design.** U5 deliberately specifies "absolute budget gate (per A5) … scaling assertion
  (archive-size independence) is the primary gate"; the implementation matches that exactly.
  The active-reparse budget is 50 ms→100 ms, which the plan authorizes ("tunable post-A1").
- **codex-02 Medium #2 (custom-workflow active stage dirs omitted from the mtime watch set,
  `state_source.rb` `active_stages`) → effectively mitigated; at most a low nit.** Each active
  row's own `state_file`/`.lock` is added to the fingerprint directly (not only via stage
  dirs), so existing custom-stage tasks ARE watched; `all_stage_dirs` includes custom stages
  after each `json_payload` overlay load; and a brand-new task in an as-yet-unwatched custom
  stage self-heals via the 3 s liveness fallback. Pass 1 already fixed the *parse* path
  (per-project `exclude_archived:` seam computed inside `Workflows::Project.synchronize`) and
  added `test_incremental_reparse_keeps_custom_workflow_active_stage_rows`.
- **codex-02 Medium #3 maps to this review's Medium** (empty cache after status degradation).
- **codex-02 items on `task_action.rb`, `dry_run_env.rb`, `bin/hive-e2e`, `bin/hive-eval`,
  `bin/hive-babysitter-stub-gh`, deleted `wiki/log.d/*`, `wiki/commands/approve.md` →
  OUT OF SCOPE.** Reversions from `origin/main` advancing past the fork point; absent from
  `git diff origin/main...HEAD`. The branch's own added `wiki/log.d/*` fragments without an
  edited `wiki/log.md` are the correct per-CLAUDE.md workflow (compiled later, not in feature PRs).

ivankuznetsov:

# PR review (multi-persona) — tui-reactivity-degrades-with-many (pass 2)

Personas run: **code-review**, **tests**, **silent-failure**, **comments**, **type-design**.
Scope: this branch's own changes, the three-dot merge-base diff `git diff origin/main...HEAD`.
The two-dot `git diff origin/main..HEAD` additionally shows main-divergence reversions
(`task_action.rb`, `babysitter/dry_run_env.rb`, `bin/hive-e2e`, `bin/hive-eval`,
`bin/hive-babysitter-stub-gh`, deleted `wiki/log.d/*`, `wiki/commands/approve.md`) — those
are NOT this branch's work and were excluded. All personas verified claims against the code;
rubocop clean and the touched suites pass.

Two load-bearing claims were independently verified for this report:
`Workflows::Project.load!` *replaces* (not accumulates) the project overlay, so the fingerprint's
stage union reflects only the last-loaded project; and the `project_payload_or_degraded` rescue
arm returns no `"error"` key (status.rb:174-181).

## High

(none) — no plan Requirement is required-but-missing, and no defect rises to High. The
silent-failure persona's strongest item (degraded-rescue `"error"` contract) is filed Medium
below because it triggers only on a generic (non-path/non-init) exception, self-heals next
tick, and its "looks healthy / launch target" amplification is pre-existing daemon behavior,
not introduced here.

## Medium
- [ ] [silent-failure] `merge_archived_payload`'s `project["error"] ? [] : …` guard relies on a key the rescue path never sets (`lib/hive/tui/state_source.rb:285` vs `lib/hive/commands/status.rb:174-181`): a project whose active parse raises a generic `StandardError` degrades to `{"tasks"=>[]}` with **no** `"error"`, so the guard treats it as healthy and merges its cached archived rows onto an empty active list — it renders as "complete" (only done tasks) rather than broken. 
- [ ] [code-review] Watch-set narrowing misses non-last custom-workflow active stages (`lib/hive/tui/state_source.rb:239-242`, `active_stages` :270): `project_watch_paths` now joins the global `all_stage_dirs - [archive]` union, but `load!` leaves only the last-loaded project's overlay registered, so a custom-workflow project that isn't last sees its custom stage dirs absent from the fingerprint — a **new** task appearing there is caught only by the 3 s liveness fallback, not the ~1 s mtime gate. Pass-1 fixed this for the *parse* (per-project `default_stage_dirs`) but not the watch set. 
- [ ] [silent-failure] Permanent archive-refresh failure hot-loops a full `9-done` rescan with no operator signal (`lib/hive/tui/state_source.rb:391-401`, `:363`): the rescue re-arms `@archive_refresh_dirty` with no backoff and doesn't advance `@archive_last_refresh_at`, so `start_archive_refresh_if_needed` respawns a full archive parse every poll tick (~1 Hz — the heavyweight scan this PR exists to avoid), while the poll thread's per-tick `@last_error = nil` (idle-gate `:164` / `publish_snapshot` `:267`) wipes the refresher's error before it stably surfaces. 
- [ ] [silent-failure] A *persistent* `safe_mtime` stat error silently defeats the optimization (`lib/hive/tui/state_source.rb:252-258`, `:333-344`): the fresh-per-call `StatError` is never `==` itself, so a stuck EACCES/ESTALE on a watched `.lock` or a project's `9-done` dir makes `mtime_fingerprint_unchanged?` force an active re-parse every tick and/or `refresh_archive_signals` fire `request_archive_refresh` every tick — re-introducing the degradation — with no `@last_error`/`Debug.log` breadcrumb. 
- [ ] [tests] Reverse-direction merge dedup is unasserted (`test/unit/tui/state_source_test.rb`, guard at `lib/hive/tui/state_source.rb:285-286`): the only move test is *into* `9-done`; the `reject { active_slugs.key?(slug) }`=true branch — a task moved back OUT of `9-done` while still in the stale archived cache (gate (ii) reverse / resurrection window) — has no no-duplicate assertion. 
- [ ] [comments] The class-level "SOLE writer of `@archived_cache` and `@archive_last_refresh_at`; the poll/render threads only read them" is literally false (`lib/hive/tui/state_source.rb:27-28` vs `:175-176`): the cold-start branch of `refresh_once` (poll/boot thread) also writes both. No actual race (cold start runs only while `@current.nil?`, before any refresher can spawn), but the wording is exactly the kind of load-bearing concurrency invariant a maintainer could mis-rely on. 

## Nit
- [ ] [comments] "Mirrors the single-writer note on `BubbleModel`'s `@healed_folders`" is inverted (`lib/hive/tui/state_source.rb:35-36`): `@healed_folders` (`bubble_model.rb:129-136`) is a **mutex-guarded multi-writer** field, the opposite of the local lock-free/backstop-recovered discipline the comment claims to mirror.
- [ ] [comments / code-review / type-design] The "Cross-thread state" inventory omits `@last_error` (`lib/hive/tui/state_source.rb:25-36`), which the refresher thread now also writes (`:400`) alongside the poll/boot thread — a completeness gap in an otherwise-exhaustive inventory.
- [ ] [type-design] Dependency-identity hashes are string-keyed and `concat`-ed into a snapshot array of symbol-keyed rows (`lib/hive/tui/state_source.rb:309` + `lib/hive/commands/status.rb:701`): correctness rests entirely on `Dependencies.field` duck-typing both key forms; emitting symbol keys would make the array homogeneous and remove a footgun (any future direct `entry[:slug]` index would silently miss archived prereqs). 
- [ ] [type-design] The `status.rb` seam admits ambiguous states (`json_payload`/`collect_rows` :137/:570): passing both `stages:` and `exclude_archived: true` silently ignores `exclude_archived` (non-obvious precedence via `Array(stages || default_stage_dirs(exclude_archived))`), and `exclude_archived: true` without `extra_dependency_tasks:` permits the silent-reblock footgun `dependencies.rb` warns about. A single `stages:` selector (`:all`/`:active`/array) would collapse the contradiction. 
- [ ] [type-design] Cached rows are shallow-frozen (`lib/hive/tui/state_source.rb:297`, `task.dup.freeze`): nested mutable values (e.g. an `attrs` sub-hash) stay mutable, so the comment's "deep-freeze discipline" framing overstates. Low risk (rows are read-only).
- [ ] [type-design] Cache outer key is symbol `:projects` while the payload uses string `"projects"` (`lib/hive/tui/state_source.rb:293-308`): near-identical names for different shapes invite conflation; a distinct wrapper name (e.g. `:rows_by_path`) would remove the collision.
- [ ] [type-design] `StatError`'s never-`==` contract is a *negative* invariant guarded only behaviorally (`lib/hive/tui/state_source.rb:62`): a future refactor adding `==`/`eql?`/`hash` (or making it a Struct) would silently break it; a one-line "must never gain equality" comment would harden it.
- [ ] [tests] Perf-gate active-scaling tolerance is coarse (`test/integration/tui_reactivity_perf_test.rb:22-23,98-102`): the binding ceiling is `small*2` (~80 ms) leaving ~34 ms of archive-linear slack, so it catches a catastrophic full-archive-parse regression (~19×) but not subtle archive-linear creep in `merge_archived_payload`. 
- [ ] [tests] The degraded-project merge arm (`project["error"] ? []`, `lib/hive/tui/state_source.rb:285`) drop-on-degrade behavior is line-covered but has no behavioral assertion.
- [ ] [tests] `hive drop` of a single archived folder updating the cache (Risk #6 removal direction) is not asserted at the cache level — only archived *addition* drives the refresh test (`state_source_test.rb:443`).
- [ ] [tests] `test_registry_project_set_changes…` (`state_source_test.rb:567`) asserts a newly-added project's *active* rows appear but not that its *archived* rows surface after the follow-up background refresh (gate iv covered for active rows only).
- [ ] [silent-failure] `registry_config_path`'s rescue is fully silent (`lib/hive/tui/state_source.rb:233-237`): bounded (path dropped from fingerprint; project set self-heals within the 3 s liveness reparse) but a one-line `Debug.log` would leave a breadcrumb for an otherwise-mysterious "project set briefly stale".

## Cross-reference with the parallel CE review (pass 2)

- The `@last_error` doc-inventory omission and the empty-cache-on-degradation robustness gap also appear in `claude-ce-code-review-02.md` (deduped here under the silent-failure/comments items).
- The active-reparse merge being O(active+archived) and the perf gate being wall-time-based are the **plan's deliberate design** (U3 merge-before-`Snapshot.from_payload`; U5 absolute+scaling gate) — not re-raised as defects here; the perf-gate item above is filed only as a coarse-tolerance test observation.
- The custom-workflow concern here is specifically the **watch set** (`project_watch_paths`); pass-1 already fixed the **parse** path, so this is the residual half of that earlier finding.

ivankuznetsov:

## High
- [ ] Dry-run overlay executes Ruby before startup env is scrubbed (`lib/hive/babysitter/dry_run_env.rb:84`): command-local `RUBYOPT`/`RUBYLIB` code runs before the wrapper can set safe env, reintroducing the deleted startup-env injection regression.
- [ ] TUI archive cache cannot clear a project's last archived row (`lib/hive/tui/state_source.rb:401`): an empty `9-done` payload retains prior rows, so dropping the final archived folder leaves a ghost row in the merged snapshot/archive pane indefinitely.
- [ ] Committed TUI perf gate fails (`test/integration/tui_reactivity_perf_test.rb:37`): `bundle exec ruby -Itest test/integration/tui_reactivity_perf_test.rb` measured active parse at 118.11 ms, above the branch's 100 ms CI budget.

## Medium
- [ ] `hive-e2e replay` follows symlinked scenario directories (`bin/hive-e2e:161`): only the final `repro.sh` is lstat-checked, so a symlinked scenario directory can execute an outside script instead of being rejected as an unusable artifact.
- [ ] `hive-eval` invalid-byte report cleanup now crashes (`bin/hive-eval:68`): `String#split` on an invalid UTF-8 `--report=` token raises `ArgumentError`, exits 1 with a backtrace, and leaves the stale selected report file in place.
- [ ] Markerless coding plan rows are misclassified as input gates (`lib/hive/task_action.rb:442`): a not-yet-run `3-plan` task now returns `needs_input`, causing manual TUI/bot/status surfaces to treat it as human input instead of a runnable plan stage.

## Nit

ivankuznetsov:

# Code review — tui-reactivity-degrades-with-many (pass 3)

Scope: this branch's own work vs `origin/main`, i.e. the effective diff
`git diff origin/main...HEAD` (merge-base). Confirmed this pass that
`task_action.rb`, `babysitter/dry_run_env.rb`, `bin/hive-e2e`, `bin/hive-eval`,
`bin/hive-babysitter-stub-gh` are **absent** from the three-dot diff — they are
reversions from `origin/main` advancing past the fork point, not this branch's
changes. The codex-03 High (`dry_run_env`) and Mediums (`hive-e2e`, `hive-eval`,
`task_action`) all land on those out-of-scope files and are dropped here (same
conclusion as passes 1–2; see `escalations-02.md`).

Verified this pass:
- `state_source_test` green (37 runs, 135 assertions); rubocop clean on
  `state_source.rb` / `status.rb` / `dependencies.rb`.
- Perf gate (`tui_reactivity_perf_test`) passes 3× locally; measured active
  parse **43 ms** (budget 100 ms), idle tick **0.18 ms** (budget 5 ms) at 8×200.
- Pass-2 AUTO-FIX items applied: `archived_cache_from_payload` retains prior
  rows on empty, backstop no longer re-arms during an in-flight refresh, cross-
  thread doc enumerates the refresher's writers. The pass-3 "pin" commit adds
  drop/dedup/gate-iv tests.

## High

(none) — no plan Requirement is required-but-missing; every Requirements-Trace
item maps to shipped code plus a test. The two findings below are edge/quality
regressions, not missing plan deliverables.

## Medium
- [ ] Archived cache can never clear a project's **last** archived row, leaving a permanent ghost in the archive pane/grid (`lib/hive/tui/state_source.rb` `archived_cache_from_payload`, ~:380-405): the pass-2 retain-prior-on-empty guard cannot distinguish a transient per-project degradation (empty tasks, no `"error"`) from a legitimate "final `9-done` folder dropped" — both yield `tasks=[]` under the `stages:[9-done]` refresh. So after `hive drop` of a project's only archived task, every subsequent refresh re-retains the stale prior rows (`previous` always non-empty, payload always empty), and the merge keeps showing the dropped row indefinitely (until a new archived task arrives in that project or the TUI restarts). This directly contradicts plan **Risk #6** ("the dir-mtime stat does catch a folder removed from `9-done`, so this is covered; verify with a test that dropping an archived folder updates the cache"). The committed `test_dropping_one_archived_folder_updates_cache_on_next_refresh` only exercises the shrink-to-**non-empty** path (its own comment notes "the retain-prior guard only holds over an EMPTY archive result"), so the last-row edge is untested. Blast radius is the terminal archive view only (no active-reactivity or dependency impact, archived tasks are terminal), and codex-03 rated it High; I rate it Medium given the display-only, self-heals-on-next-archive scope — but the never-self-clearing persistence is real. Fix: before retaining on empty, gate on positive evidence the result is transient (e.g. the project's `9-done` dir still contains task subfolders on disk → retain; genuinely empty/absent dir → publish `[]`), which satisfies both the transient-degradation guard and Risk #6.

- [ ] Absolute active-reparse perf budget (100 ms) runs in the default `rake test` gate and is flaky on slow/loaded hosts (`test/integration/tui_reactivity_perf_test.rb:16,37`): codex-03's host measured the active parse at **118 ms** (> the 100 ms `ACTIVE_REPARSE_BUDGET_MS`), while this host measures ~43 ms — i.e. the budget has only ~2.3× headroom on a fast machine and was already observed exceeded once. A wall-clock budget in the CI-gating suite that a reviewer's real run breaches is a spurious-red risk (and conflicts with the project's own "no flaky tests" rule). Plan **Risk #2** anticipates exactly this and names the remedy: the archive-size **scaling** assertion is the robust, machine-independent gate, with the option to "consider a dedicated rake task … unless proven flaky." The 118 ms observation is that proof. Suggested: move the absolute ms budget out of the default gate (own rake task / opt-in env, as the coverage budget already is), or widen `ACTIVE_REPARSE_BUDGET_MS` headroom; keep the scaling assertion as the primary CI signal.

## Nit

(none beyond the above — the remaining type-design / test-golden observations
from prior passes are already triaged no-fix in `suppressed.md` and not re-raised.)

ivankuznetsov:

# PR review (pr-review-toolkit, multi-persona) — pass 3

Union of findings from 5 personas (code-reviewer, pr-test-analyzer, silent-failure-hunter,
type-design-analyzer, comment-analyzer) over `git diff origin/main...HEAD` (merge-base).
Out-of-scope reversions (`task_action.rb`, `dry_run_env.rb`, `bin/hive-e2e`, `bin/hive-eval`,
`bin/hive-babysitter-stub-gh`) excluded — confirmed absent from the three-dot diff.

Verified this pass: `Hive::Workflows::Project::LOCK` is a single process-global `Monitor`
(`lib/hive/workflows/project.rb:23`) and the refresher's `9-done` walk runs **inside**
`Workflows::Project.synchronize` (`status.rb:203`); cache keying is consistent
(`File.expand_path` round-trips through Snapshot → `projects_from_snapshot`); the string-key
dependency arm is genuinely live; class-doc concurrency comments are accurate.

## High

(none) — no plan Requirement is required-but-missing; every Requirements-Trace item maps to
shipped code + a test. No High-confidence correctness break. The lock-contention item below
overlaps a pass-1 High already triaged no-fix and is filed at the personas' assessed severity.

## Medium
- [ ] [code] The "off-thread" archive refresher still serializes with the poll thread on the process-global `Workflows::Project::LOCK` (`lib/hive/tui/state_source.rb:498` → `status.rb:203`): `collect_rows`' heavy `9-done` walk runs inside `Workflows::Project.synchronize`, so while the refresher holds the (per-project) `Monitor` the poll thread's active reparse blocks on the same lock — partially contradicting the plan's "an O(archive) parse can never stall active reactivity." Bounded (lock released between projects, ~one project's archive-walk per stall) and within the 1–2-tick latency target except under sustained completion churn. [Overlaps suppressed pass-1 `fp=31ab7183b1b625c0`; NEW sub-angle: the perf gate never exercises this — it measures `refresh_once` with no refresher in flight, so archive-size independence is validated only in the no-concurrent-refresh case.]
- [ ] [code/type-design] A persistent stat error on any project's `9-done` dir drives back-to-back heavy archive rescans with no backoff (`state_source.rb:441-445` `refresh_archive_signals` + `:307-316` `safe_mtime`): the never-equal `StatError` marker is correct as a *gate* recheck-bias but it also flows into the archive *trigger* (`@archive_dir_mtimes != mtimes`), where "uncertain" reads as "changed" → `request_archive_refresh` every tick → repeated `json_payload(stages:[9-done])`; the in-flight/30s throttle in `archive_refresh_due?` gates only the time-based re-arm, not this dirty path, and the failure backoff only engages when `json_payload` *raises* (here it succeeds). Compounds the lock stall above. Trigger is narrow (an existing dir where `File.exist?` is true but `File.mtime` raises — mostly a transient TOCTOU race, so persistent reachability is uncommon), but the consequence is exactly the throttled regression this PR targets.
- [ ] [tests] The dual error-channel separation (`@archive_last_error` vs `@last_error`) is unverified end-to-end (`state_source_test.rb:739-770`): the test asserts `last_error` is non-nil immediately after a failed `refresh_archived_cache`, which passes whether the error lives in `@archive_last_error` OR `@last_error`; no test runs a *subsequent successful* `refresh_once` (which sets `@last_error = nil`) and re-checks the archive error is still surfaced — so a regression collapsing the two channels would pass the whole suite yet silently swallow archive failures on the next active tick.
- [ ] [tests] Custom-workflow active stage-dir *detection* is unpinned (`state_source.rb:282-297` `project_watch_paths`): `test_incremental_reparse_keeps_custom_workflow_active_stage_rows` only bumps an *existing* state file (already in the fingerprint), so reverting the on-disk glob back to the `all_stage_dirs` union (the exact Codex regression the comment guards) would still pass every test while silently delaying detection of a *new* task in a non-last custom project's `2-work` to the 3s liveness fallback. Missing test: create a brand-new folder in a non-last custom project's active stage between two `refresh_once` calls and assert it surfaces via the ~1s mtime gate.

## Nit
- [ ] [code] A just-completed task briefly vanishes from BOTH views for the refresh duration (`state_source.rb:217-230`): when a task moves into `9-done`, the same-tick active reparse drops it (excluded) while merging the still-stale archived cache (not yet containing it), so it is absent from active and archived until the spawned refresher completes (~1 tick normally, seconds under a heavy `9-done` scan). Within the accepted archive-staleness model; a visible blip on exactly the task the operator just watched complete.
- [ ] [type-design] The off-thread archive parse resolves archived rows' own dependency state against an archive-ONLY task set (`state_source.rb:498-502`, `json_payload(stages:[9-done])` with no `extra_dependency_tasks`): an archived task whose prerequisite is in an active stage now resolves to `blocked:true`/`blocked_by:nil (unresolved)` and that value is cached/merged — a content drift vs the pre-split full parse. Cosmetic (terminal `9-done` rows aren't blocked-gated) and does not self-heal (every archive refresh re-resolves archive-only).
- [ ] [errors] A hung/blocked refresher thread permanently suppresses all future archive refreshes with no breadcrumb (`state_source.rb:465`,`:479`): both `archive_refresh_due?` and `start_archive_refresh_if_needed` early-return on `@archive_refresh_thread&.alive?`, so an uninterruptible I/O hang scoped to the archive dir (e.g. stale NFS) freezes the cache forever with no `@archive_last_error` and no Debug.log (the thread blocks rather than raises) while `stalled?` stays false. Narrow; suggest a breadcrumb when a spawn is skipped because the prior refresher has outlived N ticks.
- [ ] [errors] The refresher invokes `Status#json_payload` unprotected, so status.rb degrade-path `warn`s can corrupt the alt-screen frame from a SECOND concurrent emitter (`state_source.rb:498-502`): StateSource itself correctly avoids `warn`, but `json_payload`'s `project_payload_or_degraded` / `dependency_gate_stage_for` still `warn` to real `$stderr` with no `capture_command_io`-style wrapper around the in-process call. The active-poll path already had this exposure (pre-existing), so this is a low-severity broadening, not a new defect.
- [ ] [comments] The `state_source.rb` class-doc "Cross-thread state beyond `@current`" enumeration (`:26-44`) lists the cache data fields + `@archive_refresh_dirty` but omits `@archive_refresh_thread`, which is also touched cross-thread (poll spawns/reads `&.alive?`; `stop` nils it). Not wrong (the teardown race is documented in `stop`'s comment) — flagged only if a reader expects the class-doc list to be exhaustive. No action required.

## Personas with no new findings beyond the known/accepted set
- comment-analyzer: in-scope comments are accurate after prior-pass fixes; the `dependencies.rb` `field` string-key doc is a model accurate comment (verified the full production trace). Only the optional class-doc nit above.

PR #610, opened 2026-06-27
feat: add /waiting command and daily pending-answer digest

ivankuznetsov:
# feat: add /waiting command and daily pending-answer digest

## Summary

Operators could not see at a glance which tasks were blocked on *their* answer. The "needs my input" tasks were buried in the full `/status` queue, and nothing proactively surfaced them — a human gate (e.g. `review_waiting` / `fix_guardrail`) got exactly one Telegram push when it first appeared and then deduped forever, so a task genuinely needing a decision could go quiet for hours.

This PR adds two surfaces over the same "waiting on you" set:

- **`/waiting`** — an on-demand Telegram view, the `/status` queue filtered to only the tasks needing a human answer, each with its working inline button.
- **`hive answer-digest`** — an opt-in daily push (09:00 local) listing the same set with per-task buttons; silent when nothing waits, one per local calendar day, idempotent across daemon restarts.

Both render one-tap action buttons derived from the same `RowActions` resolver the original first-appearance push uses, so a tap from either surface does exactly what the original push button did.

The existing reminder behavior is deliberately untouched. The brainstorm started from "add another reminder" and pivoted to "make the waiting set discoverable instead." The one-shot first-appearance push, the recovery-reminder path, and the completed-work `hive digest` all keep their current behavior; this PR adds parallel surfaces alongside them.

```mermaid
graph TB
  status["hive status --json<br/>(one row per task)"]
  wr["WaitingRows<br/>select + button_for"]
  waiting["/waiting<br/>on-demand Telegram view"]
  digest["hive answer-digest<br/>daily Telegram push"]
  sched["AnswerDigestScheduler<br/>09:00 local, idempotent"]
  status --> wr
  wr --> waiting
  wr --> digest
  sched --> digest
```

### What changed

| Piece | Role |
|---|---|
| `Hive::Bot::WaitingRows` | Shared selector: filters status rows to the 6 needs-input kinds, sorts most-urgent-first, and builds the matching inline button by reusing `RowActions.resolve(row).primary`. Single source of truth for both surfaces, with bidirectional drift guards and per-row rescue. |
| `/waiting` command | New router intent + slash handler. Reuses the `/status` renderer over the filtered set; empty set replies "Nothing is waiting on you." |
| `hive answer-digest` | New CLI command. Fetches status, selects the waiting set, posts one Telegram message with per-task buttons. Supports `--dry-run`, `--json`, `--date`. Silent (no send, exit 0) when nothing waits. Publishes a versioned `hive-answer-digest.v1` JSON envelope for agent consumers. |
| `Hive::Daemon::AnswerDigestScheduler` | Mirrors `DigestScheduler`: fires `hive answer-digest` once per local calendar day at/after the configured hour, with catch-up, failure backoff, a child-timeout cap, and an `answer_digest_state.json` cursor for restart idempotency. |
| `answer_digest` config block | New `enabled` (default `false`) and `hour` (default `9`, bounds `0..23`) keys, with validation and a documented template stanza. |

### Design decisions

- **One selection module, byte-identical buttons.** `/waiting` and the digest both go through `WaitingRows`, which derives buttons from the same `RowActions` resolver the first-appearance push uses, so there is no second definition of "what does the button do" to drift.
- **No de-dup logic for "one gate per task."** `hive status --json` already emits exactly one row per task with a single chosen marker, so "one line per task, most-urgent gate first" falls out of the source; `WaitingRows` only orders *across* tasks for display.
- **Daemon-on plan-pause suppression.** A `plan_waiting` row is dropped when its project has `daemon.enabled`, matching the dispatcher's existing suppression. The per-project lookup is an injected, memoized callable so the module stays pure and testable.
- **Silent-empty still counts as fired.** An empty snapshot exits 0 with no send; the scheduler advances its cursor (before clearing backoff) and does not retry, preserving the one-per-day contract without spamming.
- **Opt-in.** The digest defaults off; turning it on is a config change plus SIGHUP.

## Test plan

- `bundle exec rake test` — green.
- `bundle exec rake coverage` — 100% line coverage gate passes (new rescue/fallback/backoff/empty/dry-run branches each covered).
- `hive answer-digest --help` interpretable (exit 0).
- New / expanded unit tests:
  - `test/unit/bot/waiting_rows_test.rb` — selection set, ordering, per-row rescue, daemon-on `plan_waiting` drop, button byte-equality with the push.
  - `test/unit/commands/answer_digest_test.rb` — empty/non-empty/dry-run/overflow/`--json` envelope, schema-valid error envelopes, long-slug (>64-byte callback) compaction case.
  - `test/unit/daemon/answer_digest_scheduler_test.rb` — tick/complete/backoff, restart idempotency, cursor-before-backoff ordering, disabled.
  - `test/unit/daemon/dispatcher_test.rb` — daemon wiring, fatal-dedup re-arm, single global digest slot contention.
  - `test/unit/{bot/router_test,bot/slash_handlers_test,bot/supervisor_test,bot/notification_builders_test,commands/daemon_test,config_test,cli_test}.rb` — routing, render branch, config bounds, CLI registration.
  - `test/unit/schema_files_test.rb` — `hive-answer-digest.v1` required-key-drift + JSONSchemer round-trip guardrails matching every sibling schema.

Telegram delivery itself is exercised via a capturing `StubTelegram`; the runnable `--dry-run --json` read path is verified side-effect-free (no `.env` load, no chat/token resolution, no send). A `vhs` terminal capture of the dry-run digest is attached in `media/` (Screenote not connected, so no hosted URLs).

## Review summary

Three CE review passes (Claude + Codex personas + pr-review-toolkit) ran over the feature diff. The bulk of findings were auto-fixed and verified across the passes:

- **Reliability** — added the missing `answer-digest` `child_verb_timeouts` cap (a hung child no longer pins the shared digest slot); advanced the scheduler cursor before clearing backoff so a `write_state` fault can't re-fire the daily digest; added a `@delivered` guard + EPIPE/JSON rescue so a post-send emit failure can't re-send.
- **Contract** — published-schema reconciliation: `schema`/`schema_version` parity with `hive digest`, `JSON_USAGE_ERROR_CONTRACTS` registration, exit-code reconciliation toward the `EnvelopeEmitter` convention, `count`/`tasks[]` in the envelope, and the new per-schema drift + round-trip tests.
- **Correctness/observability** — logger built/injected *before* selection so silently-dropped human-blocking rows are logged in the daemon path; `/waiting` and digest plan-pause suppression made identical (shared `find_project` fallback); resolver rescue broadened to `Psych`/`SystemCallError` and memoized once per project.
- **Maintainability** — single-sourced the role-emoji map and button builder, collapsed duplicated dispatcher reap branches, made the `NEEDS_INPUT_KINDS` drift guard bidirectional, and removed dead delegators.

The branch was merged up to `origin/main` (including #608's markerless-3-plan reclassification, which the `WaitingRows` filter keys on) and the waiting set re-validated; the `origin/main..HEAD` phantom-revert noise from the stale base is gone.

**Known limitation (escalated, unresolved by design choice):** digest inline buttons whose Telegram callback exceeds 64 bytes (most needs-input kinds for long project slugs) are not tap-actionable across the `hive answer-digest` subprocess → bot boundary, because the callback-compaction registry is in-process only. The short `answer:` callback survives; for the rest the digest is informational and the action is taken in-chat via `/waiting` (whose buttons are in-process and fully tap-actionable). This is documented in `wiki/commands/answer-digest.md`. The architectural fix (shared/persisted registry vs. id-keyed ≤64B callbacks) was escalated for an operator decision across all three passes and left open.

## Linked task

- Hive task slug: `bot-never-reminds-about-human-260625-f70e` ("Needs Input Reminder Gap" — surface tasks waiting on a human answer).
- Originating evidence: tasks 9400 (#588) and 9512 (#590) tripped the `review_waiting` / `fix_guardrail` gate and sat in `6-review` for hours with no follow-up ping — the asymmetry that motivated this work.


ivankuznetsov:

## High
- [ ] Dry-run shims no longer scrub Ruby startup env before executing Ruby: `lib/hive/babysitter/dry_run_env.rb:84` now generates a Ruby shim, and `bin/hive-babysitter-stub-gh:4` requires load-path libraries before any scrub, so caller-controlled `RUBYOPT`/`RUBYLIB` can execute code before the dry-run git/gh safety gates.

## Medium
- [ ] `hive-e2e replay` can follow a symlinked scenario directory outside the run artifact: `bin/hive-e2e:161` only `lstat`s the final `repro.sh`, so intermediate symlinks bypass the containment check and replay can exec an external script.
- [ ] `bin/hive-eval` no longer cleans up/report usage errors for invalid-byte report arguments: `bin/hive-eval:96` rescues only `OptionParser::ParseError`, so the invalid UTF-8 `ArgumentError` path escapes and leaves stale selected reports behind.

## Nit
- [ ] Deleted historical wiki log fragments: the diff removes existing `wiki/log.d/20260626T083614Z-e2e-replay-parent-symlink.md`, `wiki/log.d/20260626T085149Z-hive-eval-invalid-byte-usage.md`, and `wiki/log.d/20260626T163650Z-babysitter-ruby-startup-env.md` instead of only appending the new fragment, which loses project changelog history.

ivankuznetsov:

# Code Review — bot-never-reminds-about-human-260625-f70e (pass 1)

Scope: feature commits only (`git diff $(git merge-base origin/main HEAD)..HEAD`). `origin/main` is 6 commits ahead of the branch base, so the raw `origin/main..HEAD` diff also shows reversed babysitter/e2e/eval/task_action noise that is **not** this branch's work — excluded from review.

Intent: add `/waiting` (filtered `/status`), a `hive answer-digest` daily digest subprocess with tap-actionable inline buttons, and an `AnswerDigestScheduler`. Buttons must be byte-identical/tap-actionable vs the original needs_input push; the reminder path, completed-work `hive digest`, and `/status` rendering stay unchanged.

Reviewers: correctness, reliability, adversarial, cli-readiness, testing, maintainability, project-standards, agent-native, learnings.

## High
- [ ] Digest inline buttons with >64-byte callbacks are dead on tap (cross-process compaction registry): `hive answer-digest` runs as a separate subprocess and builds buttons via `WaitingRows.button_for` → `NotificationBuilders.button` → `compact_callback` (`notification_builders.rb:667`), which stores the long→`#prefix:digest` mapping only in that process's in-memory `@registry` then exits; the long-running bot that handles the tap can't resolve it and routes to `:callback_expired` (`router.rb:246`). For this project's ~40-char slugs `approve_plan:`(66B)/`findings:accept_all:`(75B)/`rerun:…:develop`(70B) all exceed Telegram's 64-byte cap, so 5 of the 6 waiting kinds yield "button expired" — directly defeating plan A4. Flagged independently by correctness, adversarial, testing, and learnings; the router comment at `router.rb:243` already notes these callbacks "routinely exceed" 64 bytes. Fix: keep digest callbacks ≤64B (id-keyed) or persist/share the registry across the subprocess→bot boundary (`waiting_rows.rb:58`, `answer_digest.rb:99`).
- [ ] New `answer-digest` verb has no `child_verb_timeouts` cap yet shares the single global digest slot: it is dispatched through `dispatch_digest` with `kind: :digest` (`dispatcher.rb`), but `child_verb_timeouts` only caps `"digest"` (`config.rb:342`), so `answer-digest` falls back to `child_timeout_sec: 0` (disabled). A hung child (black-holed Telegram socket) pins the shared slot and leaves the scheduler's in-memory `@pending` marker set forever, silently disabling **all** future digests AND answer-digests until daemon restart — exactly the failure the comment at `config.rb:329-339` says the 3600s digest cap was added to prevent. Add `"answer-digest" => <cap>`.

## Medium
- [ ] `/waiting` and the digest can disagree on plan-pause suppression: `AnswerDigest#daemon_enabled_resolver` (`answer_digest.rb:165`) reads only `row.project_path` and returns false when empty, while `Supervisor#waiting_row_project_path` (`supervisor.rb:1528`) falls back to `Config.find_project(row.project)["path"]`. For a snapshot row lacking `project_path`, the digest surfaces a daemon-auto-approved `plan_waiting` row the bot would hide — the two surfaces are meant to be identical (correctness + maintainability agree).
- [ ] `button_for` has no per-row isolation: `deliver_waiting_digest`'s `visible.filter_map { WaitingRows.button_for(row) }` (`answer_digest.rb:99`) calls the raw `RowActions.resolve` + `ROLE_EMOJI.fetch` with no rescue, so one malformed row aborts the entire digest and trips scheduler backoff — strictly less defensive than `Supervisor#status_action_button`'s per-row `begin/rescue`. Wrap per row.
- [ ] `answer-digest --json` envelope omits `schema`/`schema_version` that sibling `hive digest` emits (`answer_digest.rb:142`), breaking hive's versioned self-identifying JSON contract — a consumer can't tell digest from answer-digest output or detect a future shape change.
- [ ] Thor argv errors (unknown flag / extra positional) under `--json` emit no JSON envelope — verified empty stdout + bare stderr + exit 64 (sibling `hive digest` emits a structured envelope). The in-command `parse_argv` ConfigError path never runs because Thor parses argv first; add an `answer-digest` entry to `JSON_USAGE_ERROR_CONTRACTS` in `bin/hive`.
- [ ] `--date` is misleading: it neither scopes the waiting set (always the live snapshot) nor provides idempotency (dedup lives solely in the daemon state file), unlike `hive digest --date`. Two same-date invocations both send (`answer_digest.rb:96`). Clarify the help/long_desc or honor it as a real dedup guard.
- [ ] `AnswerDigestScheduler` is ~100 lines near-verbatim of `DigestScheduler` (`answer_digest_scheduler.rb`); every method except `tick` is byte-identical save the state key/event names, and the explanatory comment on `#complete`'s nil-exit guard was already dropped. Extract a shared base/mixin so persistence/backoff fixes can't drift across the two.
- [ ] `WaitingRows::ROLE_EMOJI` (`waiting_rows.rb:27`) is a byte-identical second copy of `Supervisor#status_action_emoji` (`supervisor.rb:1500`); a new primary-capable `RowActions` role added to one map raises `KeyError` only on whichever surface renders it first. Single-source the map.
- [ ] Test gap masks the headline defect: every digest/button fixture uses short slugs (≤55B) so `compact_callback` never fires; add a digest test with a realistic ≥40-char slug that forces compaction and asserts the bot can resolve the tapped token from a fresh (empty) registry — i.e. the actual subprocess→bot handoff (`answer_digest_test.rb:113`, `waiting_rows_test.rb:94`).

## Nit
- [ ] `answer-digest` long_desc omits the exit-code contract (0 = sent/empty-silent, 78 = unconfigured, 1 = status unavailable) and the `--json` envelope key/`reason`-value shapes a daemon/agent caller must branch on (`cli.rb:700`).
- [ ] `WaitingRows.button_for` duplicates `Supervisor#status_action_button`; the new shared module was the natural home to unify both button builders, but `/waiting` still renders via the supervisor copy — the same conceptual button now has two implementations.
- [ ] Inconsistent resolve discipline inside one small module: `select` guards `RowActions.resolve` with a rescue→nil wrapper, `button_for` calls it raw (`waiting_rows.rb:52,61`). Pick one.
- [ ] `answer_digest.rb` calls `NotificationBuilders.display_title` (`:118`) without requiring `hive/bot/notification_builders` — it loads only transitively via `waiting_rows`. Add the explicit require so the dependency is load-order-robust.
- [ ] Reap paths hardcode per-stage `if entry.stage == X::STAGE` complete-branches (`dispatcher.rb:622-642`, `759-776`) instead of reusing the existing `global_digest_scheduler(stage)` helper; a third digest type needs four more copy-paste blocks.
- [ ] `answer-digest --json` reports `button_count` (capped at 10, post-`filter_map`) not the true `rows.size`, and nils the task list on real sends, so an agent polling JSON can't learn how many / which tasks wait (`answer_digest.rb:142`). Add `count` + a `tasks[]` array.
- [ ] Wiki: `WaitingRows` and `/waiting` not added to the `wiki/modules/bot.md` module map (its `source:` is `lib/hive/bot/`); `wiki/index.md` also lists `answer-digest` out of the file's otherwise-alphabetical order.
- [ ] `status unavailable` upstream failure maps to generic exit 1 / `error_kind: "internal"` (`answer_digest.rb:90`), indistinguishable from an internal crash for a `--json` agent (the daemon retries on any non-zero, so it's unaffected).
- [ ] Branch is 6 commits behind `origin/main`, including #608 which reclassifies markerless 3-plan from `plan_waiting`→`ready_to_run` — directly touching the needs-input gate this feature filters on. Rebase and re-run the 100% coverage gate before merge (the new rescue/fallback branches in `WaitingRows`, `waiting_daemon_enabled_resolver`, and `daemon_enabled_resolver` must each be covered).

ivankuznetsov:

# PR Review Toolkit — bot-never-reminds-about-human-260625-f70e (pass 1)

Scope: feature commits only (`git diff $(git merge-base origin/main HEAD)..HEAD`, base `16fff941`). `origin/main` is 6 commits ahead of the branch base, so the reversed babysitter/e2e/eval/task_action churn in the raw `origin/main..HEAD` diff is excluded. Personas run: code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier.

## High
- [ ] [code-reviewer / pr-test-analyzer / comment-analyzer] Digest inline buttons are dead on tap for >64-byte callbacks: `hive answer-digest` runs as a separate subprocess and compacts long callbacks via `NotificationBuilders.compact_callback` (`notification_builders.rb:663-693`) into `#prefix:digest` tokens stored only in that process's in-memory `@registry`, then exits; the long-running bot resolving the tap misses → router `:callback_expired` "button expired" (`router.rb:240-246`). Measured for project `hive` + this branch's 41-char slug: `approve_plan:`=66B, `findings:accept_all:`=75B (urgency rank 0, sorted first), `rerun:…:develop`=70B, generic rerun=65B — only short-slug brainstorm `answer:` (53B) survives. The 3600s registry TTL means even an earlier in-process push is already evicted by the daily 09:00 fire, so the miss is near-certain for the digest's entire target population. Directly violates plan A4 (`answer_digest.rb:99`, `waiting_rows.rb:52-59`).

## Medium
- [ ] [pr-test-analyzer / code-reviewer] The tap-actionable-button guarantee is untested: every digest/waiting fixture uses short slugs (≤55B) so `compact_callback` never fires; the suite is green while the real long-slug failure mode is uncovered. Add a ≥40-char-slug case asserting (a) `callback_data.bytesize <= 64` and (b) a fresh `NotificationBuilders` registry (simulating the bot process) still resolves the token — (b) fails today and pins the High finding (`answer_digest_test.rb:113`, `waiting_rows_test.rb:82,94`).
- [ ] [silent-failure-hunter] `WaitingRows.resolve` rescues `StandardError → nil` and `select` drops the row (`waiting_rows.rb:61-65`, consumed at `:41`) with no log, so a task that genuinely needs input silently vanishes from `/waiting` and the digest — and `waiting_rows_test.rb:102-116` codifies the silence by asserting the drop with no log assertion. For a feature named *bot-never-reminds-about-human*, thread a logger and emit an error event before returning nil (keep the rescue).
- [ ] [silent-failure-hunter / code-simplifier] `answer_digest.rb#daemon_enabled_resolver` (`:165-176`) has diverged from its supervisor twin (`supervisor.rb:1513-1537`): it lacks the `Config.find_project(row.project)["path"]` fallback for a blank `row.project_path` AND swallows `ConfigError` with no log (and `@logger` is nil during the selection phase) → for a row with blank project_path the digest includes a `plan_waiting` row the bot's `/waiting` hides (two surfaces meant to be identical disagree), and a broken project config is invisible from the digest path.
- [ ] [type-design] `AnswerDigest::Result = Data.define(...)` (`answer_digest.rb:20`) is an unvalidated 7-tuple: illegal states are representable (`sent:true, chat_id:nil, reason:"empty", dry_run:true`), `sent` is fully derivable from `reason`, and `reason` is stringly-typed with no closed-vocabulary constant yet emitted verbatim into the JSON contract (`:147`). Derive `sent` from `reason`, add `REASONS = [nil,"empty","dry_run"]`, and validate cross-field rules in a constructor (or expose `empty`/`dry_run`/`sent` factories) — the bar the sibling `RowActions::Resolution` already sets.
- [ ] [type-design] `NEEDS_INPUT_KINDS` and `URGENCY_RANK` (`waiting_rows.rb:9-25`) hand-copy a subset of `RowActions::KINDS` with no derivation or load-time check; a new upstream needs-input kind is then silently excluded from `/waiting` and the digest (`select` at `:42`). Add a boot-time assertion `(NEEDS_INPUT_KINDS - RowActions::KINDS).empty?` and `URGENCY_RANK.keys.sort == NEEDS_INPUT_KINDS.sort` so drift fails loud.
- [ ] [type-design / silent-failure-hunter] `ROLE_EMOJI` (`waiting_rows.rb:27-36`) is a third verbatim copy of the role vocabulary and `button_for`'s `.fetch` (`:58`) is unguarded inside an un-rescued `filter_map` (`answer_digest.rb:99`); since `button_for` calls the raw `RowActions.resolve` and `call` only rescues `Hive::Error`, adding a primary-capable role to `RowActions::ROLES` turns into a `KeyError` that aborts the entire digest send — whereas the sibling `status_action_button` rescues per-row. Single-source the emoji map off `ROLES` (fail-loud at load) and wrap the per-row build.
- [ ] [comment-analyzer] Wiki overstates the digest buttons: `wiki/commands/answer-digest.md:46-49` ("callbacks match the original push-notification buttons byte-for-byte") and `wiki/commands.md:92-94` ("matching the original push notifications") give no caveat that the cross-process compaction registry makes long-callback buttons dead on tap — the router comment at `router.rb:240-245` already documents exactly this hazard. Add the caveat.
- [ ] [comment-analyzer] `wiki/commands/answer-digest.md:27` documents `--date` as used "for scheduler idempotency," but the command neither scopes the (always-live) snapshot nor dedups on it — idempotency lives solely in the daemon's `answer_digest_state.json`, and `tick`/`complete` never consume the command's `--date`. Reword to "echoed into the JSON `date` field; does not scope data or dedup."

## Nit
- [ ] [pr-test-analyzer] No test asserts the `--json` envelope of a successful real send (`sent:true`, `chat_id`, `button_count`, `message:nil`) — the automation-facing contract for the live path (`answer_digest.rb:142-153`).
- [ ] [pr-test-analyzer] No test asserts the per-project-path resolver cache actually dedupes `Hive::Config.load` (two rows sharing a `project_path` → one load); a regression to N loads/N YAML parses per digest would be uncaught (`answer_digest.rb:165-176`, `supervisor.rb:1513-1527`).
- [ ] [silent-failure-hunter] `WaitingRows.daemon_plan_pause?` rescues `StandardError → false` unlogged (`waiting_rows.rb:67-71`) — fail-open (over-reminds) so low severity, but a real bug in the daemon-enabled check stays invisible; log once a logger is threaded.
- [ ] [silent-failure-hunter] `AnswerDigestScheduler#parse_date` swallows a well-formed-JSON-but-malformed date value → nil with no log (`answer_digest_scheduler.rb:134-138`); `read_state` only logs JSON/IO faults, so a corrupt `last_fired_date` self-heals in the dark. Add a one-line `log_state_unreadable`.
- [ ] [silent-failure-hunter] In `--json` mode a non-`Hive::Error` `StandardError` (e.g. Faraday) escapes `call` with no JSON error envelope (only `Hive::Error` is rescued, `answer_digest.rb:54-57`) — harmless for the daemon (reads exit code) but degrades a manual `--json` run to a raw backtrace.
- [ ] [silent-failure-hunter] The dispatcher's `:fatal` rescues around `answer_digest_scheduler.tick`/`complete` don't dedup, unlike `brainstorm_answers_pending?`, so a persistently-raising scheduler emits `:fatal` every ~30s poll — observable but noisy (`dispatcher.rb:228-236`).
- [ ] [type-design] `urgency_rank`'s fetch-with-default (`waiting_rows.rb:73-75`) is unreachable dead defensiveness that would silently down-rank a future-drifted kind; the load-time assertion above makes it provably dead so it can become an honest `URGENCY_RANK.fetch(kind)`.
- [ ] [type-design] `AnswerDigestScheduler` in-memory state is under-typed: `@pending` is a Hash used as an at-most-one Set and `@failure` an untyped `{count:,next_eligible_at:}` Hash (`:23-24,82-93`); a `@failure={}` would make `backed_off?` do `now < nil`. Model `@pending` as a nilable date and `@failure` as `Data.define(:count,:next_eligible_at)` (only if touching both schedulers).
- [ ] [type-design] "Already fired today" is split across durable `last_fired_date` and non-durable `@pending`, so a daemon crash/restart between dispatch and `complete` re-fires a duplicate user-visible Telegram digest (`answer_digest_scheduler.rb:32-62`); inherited from `DigestScheduler` but the type design doesn't prevent it — persist the in-flight date or document the restart-window duplicate.
- [ ] [comment-analyzer] `wiki/commands/bot.md:104-107` ("byte-identical to the push-notification buttons") is defensible for the in-process `/status`/`/waiting` surfaces, but the compaction/expiry behavior (why a shown button can later reply "expired") is undocumented anywhere; add a one-line note.
- [ ] [code-simplifier] `AnswerDigestScheduler` and `DigestScheduler` share ~70 verbatim lines (backoff/`record_failure`/`read_state`/`write_state`/`parse_date`/`complete`/`cancel`/`pending?`); extract a `DailyDigestScheduler` base parameterized by state-key/filename/event-symbols (keep event names + state keys distinct — asserted in tests). Heavier change; don't auto-apply.
- [ ] [code-simplifier] `Supervisor#status_action_emoji` (`supervisor.rb:1500-1511`) is byte-identical to `WaitingRows::ROLE_EMOJI`; make it a one-line delegator (the direct unit test at `supervisor_test.rb:1664` keeps the method alive).
- [ ] [code-simplifier] `Supervisor#status_action_button` (`supervisor.rb:1472-1498`) duplicates `WaitingRows.button_for`; delegate the core to the shared builder while keeping the supervisor's per-row rescue + pre-captured log locals.
- [ ] [code-simplifier] `dispatcher.rb` reap sites hand-roll four copies of the per-stage "complete + isolate disk I/O" block (`:622-642`, `:759-776`) though `global_digest_scheduler(stage)`/`global_digest_action(stage)` already exist; collapse to one `complete_digest_scheduler_for(entry, now:)` helper (preserve the `:fatal` message text — asserted at `dispatcher_test.rb:2240,2268`). Leave the tick-loop fold (4b) as-is.
- [ ] [code-simplifier] The two per-project `daemon.enabled` memoizers (`answer_digest.rb:165-176`, `supervisor.rb:1513-1537`) share a cache/fetch/rescue skeleton; lift it into `WaitingRows.daemon_enabled_resolver(project_path:, on_error:)` with injected hooks — but keep the two surfaces' path-resolution and logging distinct (unifying them is the behavioral divergence flagged above, not a pure refactor).

ivankuznetsov:

## High
- [ ] `lib/hive/task_action.rb:442` markerless coding plan rows now classify as `needs_input`: dispatcher treats any coding `3-plan` `needs_input` row as plan approval, `PlanApproval` rejects marker `:none`, and `/waiting`/digest suppress the row, so freshly promoted plan-stage tasks can wedge invisibly instead of running the plan agent.
- [ ] `lib/hive/commands/answer_digest.rb:144` builds digest buttons with the bot's in-memory callback compactor from a short-lived subprocess: long approve/review/rerun callbacks are sent as `#...` tokens that the live Telegram bot cannot resolve, violating the plan-required working per-task digest buttons.
- [ ] `lib/hive/babysitter/dry_run_env.rb:84` generates Ruby shims instead of the prior shell wrapper that scrubbed Ruby startup variables: `RUBYOPT`/`RUBYLIB` can execute caller-controlled code before the dry-run `git`/`gh` guards run, and `bin/hive-babysitter-stub-gh:4` now adds pre-gate requires.

## Medium
- [ ] `lib/hive/daemon/answer_digest_scheduler.rb:43` keeps the in-flight answer-digest marker only in memory and writes `last_fired_date` only on completion: a daemon restart or state-write failure after Telegram delivery can dispatch the same local-day digest again, breaking the one-per-day restart-idempotency requirement.
- [ ] `bin/hive-e2e:161` no longer rejects symlinked replay directories: lexical containment still passes when `scenario_dir` is a symlink, so `hive-e2e replay` can execute an outside `repro.sh` that used to be rejected.

## Nit
- [ ] `test/unit/task_action_test.rb:71` now asserts `needs_input` while the assertion message says markerless plan folders "must remain runnable": the test documents the opposite of the behavior it enforces, masking the plan-stage wedge above.

ivankuznetsov:

# Code Review — bot-never-reminds-about-human-260625-f70e (pass 2)

Scope: `git diff origin/main..HEAD`. Plan-aligned: `/waiting` command, the
separate 9am "pending answers" digest (`hive answer-digest`), and the
`AnswerDigestScheduler`. Verified locally: affected unit suites green
(`waiting_rows`, `answer_digest`, `answer_digest_scheduler`, `dispatcher`,
`supervisor`, `config`, `router`, `slash_handlers`, `daemon`, `cli`,
`notification_builders`), `rubocop` clean on touched `lib/` files, and
`hive answer-digest --help` interpretable (exit 0). Byte-identical button reuse,
silent-empty, dry-run, idempotency, and daemon wiring all check out. The
findings below are integration/observability/contract issues, not blocking
logic defects.

## High
- [ ] Branch is ~8 PRs behind `origin/main` (#598–#608); the `origin/main..HEAD` diff is contaminated with phantom reverts of those merged fixes (e.g. `lib/hive/task_action.rb` from #608, `bin/hive-babysitter-stub-gh`/`bin/hive-e2e`/`bin/hive-eval`/`lib/hive/babysitter/dry_run_env.rb` patrol fixes #598–#602, several deleted `wiki/log.d` fragments) — none of which this branch authored. The feature was built/tested on a stale base, and #608 specifically changes how a markerless 3-plan row classifies (`ready_to_run` vs `plan_waiting`), which is exactly what the new `WaitingRows` filter keys on to decide whether such rows surface in `/waiting`/the digest. Rebase/merge `origin/main` and re-validate the waiting set before landing.

## Medium
- [ ] `Hive::Commands::AnswerDigest#call` runs `WaitingRows.select(..., logger: @logger)` (and `button_for`, `daemon_enabled_resolver`) while `@logger` is still `nil` — the logger is only built lazily inside `deliver_waiting_digest`, which runs *after* selection. In the daemon/CLI path the per-row drop logging (`:poll_failure` / `:status_button_failed`) that the code comments claim makes a silently-dropped human-blocking row observable never fires, defeating the feature's stated "never silently forget a waiting task" goal (`/waiting` in the supervisor is unaffected — it has a real logger). Build/inject the logger before selection.
- [ ] Published v1 contract `schemas/hive-answer-digest.v1.json` (line 5) states a mid-delivery Telegram send error "surfaces on stderr + a non-zero exit code without an envelope," but `call`'s `rescue StandardError` emits a JSON error envelope (`error_kind:"internal"`) under `--json` before re-raising (the code comment even says "the agent parsing stdout still gets a structured error"). The prose is wrong for the always-`--json` daemon path; the emitted envelope is itself schema-valid, so the description, not the code, should change.

## Nit
- [ ] Each waiting row is resolved twice — `WaitingRows.select` calls `resolve(row)` and `WaitingRows.button_for` independently calls `RowActions.resolve(row)` again (`lib/hive/bot/waiting_rows.rb:57,82`). Harmless (resolve is cheap) but redundant; the resolution could be threaded through.
- [ ] `daemon_enabled_resolver` (the supervisor twin at `supervisor.rb` and the answer-digest twin at `answer_digest.rb:243`) returns `false` on `Hive::ConfigError` without populating the memo cache, so a project with a broken config is re-loaded and re-logged once per waiting row of that project rather than once.
- [ ] An empty waiting set under `--dry-run` emits `"message": ""` (json_payload: `@dry_run ? result.message : nil`), while the SuccessPayload `message` description says it is `null` on an empty waiting set — type-valid (a string), but a minor description drift for the empty+dry-run case.

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — bot-never-reminds-about-human-260625-f70e (pass 2)

Personas run: `code-reviewer`, `silent-failure`, `type-design`, `pr-test`,
`comment`, `code-simplifier`. Diff reviewed against the **merge-base** (the
branch is behind `origin/main`, so `origin/main..HEAD` is polluted with phantom
reverts of #598–#608 — those non-authored files were excluded). Overall: a
high-quality, heavily-tested PR with no logic bugs found by any persona; the
findings cluster around one real reliability gap (a nil logger in the production
answer-digest path), a published-schema doc inaccuracy, a few type/test
hardening gaps, and dedup nits.

## High
- [ ] [silent-failure] A waiting row whose `RowActions.resolve` raises is silently dropped from the answer-digest with zero trace (`waiting_rows.rb:95-106` via `answer_digest.rb:137`): the daemon-spawned `hive answer-digest --json` is built with no logger (`cli.rb:736`) and `logger(cfg)` is created lazily only inside `deliver_waiting_digest` (`answer_digest.rb:155`), *after* selection — so every `logger&.event(:poll_failure …)` is a no-op. The dropped row never enters `count`/`tasks[]`, emits no log, the envelope stays `ok:true`/exit 0, and the scheduler then marks the day delivered → the task is never retried — the exact "human-blocking task forgotten" failure this feature exists to prevent. Fix: build/inject the logger before selection, or surface dropped rows in the JSON envelope; consider narrowing the broad `rescue StandardError` so genuine contract bugs propagate.

## Medium
- [ ] [silent-failure] A malformed row also silently loses its inline button (`waiting_rows.rb:76-93` `button_for`, same nil-logger path): the task still shows in the digest *text* but is un-tappable, and `button_count < count` with no logged reason — surface the per-row button-build failure in the envelope.
- [ ] [comment] `schemas/hive-answer-digest.v1.json:5` claims a mid-delivery Telegram send error surfaces "without an envelope," but under `--json` (the contract this schema documents) `call`'s `rescue StandardError` emits an `internal` ErrorPayload (exit 1) then re-raises (`answer_digest.rb:94-100`); contradicts the schema's own `error_kind` enum and the CLI long_desc. Published v1 contract prose should be corrected.
- [ ] [code-reviewer] `stage_approval` rows (ready-to-X awaiting a manual `/approve`) are excluded from `NEEDS_INPUT_KINDS` (`waiting_rows.rb:9-16`), so on a `daemon.enabled:false` project a genuinely human-blocking task is absent from both `/waiting` and the digest — asymmetric with `plan_waiting` (included, daemon-suppressed). Confirm this is intended scope vs. an oversight (the wiki documents the exclusion).
- [ ] [type-design] `Result` (`answer_digest.rb:37-52`) enforces the reason↔chat_id half of its state machine but never reconciles `dry_run` with `reason`: illegal states (`reason:"dry_run", dry_run:false`; a "sent" result with `dry_run:true`) construct cleanly. Add the two missing guards.
- [ ] [type-design] `tasks` is an unvalidated Array of string-keyed Hashes (`answer_digest.rb:36,165-174`) whose `{project,slug,id,title,stage,pr}` shape is asserted in three disconnected places (producer, JSON `$defs/Task`, schema) with no Ruby link; promote to `AnswerDigest::Task = Data.define(...)` so `Result` can assert `tasks.all?(Task)` — matches the project's own `Resolution`/`Action` idiom.
- [ ] [pr-test] The `--json` overflow contract is untested: the only overflow test runs `json:false`, and the lone JSON test uses 1 row (where `count==button_count==tasks.length`). A regression capping `count`/`tasks` to the 10-row display slice would make a `--json` agent under-report the backlog yet pass every test. Add an 11-row `json:true` test asserting `count==11`, `button_count==10`, `tasks.length==11`, and the `⏳ Waiting on you (11)` header.
- [ ] [pr-test] `status_keyboard`'s >10-button cap (`supervisor.rb:1464` `.first(QUEUE_DISPLAY_CAP)`) is untested (only ≤9 actionable rows exercised); dropping the cap would attach >10 buttons (Telegram rejects an oversized keyboard) with no failing test. Add a `status_keyboard` test with >10 rows asserting `keyboard.flatten.length == 10`.
- [ ] [comment] `waiting_rows.rb:98-105` `resolve` comment asserts the drop is "logged" so a needs-input row "must not vanish from /waiting and the digest without a trace" — true for `/waiting` (real logger) but false on the nil-logger digest path; scope the comment like the honest sibling at `answer_digest.rb:252-254`, or note the digest relies on the subprocess exit/envelope.

## Nit
- [ ] [code-reviewer] `answer_digest.rb:198-200` — the bare-text success line counts `button_count` (capped at 10), so with >10 waiting tasks it prints "sent 10 waiting tasks" while N>10 wait; use `result.count`.
- [ ] [type-design/comment] Empty waiting set under `--dry-run --json` emits `message:""` (because `json_payload` nulls on the `@dry_run` ivar, `:220`) while the schema says message is `null` on an empty set; type-valid but a description/behavior mismatch — derive the nulling from `result`.
- [ ] [type-design] `Result#tasks` is not frozen (sibling `Resolution` freezes `actions`), leaving the value object shallowly mutable; also `button_count`/`count` lack a non-negativity / `button_count <= count` guard.
- [ ] [type-design] `waiting_rows.rb` lacks `# frozen_string_literal: true`, so `ROLE_EMOJI`'s String values are individually mutable despite the design assuming immutability — add the magic comment or freeze the values. (The load-time drift guards themselves are exemplary.)
- [ ] [code-simplifier] `answer_digest.rb:91-101` — the `rescue Hive::Error` and `rescue StandardError` branches have byte-identical bodies and `Hive::Error < StandardError`; collapse to one `rescue StandardError`.
- [ ] [code-simplifier] `answer_digest_row_project_path` (`answer_digest.rb:266-273`) is character-identical to `waiting_row_project_path` (`supervisor.rb`); extract `WaitingRows.row_project_path(row)`.
- [ ] [code-simplifier] `daemon_enabled_resolver` (`answer_digest.rb:243-259`) and `waiting_daemon_enabled_resolver` (`supervisor.rb`) differ only in a log `source:` string; extract a shared builder taking `source:` (do alongside the prior item).
- [ ] [silent-failure] `daemon_enabled_resolver`'s `Hive::ConfigError → false` fail-open masks a malformed project config invisibly in production (nil logger); surface the config-degradation count in the envelope so a misconfig is observable.
- [ ] [comment] `waiting_rows.rb:111-118` `daemon_plan_pause?` comment ("log once so a real bug … isn't invisible") is inaccurate on the nil-logger digest path; benign (fail-open over-reminds) but add the "logged only when a logger is present" caveat.
- [ ] [pr-test] Smaller test gaps: cross-surface button-callback byte-equality is only asserted indirectly (no single test derives the push and `/waiting`/digest button from one row, nor the compacted >64-byte-callback case); the real-send plain-text stdout + singular/plural split is unverified; the digest's one-button-per-row nesting isn't pinned (tests `.flatten`); `clamp_hour` out-of-range boundaries and a permanently-stuck `@pending` wedge are unasserted; daemon wiring asserts the scheduler instance but not that `enabled`/`hour` were forwarded.
- [ ] [code-simplifier] `dispatcher.rb` `global_digest_action` returns `"digest"` for any non-answer-digest stage (loose default) while its sibling `global_digest_scheduler` returns `nil` for unknown stages (strict); harmless but a strict `case` would read more consistently. The digest/answer-digest unification helpers are otherwise well-factored, not over-abstracted.
- [ ] [code-simplifier/code-reviewer] Each visible row is resolved twice (`select` `resolve` + `button_for` `RowActions.resolve`, `waiting_rows.rb:57,82`); pure/cheap (≤10 re-resolves) and the clean `select → [rows]` contract is worth keeping — recommend leaving as-is, noted for completeness.

ivankuznetsov:

## High
- [ ] Answer-digest buttons expire for long callbacks: `lib/hive/commands/answer_digest.rb:179` builds buttons in the short-lived CLI process, so callbacks compacted by `NotificationBuilders` are stored only in that process's memory and the long-running bot later routes the `#...` token to "button expired", violating the plan's tap-actionable per-task digest button requirement.

## Medium

## Nit

ivankuznetsov:

# Code Review — bot-never-reminds-about-human-260625-f70e (pass 3)

Scope: `git diff origin/main..HEAD` (now includes the `origin/main` merge that
resolved pass-2 High #1's stale-base / phantom-revert contamination). Reviewed
the `/waiting` command, the `hive answer-digest` command + v1 schema, the
`AnswerDigestScheduler`, the dispatcher "collapse digest reap branches / dedup
scheduler fatals" refactor, and the config/CLI wiring. Verified by direct trace:
the `answer-digest` verb→`child_verb_timeouts` (3600s) mapping is correct
(`argv_verb` = argv[1]); `Pr.number` returns the `#NNN` string the schema
declares; the reap/gate refactor preserves digest semantics (single-slot
arbitration, fatal dedup, CONFIG-drop guard, SIGHUP reconfigure all self-heal);
the load-time drift guards and `select` sort are sound. Three parallel finder
agents + manual review found no new High-severity logic defect — the one High
below is the standing, still-unanswered cross-process button escalation.

## High
- [ ] Digest inline buttons still expire when tapped (re-raise of unanswered escalation Q1, passes 01+02; also codex pass-03 High): `hive answer-digest` runs as a separate subprocess, so `WaitingRows.button_for` → `NotificationBuilders.button` → `compact_callback` (`lib/hive/bot/notification_builders.rb:669-687`) stores any >64-byte callback's `#prefix:digest` mapping only in *that subprocess's* in-process `@registry` (`@registry ||= {}`, :685), which exits when the command returns; the long-running bot resolves the tapped token against its own empty registry, misses, and routes to "button expired" — so for project `hive`'s ~40-char slugs (`approve_plan:`, `findings:accept_all:`, `rerun:…` all exceed Telegram's 64-byte cap) ~5 of the 6 waiting kinds yield dead digest buttons, violating plan A4/R7 ("buttons tap-actionable, byte-identical to the original push"), which the plan did not defer. Still shipped as a documented limitation (`wiki/commands/answer-digest.md`) rather than fixed; needs an operator decision on the path (persist/share the registry across the subprocess→bot boundary, keep digest callbacks ≤64B, or formally accept the limitation).

## Medium
- [ ] `hive answer-digest --json` is not pipe-safe and breaks its own documented exit-code contract under a closed reader: `emit` (`lib/hive/commands/answer_digest.rb:227-236`) and `emit_error_envelope` (`:270-280`) do a bare `@output.puts JSON.generate(...)` with no `rescue Errno::EPIPE, JSON::GeneratorError`, unlike every sibling `--json` command (`status.rb:1037`, `run.rb:505`, `connect.rb:155`, `metrics.rb:140`, `disconnect.rb:155`), and `bin/hive`'s top-level `begin/rescue` (`bin/hive:248-263`) catches only `Thor::Error`/`Hive::Error` — not `SystemCallError`. So `hive answer-digest --json | head` (reader closes early): on a successful send `emit` raises `EPIPE`, the `rescue StandardError` re-tries the write in `emit_error_envelope` (also raising), and the unhandled `EPIPE` escapes to a "Broken pipe" backtrace + nonzero exit; with a bad `--date` the documented exit 78 is likewise replaced by a generic crash. (Daemon path is unaffected — child stdout is a file.)

## Nit
- [ ] `tasks[].stage` is emitted as `""` (empty string) for a stageless row via `row.stage.to_s` (`lib/hive/commands/answer_digest.rb:206`) while the sibling `project`/`slug`/`id` fields emit `null` (raw `row.X`); schema-valid (`stage` is `["string","null"]`) but inconsistent with the schema's "stage or null" intent, so a `--json` consumer branching on `stage === null` mis-classifies the stageless case. Real status rows always carry a stage, so this is latent.
- [ ] A `--dry-run` daemon advances the *real* `answer_digest_state.json` cursor: `reap_dry_run` synthesizes an exit-0 pseudo-child and calls `complete_digest_scheduler_for` → `AnswerDigestScheduler#complete(exit_code: 0)` → `write_state("last_fired_date" => today)` (`lib/hive/daemon/answer_digest_scheduler.rb:61`), so testing wiring with `hive daemon start --dry-run` at/after 09:00 marks the day done and a subsequent real daemon that day sends nothing. Inherited verbatim from `DigestScheduler`/`digest_state.json` (consistent with the U4 design), but it is a new surface for the pending-answers reminder.
- [ ] An all-failed keyboard is still sent as `reply_markup: []`: `deliver_waiting_digest` builds `keyboard = buttons.map { [button] }` (`answer_digest.rb:180`), which is `[]` if every visible row's `button_for` hits its per-row rescue and returns nil; `Telegram#send_message`'s `reply_markup && ...` guard treats `[]` as truthy and ships an empty `inline_keyboard`, which Telegram may reject (400 → `error_kind:internal`, exit 1, scheduler backoff). Low-probability (needs every needs-input row's button build to raise), but the `[] vs nil` distinction is unguarded here (the supervisor's `status_keyboard` returns nil when empty).
- [ ] When both `digest` and `answer_digest` are enabled, the day's 09:00 answer-digest can arrive hours late after a daemon down-window: `run_digest_scheduler_tick(@digest_scheduler, …)` always runs before the answer-digest tick (`dispatcher.rb:223-224`) and both share the single `kind: :digest` slot, so a multi-day `DigestScheduler` catch-up keeps winning the slot and the answer-digest is `cancel`-and-re-owed every tick until the catch-up drains. Self-heals the same day (no day permanently lost), so severity is low.

ivankuznetsov:

# PR Review Toolkit — bot-never-reminds-about-human-260625-f70e (pass 3)

Union of six persona agents (code-reviewer, silent-failure-hunter,
comment-analyzer, pr-test-analyzer, type-design-analyzer, code-simplifier) over
`git diff origin/main..HEAD`. Items already tracked in the standing escalations /
`suppressed.md` (digest button expiry across the subprocess→bot boundary;
restart-window duplicate from in-memory `@pending`; scheduler↔DigestScheduler
duplication; double row-resolve; `stage_approval` exclusion; the
`daemon_enabled_resolver` ConfigError fail-open itself) are deliberately omitted.
Key claims were verified against the code before listing.

## High
- [ ] [pr-test-analyzer] The dispatcher fatal-dedup (`@digest_scheduler_fatal_signatures`) — the core new logic of the "dedup scheduler fatals" refactor — has zero behavioral coverage: `dispatcher_test.rb:747/761` only assert a single raise logs one `:fatal`. None assert same-fault-across-ticks logs once, clean-run re-arms the dedup, or a distinct fault re-logs; a regression that never re-arms (silent forever) or never dedups (per-poll `:fatal` spam — the thing the refactor exists to prevent) passes today. Mirror the sibling brainstorm-dedup test (`dispatcher_test.rb:3395`+`:3414`).

## Medium
- [ ] [silent-failure] `AnswerDigestScheduler#complete` sets `@failure = nil` (answer_digest_scheduler.rb:56) before `write_state` (:61), so a cursor-write fault (ENOSPC/EROFS on the atomic rename) clears backoff and leaves `last_fired_date` un-advanced → the day re-fires every poll, re-sending a real Telegram digest in an unbounded loop whose only signal is the (deduped, never-re-arming) fatal above. Inherited verbatim from `DigestScheduler#complete` (:82/:87), so it also affects the shipped digest; surface for a shared fix (advance cursor before clearing backoff; engage backoff on write failure).
- [ ] [silent-failure] `WaitingRows.daemon_enabled_resolver` rescues only `Hive::ConfigError` (waiting_rows.rb:164), but `Hive::Config.load` (config.rb:585) does NOT wrap `Psych::Exception`/`Errno::*` from a malformed/unreadable project `.hive-state/config.yml` (only the global loader does, :809-814); those raw faults escape the memoizing `cache.fetch`, fall through to `daemon_plan_pause?`'s generic rescue → logged under the wrong `source:"waiting_daemon_pause"` and re-loaded+re-logged for every `plan_waiting` row, violating the method's "loads/logs once per project" contract.
- [ ] [code-reviewer/silent-failure] The digest is sent (answer_digest.rb:188-193) before `emit`/`json_payload` run, all under one broad `rescue StandardError` that emits an `error_kind:"internal"` envelope and re-raises; any post-send fault (broken stdout pipe; `JSON.generate` failure) mislabels a delivered digest as a failure and, on the daemon path, drives a non-zero exit → backoff → re-send of an already-delivered digest. Compounded by `emit`/`emit_error_envelope` lacking the `Errno::EPIPE`/`JSON::GeneratorError` guard every sibling `--json` command has (`status.rb:1037`, `run.rb:505`, `connect.rb:155`, `metrics.rb:140`).
- [ ] [pr-test-analyzer] No test runs the real digest and answer-digest schedulers in the same tick, so the single-global-slot ordering (`run_digest_scheduler_tick(@digest_scheduler)` claims the slot first; answer-digest then blocks+cancels) is unverified — `make_dispatcher` defaults both off and contention is faked via `ctrl.record_dispatch`; swapping the two tick calls (dispatcher.rb:223-224) would flip the winner with no failing test.
- [ ] [pr-test-analyzer] The digest command's positive daemon-enabled `plan_waiting` suppression is not asserted end-to-end (only fail-open `answer_digest_test.rb:350` + resolver-in-isolation `:412`); a regression that stopped wiring the resolver into `WaitingRows.select` for the digest path would pass — the `/waiting` surface has this coverage (`supervisor_test.rb:1773`) but the digest does not.
- [ ] [code-simplifier] Dead delegator `Supervisor#waiting_row_project_path` (supervisor.rb:1495) has no production caller (only its def + `supervisor_test.rb:1681`); the `row_project_path` comment (waiting_rows.rb:139) overstates "single source for the supervisor and answer-digest delegators" — there is no answer-digest delegator and the supervisor one is dead.
- [ ] [comment-analyzer] The drift-guard comment (waiting_rows.rb:46-53) claims the guards prevent "a new upstream needs-input kind [being] silently excluded from /waiting and the digest," but the subset check `(NEEDS_INPUT_KINDS - RowActions::KINDS).empty?` (:54) does NOT catch a newly-ADDED upstream needs-input kind (it stays silently excluded) — false assurance about the feature's core "never silently forget a waiting task" invariant (the code is correct; only the comment overclaims).

## Nit
- [ ] [code-reviewer] `/waiting` renders through `render_queue`, whose header reads "N active task(s)" (supervisor.rb:1440), inconsistent with the command's intent and the digest's own "⏳ Waiting on you (N)" header (answer_digest.rb:212).
- [ ] [code-reviewer] `render_digest` reads `row.stage` unguarded (answer_digest.rb:215) while sibling `task_descriptor` (:206) guards with `respond_to?(:stage)`; a stageless row crashes the digest into the `call` rescue (exit 1 → backoff) instead of degrading.
- [ ] [code-reviewer] The shared `kind: :digest` slot serializes the two digests (dispatch_digest sets `kind: :digest`; `can_dispatch_digest?` allows one); on a multi-day catch-up the 09:00 answer-digest can lag by the shipped digest's runtime — self-healing, but undocumented coupling.
- [ ] [silent-failure] `emit_error_envelope` (answer_digest.rb:270-280) has no guard of its own and calls `SCHEMA_VERSIONS.fetch`/`@output.puts`; if either raises, the replacement exception masks the original error before the bare `raise` (:135) can re-raise it.
- [ ] [silent-failure] `dispatch_digest`'s `running_task?`-gated `complete(exit 1)` (dispatcher.rb:1195-1198) double-counts backoff if a spawn succeeds but raises before `record_dispatch` — the later reap calls `complete` again on the same child (narrow window; the comment's reasoning doesn't cover it).
- [ ] [silent-failure] All `WaitingRows` selection entry points default `logger: nil` (select/button_for/resolve/daemon_plan_pause?/daemon_enabled_resolver); every current caller injects a real logger, but a future caller that forgets gets exactly the silent human-blocking-row drop this feature exists to prevent — consider a required logger kwarg on the selection surfaces.
- [ ] [type-design] `Task` (answer_digest.rb:41) is a plain `Data.define` with no initialize, so `Task.new(title: nil, …all nil)` is valid even though the schema marks `title` non-null; `Result`'s `tasks.all?(Task)` checks only class membership (safe today only because `display_title` always returns a non-empty string).
- [ ] [type-design] `Result` leaves illegal states representable: it never asserts `count == tasks.size`, and `reason:"empty"` with `count > 0` / non-empty `tasks` / non-blank `message` (and `reason:"dry_run"` with empty tasks) all pass the guards (answer_digest.rb:44-81), so a self-contradictory JSON envelope is constructible.
- [ ] [type-design] The sent-`chat_id` guard uses `chat_id.to_s.empty?` (answer_digest.rb:50), accepting integer `0` as a valid chat (Telegram chat ids are never 0); `Sender.resolve_chat_id`'s `blank?` is likewise 0-blind, so a `chat_id_allowlist: [0]` misconfig flows through as "sent to chat 0."
- [ ] [type-design] `Result.new(tasks: nil)` raises `NoMethodError` from `tasks.all?` (answer_digest.rb:73) instead of the clean `ArgumentError` every other guard produces — boundary-contract asymmetry.
- [ ] [comment-analyzer] `row_project_path` comment (waiting_rows.rb:134) lists `/status` as a consumer, but `/status` never reaches it (only `/waiting` + the digest do, via `daemon_enabled_resolver`).
- [ ] [comment-analyzer] The reap comment (dispatcher.rb:597-601) still describes only the singular "digest" pseudo-project, but the guard it explains was generalized in this PR to `!global_digest_stage?(entry.stage)` (:603), which also protects the new `answer_digest` phantom.
- [ ] [comment-analyzer] CLI exit-codes block (cli.rb:725-729) lists 0/1/64/78 and never 70, while the schema `exit_code` enum includes 70 (`hive-answer-digest.v1.json:120`); 70 is unreachable for this command, so schema over-lists or CLI under-lists — they disagree.
- [ ] [comment-analyzer] Schema top-level description (hive-answer-digest.v1.json:5) groups `config` and `status_unavailable` under the heading "Usage errors," mislabeling two of the three error categories (only the Thor `usage` path is a usage error).
- [ ] [comment-analyzer] Schema `button_count` description ("attached to the sent message", ~:56) is inaccurate for the `--dry-run` envelope, where `button_count` is populated but nothing is sent.
- [ ] [comment-analyzer] wiki/commands/answer-digest.md:32 says `Hive::EnvFile.load!` runs "for real sends only," but the code is `@env_loader.load! unless @dry_run` — it also runs on a non-dry-run empty snapshot (no send); only the token read is send-only.
- [ ] [code-simplifier] `button_for` hoists 4 locals (project/slug/marker/row_action, waiting_rows.rb:80-83) used only by its rescue, unlike the two sibling rescues (resolve :106, daemon_plan_pause? :120) that inline `(row.x if row.respond_to?(:x))`.
- [ ] [code-simplifier] The dispatcher `global_digest_stage?`/`global_digest_action`/`global_digest_scheduler` trio (dispatcher.rb:1210-1231) spells the digest-stage list three times; a single frozen `{DIGEST_STAGE=>"digest", ANSWER_DIGEST_STAGE=>"answer_digest"}` map collapses `global_digest_stage?` (→`.key?`) and `global_digest_action` (→`[stage]`).
- [ ] [code-simplifier] `Hive::Pr.number(row.respond_to?(:pr_url) ? row.pr_url : nil)` is duplicated verbatim in `task_descriptor` (answer_digest.rb:207) and `pr_label` (:224); extract a private `pr_number(row)` helper.
- [ ] [code-simplifier] `waiting_rows(_cfg)` (answer_digest.rb:165, called at :118) ignores its `cfg` argument — drop the param and argument.

PR #611, opened 2026-06-27
feat(review): add ad-hoc PR review entry point

ivankuznetsov:
## Summary

Hive's autonomous review loop could only run against PRs Hive itself opened (the
normal `4-execute → 5-open-pr` pipeline) or PRs patrol opened. Reviewing someone
else's PR meant doing it by hand. This branch adds `hive review --pr 197`: from
inside any repo registered with `hive init`, it spins up the full stage-6 review
loop against an existing GitHub PR, posts reviewer findings as PR comments, and
parks the task at `6-review` for you to act on.

The feature is an assembly of machinery that already exists rather than new review
logic. It reuses the project registry, patrol's synthetic-task pattern, the
babysitter's fork-agnostic PR-head fetch, the unchanged review stage, and the
existing GitHub comment mirror. The bare form `hive review 197` is untouched and
still resolves task id `197`; ad-hoc review is reachable only through the
disambiguated `--pr` flag.

```mermaid
flowchart TB
  CLI["hive review --pr 197"] --> AR["Commands::AdhocReview"]
  AR --> RP["Config.registered_project!\n(cwd or --project)"]
  AR --> ID["Pr.identifier_to_number\n197 / #197 / URL"]
  AR --> GH["Gh.pr_metadata\n(base, head, fork, state)"]
  AR --> COL{"PR already owned\nby a hive task?"}
  COL -->|yes| REFUSE["refuse: never shadow"]
  COL -->|no| WT["Worktree.materialize_pr\n→ worktree_root/<slug>"]
  WT --> SC["write sidecars\nsource: ad-hoc"]
  SC --> SA["StageAction run_at 6-review"]
  SA --> REV["Stages::Review.run!\n(unchanged)"]
```

### What's new, and why this shape

| Piece | Decision |
|---|---|
| Identifier parsing (`Pr.identifier_to_number`) | `Pr.number` only matched `/pull/` URLs and returned nil for `197`/`#197`. A dedicated normalizer accepts all three forms (and rejects non-positive numbers) while leaving `Pr.number`'s semantics untouched, since status/TUI/bot depend on its nil-for-non-URL behavior. |
| Project resolution (`Config.registered_project!`) | Resolves the target from cwd (longest registered-path prefix wins for nested checkouts) or `--project NAME`, and requires `.hive-state` to exist. Not invited → fails loudly with a `hive init` hint; a registered-but-deleted `.hive-state` gets its own distinct message. No ephemeral mode. |
| Worktree (`Worktree.materialize_pr`) | The babysitter's fetch-pattern (`git fetch origin +pull/N/head` then `worktree add -B`) extracted into a shared helper that the babysitter now delegates to, using the non-interactive fetch env so the new CLI entry point can't hang on a credential prompt. Fork-agnostic, placed at the normal `worktree_root/<slug>` so the task is a first-class `hive status` / `hive drop` citizen. |
| Synthetic task (`Commands::AdhocReview`) | Mirrors patrol's handoff: creates `6-review/adhoc-review-pr-N/` directly, skipping stages 1–5, with `meta.yml` / `idea.md` / `task.md` / `pr.md` / `worktree.yml` and an empty `reviews/`, tagged `source: ad-hoc`, `workflow: coding`. A create-phase failure tears down both the task folder and the worktree/branch/ref so a retry can't wedge. |
| Refuse-to-shadow | Collision detection scans both the reserved slug across stage dirs and every task's `pr.md` `pr_number`. If another hive task already owns the PR, the command refuses (naming that slug) rather than creating a second folder that would collide on the worktree path and double-post comments. Re-running reuses the existing task (validated as `source: ad-hoc` with a matching `pr_number`). |
| Review-stage `source: ad-hoc` | `reviewer_specs_for` uses `review.adhoc.reviewers` when set, else falls back to `review.reviewers`. Both the Phase-1 CI-fix and the Phase-4 fix agent are gated off by default — accepted findings pause as `REVIEW_WAITING reason=adhoc_fix_disabled` instead of committing to someone else's branch. `review.adhoc.fix: true` opts back in. A clean ad-hoc review is classified as non-advancing so the daemon parks it at `6-review` instead of finalizing it. |

CLI wiring delegates to the existing `StageAction` `run_at` idempotent path, so
standard markers (`REVIEW_COMPLETE` / `REVIEW_WAITING`) and the `--json` envelope
come for free — including for ad-hoc create/setup failures.

### Scope boundary (v1)

The review stage resolves its diff base only from the project default branch; there
is no per-task base override today. v1 relies on the common case (PR base ==
default branch) and **records** the PR's `baseRefName` in `pr.md` for transparency.
Honoring a non-default PR base is a follow-up. GitHub PRs only (every PR op shells
`gh`). No auto-advance, auto-merge, or auto-cleanup — the task stays parked at
`6-review`; teardown is an explicit `hive drop adhoc-review-pr-N`. Re-running after
the author pushes new commits reviews the pinned first-run head; the documented
workaround is `hive drop` + re-run (see `wiki/cli.md`).

## Test plan

New coverage spans each implementation unit, using real git fixtures and the
established `gh`/`git` stub harness (no AI-provider stubs):

- **PR identifier parsing** — `197` / `#197` / full `/pull/N` URLs, plus rejection
  of junk and non-positive numbers (`test/unit/pr_test.rb`).
- **Project resolution** — cwd and nested-subdir resolution, `--project NAME`,
  the not-invited error, the registered-but-missing-`.hive-state` error, and
  relative `hive_state_path` handling (`test/unit/config_test.rb`).
- **`gh pr view` metadata** — happy path, fork (`isCrossRepository`) PRs, the
  non-Hash-JSON guard, and `--project` → `chdir` scoping (`test/unit/gh_test.rb`).
- **PR-head materialization** — against a local bare-repo fixture, with the
  babysitter test kept green through the shared-helper refactor
  (`test/unit/worktree_test.rb`, `test/unit/babysitter/...`).
- **Ad-hoc enqueue** — fresh create, reuse, refuse-to-shadow (slug-owner and
  pr-owner precedence), fork PRs, the create-failure worktree cleanup, the
  `--json` error envelope, and lock-contention arms
  (`test/unit/commands/adhoc_review_test.rb`).
- **Review-stage behavior** — reviewer selection, the CI-fix / fix-phase ad-hoc
  gates, fail-closed-on-read-error diagnostics, the empty-reviewer warning, and
  the non-advancing clean-review classification
  (`test/unit/stages/review/run_reviewers_test.rb`, `test/unit/task_action_test.rb`,
  `test/unit/daemon/policy_test.rb`).
- **CLI routing + integration** — `--pr` vs bare target vs both, `--pr` + `--from`
  guard, and the CLI → enqueue → `StageAction` path
  (`test/unit/cli_test.rb`, `test/integration/adhoc_review_test.rb`,
  `test/integration/run_review_test.rb`).

Full suite green (0 failures) and rubocop clean across all touched files; the repo's
100%-line-coverage gate is satisfied.

## Review summary

Three compound-engineering review passes ran against this branch (claude + codex +
the multi-persona PR toolkit). No High-severity defect survived triage; every
Requirements-Trace item (U1–U8) is implemented and exercised.

- **Pass 1** surfaced the headline risks: the `--json` error envelope was bypassed
  on the ad-hoc create path, a create-phase failure leaked the git worktree and
  wedged retries, and the `review` positional going optional regressed the
  no-target usage error. All auto-fixed.
- **Pass 2** caught the safety gaps: Phase-1 CI-fix ran before the ad-hoc fix gate
  (so a configured `review.ci.command` could auto-commit on someone else's PR), the
  fetch skipped the non-interactive env, `--pr 0` parsed as a valid number, and the
  relative-`hive_state_path` readers disagreed. All auto-fixed; the reuse
  stale-head limitation was documented rather than changed (matches the plan's
  "re-run = next pass on same worktree" semantics).
- **Pass 3** caught the daemon auto-advancing a *clean* ad-hoc review past
  `6-review` toward finalize — contradicting the explicit "stays parked" guarantee
  and the "don't finalize someone else's PR" non-goal — plus diagnostic-accuracy and
  type-design polish (`Data.define` for `PrMetadata`, fail-closed gate messages,
  helper de-duplication). All auto-fixed.

The one escalation raised (pass 1: what a red-CI ad-hoc review should do) was
resolved against the acceptance bar — gate the CI-fix agent off and review the PR
as submitted — since blocking on red CI would make ad-hoc review impossible for
exactly the problematic PRs a maintainer most wants to review. Remaining
RESOLVED/NO-FIX items are out-of-scope (already fixed on `origin/main` and surfaced
only by a two-dot diff view) or deliberate v1 boundaries (PR `state` recorded for
transparency, not enforced).

## Linked task

- Hive task: **#25** — `Ad hoc PR reviews`
- Slug: `explore-ad-hoc-mode-where-260528-ddf6`
- Pipeline: planned → executed → 3 review passes (all fix-success) → finalize
- Plan / brainstorm artefacts live in `hive-private` (gitignored in this repo)

---


ivankuznetsov:

## High
- [ ] `lib/hive/commands/adhoc_review.rb:35` fetches PR metadata from the process cwd instead of the resolved project: `hive review --project B --pr N` run from repo A can query repo A's PR metadata and then fetch/materialize repo B's `pull/N/head`, violating the project override and reviewing or failing against the wrong PR.
- [ ] `lib/hive/stages/review.rb:192` runs CI-fix before the ad-hoc fix gate: with default `review.adhoc.fix: false`, a failing configured CI command can still spawn the CI fix agent and edit/commit in someone else's PR worktree before the later Phase 4 short-circuit.
- [ ] `lib/hive/cli.rb:556` hands StageAction only the deterministic slug after cwd-based ad-hoc enqueue: if any other registered project already has `adhoc-review-pr-N`, the command creates/reuses the cwd task but then slug resolution becomes ambiguous instead of honoring the cwd-selected project.

## Medium
- [ ] `lib/hive/commands/adhoc_review.rb:38` reuses any existing `6-review/adhoc-review-pr-N` folder without validating `source: ad-hoc` and matching `pr_number`: a normal or wrong-PR task with that slug can be silently treated as the ad-hoc review task.
- [ ] `lib/hive/commands/adhoc_review.rb:110` cleans up only the task folder after post-materialization failures: if `worktree add` succeeds but head verification or sidecar writes fail, the orphaned worktree/branch remains and the next retry can wedge on the existing path/ref.
- [ ] `lib/hive/cli.rb:551` runs ad-hoc setup before StageAction owns the JSON response: `hive review --pr ... --json` emits plain stderr/no standard envelope for gh/auth/collision/worktree setup failures.

## Nit

ivankuznetsov:

# Review — ad-hoc PR review (`hive review --pr <n>`)

Scope: `git diff origin/main..HEAD`. All plan requirements (U1–U8) are
implemented; full suite (`adhoc_review`, `pr`, `gh`, `worktree`, `config`,
`cli`, `run_reviewers`, integration) is green and rubocop is clean. Findings
below are real defects/risks, not plan-scope gaps.

## High

(none — every Requirements-Trace item is implemented and exercised by tests.)

## Medium
- [ ] `--json` error envelope is bypassed on the ad-hoc create path: `AdhocReview` accepts `json:` but **never uses `@json`** (`lib/hive/commands/adhoc_review.rb:20-23`), so a create-phase failure (bad identifier, not-invited, refuse-to-shadow collision) propagates as a bare `Hive::Error` to `bin/hive:259`, which only `warn`s — `emit_json_usage_error` runs solely in the `Thor::Error` arm (`bin/hive:254-258`). Verified empirically: `hive review --pr abc --json` prints a plain stderr line (exit 78), whereas `hive review <bad-slug> --json` emits the `hive-stage-action` JSON envelope (via `StageAction`'s own rescue). Plan U7 promises "`--json` standard envelope" for exactly the "not in an invited repo → loud error" case. No test covers it (`test_review_pr_outside_invited_repo_fails_before_creating_state` omits `--json`).
- [ ] Making the `review` positional optional silently regresses the no-target `--json` error: `def review(target = nil)` (`lib/hive/cli.rb:545`) now raises `Hive::InvalidTaskPath` *inside the method body* for the missing-target case (`:564`), so `hive review --json` (no target) bypasses the JSON usage envelope it previously got when Thor raised a `Thor::Error` for the missing positional. Same root cause as above — fold the fix together.
- [ ] Create-failure cleanup leaks the git worktree and wedges the task: `create_task!`'s rescue removes only `task_folder`, not the worktree `materialize_pr` already added at `canonical_root/<slug>` (`lib/hive/commands/adhoc_review.rb:110-120`). Any failure *after* `worktree add` — most plausibly `verify_head!` firing on a benign PR re-push between `Gh.pr_metadata` and the fetch (`:114`/`:122-127`), or a sidecar write error — leaves an orphan worktree; the next `hive review --pr <n>` finds `reusable_folder?` false, re-enters `materialize`, and `git worktree add` fails with "already exists", wedging the PR until a manual `git worktree remove`. This is precisely the failure mode the babysitter guards against with `remove_existing!` (`lib/hive/babysitter/worktree.rb:20,40-50`); the ad-hoc path has no equivalent. The unit test for this path stubs `materialize_pr` so no real worktree is created, masking the leak (`test/unit/commands/adhoc_review_test.rb:138-154`).

## Nit
- [ ] `verify_head!` hard-aborts on a head mismatch (`adhoc_review.rb:122-128`) even though the common trigger is a PR re-pushed between metadata fetch and `git fetch`, where the materialized head is the *newer, still-valid* PR tip — arguably the better thing to review. Combined with the worktree-leak above, a transient race becomes a persistent wedge. Consider re-fetch/accept-newer, or at least removing the worktree on this failure.
- [ ] Dead method: `Hive::Babysitter::Worktree#run_git!` (`lib/hive/babysitter/worktree.rb:52-57`) is no longer called after the refactor delegated `materialize` to `Hive::Worktree.materialize_pr`; `remove_existing!` shells `Open3` directly. The diff also changed its `err.to_s.strip` → `err.strip`, diverging from the live `run_materialize_git!` helper (`lib/hive/worktree.rb:323`) for no benefit. Delete it.
- [ ] `AdhocReview` resolves the project from `Dir.pwd` (or `--project`) but forwards only the raw `options[:project]` (possibly `nil`) to `StageAction` (`lib/hive/cli.rb:556-562`), which re-resolves the slug independently across *all* registered projects. A same-named `adhoc-review-pr-<n>` slug in two registered repos could resolve to the wrong task on the second hop; forward the already-resolved project name for consistency.
- [ ] Misleading parameter name: `Hive::Worktree.materialize_git_stdout!(repo_root, *args)` (`lib/hive/worktree.rb:330`) is invoked with the worktree `path`, not a repo root, for `rev-parse HEAD` (`:296`). Rename the param (e.g. `dir`) to avoid implying it's the repo root.

ivankuznetsov:

# PR Review Toolkit — ad-hoc PR review (`hive review --pr <n>`)

Union of findings from five personas (code-reviewer, silent-failure-hunter,
pr-test-analyzer, type-design-analyzer, comment-analyzer) over
`git diff origin/main..HEAD`. Plan scope boundaries respected: the deferred
"non-default PR base" (reviewers diff the project default branch; `base_ref_name`
is recorded for transparency only) is **not** raised as a defect.

## High
- [ ] [code-reviewer/silent-failure/tests] `create_task!` failure leaks the git worktree/branch/ref and wedges all retries — `lib/hive/commands/adhoc_review.rb:110-120`: the rescue (`:117-119`) removes only `task_folder` under `.hive-state`, never the worktree `materialize` created at `canonical_root/<slug>` (`:101-108`), nor the `hive/review/pr-N` branch or `refs/hive/review/pr-N`. `materialize_pr` (`lib/hive/worktree.rb:285-299`) runs `git worktree add` without a `remove_existing!` step, so after a `verify_head!` head-race failure (or any `write_sidecars` error) the next `hive review --pr N` hits "already exists" and is stuck until a manual `git worktree prune` — the exact wedge `Babysitter::Worktree#remove_existing!` (`lib/hive/babysitter/worktree.rb:40-50`) was written to prevent. The only test for this path stubs `materialize_pr` so the real leak is invisible (`test/unit/commands/adhoc_review_test.rb:138-154`).

## Medium
- [ ] [silent-failure/type-design] `adhoc_task?` read-failure fails OPEN on the fix gate — `lib/hive/stages/review.rb:1300-1302,1352-1359`: `adhoc_fix_enabled?` is `!adhoc_task?(task) || …`, so a transient `SystemCallError` reading `task.md` makes `adhoc_task?` return `false` → `!false == true` → the fix agent runs and auto-commits on a genuine ad-hoc/fix-off task. Safe direction for reviewer *selection*, wrong direction for the safety gate; should fail closed (treat unknown as ad-hoc/fix-disabled) and the warning should disclose the fix-gate consequence.
- [ ] [tests/code-reviewer] `--json` error envelope is never emitted on the ad-hoc create/error path — `lib/hive/commands/adhoc_review.rb:23` stores `@json` but it is **never used**; `enqueue` just raises `Hive::Error` subclasses. `bin/hive:259` only `warn`s for `Hive::Error` (JSON is emitted solely in the `Thor::Error` arm), so `hive review --pr bad --json`, not-invited, and collision all emit no envelope — asymmetric with the success path and with `hive review <bad-slug> --json`. Untested: every `--pr` failure test omits `--json` (`test/integration/adhoc_review_test.rb:137-149,156-164`).
- [ ] [code-reviewer] `gh pr view` resolves against `Dir.pwd`, not the project — `lib/hive/gh.rb:117` (`pr_metadata`) calls `capture3` with no `chdir`/`--repo` (`:434-441`), while `materialize` fetches from the resolved `project_root`. With `hive review --pr N --project NAME` run from outside that checkout, gh looks up the PR in the wrong repo (or errors), yielding a wrong PR or a `verify_head!` mismatch. The no-`--project` path is safe (cwd is always inside the project).
- [ ] [silent-failure] Blanket `rescue ArgumentError => Hive::InvalidTaskPath` spans the whole `enqueue` body — `lib/hive/commands/adhoc_review.rb:46-47`: intended only for `Pr.identifier_to_number`'s "invalid identifier", it also catches any `ArgumentError` from Config/gh/YAML/`Time.iso8601`, re-labeling a genuine programmer bug as a USAGE error (exit 64). Normalize the identifier before the begin/rescue, or rescue only around that one call.
- [ ] [silent-failure] An unrelated unreadable/racing `pr.md` aborts the entire enqueue — `lib/hive/commands/adhoc_review.rb:83-91` globs every `stages/*/*/pr.md` and calls `Gh.pr_frontmatter` (`lib/hive/gh.rb:370-381`), which rescues only `Psych::Exception`; a permission error or `Errno::ENOENT` TOCTOU race on *another* task's file propagates and is reshaped to an opaque `InternalError`, blocking ad-hoc review for every PR. Rescue per-file and skip (ideally `warn` which file).
- [ ] [silent-failure/type-design] `verify_head!` silently no-ops when the expected head is empty — `lib/hive/commands/adhoc_review.rb:122-129`: `head_ref_oid` is `doc["headRefOid"].to_s` (`lib/hive/gh.rb`), so a missing/renamed field becomes `""` and `return if expected.empty?` disables the whole "worktree matches the PR GitHub reported" guard with no `warn`. At minimum warn; ideally treat an empty head from gh as an error.

## Nit
- [ ] [all personas] Dead code: `Babysitter::Worktree#run_git!` (`lib/hive/babysitter/worktree.rb:52-57`) has no caller after the refactor (`materialize` delegates to `Worktree.materialize_pr`; `remove_existing!` uses `Open3` directly). The diff even edited it (`return out`, `err.strip`); it survives the coverage gate only via a direct `.send` test. Delete it.
- [ ] [code-reviewer/tests] `Gh.pr_metadata` is fetched (`adhoc_review.rb:35`) before the `reusable_folder?` short-circuit (`:38`), so every re-run of an already-enqueued review needlessly pays a `gh pr view` round-trip and fails if gh is offline. Move the fetch after the reuse check.
- [ ] [silent-failure] Reuse path neither re-materializes nor re-runs `verify_head!` (`adhoc_review.rb:38,54-56`); if the PR head advanced since first enqueue, the user silently re-reviews a stale worktree with no indication (`reused`/`--json` never surfaces it).
- [ ] [code-reviewer] `TaskCounter.next!` runs inside `write_sidecars` (`adhoc_review.rb:131-148`) before later sidecar writes; a failure removes the folder but leaves the global counter advanced → task-id gap. Cosmetic.
- [ ] [code-reviewer] `registered_project!` gates on the *expanded* `hive_state_path` but returns `project` carrying the raw stored value (`lib/hive/config.rb:776,926-944`), which `AdhocReview` then uses in `Dir.glob`/`File.join`; a hand-edited *relative* `hive_state_path` would let the gate and consumer disagree (resolves against `Dir.pwd`).
- [ ] [type-design] `PrMetadata` is an anemic mutable `Struct` (`lib/hive/gh.rb:32-33`) unlike its sibling result types (`CommandStatus`/`PushResult`/`ScanResult` carry predicates); a `head_known?`/`cross_repository?`/`open?` (or `Data.define`) would name the `verify_head!` skip and give the unused fields a purpose. Likewise `materialize_pr` and `enqueue` return bare hashes accessed via `fetch(:head_sha)` against the module's result-Struct idiom (`lib/hive/worktree.rb:294-299`, `adhoc_review.rb:43`).
- [ ] [type-design] `review.adhoc.reviewers` nil-vs-`[]` tri-state (`nil`=inherit `review.reviewers`, `[]`=run none) is load-bearing but undocumented at the DEFAULTS site (`lib/hive/config.rb:144-147`) — unlike `review.reviewers`, which has an explanatory comment; an operator setting `[]` expecting "inherit" gets zero reviewers silently.
- [ ] [type-design] `source: ad-hoc` routing duplicates the literal: producer defines `SOURCE = "ad-hoc"` (`adhoc_review.rb:17`) but the consumer hardcodes `"ad-hoc"` (`lib/hive/stages/review.rb:1354`); a rename silently stops recognizing ad-hoc tasks. Reference the constant.
- [ ] [type-design] Captured-but-unconsulted fields persisted to `pr.md`: `state` (no guard/warn — a CLOSED/MERGED PR is still reviewed) and `is_cross_repository` (fork flag never consulted). Informational-only is acceptable per v1 scope, but worth a comment or a non-OPEN-state warning.
- [ ] [comment] `lib/hive/pr.rb:4-8` module banner still frames the module as display-only "PR-URL helpers" and omits the new command-input parser `identifier_to_number`; and that method (`:36-47`) has no doc-comment while both siblings do — its key property (raises on junk, opposite of `number`'s nil-return) is captured only in the wiki.
- [ ] [comment] `adhoc_task?` (`lib/hive/stages/review.rb:1352-1359`) is a near-clone of `patrol_task?` but carries none of its rationale (why `strip`+`casecmp?`, why `SystemCallError`-only rescue + warn); add a one-line cross-reference.
- [ ] [comment] `wiki/modules/config.md:225` mis-attributes the `review.adhoc.permissions` rejection to `validate_review_adhoc!`, which only checks `reviewers`/`fix`; the rejection actually happens in the permissions validator (`reject_unsupported_permissions_at!`). Move the sentence to validator item 8.

ivankuznetsov:

## High
- [ ] Ad-hoc fix-off does not gate CI-fix: `lib/hive/stages/review.rb:189` runs `CiFix.run!` before the `adhoc_fix_enabled?` gate at `lib/hive/stages/review.rb:496`, and `CiFix` can spawn a mutating fix agent (`lib/hive/stages/review/ci_fix.rb:113`), so a red-CI ad-hoc review can edit/commit in the PR worktree despite the default comments-only/fix-off requirement.

## Medium

## Nit

ivankuznetsov:

# Review: ad-hoc PR review (`hive review --pr <n>`) — pass 2

Scope: `git diff origin/main..HEAD` (U1–U8). Verified against the plan's
Requirements Trace, Scope Boundaries, and Risks. rubocop clean on all 8 changed
lib files; targeted unit + integration suites all green
(adhoc_review unit/integration, worktree, gh, pr, config, cli, run_reviewers,
run_review). Every Requirements-Trace item is implemented or explicitly
deferred-and-documented (non-default base, recorded via `base_ref_name`). No
High-severity findings.

## High

(none)

## Medium

- [ ] `AdhocReview#verify_head!` fails-loud on a benign concurrent PR re-push (`lib/hive/commands/adhoc_review.rb:207-219`): the materialized worktree comes from `git fetch origin +pull/<n>/head` (authoritative current head), while `metadata.head_ref_oid` is the snapshot from the earlier `gh pr view`. If the PR is re-pushed in the window between those two calls, the worktree is *correct* but the check raises `WorktreeError`, triggers `cleanup_failed_task!`, and forces the operator to re-run — even though nothing is actually wrong. It self-heals on retry and is documented, but the tool cannot distinguish "wrong PR materialized" from "PR moved forward," so the check trades a real (if rare) spurious-failure on valid input for little added protection; consider warn-and-proceed (the fetched head is canonical) or re-reading metadata after materialize.

## Nit

- [ ] Hard-kill mid-create leaves a partial `6-review/adhoc-review-pr-<n>/` folder that wedges retry with a confusing message (`lib/hive/commands/adhoc_review.rb:174-184`, `98-108`): `create_task!` mkdir's `reviews/` and materializes before `write_sidecars`, but `cleanup_failed_task!` only runs on a Ruby exception — a SIGKILL/OOM/power-loss in that window orphans a folder with no `pr.md`. The next `--pr <n>` then hits `validate_reusable!`, reads empty frontmatter, and raises `CollisionError: … not an ad-hoc review for PR #N (source="", pr_number=0); run hive drop …`. The drop hint makes it recoverable, but the `source=""`/`pr_number=0` wording misdescribes a half-written ad-hoc task; babysitter avoids this entirely by calling `remove_existing!` proactively before materialize.
- [ ] `validate_reusable!` does not tolerate a SystemCallError on the reuse-folder's own `pr.md` (`lib/hive/commands/adhoc_review.rb:98-102`): it calls `Hive::Gh.pr_frontmatter` directly (which rescues only `Psych::Exception`), whereas the cross-task ownership scan deliberately wraps the same call in `read_pr_frontmatter_for_scan` to skip unreadable sidecars (`:142-147`). An unreadable (e.g. chmod-000) reuse `pr.md` therefore aborts the command (wrapped as `InternalError`) instead of degrading; the asymmetry with the sibling scan reads as an oversight.
- [ ] `adhoc_review.rb` references `Hive::Schemas::ErrorEnvelope` (`:319`) without `require "hive/schemas"` at the top of the file; it resolves only because the umbrella `require "hive"` loads schemas first (confirmed reachable + green in the integration tests). This matches existing repo convention, but an explicit require would make the dependency robust against future load-order changes.

ivankuznetsov:

# pr-review-toolkit multi-persona review — pass 2

Feature: ad-hoc PR review (`hive review --pr <n>`). Personas run: code-reviewer,
silent-failure-hunter, type-design-analyzer, comment-analyzer, pr-test-analyzer,
code-simplifier. Union of findings below (deduplicated; corroborating personas
listed in brackets). No persona surfaced a Critical/breaking defect; the
implementation is solid and well-tested.

## High

(none — no persona reported a critical or merge-blocking defect)

## Medium

- [ ] [silent-failure, comments, tests] Ad-hoc task that resolves to **zero reviewers passes silently** (`lib/hive/stages/review.rb` ~1172-1183): `run_reviewers`' empty-specs early return warns only for `patrol_task?`; an ad-hoc task with `review.adhoc.reviewers: []` (or `nil`-inherit when `review.reviewers` is also empty) returns `:ok`, stamps `REVIEW_COMPLETE`, and may post a "clean" GitHub comment with no reviewer ever run — indistinguishable from a real clean pass. The patrol sibling already treats the identical case as warn-worthy; ad-hoc deserves the same `elsif adhoc_task?(task)` warn.
- [ ] [comments/reliability] `Worktree.materialize_pr` fetch is **unbounded** (`lib/hive/worktree.rb:292`,`:322`): `run_materialize_git!` calls bare `Open3.capture3("git", …)` with no env, so it skips `NONINTERACTIVE_FETCH_ENV` (`GIT_TERMINAL_PROMPT=0`, `GIT_HTTP_LOW_SPEED_*`) — the new *foreground* `hive review --pr` can hang on a credential prompt or stalled remote, the exact failure the sibling `fetch_origin_branch` (`:8-18`) documents 12 lines to prevent. Either thread the env through or document why it is deliberately left unbounded.
- [ ] [type-design] `AdhocReview::CollisionError` re-invents `Hive::DestinationCollision` (`lib/hive/commands/adhoc_review.rb:17`,`:331`): both map to `error_kind "destination_collision"`, but `Hive::Schemas::ErrorEnvelope` populates the structured `payload["path"]`/owner field only for `DestinationCollision`, so the ad-hoc collision `--json` envelope ships prose-only and `error_kind_for` hand-duplicates the central mapping — subclass/reuse `DestinationCollision` to restore the structured surface agents consume.
- [ ] [type-design] `materialize_pr` returns a loose `{path:, branch:, head_sha:}` Hash for a contract crossing two modules (`lib/hive/worktree.rb:285`): `adhoc_review.rb` symbol-fetches `:head_sha` for the `verify_head!` cross-check while `babysitter/worktree.rb:27` immediately re-wraps and discards it — a `MaterializedWorktree` Struct (sibling to `PrMetadata`/`PushResult`) would express the head-SHA contract once.
- [ ] [simplify] `registered_project!` resolves `project_hive_state_path(project)` only to discard it, while `enqueue` re-reads raw `project.fetch("hive_state_path")` (`lib/hive/config.rb:920-937`): a configured *relative* `hive_state_path` (passed through unexpanded by `registered_projects`, `:782`) is validated as absolute but used raw at create-time — latent inconsistency plus the identical `ConfigError` message raised twice; have `registered_project!` return the resolved path as the single source of truth.
- [ ] [simplify] `run_materialize_git!` and `materialize_git_stdout!` are byte-for-byte identical except the success return (`return` vs `return out`) (`lib/hive/worktree.rb:322-336`): collapse into one `materialize_git!(dir, *args)` returning stdout; fetch/worktree-add callers ignore the return.
- [ ] [simplify] `adhoc_task?` is an explicit near-clone of `patrol_task?` (`lib/hive/stages/review.rb:1336-1371`): same read → `source.to_s.strip.casecmp?` → `rescue SystemCallError` → warn; the only deltas (`on_read_error` direction, warn text) parameterize cleanly into a shared `task_source_is?` helper (the in-code comment already concedes the clone).
- [ ] [tests] The `--project`/cwd → "query the right repo, not cwd" behavior is **asserted nowhere** (`test/unit/gh_test.rb`, `test/unit/commands/adhoc_review_test.rb`): every `pr_metadata`/`enqueue` stub swallows `**_kwargs` and never checks `chdir:`, so dropping `chdir: project_root` would silently regress the headline feature with green tests.
- [ ] [tests] Collision `--json` envelope is untested (`lib/hive/commands/adhoc_review.rb:333`): the JSON-envelope integration tests cover only `error` (config) and `invalid_task_path` (usage); the `destination_collision` kind could ship wrong unnoticed.
- [ ] [tests] `verify_head!` empty-head skip branch has no asserting test (`lib/hive/commands/adhoc_review.rb:210`): every fixture supplies a non-empty head, so a regression flipping the skip/raise guard wouldn't be caught.
- [ ] [tests] Explicit `review.adhoc.reviewers: []` (the documented foot-gun) → zero-reviewers path is untested (`test/unit/stages/review/run_reviewers_test.rb`): only the `nil`→inherit fallback is covered, leaving the silently-unreviewed outcome and the patrol-vs-adhoc warn asymmetry unpinned.

## Nit

- [ ] [code-reviewer, silent-failure] `cleanup_failed_task!` runs `git worktree prune` before `FileUtils.rm_rf(worktree_path)` (`lib/hive/commands/adhoc_review.rb:196-205`), and `materialize_pr` has no defensive pre-clean (unlike babysitter's `remove_existing!`): if `worktree remove --force` fails, the dir survives prune so stale `.git/worktrees/<slug>` metadata lingers and the next run wedges on "already exists in the worktree list" (self-heals on the *following* retry). Reversing to `rm_rf` then `prune` cleans it in one shot.
- [ ] [silent-failure] `cleanup_failed_task!` best-effort guarantee is scoped to git *non-zero exits* only (`lib/hive/commands/adhoc_review.rb:196-205`): if a `capture3` itself raises (spawn-time `SystemCallError`) it propagates out and the `create_task!` `rescue … ; raise` never re-surfaces the real cause — wrap the git calls so cleanup can never raise.
- [ ] [silent-failure] `emit_error_envelope` / `emit_review_usage_error` rescue `Errno::EPIPE, JSON::GeneratorError` but not `IOError` (`lib/hive/commands/adhoc_review.rb:326`, `lib/hive/cli.rb:1119`): on a fully-closed stdout the `puts` raises `IOError`, the subsequent `raise` of the real error is skipped, and the user sees `IOError` instead of the underlying `GhError`/`CollisionError`.
- [ ] [code-reviewer] `registered_project!` raises "not a hive-invited repo — run `hive init` here first, or pass --project NAME" even when an explicit `--project NAME` typo is the cause (`lib/hive/config.rb` ~925) — misleading wording for the `--project` path.
- [ ] [type-design] `PrMetadata` normalization (`number.to_i`, `isCrossRepository == true`, `.to_s`) lives only in the `pr_metadata` factory, not the type (`lib/hive/gh.rb:32`): the keyword-init Struct permits unnormalized instances; a `PrMetadata.from_gh_json(doc)` would co-locate the coercions with the type.
- [ ] [type-design] `write_worktree_pointer` hand-writes `worktree.yml` (`path/branch/created_at/execute_base_head`), duplicating the canonical pointer schema already owned by `Worktree#write_pointer!` (`lib/hive/commands/adhoc_review.rb:273` vs `lib/hive/worktree.rb:182`) — two independent writers of a contract read by execute/babysitter can drift.
- [ ] [type-design] The review PR head SHA is stored under key `execute_base_head` (`lib/hive/commands/adhoc_review.rb:280`) — semantically the *execute* stage's concept on a review-only task; compatibility-driven (execute.rb reads it) but worth a comment or neutral alias.
- [ ] [type-design] `PrMetadata#state` / `#is_cross_repository` are captured and written to `pr.md` but never gate logic (`lib/hive/gh.rb`, `adhoc_review.rb:294-295`): nothing checks `state`, so a CLOSED/MERGED PR is still materialized and reviewed — informational only as written (not plan-required; flag only if an OPEN-state guard was intended).
- [ ] [type-design] `Pr.identifier_to_number` regex `/\A#?(\d+)\z/` accepts `0`/`#0` (`lib/hive/pr.rb:47`), producing slug `adhoc-review-pr-0` that only fails later at `gh pr view 0`; a `> 0` check would reject it at the usage boundary.
- [ ] [type-design] The `source: ad-hoc` fact is written into three sidecars (`idea.md`/`task.md`/`pr.md`) from one `SOURCE` constant, two of which are independently load-bearing (`task.md` drives `adhoc_task?` routing, `pr.md` drives `validate_reusable!`) — co-written so they can't drift, but diffuses one authoritative marker across three files.
- [ ] [comments] "Mirror `Hive::Babysitter::Worktree#remove_existing!`" (`lib/hive/commands/adhoc_review.rb:186`) overstates the analogy — `cleanup_failed_task!` is a superset (also deletes the task folder, branch, and ref); the enumerated list that follows is accurate, but "Mirror" implies a parity that doesn't exist.
- [ ] [tests] `read_pr_frontmatter_for_scan` `SystemCallError` skip-and-warn (`lib/hive/commands/adhoc_review.rb:144`) has no asserting test — no fixture makes a sidecar unreadable mid-scan.
- [ ] [tests] The `rescue StandardError => InternalError` wrap and its `--json` envelope (`lib/hive/commands/adhoc_review.rb:58`) is unexercised — only `Hive::Error` reaches the tested error path.
- [ ] [tests] `test_review_pr_creates_task_and_runs_review_stage` stubs `Hive::Stages::Review.run!` wholesale (`test/integration/adhoc_review_test.rb:75`), so "runs the review stage" is a call-ran assertion; the real CLI→AdhocReview→StageAction→`run!` chain is never exercised end-to-end (mitigated by sidecar/marker assertions + real fix-gate coverage elsewhere).
- [ ] [tests] `project_hive_state_path` relative and nil/empty branches are effectively unreachable from tests (`lib/hive/config.rb:944`) because `register_project` always stores an absolute path — only the absolute branch is exercised.
- [ ] [tests] `idea.md`/`task.md` `created_at`/`started_at` are written from `now:` but unverified (only `worktree.yml`'s `created_at` is asserted) (`test/integration/adhoc_review_test.rb:95`).
- [ ] [simplify] `emit_error_envelope` / `emit_review_usage_error` triplicate StageAction's `ErrorEnvelope.build(schema: "hive-stage-action", verb: "review") → puts → rescue EPIPE/GeneratorError` construct (`adhoc_review.rb:318`, `cli.rb`) — a shared `emit_stage_action_envelope` helper would unify it (defensible as-is).
- [ ] [simplify] `existing_slug_owner` and `existing_pr_owner` both end in `Dir.glob(...).filter_map{…}.first` building the same `{stage:, slug:, folder:}` hash (`lib/hive/commands/adhoc_review.rb:118-135`) — extract `owner_for(folder)`.
- [ ] [simplify] `path_prefix?` re-implements the `path == prefix || start_with?("#{prefix}#{SEPARATOR}")` containment idiom already in `worktree.rb`'s `validate_pointer_path` (`config.rb:939-942` vs `worktree.rb:307`) — minor cross-module duplication.
- [ ] [simplify] `write_idea_md`'s `<<~MD` heredoc body is indented inconsistently with its `write_task_md`/`write_pr_md` siblings (`lib/hive/commands/adhoc_review.rb:240`) — cosmetic only (squiggly heredoc).

ivankuznetsov:

## High
- [ ] `gh` dry-run passthrough still honors agent-controlled proxy/TLS env: `bin/hive-babysitter-stub-gh:458` now forwards allowlisted real-`gh` reads without clearing `HTTPS_PROXY`/`ALL_PROXY` or `SSL_CERT_FILE`/`SSL_CERT_DIR`, so a dry-run agent can MITM authenticated GitHub reads.

## Medium
- [ ] Ad-hoc `--project` handoff loses relative `hive_state_path` normalization: `lib/hive/cli.rb:561` hands the new slug to `StageAction`, whose `TaskResolver` still joins the raw registry value at `lib/hive/task_resolver.rb:139`, so `hive review --pr N --project NAME` can create the task under the project root and then fail to run it from another cwd.
- [ ] Dry-run skip logs can write through hardlinks again: `bin/hive-babysitter-stub-gh:17` and `bin/hive-babysitter-stub-git:22` removed the `nlink == 1` guard, letting an agent append skipped-command audit text to another same-user file via a hardlink.

## Nit

ivankuznetsov:

# Code Review — ad-hoc PR review (`hive review --pr <n>`) — pass 3

Scope reviewed: `git diff origin/main...HEAD` (branch contribution since merge base
`03c3d34`; the two-dot `origin/main..HEAD` view also shows unrelated deletions from
origin/main's divergence, which are not part of this branch).

Overall: the feature is an assembly of existing machinery as the plan intends. All
Requirements-Trace items are implemented (U1–U8), error/`--json` envelopes are
symmetric with `StageAction`, refuse-to-shadow / reuse / cleanup are correct, and
tests use real git fixtures rather than over-mocking (gh/git stubs only — not AI
providers). `bundle exec rake test` green on all touched suites; rubocop clean on
all changed `lib/` files. No High-severity defects found.

## High
- [ ] (none)

## Medium
- [ ] `Hive::Worktree.materialize_pr` expands `path` with `File.expand_path` (resolves against `Dir.pwd`), but the babysitter's `remove_existing!`/`path` still pass the *raw* value to `git -C repo_root` (`lib/hive/worktree.rb:285`, `lib/hive/babysitter/worktree.rb:30`): for a hand-edited **relative** `hive_state_path`, `registered_projects` returns it verbatim, so the babysitter would `worktree add` under the daemon's cwd while `remove_existing!` looks under repo_root — a remove/add mismatch that can orphan or wedge the worktree, diverging from the plan's "keep babysitter behavior identical" contract; the kept babysitter test uses an absolute `hive_state_path` so it can't catch it (real-world impact is low since `hive init`/`register_project` only ever write absolute paths).

## Nit
- [ ] Reuse path returns before any `gh pr view`/re-fetch and pins to the first-run head with no staleness warning, so re-running `hive review --pr N` after the author pushes new commits silently re-reviews the original head — already documented as a known limitation in `wiki/cli.md:114` (workaround: `hive drop` + re-run); flagging only so the no-warning trade-off is a conscious sign-off, not a gap.
- [ ] `verify_head!` (`lib/hive/commands/adhoc_review.rb:292`) raises a hard `WorktreeError` and tears down the freshly-created task when a benign PR re-push lands between metadata fetch and `materialize`; it self-corrects on the next run, but a warn-and-proceed (review the head actually materialized) would avoid discarding a just-built worktree on a transient timing race — current behavior is deliberate per the inline comment, so this is optional.
- [ ] `--from` is silently ignored when `--pr` is supplied (`lib/hive/cli.rb`); harmless since `--from` only disambiguates same-slug tasks, but an explicit `--pr` + `--from` combo accepts and drops `--from` without comment.

ivankuznetsov:

# PR Review Toolkit — multi-persona review — ad-hoc PR review (`hive review --pr <n>`) — pass 3

Personas run (union captured below): `[code]` general code review, `[tests]` test
coverage, `[comments]` comment accuracy, `[errors]` silent-failure hunt,
`[type-design]` type/data-shape design, `[simplify]` simplification. Reviewed the
branch's actual contribution (three-dot `origin/main...HEAD`). Rubocop clean and all
touched test suites pass. The branch is unusually polished; most error handling is
deliberately narrowed + documented. The one High below was independently surfaced by
`[code]`/`[type-design]` and verified directly against `daemon/policy.rb` +
`task_action.rb`.

## High
- [ ] [code/type-design] Daemon auto-advances a clean ad-hoc review past 6-review on daemon-enrolled projects: `:review_complete` → `READY_TO_ARTIFACTS` (`task_action.rb:86`) is in `Policy::ADVANCE_ACTIONS` (`daemon/policy.rb:63`) with no source/provenance gate (`advance_action?` :213), so a clean PR reaching `REVIEW_COMPLETE` is dispatched `hive artifacts <slug>` and advanced toward finalize — contradicting the Requirements-Trace guarantee "Stays parked at 6-review; no auto-advance" and the explicit non-goal "Finalizing someone else's PR" (plan's mitigation note only covered the PrMergeWatcher/merge path, not the artifacts-advance path; triggers when the project is daemon-enrolled AND the review is clean, since findings + fix-off park at REVIEW_WAITING instead).

## Medium
- [ ] [code/comments] The `elsif adhoc_task?(task)` zero-reviewer branch comment "(nil inherits review.reviewers and never reaches here.)" is factually wrong (`review.rb:1197-1202`): with the DEFAULT `review.reviewers: []`, a nil `review.adhoc.reviewers` falls back to an empty set and DOES reach this branch, so the warning fires and points the operator at `review.adhoc.reviewers` when the actually-empty knob is `review.reviewers` — a normal review on the same project is silent but the ad-hoc one is noisy and misdirects.
- [ ] [errors/code] CI+fix gate fails closed for a NORMAL task on a transient state-file read error: `adhoc_fix_enabled?` calls `adhoc_task?(on_read_error: true)` (`review.rb:1320-1325`), so a `SystemCallError` reading `task.md` makes a non-ad-hoc task skip CI (`:200`) and stop at `REVIEW_WAITING reason=adhoc_fix_disabled` (`:507`) — the reason string actively misdescribes why a normal task halted; low probability (task.md was just read for the entry marker) but the blast radius reaches ordinary tasks.
- [ ] [type-design] PR `state` is captured (`gh.rb:137`) and written to `pr.md` (`adhoc_review.rb:380`) but no consumer gates on it, so `hive review --pr <merged/closed>` materializes a worktree and runs a full review of an already-merged PR (and with `review.adhoc.fix: true` spawns a fix agent on it) — a one-line refuse/warn on `metadata.state != "OPEN"` in `enqueue` would close it.
- [ ] [tests] The PR-number half of the reuse refuse-to-shadow guard is unasserted: `validate_reusable!` gates on `source.casecmp?(SOURCE) && owned_pr == pr_number.to_i` (`adhoc_review.rb:115`), but no test exercises a `source: ad-hoc` reuse folder whose `pr_number` differs from the requested PR — deleting `&& owned_pr == pr_number.to_i` leaves every test green while a corrupted/hand-edited `pr.md` would be silently reused against the wrong PR's worktree.

## Nit
- [ ] [tests] `Gh.pr_metadata`'s non-Hash-JSON guard (`gh.rb:129`, `/expected Hash/`) is unasserted, unlike its sibling `pr_stats` which has `test_pr_stats_raises_when_json_is_not_a_hash`; add a `pr_metadata` test returning `"[]"`.
- [ ] [code] `TaskCounter.next!` in `write_meta` (`adhoc_review.rb:316`) is not tolerant of `Hive::ConcurrentRunError`, unlike the patrol handoff it mirrors (which rescues it and proceeds with `id: nil`); under >30s lock contention it discards the already-completed fetch+worktree and fails the enqueue.
- [ ] [code] `Gh.pr_metadata` is called with only `chdir:` (`adhoc_review.rb:56`), so `cfg` is nil and the `gh pr view` uses the default `network_timeout_sec` instead of the project's configured timeout; sibling gh helpers thread `cfg:`.
- [ ] [errors/code] `registered_project!` raises the identical "not a hive-invited repo — run `hive init`" message for two distinct failures (project-not-found vs registered-but-`.hive-state`-dir-deleted, `config.rb:929`/`:936`); the second case's remediation is misleading — differentiate the messages (and the duplicated literal can be hoisted).
- [ ] [errors] `emit_error_envelope`'s `rescue Errno::EPIPE, JSON::GeneratorError` has an empty body (`adhoc_review.rb:411`): the `EPIPE` arm is justified, but the `JSON::GeneratorError` arm silently hides a non-serializable-payload bug (the real error still re-raises) — add a `warn` to that arm (same shape at `cli.rb` `emit_review_usage_error`).
- [ ] [errors] `verify_reusable_worktree!` reports an *unreadable* `worktree.yml` (permission error swallowed by `read_pointer` → nil) as "worktree is missing (no worktree.yml path)" (`adhoc_review.rb:137`), conflating "pointer unreadable" with "worktree gone" and pointing the operator at `hive drop` for the wrong cause.
- [ ] [errors] `enqueue` reimplements the house `EnvelopeEmitter` pattern inline (`adhoc_review.rb:60-67` + `403-414` duplicate `lib/hive.rb:240-259`) instead of `include`-ing it, so the two error-envelope implementations can drift.
- [ ] [type-design] `Gh::PrMetadata` is a mutable `Struct` with public writers (`gh.rb:32`); the repo uses immutable `Data.define` widely (Ruby 3.4) — `Data.define` keeps `pr_metadata` as the single validated constructor and makes the value tamper-proof downstream.
- [ ] [type-design] Sidecar `pr.md` writes `pr_number: metadata.number` (gh-sourced, `doc["number"].to_i` coerces absent→0) while slug/branch/reuse-check use the validated parsed `pr_number` (`adhoc_review.rb:307` vs `:87`/`:115`/`:202`); a malformed gh response could write `pr_number: 0` under an `…-pr-N` slug and mis-fire reuse detection — thread the parsed `pr_number` or assert `metadata.number == pr_number`.
- [ ] [type-design] `enqueue`'s `reused:` flag (`adhoc_review.rb:59`/`:82`) is dead outside tests — the CLI (`cli.rb:556-566`) reads only `:slug`/`:project`, so a reuse vs fresh-create is indistinguishable in CLI/`--json` output.
- [ ] [simplify] `run_materialize_git!` and `materialize_git_stdout!` (`worktree.rb:329`/`:338`) differ only cosmetically (`Open3.capture3({}, …)` ≡ `capture3(…)`, and `return out` covers the value-ignoring callers); collapse into one helper returning `out`.
- [ ] [simplify] `File.join(hive_state_path, "stages", REVIEW_STAGE, slug)` is computed identically in `reuse`, `reusable_folder?`, and `create_task!` (`adhoc_review.rb:81`/`:91`/`:206`) — extract a `review_task_folder` helper to remove drift risk.
- [ ] [simplify] The `:folder` key built into the collision-owner hashes (`existing_slug_owner` :154, `existing_pr_owner` :165) is never read by any consumer (`raise_collision!`/`refuse_if_owned!` read only `:slug`/`:stage`) — dead data, removable.
- [ ] [tests] Minor assertion gaps: `verify_reusable_worktree!`'s "no worktree.yml path" arm, an ad-hoc fix-off task reaching `REVIEW_COMPLETE` cleanly, the fix-opt-in path's *final marker* (the opt-in test asserts only that the finding was accepted), and a malformed `review.adhoc.reviewers` *entry* (vs the array-shape) under that label — all low-risk but currently relying on inverse/line-coverage rather than direct assertions.

PR #614, opened 2026-06-27
Hive patrol: gh dry-run passthrough keeps proxy trust env

ivankuznetsov:
## Summary

Security hardening of the babysitter dry-run `gh` stub (`bin/hive-babysitter-stub-gh`).

After allowlisting a read-only `gh` command, the stub already scrubbed pager,
editor, browser, host, repo, config, and enterprise-token environment variables
before `exec`ing the real `gh` binary — but it left proxy and TLS trust-store
variables intact. A dry-run agent could therefore route an otherwise-allowed
`gh pr view` / `gh api` read through an agent-controlled proxy with an
agent-controlled CA, bypassing the host-redirect guard and capturing the GitHub
credentials `gh` uses for the read.

The fix extends that same exec-seam scrub list to delete both upper- and
lower-case proxy variables (`HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY`,
…) and TLS trust-store override variables (`SSL_CERT_FILE`, `SSL_CERT_DIR`, …)
before passthrough, consistent with the stub's established deny-all clean-slate
design.

## Test plan

- `test/unit/babysitter/dry_run_env_test.rb` extends
  `test_gh_stub_scrubs_exec_influencing_environment_before_passthrough` to set
  hostile proxy/CA env and assert each variable is recorded as unset for an
  allowed `gh` passthrough.
- Regression run from the worktree:

  ```
  ruby -Itest test/unit/babysitter/dry_run_env_test.rb \
    -n test_gh_stub_scrubs_exec_influencing_environment_before_passthrough
  # 1 runs, 30 assertions, 0 failures, 0 errors, 0 skips
  ```

- Full gates: `bundle exec rubocop` (lint) and `bundle exec rake test` (suite)
  both passed.

Diff vs `main`: 2 files changed, 19 insertions(+), 1 deletion(-).

## Review summary

- 6-review flow completed (pass 1). High: none. Nit: none.
- One Medium ("preserve trusted proxy/CA settings for gh passthrough") was
  triaged RESOLVED / NO-FIX and suppressed: the patrol task Recommendation
  explicitly sanctioned deleting these vars, the stub follows an established
  deny-all clean-slate design (it scrubs even the captured trusted-config dir and
  uses a fresh empty config dir), and no proxy/CA support exists anywhere in
  `lib/`/`bin/`, so the corporate-proxy scenario is out of scope.
- No escalations; no open user questions.

## Linked task

[...]

PR #615, opened 2026-06-28
fix(diagnose): surface on-disk evidence instead of bare fallback

ivankuznetsov:
## Summary

Tapping **Refresh diagnosis** in the Telegram bot (or running `hive status --diagnose <slug> --json`) for a failed task could reply with the bare, unhelpful `No diagnostic available for <slug>.` even when the task folder still held logs, a `diagnostics/red-status.md`, and a marker. This happened whenever the task's marker had rotated away from a red-recovery state between the failure and the tap, so `TaskAction#diagnostic` resolved to `nil` while the evidence sat untouched on disk.

A new shared resolver, `Hive::DiagnosticEvidence`, now backs both surfaces. When the diagnostic is nil but evidence exists, the bot and CLI surface a one-line summary (marker name plus the last meaningful log line) and a pointer to its source, and never emit the bare string for a task that has any marker or log.

### How evidence is resolved

`DiagnosticEvidence.summarize(folder:, marker_summary:, state_file:)` walks the task folder in priority order and returns the first tier that yields content, or `nil` when the folder holds neither a marker nor a log:

| Priority | Source | Surfaced summary | Pointer |
|---|---|---|---|
| 1 | `diagnostics/red-status.md` | frontmatter `summary:`, else first body line | the red-status file |
| 2 | newest meaningful `*.log` | `marker: last non-blank log line` | that log file |
| 3 | current marker file | marker name plus display attrs | the state file |

The shared resolver keeps the bot and CLI in lockstep without duplicating the glob, tail, frontmatter, and containment logic — those live once in `Hive::DiagnosticHelpers`, shared with `Diagnostic`, instead of reaching into its privates.

### Design notes

- **The `--write` budget gate is untouched.** The resolver is additive and only runs on the read path (or after the gate), so a healthy, non-red task still refuses to spawn the diagnosis agent. The green-task refusal test stays green.
- **Defensive by construction.** Missing folders, unreadable files, binary log tails, FIFOs/devices, symlinked evidence dirs, and deeply-nested YAML all degrade to a lower tier or `nil` rather than raising or blocking, so a Refresh reply can never crash or wedge the bot reaper. Surfaced lines are redacted through `Hive::SecretPatterns.redact` and truncated to 120 characters with an ellipsis.
- **New envelope field.** `hive-status-diagnose` v2 gains `marker_summary` (marker name plus display attrs), present even when `diagnostic` is null, so bots and agents get a cheap evidence prefix. The human CLI output now prints the resolved summary plus pointer in place of `no red-status diagnostic for <slug>`.

## Test plan

- `ruby -Itest test/unit/diagnostic_evidence_test.rb` — every resolution tier, secret redaction, tail-boundary secret suppression, truncation, invalid-UTF-8 tails, FIFO/symlink/large-tail/nested-YAML defensive paths, and the R4/R8 data-driven sweep asserting the net holds (never emits `No diagnostic available`) across `ERROR`, `REVIEW_ERROR`, `REVIEW_STALE`, `REVIEW_CI_STALE`, `EXECUTE_STALE`, stale-agent (orphaned + died), `PLAN_MISSING_OUTPUT`, the `FINALIZE_*` markers, and a green-rotated folder that still has logs.
- `ruby -Itest test/unit/bot/supervisor_test.rb` — the flipped empty-success test plus marker-tier, red-status-tier, and empty-folder cases, each guarded by `refute_includes "No diagnostic available"`.
- `ruby -Itest test/unit/commands/status_diagnose_test.rb` — nil-diagnostic read path emits an evidence-backed `diagnostic` Hash + `marker_summary`; JSONSchemer round-trips (incl. the 120-char boundary) and key-set parity with `Diagnostic#to_h`; the green-task `--write` refusal stays untouched.
- `ruby -Itest test/unit/commands/status_test.rb test/unit/markers_test.rb test/unit/task_action_test.rb test/unit/schema_files_test.rb` — human CLI output, `Markers.summary(:none) → nil` end-to-end, the lifted helpers, and the schema field.
- R7 guard suites (`notification_builders_test`, `callback_handlers_test`, `slash_handlers_test`) confirm no other surface regresses the bare string.
- `rake coverage` (100% line coverage) and `rubocop` on the changed lib files, both clean.

## Review summary

Three review passes (CE code review, pr-review-toolkit, and codex) drove the branch; all findings were auto-fixed or resolved against the plan, brainstorm, and project conventions — no open escalations. Substantive fixes landed during review:

- **Never-block hardening.** Gated every evidence read on `File.file?`/`contained?` so a FIFO, device node, or directory named `*.log`/`*.md` can no longer block `File.open`/`File.read` forever and wedge the no-timeout bot reaper.
- **Never-raise hardening.** Broadened the parse-site and boundary rescues to cover `SystemStackError`/`NoMemoryError` (deeply-nested red-status YAML) and `LoadError`, keeping `summarize`'s never-raise contract intact on the exact defensive path it exists for; the pre-existing sibling gap in `Diagnostic#diagnostic_frontmatter` was fixed alongside.
- **Containment / secret-leak fixes.** Rejected symlinked evidence directories that resolved into a trusted root (host-file leak), dropped the partial first line on a mid-file tail seek so a boundary-split credential can't bypass prefix-anchored redaction, and routed `detail`/path fields through `redact` for symmetry.
- **DRY consolidation (plan R-5).** Lifted the shared caps, `tail_file`/`truncate`/`safe_mtime`/`utf8`, and the symlink-containment helpers into `Hive::DiagnosticHelpers`; extracted `Hive::Markers.summary` so the three marker-summary call sites stop drifting; derived the state-file priority list from `Hive::Task::STATE_FILES`.
- **Test strengthening.** Drove the marker-only path through real `Status`/`TaskAction` instead of echoed literals, added JSONSchemer round-trips and key-parity assertions on the evidence producer, and pinned the symlink/large-tail/boundary-secret regressions.

Out-of-scope codex findings (babysitter, router, TUI, init) were correctly traced to a polluted two-dot diff — the branch trailed `origin/main` and touches none of those files — and resolved no-fix. One nit (a `Struct` value-object refactor) is suppressed as scope-expanding.

## Linked task

- Hive task: `if-something-fail-in-hive-260603-baa7` (branch `if-something-fail-in-hive-260603-baa7`).
- Seed example: the bare `No diagnostic available for telegram-attachments-should-be-supported-260530-8f77.` reply, reproduced deterministically by the Unit 4 marker sweep rather than a single historical envelope.

---


ivankuznetsov:

## High

## Medium
- [ ] `DiagnosticEvidence` follows symlinked evidence outside task/log roots: `lib/hive/diagnostic_evidence.rb:80` accepts `*.log` paths and `lib/hive/diagnostic_evidence.rb:145` opens them without the existing `File.realpath` containment guard, so a symlink under `logs/` or `diagnostics/red-status.md` can leak arbitrary readable host-file content through CLI JSON or the Telegram fallback.

## Nit

ivankuznetsov:

# Code Review — if-something-fail-in-hive (pass 1)

Scope: `git diff origin/main..HEAD` (merge-base `03c3d34`). Reviewed the diagnose
evidence-fallback feature: `lib/hive/diagnostic_evidence.rb` (new),
`lib/hive/bot/supervisor.rb`, `lib/hive/commands/status.rb`, `lib/hive.rb`,
`schemas/hive-status-diagnose.v2.json`, and tests. Reviewed with the
`/ce-code-review` persona set (correctness, maintainability, testing, security,
api-contract, reliability), report-only.

Verification done during review (all green): the new + modified test files pass
(`diagnostic_evidence_test`, `bot/supervisor_test`, `commands/status_diagnose_test`,
`commands/status_test`, `schema_files_test`, plus the R7 guard suites
`notification_builders`/`callback_handlers`/`slash_handlers`); `rubocop` clean on
the three changed lib files; the synthesized `evidence_diagnostic` Hash validates
against the inlined `$defs/Diagnostic` schema; `Markers.current` never returns nil
(so `marker_summary` is nil-safe); the Telegram reply is sent with no `parse_mode`
(plain text, no HTML-injection risk). Requirements R1–R8 all appear implemented.

Note on diff noise: the branch is **2 commits behind `origin/main`** (#613, #614),
so the diff shows phantom "reversions" of babysitter hardening
(`bin/hive-babysitter-stub-*`, `dry_run_env_test.rb`, wiki babysitter pages). The
branch touches **zero** babysitter files; a normal 3-way merge/rebase keeps main's
hardening, so there is no real regression. Excluded from findings; see the Nit on
rebasing.

## High

- [ ] None. No correctness, security, or requirements-coverage defect rises to High; the feature implements R1–R8 and the resolver's nil/empty/exception handling is solid.

## Medium

- [ ] Marker-summary join idiom is triplicated and will drift (`lib/hive/commands/status.rb:417` `marker_summary`, `lib/hive/diagnostic_evidence.rb:115` `marker_summary_from_state_file`, `lib/hive/task_action/diagnostic.rb:300` `marker_summary`): all three reimplement the identical `Hive::Markers.display_attrs(...).map{...}.join(" ")` + upcase-name idiom, so any change to attr rendering must be mirrored in three files or the bot reply, the envelope's `marker_summary` field, and `diagnostic.summary` silently diverge — exactly the duplication the plan's R-5 warned against. Extract one `Hive::Markers.summary(marker)` and have all three call it.
- [ ] `DiagnosticEvidence#inferred_task_log_dir` re-derives the canonical global log-dir layout (`lib/hive/diagnostic_evidence.rb:94`): it hardcodes `<root>/.hive-state/logs/<slug>` via its own regex, duplicating the single source of truth `Hive::Task#log_dir` (`lib/hive/task.rb:117`) that `Hive::TaskAction.log_dirs` already shares with `Diagnostic`. If the global log location ever moves, `Task#log_dir` updates but this resolver keeps probing the old path, silently finds no logs, and degrades to marker-only/nil — reintroducing the "No diagnostic available" gap this change set out to close. Expose a path-based helper (or reuse the Task path captures) so the literal lives once.
- [ ] `evidence_diagnostic` is a second hand-rolled producer of the schema-governed `Diagnostic` shape, untested against the schema (`lib/hive/commands/status.rb:425`): it manually assembles the 9-key payload that only `Diagnostic#to_h` (`lib/hive/task_action/diagnostic.rb:66`) otherwise produces, under `additionalProperties:false` with all 9 keys required. A field added to `Diagnostic#to_h`+schema would leave this path emitting an invalid/missing-key envelope, and the new tests assert individual keys but never run JSONSchemer — so the drift ships silently. Add a JSONSchemer round-trip on the nil-diagnostic (evidence) branch and/or route both producers through one builder. (Flagged by maintainability, testing, and api-contract reviewers.)
- [ ] The R4 regression sweep is tautological for the synthetic markers (`test/unit/diagnostic_evidence_test.rb` `test_no_empty_fallback_net_for_marker_and_synthetic_evidence`): every case unconditionally writes a `*.log` (so the log tier always satisfies and the marker-only fallback is never exercised), and the `PLAN_MISSING_OUTPUT`/`FINALIZE_*`/`ERROR`/`REVIEW_*` labels are passed in as literal `marker_summary:` strings the resolver just echoes — yet in production those states yield a non-nil `Diagnostic#to_h`, so the resolver is never reached for them. The test therefore does not prove the production net for the states it names. Drive the marker-only path (no log) through real `Status`/`TaskAction`, the way `test_red_recovery_markers_have_non_empty_task_action_diagnostics` does for red markers. (Per CLAUDE.md: "does this test actually test anything?")
- [ ] Tail-boundary redaction bypass can leak a secret into Telegram + `--json` (`lib/hive/diagnostic_evidence.rb:145` `tail_file`): the read seeks to EOF-8192 and starts mid-file, so a single log line longer than `TAIL_BYTES` that is also the file's last line begins after its leading identifier (`sk-`, `api_key=`, `AKIA`, `Authorization: Bearer`, PEM header). `SecretPatterns` is prefix-anchored, so `redact` no longer matches and the trailing credential portion is surfaced as the summary. Fix: when the seek lands mid-file, drop the first (partial) line of the tail before line-splitting (and optionally redact the full tail buffer as defense-in-depth).
- [ ] Read-path diagnostic semantics changed and the doc still claims parity (`lib/hive/commands/status.rb:362`; `wiki/commands/status.md:114`): the non-`--write` `--diagnose --json` path now returns a **non-null** `diagnostic` for any task with on-disk evidence — including healthy/`COMPLETE` tasks — and so diverges from the per-row `diagnostic` (still null) in `hive status --json`. The behavior is intended (R5), but line 114 still says `--diagnose` "prints the same local diagnostic payload used by `hive status --json`," which is now false for non-red rows, and the divergence/"`diagnostic == null` is no longer a health signal here" is undocumented. Update the wiki and note the deliberate cross-surface divergence.

## Nit

- [ ] Hardcoded `"Log:"` prefix mislabels non-log evidence (`lib/hive/commands/status.rb:432`, `lib/hive/bot/supervisor.rb:1319`): the highest-priority tier is `diagnostics/red-status.md` (a diagnostic artifact) and the marker tier points at a state file (`task.md`), yet both consumers render `"Log: <source_path>"`. Operators/agents are told a diagnostic artifact or state file is a log. Have `summarize` return a `kind` and let consumers pick `Diagnostic:`/`Log:`/`Marker:`. (correctness + maintainability)
- [ ] Marker-tier `source`/`source_path`/`artifact_paths` can describe a different file than the summary (`lib/hive/diagnostic_evidence.rb:23-29`, `lib/hive/commands/status.rb:441`): `marker_state_file` returns the first `*.md` in `STATE_FILE_NAMES` order carrying any marker, which for a multi-state-file folder need not be `task.state_file`; `evidence_source` then classifies a genuine marker summary as `source:"artifact"` pointing at an unrelated sibling, and `source:"marker"` becomes effectively unreachable. No control-flow harm. Thread the caller's authoritative state file into the resolver.
- [ ] `status.rb#marker_summary` omits the synthetic `PLAN_MISSING_OUTPUT`/`FINALIZE_*` names that `Diagnostic#marker_summary` emits (`lib/hive/commands/status.rb:417`): for a synthetic-state task the envelope's `marker_summary` reads e.g. `NONE` while `diagnostic.summary` reads `PLAN_MISSING_OUTPUT`. Latent only (those states yield a non-nil diagnostic, so the field isn't consumed by the bot one-liner today), but a surprise for any future consumer that treats `marker_summary` as authoritative.
- [ ] `DiagnosticEvidence` re-declares `SUMMARY_MAX`/`TAIL_BYTES`/`LOG_GLOB_CAP`/`FRONTMATTER_SCAN_BYTES` and re-implements `tail_file`/`truncate`/`safe_mtime` from `Diagnostic` (`lib/hive/diagnostic_evidence.rb:8`): `SUMMARY_MAX` in particular is coupled to the schema's `maxLength:120`; if `Diagnostic`'s cap and the schema move, the two diagnose surfaces truncate differently. Lift the input-agnostic helpers/caps into a small shared module.
- [ ] `log_candidates` applies the lexicographic cap before the mtime sort (`lib/hive/diagnostic_evidence.rb:83`): `candidates.sort.last(LOG_GLOB_CAP)` over paths mixed from two dirs (`logs/` < `stages/`) can drop the true newest-by-mtime log when >20 logs exist, surfacing a stale line. This matches the established `Diagnostic#latest_log_artifacts` idiom (intentional ISO-timestamp-name proxy), so it is consistent, but the cross-dir mix makes the drop more likely; the net still holds (an older log or marker is returned).
- [ ] `summarize` has no outer rescue; its never-raise contract is emergent across ~8 leaf rescues (`lib/hive/diagnostic_evidence.rb:16`). The CLI caller re-raises any `StandardError` from `diagnose_call` as `InternalError` (exit 70), so a future unguarded raise would crash `hive status --diagnose` rather than degrade. Add an outer `rescue StandardError; nil`, mirroring the sibling `unanswered_question_count` (`status.rb:319`).
- [ ] Branch/edge coverage gaps the 100% line gate cannot see: `tail_file`'s large-file (>8 KB) seek path is never exercised (only the `EINVAL` rewind runs); the `evidence_source` `"marker"` branch + marker-only `evidence_diagnostic` are never driven end-to-end through `Status` (both new status tests write a log → always `"artifact"`); `empty_diagnose_success_reply`'s "Task folder:" branch (folder present, no evidence) is untested; `*.log`-is-a-directory / EACCES / ELOOP degradation is untested (only the `*.md`-is-a-directory case is). Add targeted cases for the marker-only and large-tail paths.
- [ ] `marker_summary` is added to the released v2 schema `required[]` in place without a `schema_version` bump (`schemas/hive-status-diagnose.v2.json:26`). This follows the documented ADR-028 pre-1.0 carve-out (in-place edits permitted while hive has no external consumers), so it is acceptable by project convention — but the change's stated "external `--json` consumers" goal is in mild tension with ADR-028's "no external consumers" premise. Reconcile if/when external consumers become real (bump to v3, keep the v2 file frozen as v1 already is).
- [ ] Rebase the branch onto current `origin/main` before merge. It is 2 commits behind (#613, #614), so the review diff carries phantom babysitter "reversions" (hardlink guard, proxy/TLS env scrubbing). No real regression under a normal merge, but rebasing removes the noise and re-confirms the net against the latest tree.
- [ ] `marker_summary` is emitted into the JSON envelope unredacted (`lib/hive/commands/status.rb:400`); only the bot re-feeds it through `summarize` (which redacts). It is derived from hive-controlled marker `display_attrs` (status names, pids, reasons), so secret content is unlikely, but the raw field bypasses `SecretPatterns`. Low risk; route it through `redact` for symmetry if desired.

ivankuznetsov:

# PR Review (pr-review-toolkit) — if-something-fail-in-hive (pass 1)

Scope: `git diff origin/main..HEAD` (merge-base `03c3d34`). Diagnose
evidence-fallback feature: `lib/hive/diagnostic_evidence.rb` (new),
`lib/hive/bot/supervisor.rb`, `lib/hive/commands/status.rb`, `lib/hive.rb`,
`schemas/hive-status-diagnose.v2.json`, and tests. Ran the pr-review-toolkit
persona set: `code-review`, `comments`, `tests`, `errors` (silent-failure),
`type-design`, `simplify`. Union of findings below; each is prefixed with the
persona that surfaced it.

Context: the babysitter changes in the diff are a phantom artifact (the branch
is 2–4 commits behind `origin/main`, which added #613/#614 etc.); the branch
touches zero babysitter files, so they were excluded. Verified during review:
`rake coverage` = 100% line (0 failures/errors), rubocop clean, synthesized
`evidence_diagnostic` validates against the schema today. Requirements R1–R8 are
all implemented.

## High

- [ ] None. No persona found a Critical defect; the feature is correct, schema-valid today, and implements R1–R8. The Medium items below are robustness, test-strength, encoding, and doc-accuracy gaps, not merge blockers.

## Medium

- [ ] [tests] The named R4 net `test_no_empty_fallback_net_for_marker_and_synthetic_evidence` (`test/unit/diagnostic_evidence_test.rb`) is vacuous for the synthetic markers: it always plants a `*.log` (so the log tier always fires and the marker-only tier is never reached), passes the synthetic label as a literal `marker_summary:` the resolver just echoes, and `refute_includes reply, "No diagnostic available"` checks a string the test itself built — none of it exercises `diagnose_call` or `empty_diagnose_success_reply`. In production those states yield a non-nil `Diagnostic#to_h`, so the resolver is never reached, and `test_red_recovery_markers_have_non_empty_task_action_diagnostics` covers only ERROR/REVIEW_*/EXECUTE_STALE — PLAN_MISSING_OUTPUT, FINALIZE_*, and stale-agent have no production-path assertion.
- [ ] [tests] The real `evidence_diagnostic` envelope (`status.rb:425`) is asserted field-by-field but never round-tripped through `hive-status-diagnose.v2.json` via JSONSchemer, unlike every other producer in the repo; `additionalProperties:false`, the `source`/`generated_by` enums, and the `summary`/`detail` maxLengths are unguarded against drift in the actual producer (the one schema test validates a hand-built payload, not this output).
- [ ] [errors] `summarize` has no outer rescue — its "must NEVER raise" contract is emergent across ~8 leaf rescues, and the unguarded `File.directory?(root)` (line 18) plus `Dir[...]` (whose `rescue SystemCallError` does not catch the `ArgumentError` raised on a null-byte path) is a currently-reachable escape; if it raises, CLI `Status#call` re-wraps it as `InternalError` exit 70 and the bot's `reap_all.each { reply_for_child }` (no rescue around the loop) drops every other child's reply in the same tick. Inputs are trusted today, but the contract says never. Add one outer `rescue StandardError => e; warn(...); nil`.
- [ ] [errors] Leaf rescues collapse `EACCES`/`EIO`/`ELOOP` into the same silent `nil`/`[]` as benign `ENOENT`, with zero breadcrumb — the exact failure mode the sibling `Diagnostic` (`task_action/diagnostic.rb:545-555`) explicitly refuses, splitting ENOENT (silent) from other `SystemCallError` and warning, with a comment that silently invisibling a real-but-unreadable file is the worst thing a diagnostics tool can do. Split ENOENT from other errors and `warn` on the rest (`diagnostic_evidence.rb:86,111,131,141,160,166`).
- [ ] [errors] Silent wrong-tier / ambiguous-nil degradation: an unreadable high-priority tier (red-status.md or the newest log) is indistinguishable from an empty one, so the resolver silently presents a staler/lower tier as the authoritative diagnosis; and `summarize` returns the same `nil` for "no evidence" and "every read errored," so the bot says "tap Refresh diagnosis again" (a guaranteed loop if the folder is unreadable) and the CLI prints "no diagnostic evidence on disk" — both false when evidence exists but is unreadable. Surface a breadcrumb and/or a `degraded`/`tier` signal so the user-facing copy can tell the truth.
- [ ] [type-design] The `{summary:, source_path:}` return drops the tier the resolver knows, forcing `evidence_source` (`status.rb:441`) to reconstruct kind by string-comparing `source_path` to `task.state_file`; in any advanced folder `marker_state_file` resolves an earlier-stage `.md` (e.g. a stale `COMPLETE` in `task.md`) that differs from the canonical `task.state_file` (e.g. `pr.md`), so a genuine marker source is mislabeled `source:"artifact"` and pushed into `artifact_paths`. Carry an explicit `kind` (`:red_status`/`:log`/`:marker`) instead of reconstructing it.
- [ ] [type-design/code-review/comments] Hardcoded `"Log: #{source_path}"` in both `evidence_diagnostic` detail (`status.rb:431`) and the bot reply (`supervisor.rb:1319`) mislabels the two most common tiers — `diagnostics/red-status.md` (an agent verdict artifact) and the marker state file (`task.md`/`plan.md`) — as a "Log", and the CLI prints this `detail` verbatim. Flagged by three personas. Derive the prefix from the tier (`Diagnostics:`/`Log:`/`Marker:`) or use a neutral `Source:`.
- [ ] [comments] `wiki/commands/status.md:115` lead sentence still claims `--diagnose` (without `--write`) "prints the same local diagnostic payload used by `hive status --json`," which this change makes self-contradictory: the read path now synthesizes a non-null diagnostic for exactly the rows `hive status --json` reports as `diagnostic: null`. Conditionalize the claim to red-recovery rows; the appended fallback sentence never retracts the blanket assertion.
- [ ] [simplify/code-review] The "upcased marker name + `display_attrs` join" idiom is triplicated verbatim across `diagnostic_evidence.rb:115` (`marker_summary_from_state_file`), `status.rb:417` (`marker_summary`), and `task_action/diagnostic.rb:300` (`marker_summary`), all feeding the same `summary`/`marker_summary` fields — a drift trap (plan R-5). Extract `Hive::Markers.summary(marker)` next to `display_attrs` and delegate all three.

## Nit

- [ ] [code-review] On the CLI read path a markerless task with logs renders `"NONE: <log line>"` and emits `marker_summary:"NONE"` because `status.rb#marker_summary` upcases the `:none` marker, whereas the resolver omits the prefix when `marker_summary` is unsupplied — pass `nil` for `:none` for consistency between the two call paths.
- [ ] [type-design] The resolver return is a bare Hash whose invariants (both keys present, `summary` non-empty, `source_path` present) are conventional only — `summary_payload` will happily return `{summary: ""}` if all parts are blank. Consider a `Data.define(:summary, :source_path, :kind)` value object that validates non-emptiness and `kind` at construction (also fixes the two findings above).
- [ ] [type-design] Provenance inconsistency on the marker tier: `marker_signature` is computed from the `task.state_file` marker (`status.rb:435`) while a marker-tier `summary` may be built from a different file's marker (`marker_state_file`), so the `--write` idempotency signature can describe one marker while the summary describes another.
- [ ] [simplify] `SUMMARY_MAX`/`TAIL_BYTES`/`LOG_GLOB_CAP`/`FRONTMATTER_SCAN_BYTES` and `truncate`/`tail_file`/`safe_mtime` are re-declared from `Diagnostic`; `SUMMARY_MAX=120` is the schema-pinned cap now duplicated three ways (two constants + schema `maxLength`), so a future bump truncates the two diagnose surfaces differently. Lift the caps + `truncate` into one shared home and have `Diagnostic` reference them.
- [ ] [simplify] `inferred_task_log_dir` (`diagnostic_evidence.rb:94`) hand-rolls a looser copy of `Hive::Task::PATH_RE` to re-derive `<root>/.hive-state/logs/<slug>` — the exact hardcoded-layout literal the project's stage-literal guard exists to prevent. Keep the folder-string seam (the bot only has `task_folder`) but reuse `Hive::Task::PATH_RE` (referenced lazily) instead of a second regex.
- [ ] [tests] The `>8KB tail_file` seek path is never exercised (largest log fixture is 200 bytes), so `seek(-TAIL_BYTES, SEEK_END)` always raises `EINVAL` and only the rewind branch runs; a regression that mis-sized the tail window would pass green. Add a `>8192`-byte log whose final meaningful line differs from earlier content and assert it is picked.
- [ ] [tests] The bot "Task folder:" no-evidence branch (`supervisor.rb:1322-1325`) is unexercised — the graceful-no-evidence test omits `task_folder`, so the realistic case (folder exists but has no red-status/log/marker) never asserts the appended folder hint. Add a case with a real but evidence-empty `task_folder`.
- [ ] [errors] `current_marker` (`diagnostic_evidence.rb:131`) rescues `ArgumentError` (broad — would swallow a genuine programmer-error `ArgumentError` as "no marker" if `Markers.current`/`parse_attrs` is refactored) and `EncodingError` (unreachable, since `File.read(..., encoding: "UTF-8")` tags rather than transcodes). Drop `EncodingError` and comment why `ArgumentError` is caught (the invalid-byte `scan`).
- [ ] [errors] `evidence_diagnostic` backfills `updated_at` with `Time.now` when `safe_mtime` returns nil (`status.rb:437`); when the evidence file exists but its mtime can't be stat'd (EACCES), stale evidence is stamped as just-generated. `warn` on the non-ENOENT branch of `safe_mtime`, consistent with the sibling convention.
- [ ] [comments] `wiki/commands/status.md:115` understates the log source — `DiagnosticEvidence.log_dirs` also globs the inferred `.hive-state/logs/<slug>/*.log` (the primary run-log dir for coding tasks), but the wiki implies only an in-folder `logs/*.log`, sending a maintainer to the wrong place.
- [ ] [comments] The new `lib/hive/diagnostic_evidence.rb` ships with no rationale comment on its two rot-prone spots: `inferred_task_log_dir`'s on-disk-layout coupling and the `summarize` tier-precedence (red-status → newest log → marker). One line each would mark them intentional.
- [ ] [comments] `evidence_diagnostic` (`status.rb:425`) synthesizes a schema-governed Diagnostic shape with no breadcrumb tying it to the schema or `TaskAction#diagnostic`, an outlier in a file otherwise meticulous about schema coupling — a one-line "keep in sync with the hive-status-diagnose `diagnostic` shape" comment would prevent a silent missed-field drift.
- [ ] [comments] The comment above the empty-success branch (`supervisor.rb:1284-1290`) is still accurate but under-describes the branch — it no longer dead-ends at the bare string; append "…which `empty_diagnose_success_reply` now backfills from on-disk evidence before falling back to a try-again message."
- [ ] [code-review/simplify] `safe_mtime` is implemented in both `status.rb:447` and `diagnostic_evidence.rb` (with slightly different nil-guarding); fold into a single shared helper alongside the constants consolidation above.

ivankuznetsov:

## High
- [ ] `gh` passthrough keeps attacker-controlled proxy/TLS environment: `bin/hive-babysitter-stub-gh:458` no longer deletes `HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY` or `SSL_CERT_*`, so allowlisted authenticated `gh` reads can be redirected through an agent-controlled MITM.
- [ ] Bare Telegram text no longer captures ideas: `lib/hive/bot/router.rb:163` now falls through to `:unknown` outside conversations/drafts, removing the shipped "send any message to capture an idea" workflow outside the diagnose-fallback scope.

## Medium
- [ ] Dry-run skip logs accept hardlinks again: `bin/hive-babysitter-stub-git:21` and `bin/hive-babysitter-stub-gh:17` only check regular-file ownership, so a hardlinked `.babysitter-dry-run-skipped.log` can redirect audit appends into another same-UID file.
- [ ] TUI polling scales with archive size again: `lib/hive/tui/state_source.rb:106` calls full `Status#json_payload` and then fingerprints every row, so `9-done` archives are reparsed/statted on the hot path after the archive cache and scale tests were removed.
- [ ] Diagnose evidence can miss the newest log: `lib/hive/diagnostic_evidence.rb:122` keeps the lexicographically last 20 logs before sorting by mtime, so a newer log with an earlier filename is ignored despite R2 requiring newest-log evidence.

## Nit
- [ ] Closed stdin can crash `hive init` prompts: `lib/hive/commands/init/prompts.rb:187` lets `tty?` `IOError` escape instead of treating the stream as non-interactive and using defaults.

ivankuznetsov:

# Code Review — if-something-fail-in-hive (pass 2)

Scope: this branch's real work only — `git diff 03c3d342...HEAD` (merge-base / three-dot).
The two-dot `origin/main..HEAD` is polluted because the branch trails `origin/main`;
the in-scope change is the 7 diagnose commits (16 files): the new `Hive::DiagnosticEvidence`
+ `Hive::DiagnosticHelpers`, the bot empty-success backfill, the CLI nil-diagnostic enrichment,
the `Hive::Markers.summary` extraction, and the `hive-status-diagnose` schema field.

Intent: a "no-empty-fallback net" so `hive status --diagnose` and the Telegram bot never emit
the bare `No diagnostic available for <slug>.` when on-disk evidence (red-status / log / marker)
exists. Tests pass (179 in-scope) and rubocop is clean; findings below are semantic.

Two High findings were independently reproduced live against the loaded library (not theory).

## High
- [ ] **`summarize` hangs forever on a non-regular file (FIFO / device / dir) named `*.log` or `*.md` in the evidence dirs** (`lib/hive/diagnostic_evidence.rb:106` `latest_log_summary`→`last_meaningful_line:200`→`DiagnosticHelpers.tail_file` `File.open`, and `:178` `current_marker`→`Markers.current` `File.read`): the red tier guards with `File.file?` (`:70`) but the log and marker tiers do not, so a blocking `open(2)` on a FIFO never returns. CONFIRMED: `mkfifo logs/newest.log` then `summarize` timed out (no return in 8s). This is not a raise — the outer `rescue StandardError` (`:60`) cannot catch a block — so it wedges the bot reaper loop *permanently* (every future diagnose/dispatch reply lost, strictly worse than the one-tick drop this PR set out to fix) and makes `hive status --diagnose` hang un-reapably. Directly violates the module's stated never-block contract and plan R-4 ("tolerate a folder with junk"). Fix: gate every evidence read on `File.file?(path)` before open in `latest_log_summary` and the marker tier (also closes the device-node→unbounded-read/`NoMemoryError` variant).
- [ ] **Symlinked `logs/` directory escapes the realpath containment guard and leaks arbitrary host `*.log` contents** (`lib/hive/diagnostic_evidence.rb:241` `evidence_roots` / `:133` `log_dirs` / `:245` `realpath_or_expand`): when `logs/` (or the global `.hive-state/logs/<slug>` leaf) is itself a *directory* symlink, `realpath_or_expand` resolves it to its target and adds that target as a TRUSTED root, so `contained?` (`:231`) then approves files under it. CONFIRMED: with `logs -> /outside`, `summarize` returned the outside file's content surfaced into `:summary`/`:source_path` (`kind: :log`), reaching CLI JSON (`status.rb#evidence_diagnostic`) and Telegram (`supervisor.rb#empty_diagnose_success_reply`). This defeats exactly the protection the code comment at `:225-230` claims (R-6 redaction is prefix-anchored, so non-credential-shaped lines leak verbatim). The existing test only covers a symlinked `*.log` *file* (which the guard does catch). Fix: reject a log dir whose own realpath differs from its lexical `expand_path` (i.e. the dir itself is a link), and mirror the hardening in the pre-existing `lib/hive/task_action/diagnostic.rb#diagnostic_roots`.

## Medium
- [ ] **The most dangerous boundary-split-secret branch is untested** (`lib/hive/diagnostic_helpers.rb:47` `drop_first_line` no-newline outcome): the R6 test `test_secret_split_across_tail_boundary_does_not_leak` writes `sk-…\n`, so the trailing newline lands in the tail window and only the newline-present path runs. The no-newline case (a single log line longer than `TAIL_BYTES` with no newline — the scenario the comment calls out as "never safe to surface") is never exercised; a regression that returned the partial buffer would leak a prefix-anchored secret's tail with no failing test. Add a sibling test that writes `sk-#{'A'*9000}` (no `\n`) and asserts the log tier is skipped and the summary contains neither `sk-` nor a run of `A`.
- [ ] **Symlink-escape refusal is only pinned for the log-*file* tier** (`test/unit/diagnostic_evidence_test.rb`): the actually-exploitable symlinked-*directory* case (see High #2) and the `contained?` refusals on the red-status tier (`diagnostic_evidence.rb:71`) and marker tier (`:154`) have no test, so a refactor that opened those holes would stay green. Add directory-symlink and red-status/marker symlink regression tests asserting host content never reaches `source_path`/`summary`.

## Nit
- [ ] **`marker_summary` promoted to `required` under `additionalProperties:false` without a `schema_version` bump** (`schemas/hive-status-diagnose.v2.json:26`): `lib/hive.rb:12-14` documents that *adding* keys is non-breaking, but making one required is stricter — a strict validator pinned to a pre-change v2 copy rejects new envelopes (additionalProperties), and the new schema's `required` rejects an older binary's envelope. No in-tree consumer schema-validates (`supervisor.rb:1280` checks only `schema ==`) and the envelope is ephemeral, so blast radius is external strict validators. Either relax to optional / `additionalProperties:true`, or mint v3 (as v2 was minted for `id`/`display_name`).
- [ ] **`evidence_diagnostic` `detail` is not capped at `DETAIL_MAX` (4000)** (`lib/hive/commands/status.rb:449`): unlike the canonical `Diagnostic#to_h` (`diagnostic.rb:81`, `truncate(redact(detail), DETAIL_MAX)`), the hand-built `"#{source_label}: #{source_path}"` is unbounded; a deeply nested `source_path` (~4000+ chars, within `PATH_MAX`) emits an envelope that fails its own schema `detail.maxLength`. Round-trip tests use short tmpdir paths so the branch is never stressed. Wrap in `Hive::DiagnosticHelpers.truncate(..., Hive::TaskAction::Diagnostic::DETAIL_MAX)`.
- [ ] **Dead `TAIL_BYTES` aliases** (`lib/hive/diagnostic_evidence.rb:25`, `lib/hive/task_action/diagnostic.rb:27`): both are unreferenced after the tail logic moved into `DiagnosticHelpers` (every tail read goes through `DiagnosticHelpers.tail_file`, which uses the helper's own constant). The block comments claim they prevent drift, but an unused alias guards nothing. Delete both.
- [ ] **`present(state_file)` applies a display-text normalizer to a filesystem path** (`lib/hive/diagnostic_evidence.rb:52` → `one_line:269`, `gsub(/\s+/, " ").strip`): a project root containing a tab or consecutive spaces is silently mangled, breaking the marker read and corrupting the emitted `source_path`. Impact is narrow (whitespace runs only), but the `present` name reads like a blank-check and hides the mutation. Prefer `state_file unless state_file.to_s.strip.empty?` for the path, reserving `present()` for display strings.
- [ ] **`Hive::Markers.summary(:none) → nil` ("null not NONE") is asserted only via hand-built literals** (`lib/hive/markers.rb:173`; `schema_files_test`/`status_test` hard-code `marker_summary: nil` rather than producing it): a regression returning `"NONE"` would pass the whole suite. Relatedly, the `:none can't occur` claim in `diagnostic.rb:303#marker_summary` is an unenforced invariant — the old inline code rendered `"NONE"`, the new delegation degrades to `""` (`redact(nil)`) if ever violated. Add a direct `Markers.summary(:none)`/markerless-`Status#diagnose` assertion.
- [ ] **`summarize`'s never-raise contract leans on a lazy `require "hive/task"`** (`lib/hive/diagnostic_evidence.rb` `inferred_task_log_dir`): a `LoadError` is not a `StandardError`, so it would escape the `:60` rescue and (on the bot reap path) drop sibling replies. Unreachable today because both production entry points eager-require `hive/task` first, but a future caller that invokes `summarize` cold would reintroduce the escape — consider hoisting the require or rescuing `LoadError` at the boundary.
- [ ] **Two registration mechanisms for the same modules** (`lib/hive.rb` autoloads `DiagnosticEvidence`/`DiagnosticHelpers` while `supervisor.rb`/`status.rb`/`diagnostic.rb` also eager-`require` them): harmless and idempotent, but the loading story is split — pick the autoload seam *or* the explicit requires for legibility.

ivankuznetsov:

# PR Review Toolkit — if-something-fail-in-hive (pass 2)

Multi-persona review (code-reviewer, silent-failure-hunter, pr-test-analyzer,
comment-analyzer, type-design-analyzer, code-simplifier) over this branch's real
work — the merge-base / three-dot diff `git diff 03c3d342...HEAD` (the two-dot
`origin/main..HEAD` is polluted because the branch trails `origin/main`). In-scope:
the new `Hive::DiagnosticEvidence` + `Hive::DiagnosticHelpers`, the bot empty-success
backfill, the CLI nil-diagnostic enrichment, `Hive::Markers.summary`, and the
`hive-status-diagnose` schema field. Tests pass and rubocop is clean.

The two High findings were reproduced live by multiple personas (and confirmed in pass 1).

## High
- [ ] [silent-failure-hunter] **A FIFO / non-regular file named `*.log` or `*.md` in the evidence dirs blocks `File.open`/`File.read` forever** (`lib/hive/diagnostic_evidence.rb:231` `contained?` feeding `latest_log_summary:106`→`DiagnosticHelpers.tail_file` and `marker_state_file:152`→`Markers.current` `File.read` at `markers.rb:61`): the log and marker tiers lack the regular-file guard the red tier has (`:70`), and the producer it mirrors (`diagnostic.rb:525 safe_diagnostic_artifact?`) leads with `!File.file?`. A blocking `open(2)` is not an exception, so it escapes every rescue (incl. the top-level `:60` and the bot reaper's `:278`), hangs `hive status --diagnose`, and permanently wedges the bot reaper thread (every other child's reply lost). Reproduced (timed out). Fix: add `File.file?(path)` to `contained?` so both tiers reject FIFOs/dirs/sockets/devices before any open.
- [ ] [code-reviewer] **A symlinked `logs/` *directory* escapes the realpath containment guard and leaks arbitrary host `*.log` contents** (`lib/hive/diagnostic_evidence.rb:241` `evidence_roots` / `:133` `log_dirs`): `realpath_or_expand` resolves a symlinked log dir into a *trusted* root, so `contained?` (`:231`) then approves every file under the symlink target — surfaced into CLI JSON (`status.rb#evidence_diagnostic`) and the Telegram reply. Reproduced (host file content appeared in `:summary`). The existing symlink tests only cover a symlinked *file* under a real `logs/`. The same flaw pre-exists in `diagnostic.rb:551 diagnostic_roots`; fix both by rejecting a symlinked evidence-dir component (`lstat` / realpath-vs-expand_path) rather than resolving it into a root.

## Medium
- [ ] [code-reviewer] **Marker tier reads each `*.md` candidate in full with no byte cap** (`lib/hive/diagnostic_evidence.rb:152` `marker_state_file`→`current_marker:178`→`Markers.current` `File.read`, `markers.rb:61`): asymmetric with the deliberately bounded red (16 KB) and log (8 KB) tiers. On the bot path (no `state_file` passed) it globs and fully reads every `folder/*.md` — e.g. a multi-MB `artifacts.md` — just to scan for a marker, on the no-timeout reaper thread; an oversized file also raises `NoMemoryError`, which escapes the never-raise `rescue StandardError`. Bound the marker read, or restrict candidates to `STATE_FILE_NAMES`.
- [ ] [code-reviewer] **Bot marker-tier `source_path` can point at a different file than the `summary` describes** (`lib/hive/bot/supervisor.rb:1312` `empty_diagnose_success_reply` calls `summarize` without `state_file:`): the marker tier falls to `marker_state_file`, which returns the *first* `*.md` carrying any non-`:none` marker, while `summary` comes from the authoritative envelope `marker_summary` — so e.g. a `COMPLETE` in `task.md` can be labelled `Marker: …/task.md` under a `REVIEW_STALE` summary from `pr.md`. The CLI added `test_threaded_authoritative_state_file_pins_marker_tier_source` to prevent exactly this; thread the state-file path through the envelope so the bot can pin it too, or document the limitation.
- [ ] [silent-failure-hunter] **Lazy `require "hive/task"` can raise `LoadError`, which the never-raise rescue does not catch** (`lib/hive/diagnostic_evidence.rb:145` `inferred_task_log_dir`, reached via `log_dirs`→`evidence_roots`/`log_candidates`): `LoadError < ScriptError`, not `StandardError`, so it escapes `summarize`'s `:60` guard and the bot reaper, violating the stated absolute contract. Shielded today only because both production callers eager-require `hive/task`. Rescue `LoadError` at the boundary or hoist the require to load time.
- [ ] [type-design-analyzer] **The `:kind` closed set {:red_status,:log,:marker} is unenforced and spread across three sites** (`diagnostic_evidence.rb:32` `SOURCE_LABELS`, `:37` `source_label` "Source" default, `status.rb:446` `kind == :marker ? "marker" : "artifact"`): a future 4th kind compiles, silently maps to `source:"artifact"` + label `"Source"`, and still passes the schema — a mislabeled payload with no failure. Centralize the kind→(source,label) mapping in one `.fetch`-ed table so an unmapped kind raises.
- [ ] [pr-test-analyzer] **The boundary-secret suppressor is never exercised** (`lib/hive/diagnostic_helpers.rb:47` `drop_first_line`, the no-newline `: ""` arm): the only boundary test plants `sk-…\n` (trailing newline → the TRUE arm), so a >`TAIL_BYTES` single line with no internal newline — the case that must collapse to `""` — is untested. A regression returning the partial buffer would leak the secret's tail with the suite still green (line coverage stays 100%). Add a no-trailing-newline >8 KB fixture asserting the tail does not appear.
- [ ] [pr-test-analyzer] **Symlink-escape `contained?` refusal is tested only for a `*.log` file**: the exploitable symlinked-*directory* case (High #2) and the red-status (`:71`) and marker (`:155`) tiers' refusals have no test, so a refactor reopening those holes stays green. Add directory-symlink + red-status/marker symlink regression tests asserting host content never reaches `source_path`/`summary`.

## Nit
- [ ] [code-simplifier] Dead `TAIL_BYTES` aliases (`lib/hive/diagnostic_evidence.rb:25`, `lib/hive/task_action/diagnostic.rb:27`): grep-confirmed unreferenced after the tail logic moved into `DiagnosticHelpers`; delete both (the sibling `SUMMARY_MAX`/`LOG_GLOB_CAP`/`FRONTMATTER_SCAN_BYTES` aliases are still used and stay).
- [ ] [comment-analyzer] Overclaiming comments: the `diagnostic_evidence.rb:18` header "summarize NEVER raises … on any error" (only `rescue StandardError` — misses `LoadError`/`NoMemoryError` and a blocking FIFO open); the `:225-230` `contained?` comment implying uniform symlink protection a symlinked *directory* defeats; and the dead-`TAIL_BYTES` alias's "prevent drift" rationale (an unused alias guards nothing). Reword each to match the actual guarantee.
- [ ] [code-reviewer] `evidence_diagnostic` `detail` is not truncated to `DETAIL_MAX` (4000) (`lib/hive/commands/status.rb:449`), unlike `Diagnostic#to_h` (`diagnostic.rb:81`); a pathologically long `source_path` emits an envelope failing its own schema `maxLength` — uncatchable by the short-path round-trip tests. Wrap in `Hive::DiagnosticHelpers.truncate(..., DETAIL_MAX)`.
- [ ] [code-simplifier] `present(state_file)`/`one_line` whitespace-collapses a filesystem path (`lib/hive/diagnostic_evidence.rb:52`) that becomes both the read target and the emitted marker-tier `source_path`; a path containing a tab or double-space is silently mangled, undercutting the `state_file:` pin contract. Use a path-safe blank check (`state_file unless state_file.to_s.strip.empty?`), reserving `present()` for display text.
- [ ] [pr-test-analyzer] `Hive::Markers.summary(:none) → nil` ("null not NONE", `markers.rb:173`) is asserted only via hand-built literals (`schema_files_test.rb:541`), never produced end-to-end; a regression returning `"NONE"` would put `NONE` in the envelope and pass the suite. Drive a markerless task through `Status#marker_summary`/`summarize` and assert null.
- [ ] [type-design-analyzer] The 9-key `Diagnostic` shape is hand-built in three producers (`Diagnostic#to_h`, `synthetic_stale_agent_diagnostic`, `Status#evidence_diagnostic`) plus the schema, synced only by a comment + the round-trip test (which catches missing keys but not a semantic mismap). Extract a shared keyword factory so a missing key is an `ArgumentError` at the call boundary.
- [ ] [type-design-analyzer] Blanket `module_function` (`diagnostic_evidence.rb:34`) makes the entire internal surface public, so an external caller can invoke `summary_payload([...], path, :anything)` and mint a result bypassing the redact/truncate + closed-`kind` invariants; `private_class_method` everything except `summarize` and `source_label`.
- [ ] [silent-failure-hunter] `current_marker`'s `ArgumentError` rescue (`diagnostic_evidence.rb:191`) degrades silently with no `warn`, unlike its adjacent `SystemCallError` branch — a corrupt-encoding state file vanishes from the evidence with zero signal. Add a one-line breadcrumb to keep the asymmetry from costing an operator a debug session.
- [ ] [comment-analyzer] The "`:none` can't occur" claim in `diagnostic.rb:309 marker_summary` is a derived, non-local invariant (enforced upstream by `diagnostic_action?`); if ever violated it degrades to a *silent empty* summary (`redact(nil)→""`), not the old visible `"NONE"` and not a crash — note the failure mode so a future `diagnostic_action?` widening re-checks it.
- [ ] [pr-test-analyzer] Tautological R7 guard at the resolver layer (`diagnostic_evidence_test.rb:327`): it builds a string from the resolver's own hash and checks it lacks a phrase the resolver structurally never emits. The meaningful R7 guard belongs at the bot/CLI render layer (correctly present for the log/red-status cases); this one is dead weight.
- [ ] [pr-test-analyzer] R7 `refute_includes "No diagnostic available"` guard is missing on the bot marker-tier and empty-folder tests — the empty-folder case is the original bug's exact scenario and should carry the explicit regression guard, not just the positive `"Couldn't find a cached diagnosis"` assertion.
- [ ] [pr-test-analyzer] The R4 stale-agent sweep covers only the `:agent_orphaned` ("never attached") variant; the `:agent_died` ("recorded pid not alive", `diagnostic.rb:127`) synthetic-diagnostic branch the plan's stale-agent requirement names is not exercised.
- [ ] [type-design-analyzer] Evidence `detail` is a one-line `"Label: path"` pointer, but the schema advertises `detail` as multi-line operator content (the real `Diagnostic#to_h` fills up to 4000 chars of log tail); a consumer reading `detail` for evidence text gets only a path. Schema-valid, but worth a one-line schema-description note on the semantic thinning.

ivankuznetsov:

# Code review — if-something-fail-in-hive (pass 3)

Scope: feature diff `git diff origin/main...HEAD` (merge-base `03c3d342`). The
branch forked before 8 later `origin/main` commits, so the two-dot
`origin/main..HEAD` diff falsely shows those unrelated commits as deletions —
this review uses the three-dot/merge-base scope: `lib/hive/diagnostic_evidence.rb`,
`lib/hive/diagnostic_helpers.rb`, `lib/hive/markers.rb`, `lib/hive/bot/supervisor.rb`,
`lib/hive/commands/status.rb`, `lib/hive/task_action/diagnostic.rb`,
`schemas/hive-status-diagnose.v2.json` + tests.

Intent: a "no-empty-fallback net" so `hive status --diagnose` and the Telegram bot
never surface an empty/`null` diagnostic when on-disk evidence exists. Tests pass
(38 + 151 affected, 0 failures) and rubocop is clean. Reviewers: correctness,
adversarial, security, reliability, testing, maintainability, api-contract,
project-standards. Mode: report-only (no code edited).

## High
- [ ] Deeply-nested YAML in `diagnostics/red-status.md` raises `SystemStackError`, escaping the never-raise boundary and killing the bot reaper thread (`lib/hive/diagnostic_evidence.rb:108-112`, boundary `:85`): `red_status_frontmatter_summary` rescues only `Psych::Exception` and `summarize` only `StandardError`, but `YAML.safe_load` on ~6 KB of nested flow YAML (well inside the 16 KB scan window) raises `SystemStackError` (verified empirically). It propagates `reply_for_child` → `reap_children` (no per-child rescue, `supervisor.rb:194`) → `reaper_loop` (`rescue StandardError` only, `supervisor.rb:273`), killing the reaper so every later child reply is silently dropped until restart; on the CLI it also bypasses the documented exit-70 envelope. Defeats the central design contract on the exact defensive path it was built for. Fix: also rescue `SystemStackError`/`NoMemoryError` at the parse site and at the `summarize` boundary; add a regression fixture. (reliability conf 100, adversarial conf 75)

## Medium
- [ ] Never-block gap: the caller-supplied authoritative `state_file` marker read bypasses the `contained?`/`File.file?` guard (`lib/hive/diagnostic_evidence.rb:78-79` → `current_marker` → `Hive::Markers.current` `File.read` at `markers.rb:61`). When `marker_summary` is blank, the authoritative path is read with no regular-file gate (`File.size(fifo)==0` passes `marker_file_oversized?`), so a FIFO/device there blocks the no-timeout reaper forever. The module header (`:25-27`) explicitly claims "every tier gates on File.file? (via contained?)" — the comment overclaims. Fix: gate the authoritative `state_file` through `File.file?`/`contained?` before reading, or narrow the comment. (4 reviewers concur)
- [ ] `STATE_FILE_NAMES` has already drifted from the canonical workflow state files (`lib/hive/diagnostic_evidence.rb:36`): `artifacts.md` (plural) never matches the coding artifacts stage's canonical `artifact.md` (singular, `workflows/coding.rb:100`), `notes.md` matches no workflow, and all content-workflow state files (research/outline/draft/critique/article.md) are absent. The drift is masked by the `Dir[*.md]` glob fallback so the marker-tier prioritization silently does nothing for those tasks. Fix: derive the priority list from `Hive::Task::STATE_FILES.values` instead of a hand-maintained literal. (maintainability conf 100, verified)
- [ ] Security-relevant containment guard duplicated rather than consolidated: `realpath_or_expand` + `path_inside?` are byte-identical in `lib/hive/diagnostic_evidence.rb:314-330` and the pre-existing `lib/hive/task_action/diagnostic.rb:559-575`. The PR lifted `tail_file`/`truncate`/`safe_mtime`/`utf8` + caps into `DiagnosticHelpers` (plan R-5) but left the two symlink-escape helpers duplicated, where they can drift and leave one surface exploitable. Fix: lift both into `DiagnosticHelpers` too. (maintainability conf 100)
- [ ] New symlink-leaf rejection regresses the existing red-status producer when `task.folder` itself is a symlink (`lib/hive/task_action/diagnostic.rb:566`, mirrored at `diagnostic_evidence.rb:321`): `return nil if File.symlink?(path)` also fires for `task.folder` (Task uses `File.expand_path`, which does not resolve symlinks), dropping it from the trusted roots so `diagnostics/red-status.md` realpaths outside every surviving root → `safe_diagnostic_artifact?` false → the agent-written verdict is silently discarded (degrades to `marker_detail`). Operator-driven (hive does not create such symlinks) but a behavior regression vs `origin/main`. Fix: restrict the guard to a containment check rather than pre-rejecting any symlinked leaf. (adversarial conf 50)
- [ ] Pre-existing sibling worth fixing alongside the High finding: `Diagnostic#diagnostic_frontmatter` shares the `SystemStackError` gap (`lib/hive/task_action/diagnostic.rb:420`, `rescue Psych::Exception, SystemCallError` only). A corrupt red-status.md on a *red* task crashes the whole `hive status --json` snapshot past the per-project `rescue StandardError` degrade (`status.rb:160`), freezing daemon auto-advance. Pre-existing on main but in the file this diff edits — fix it with finding #1 for consistency.

## Nit
- [ ] `evidence_diagnostic` emits `detail`/`source_path`/`artifact_paths` without `SecretPatterns.redact` (`lib/hive/commands/status.rb:453-463`), asymmetric with the redacted `summary` and with canonical `Diagnostic#to_h`. Not exploitable (paths are server-controlled under `.hive-state/…`), but redacting for symmetry is cheap defense-in-depth.
- [ ] No key-set parity assertion between `evidence_diagnostic` (`status.rb:454-467`) and `Diagnostic#to_h` (`diagnostic.rb:78-88`): required-key drift is caught by the JSONSchemer round-trips, but a future *optional* Diagnostic key would be silently omitted on the evidence path. Add a key-set parity test.
- [ ] No JSONSchemer round-trip exercises an `evidence_diagnostic` summary at exactly the 120-char cap (`truncate` emits 119 chars + `…` = 120 code points). The boundary passes today but is unguarded against a counting-semantics drift; existing round-trips use short summaries.

ivankuznetsov:

# PR review (pr-review-toolkit) — if-something-fail-in-hive (pass 3)

Scope: feature diff `git diff origin/main...HEAD` (merge-base `03c3d342`). The two-dot
`origin/main..HEAD` diff is misleading — the branch forked before 8 later `origin/main`
commits that two-dot renders as deletions — so this review uses the three-dot/merge-base
feature scope (`lib/hive/diagnostic_evidence.rb`, `lib/hive/diagnostic_helpers.rb`,
`lib/hive/markers.rb`, `lib/hive/bot/supervisor.rb`, `lib/hive/commands/status.rb`,
`lib/hive/task_action/diagnostic.rb`, `schemas/hive-status-diagnose.v2.json` + tests).

Personas run (read-only, no edits): code-reviewer, silent-failure-hunter, pr-test-analyzer,
comment-analyzer, type-design-analyzer, code-simplifier. Tests pass; rubocop clean.

Confirmed-good (NOT gaps — surfaced by pr-test-analyzer): the R4/R8 marker-list regression
sweep is complete (ERROR/REVIEW_ERROR/REVIEW_STALE/REVIEW_CI_STALE/EXECUTE_STALE/stale-agent/
PLAN_MISSING_OUTPUT/FINALIZE_*), the flipped/new bot tests do `refute_includes "No diagnostic
available"`, and the boundary-split-secret drop is tested end-to-end through `summarize`.

## High
- [ ] [silent-failure-hunter, pr-test-analyzer] `SystemStackError` from deeply-nested YAML in `diagnostics/red-status.md` escapes both rescues and kills the bot reaper thread (`lib/hive/diagnostic_evidence.rb:108-112`, boundary `:85`): `red_status_frontmatter_summary` rescues only `Psych::Exception` and `summarize` only `StandardError`, but `YAML.safe_load` on ~6-12 KB of nested flow YAML (inside the 16 KB scan window) raises `SystemStackError` (verified by execution) — `SystemStackError < StandardError` is false. It propagates `reply_for_child` → `reap_children` (no per-child rescue, `supervisor.rb:194`) → `reaper_loop` (`rescue StandardError` only, `supervisor.rb:273`), killing the reaper so every later child reply is silently dropped until restart (process still polls, looks healthy); the CLI also bypasses the documented exit-70 envelope. Fix: rescue `SystemStackError`/`NoMemoryError` at the parse site and at the `summarize` boundary; add a nested-frontmatter regression fixture.

## Medium
- [ ] [silent-failure-hunter, code-reviewer, comment-analyzer, pr-test-analyzer] Never-block gap: the caller-supplied authoritative `state_file` marker read bypasses the `contained?`/`File.file?` gate (`lib/hive/diagnostic_evidence.rb:77-79` → `current_marker` → `Hive::Markers.current` `File.read` at `markers.rb:61`). When `marker_summary` is blank (the markerless/`:none` case the resolver exists for), a FIFO/char-device at `state_file` blocks the no-timeout reaper forever (`File.size(fifo)==0` passes `marker_file_oversized?`). The module header (`:25-27`) claims "every tier gates on File.file? (via contained?)" — false on this path. Fix: add `return nil unless File.file?(path)` at the top of `current_marker`, or route the authoritative pin through `contained?`.
- [ ] [comment-analyzer] The comment at `lib/hive/diagnostic_evidence.rb:65-66` states `state_file` "stays nil" for the bot ("the bot only has the folder string"), but the bot reaper passes it (`supervisor.rb:1314-1322`, via the new envelope `state_file` field) — the comment contradicts code shipped in the same PR and misdescribes the exact reaper path the never-block contract protects. Fix: rewrite to "older bot envelopes predate the field and pass nil; current callers pass an authoritative state file."
- [ ] [code-simplifier, code-reviewer, comment-analyzer] `STATE_FILE_NAMES` has drifted from the canonical workflow state files (`lib/hive/diagnostic_evidence.rb:36`): `artifacts.md` (plural) never matches the coding artifacts stage's canonical `artifact.md` (singular, `workflows/coding.rb:100`), `notes.md` matches no workflow, and all content-workflow state files (research/outline/draft/critique/article.md) are absent. Masked by the `Dir[*.md]` glob fallback so the priority hint silently misfires. Fix: derive from `Hive::Task::STATE_FILES.values` (lazily, preserving the `LoadError`-guarded no-load-couple contract).
- [ ] [code-simplifier] Security-relevant containment guard duplicated verbatim, not consolidated: `realpath_or_expand` + `path_inside?` are byte-identical in `lib/hive/diagnostic_evidence.rb:314-330` and the pre-existing `lib/hive/task_action/diagnostic.rb:559-575`. R-5's de-dup goal lifted `tail_file`/`truncate`/`safe_mtime`/`utf8` into `DiagnosticHelpers` but relocated the symlink-escape helpers instead of sharing them — divergent edits silently weaken one surface. Fix: move both into `DiagnosticHelpers` and delegate.
- [ ] [pr-test-analyzer, type-design-analyzer] New symlink-leaf rejection regresses the existing red-status producer when `task.folder` itself is a symlink (`lib/hive/task_action/diagnostic.rb:566`, mirrored `diagnostic_evidence.rb:321`): `return nil if File.symlink?(path)` also fires for `task.folder` (Task uses `File.expand_path`, which doesn't resolve symlinks), dropping it from the trusted roots so `diagnostics/red-status.md` realpaths outside every surviving root and the agent verdict is silently discarded (degrades to `marker_detail`). Asymmetric/untested behavior change vs `origin/main`. Fix: restrict to a containment check rather than pre-rejecting any symlinked leaf; add a symlinked-folder test.
- [ ] [type-design-analyzer, pr-test-analyzer] `evidence_diagnostic` hand-duplicates the canonical 9-key `Diagnostic#to_h` shape coupled only by a "keep in sync" comment (`lib/hive/commands/status.rb:449-468` vs `diagnostic.rb:78-88`). The JSONSchemer round-trip catches required-key / extra-key drift but NOT optional-key drift — a future optional Diagnostic key is silently omitted on the evidence path while the canonical path emits it. Fix: extract one shared keyword-arg builder, or assert `evidence_diagnostic.keys.sort == Diagnostic#to_h.keys.sort`.
- [ ] [type-design-analyzer] `:kind` is validated only at the consumer render helpers (`source_label`/`source_kind` `.fetch`), OUTSIDE `summarize`'s never-raise boundary (`lib/hive/diagnostic_evidence.rb:332`). Not reachable with today's three kinds, but a future 4th tier minting an unmapped kind would raise `KeyError` in exactly the reaper context the contract protects. Fix: `raise ArgumentError unless KIND_RENDERING.key?(kind)` inside `summary_payload` so a bad kind degrades to nil inside the rescue.
- [ ] [silent-failure-hunter] (pre-existing, same family — fix with the High finding) `Diagnostic#diagnostic_frontmatter` shares the `SystemStackError` gap (`lib/hive/task_action/diagnostic.rb:411-420`, `rescue Psych::Exception, SystemCallError` only); a corrupt red-status.md on a red task crashes the whole `hive status --json` snapshot past the per-project `rescue StandardError` degrade (`status.rb:160`), freezing daemon auto-advance.

## Nit
- [ ] [silent-failure-hunter, type-design-analyzer] `evidence_diagnostic` would emit schema-invalid `artifact_paths: [nil]` if a future artifact-kind tier returned a nil `source_path` (`lib/hive/commands/status.rb:462`); not reachable today. Fix: `[source_path].compact`, or assert non-nil `source_path` for artifact kinds in `summary_payload`.
- [ ] [type-design-analyzer] The `summarize` result is a bare symbol-keyed Hash whose invariants (valid kind, non-blank path, non-empty summary) live in scattered producer guards + a detached `KIND_RENDERING` table; a small `Struct` value object (mirroring `Markers::State`) would co-locate rendering and validate `kind` in one constructor (`diagnostic_evidence.rb:332`).
- [ ] [comment-analyzer] The aliasing comment at `lib/hive/diagnostic_evidence.rb:29-30` ("so the schema-pinned cap can't drift from `Hive::TaskAction::Diagnostic`") overstates: only `SUMMARY_MAX` is schema-pinned (`LOG_GLOB_CAP`/`FRONTMATTER_SCAN_BYTES` are not), and the actual single source is `Hive::DiagnosticHelpers`, not `Diagnostic`. Fix the wording.
- [ ] [comment-analyzer] The schema `detail`/`summary` descriptions cite nonexistent constants `TaskAction::DIAGNOSTIC_DETAIL_MAX` / `DIAGNOSTIC_SUMMARY_MAX` (`schemas/hive-status-diagnose.v2.json:184,189`); real names are `Hive::TaskAction::Diagnostic::DETAIL_MAX`/`SUMMARY_MAX`. The PR re-touched the `detail` line. Fix the constant names.
- [ ] [code-simplifier] Redundant `!marker.none?` guard in `marker_state_file` (`lib/hive/diagnostic_evidence.rb:196`): `current_marker` already normalizes `:none` → nil, so the block can be `current_marker(path)`.
- [ ] [code-simplifier] The breadcrumb `warn "hive: diagnose-evidence: … (#{e.class}: #{e.message})"` prefix repeats ~9× (`diagnostic_evidence.rb:89,160,206,229,238,252,269,279,306`); a tiny private helper would centralize the prefix while keeping each per-site message verbatim.
- [ ] [code-simplifier] `marker_summary` local variable shadows the same-named method (`lib/hive/commands/status.rb:332`, method at `:430`); rename the local (e.g. `marker_summary_text`) for clarity.
- [ ] [pr-test-analyzer] No JSONSchemer round-trip exercises an `evidence_diagnostic` summary at exactly the 120-char cap (`truncate` emits 119 chars + `…` = 120 code points); valid today but unguarded against an off-by-one. Add one round-trip that forces a 120-char redacted+truncated summary.
- [ ] [silent-failure-hunter] Operator-facing copy: on a hard fault (e.g. EACCES storm) the reaper logs a breadcrumb to stderr/daemon.log but the operator sees only the generic "Couldn't find a cached diagnosis … Tap Refresh diagnosis again" (`supervisor.rb:1333`), which reads as transient; consider "if this persists, check daemon.log."

PR #624, opened 2026-06-29
feat(daemon): auto-retry recoverable terminal error markers

ivankuznetsov:
## Summary

When a task parks on a terminal `ERROR` marker caused by a *transient dependency outage* — a missing Codex auth token (`implementer_failed` with a `401 Missing bearer/basic auth` signature) or a Claude launcher/readiness failure (`claude_launch_failed`) — the daemon previously left it red forever, even after the underlying dependency was fixed and verified healthy. Recovery required a human running `hive markers clear`. We hit this on 2026-06-29: task 58 (Codex 401) and task 287 (Claude launcher) both stayed parked after their dependency recovered.

The daemon can now clear those two markers itself, but only behind a health probe that confirms the dependency actually works in the daemon's environment. The mechanism is deliberately equivalent to a manual `hive markers clear` + re-enqueue: it re-runs the same stage from the start, with no new resume path and no change to the daemon's forward-advance trust boundary. A global kill-switch (`daemon.auto_retry.enabled: false`) disables the behavior entirely.

## How it decides

A new `RecoverableErrorHealer` runs each tick, immediately after `StaleAgentHealer`, over parked recoverable rows. The two never contend — their marker-reason sets are disjoint. For each row:

```mermaid
flowchart TD
    A[Parked ERROR marker] --> B{auto_retry.enabled?}
    B -- no --> Z[skip — kill switch]
    B -- yes --> C{task in flight?}
    C -- yes --> Z2[skip]
    C -- no --> D{classify reason + attrs}
    D -- no signature --> N1[not retried: unknown reason]
    D -- codex_auth / claude_launcher --> E{work area safe?}
    E -- dirty/ambiguous --> N2[not retried: unsafe]
    E -- safe --> F{attempts >= 2?}
    F -- yes --> N3[park permanently: exhausted]
    F -- no --> G{health signal changed?}
    G -- no --> N4[skip — fallback only]
    G -- yes --> H{backoff elapsed?}
    H -- no --> N5[skip — backoff]
    H -- yes --> I[run health probes]
    I -- fail/timeout --> N6[not retried: probe failed]
    I -- all pass --> J[clear marker + re-run stage]
    J --> P[audit to events.jsonl + daemon log]
```

Gating layers, each fail-closed:

- **Allowlist + diagnostic signature** — only `implementer_failed` with `provider=codex` (or legacy absent) AND a `401`/`missing bearer`/`basic auth` message, or `claude_launch_failed`. Everything else (business logic, test failures, review findings, merge conflicts, unknown exit codes) stays parked.
- **Health probes** — Codex: `codex login status` + a tiny `codex exec` smoke test in the daemon's environment. Claude: wrapper-file presence + tmux readiness + binary/version ≥ minimum + `hive doctor` green (in-process, no fork, bounded by an outer timeout). Short timeouts; timeout or error means not healthy. Cached per tick/reason.
- **Changed-signal gate** — a per-reason SHA-256 fingerprint of the relevant environment (binary paths, versions, `CODEX_HOME`, auth-file mtime, wrapper mtime, doctor-row digest). A retry is allowed only when the signal differs from the last failed attempt, with a low-frequency periodic fallback so a missed signal still eventually recovers.
- **Safety guard** — never re-run a row whose worktree carries uncommitted user edits or whose brainstorm/plan holds answered user content. Unenumerated stages, uncertainty, or any inspection error ⇒ unsafe.
- **Budget + backoff** — max 2 auto-retries per task per reason, first immediate, second after 30 minutes; on exhaustion the task parks permanently for manual clearing.

Every decision — positive and throttled negatives — is audited to the daemon log (slug, stage, marker id, reason, probe results, fingerprint, attempt count, action, rationale). Retry decisions also emit to the task `events.jsonl`; off-allowlist `unknown_reason` skips are daemon-log-only to keep unrelated task timelines clean.

## Key decisions

- **New sibling component, not an extension of `StaleAgentHealer`.** That healer's contract is synchronous and I/O-free; this path needs blocking probes, fingerprint comparison, and backoff state. Shared low-level helpers (marker clear + pre-clear mtime seeding + `3-plan` requeue) are factored into a `HealerSupport` mixin so both stay consistent, and `StaleAgentHealer` behavior — including PR #626's Claude stop-hook fix-retry — is preserved.
- **`provider` recorded on the `implementer_failed` marker** so "Codex auth failure" is confirmed by both message signature *and* executor identity; classification still tolerates legacy markers without it.
- **In-memory budgets/fingerprints**, rebuilt on SIGHUP reload — mirrors `StaleAgentHealer`'s deliberate per-process budget. A restart resets counts; accepted v1 trade-off, bounded by backoff and the changed-signal gate.
- **Daemon-global health gate for v1.** The two allowlisted reasons are global CLI dependency outages, so the doctor/fingerprint gate runs with daemon-global config (`project_root: nil`). Per-project scoping is deferred to the follow-up that expands the allowlist.

Per-reason config tuning, allowlist expansion, and cross-restart budget persistence are deferred to follow-up work.

## Test plan

Unit coverage per component (classifier, probes, signal/gate, safety guard, healer orchestration, audit enums, dispatcher wiring/reload) plus an end-to-end `test/integration/daemon_auto_retry_test.rb` driving `Dispatcher#tick` against fixtures with stubbed `codex`/`claude`/doctor probes. The seven brainstorm acceptance scenarios are covered: task-58 Codex auto-clear, task-287 Claude auto-clear only after wrapper + tmux + version + doctor all pass, unknown `implementer_failed` stays parked, 2-retry + 30-min backoff with permanent park, dirty/ambiguous not retried, probe hang ⇒ not retried, kill-switch off ⇒ no behavior.

Focused daemon suite: `471 runs, 1815 assertions, 0 failures, 0 errors`. The full suite has pre-existing failures in live-agent smoke paths (Telegram scaffold expectations, live Claude/tmux trust-prompt launches) unrelated to this change.

## Review summary

Two automated review passes ran against `origin/main..HEAD` (ce-code-review, pr-review-toolkit, plus codex/claude reviewers). No High-severity defects were found in the feature logic. Notable fixes applied across the passes:

- **Anti-stranding parity** — restored the `state_file_mtime.nil?` pre-clear guard and the `marker_heal_observer_missing` diagnostic so a cleared markerless row always seeds its dispatch baseline (mirrors `StaleAgentHealer`).
- **Fail-closed hardening** — the doctor probe now treats a non-success exit code or nil rows as not-healthy; `safe_to_retry?` defaults unenumerated stages to unsafe; in-process doctor calls are wrapped in an outer timeout.
- **Scope of the post-clear rescue** — audit/requeue after a successful clear is no longer mislabeled `auto_retry_failed`.
- **Shared-helper extraction** — `requeue_plan_rerun`, marker-attr/match/key helpers, and `observe_pre_clear_mtime` were consolidated into `HealerSupport` instead of being copy-pasted, and a shared recoverable-category constant removes the triplicated `case`.
- **Merge-base hazard resolved** — the branch was reconciled with PR #626's Claude stop-hook fix-retry so the refactor re-integrates that recovery path rather than dropping it.

Two escalations were resolved by operator decision: (1) the v1 health gate stays daemon-global, with per-project scoping deferred; (2) `unknown_reason` skips audit to the daemon log only, keeping unrelated task timelines clean. Both are documented in the wiki and the code.

## Linked task

- Hive task: `make-the-hive-daemon-automatically-260629-223d`
- Branch: `make-the-hive-daemon-automatically-260629-223d`
- Origin: `brainstorm.md` (user-curated, `<!-- COMPLETE -->`) → plan (U1–U9)

---

Claude Code


ivankuznetsov:

## High
- [ ] Project health checks use daemon-global config instead of the task project config: `RecoverableErrorHealer` builds `HealthProbe` with `project_root: nil` and passes the daemon config from `Commands::Daemon` (`lib/hive/commands/daemon.rb:156`, `lib/hive/daemon/recoverable_error_healer.rb:28`, `lib/hive/daemon/recoverable_error_healer.rb:86`), so `hive doctor`/doctor-row fingerprints can ignore a parked task's actual `.hive-state/config.yml`, reviewer inventory, `.llm-wiki`, and agent overrides, allowing auto-clear despite a red project-specific health precondition.
- [ ] Claude launcher recovery does not run the required readiness detector: `HealthProbe#claude_probes` only checks wrapper file existence, tmux availability, and `claude --version` (`lib/hive/daemon/health_probe.rb:117`), so a `claude_launch_failed` marker can be cleared even when the wrapper-launched Claude session still cannot reach `ClaudeLauncher`'s ready prompt path.

## Medium
- [ ] Doctor probe treats doctor command errors with no rows as healthy: `doctor_probe` records `exit_code = doctor.call` but sets `ok` solely from missing/version rows (`lib/hive/daemon/health_probe.rb:79`), so a config error or other doctor failure returning nonzero with empty/nil rows passes the universal health precondition instead of blocking auto-retry.

## Nit

ivankuznetsov:

# Code Review — Daemon auto-retry of recoverable terminal error markers

Scope: `git diff origin/main..HEAD` (9 commits, U1–U9). Reviewed against the plan's Requirements Trace (R1–R12). All twelve requirements appear implemented with both unit and integration coverage; the new `RecoverableErrorHealer` closely mirrors the proven `StaleAgentHealer`. No High-severity defects found. Findings below are correctness-divergence and per-tick cost items.

## High

(none)

## Medium

- [ ] Missing `state_file_mtime.nil?` guard before `clear_current` in `recoverable_error_healer.rb` (`heal_row`, ~line 122): `StaleAgentHealer#heal_error_if_auto_recoverable` returns *before clearing* when `row.state_file_mtime.nil?` precisely because a cleared markerless terminal-error row takes the edit-resume path and "without a pre-clear mtime to seed as the dispatch baseline, that row can strand as first-sight `record_baseline`." The new healer omits that guard — it clears, then calls `observe_pre_clear_mtime`, which silently no-ops when `state_file_mtime` is nil (line 234) — so a nil-mtime row can be cleared without seeding the baseline and strand un-dispatched. U8's own approach calls out preserving "the pre-clear-mtime seeding that prevents first-sight stranding"; mirror the sibling's pre-clear guard.
- [ ] Claude `health_signal` fingerprint is expensive and recomputed every tick per parked row, partly defeating the changed-signal gate: `HealthSignal.claude_parts` calls `profile_version(:claude)` (→ `check_version!`, which shells out `claude --version`, `agent_profile.rb:239`) **and** `doctor_rows_digest` (a full in-process `Hive::Commands::Doctor#call`, which itself re-runs the claude version check and may shell out to qmd under a 15s timeout). Because the fingerprint is the *input* to `changed_or_fallback?`, this work runs on every daemon tick for each parked `claude_launch_failed` row — i.e. before any probe caching or gating — which is the expensive work the gate exists to avoid. Codex fingerprints stay cheap (stat + config). Consider caching the doctor/version inputs per tick or basing the claude fingerprint on cheap signals (binary path + mtime + config), reserving doctor/`--version` shell-outs for the actual probe.

## Nit

- [ ] `observe_pre_clear_mtime` (recoverable healer, ~line 233) silently returns when `@controller` lacks `observe_state_file_mtime`, whereas `StaleAgentHealer#observe_pre_clear_mtime` emits a `marker_heal_observer_missing` event for exactly that case so the baseline-stranding regression has a log signal. Mirror the event for parity/observability.
- [ ] `AutoRetrySafety::FEEDBACK_PATTERN` matches the bare word `feedback` anywhere in `plan.md`, so any plan whose body merely mentions "feedback" (e.g. a task to "gather user feedback") will be classified unsafe and blocked from plan-stage auto-retry. It fails closed (the safe direction per R8), so this is not a defect — but the heuristic is broad enough to quietly defeat `claude_launcher` recovery on 3-plan; consider anchoring it to an actual answered-feedback marker shape.
- [ ] Codex smoke probe runs `codex exec --json "Reply with OK."` via `@capture3.call(@env.to_h, *cmd)` with no `--cd`/`--add-dir`, i.e. in the daemon process CWD. The plan explicitly defers the exact smoke-command shape (Open Questions), so this is not an escalation — just confirm during that finalization that the invocation is side-effect-free and not sensitive to the daemon's working directory.

ivankuznetsov:

# PR Review — pr-review-toolkit (multi-persona)

Ran the pr-review-toolkit personas against `git diff origin/main..HEAD`: code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer, code-simplifier. Findings below are the deduplicated union, each prefixed with the surfacing persona. Severities are calibrated by the aggregator (e.g. the untested-rescue gap is held at Medium because the dispatcher's outer `begin/rescue` at `dispatcher.rb:278` already guards the tick from a crash).

## High

- [ ] [silent-failure-hunter] `doctor_probe` fails open on the universal precondition (`health_probe.rb`, doctor_probe ~L79-85): it sets `ok: failing.empty?` and never consults the captured `exit_code`. `Hive::Commands::Doctor` uses `attr_reader :rows` with `@rows = nil` until its single assignment; any rescued `ConfigError`/`KeyError`/`ArgumentError` returns `EXIT_CONFIG_ERROR` with `@rows` still nil → `Array(doctor.rows)` is `[]` → `failing.empty?` is true → the R2 doctor gate reports healthy. A broken/misconfigured doctor silently passes the gate guarding auto-clear (partially mitigated only because the reason-specific codex/claude probes still run afterward). Treat a non-success `exit_code` (or nil rows) as not-healthy.

## Medium

- [ ] [code-reviewer] Missing `return if row.state_file_mtime.nil?` guard before `clear_current` (`recoverable_error_healer.rb` heal_row, clear block): the sibling `StaleAgentHealer` (stale_agent_healer.rb:184) bails on nil mtime because clearing a markerless terminal-error row without seeding a pre-clear baseline can strand it as first-sight `record_baseline`; here `observe_pre_clear_mtime` only no-ops on nil, so the marker is still cleared with no baseline.
- [ ] [silent-failure-hunter] Broad `rescue StandardError` wraps the post-clear success path (`recoverable_error_healer.rb` ~L137): if `audit_retry`/`emit_task_event` raises *after* `clear_current` already succeeded and `@attempts[key]` was incremented, the healer logs `auto_retry_failed` — a successful auto-clear mislabeled as a failure while the retry budget is silently consumed; the sibling deliberately scopes its post-clear requeue in its own rescue to avoid exactly this.
- [ ] [silent-failure-hunter] `safe_to_retry?` `else` branch returns `[true, ...]` for any stage outside `4-execute`/`2-brainstorm`/`3-plan` (`auto_retry_safety.rb:25`): `:claude_launcher` can be classified on a marker in any stage, so those bypass stage-specific work-area verification — a fail-open default in a module whose own header calls itself "fail-closed."
- [ ] [code-reviewer] `unknown_reason` skip emits an `auto_retry_skipped` task-timeline event (via `Hive::Events.emit`, surfaced in status tail) for every terminal ERROR row NOT on the recoverable allowlist (`recoverable_error_healer.rb` heal_row `unless category`): the vast majority of genuine domain failures now get a "not retried" event on their timeline; the sibling healer emits nothing for non-matching rows. Consider suppressing the task-channel event for non-allowlisted reasons.
- [ ] [type-design] The recoverable-category enum `{:codex_auth, :claude_launcher}` is implicit and duplicated as a `case` in three modules (classifier.rb:14, health_probe.rb:54, health_signal.rb:26) with no shared constant — adding/renaming a category needs three synchronized edits, and a missed branch degrades silently to a benign-looking `else` rather than failing loudly.
- [ ] [type-design] Terminology collision across the contract boundary (`health_signal.rb` fingerprint / `health_probe.rb` probe): both name their first arg `reason:`, but the healer passes `category` (the classifier symbol) there, while `reason` everywhere else is the marker-reason string ("implementer_failed") — the healer even logs them as distinct `reason:`/`category:` fields, making it easy to pass the wrong value.
- [ ] [comment-analyzer] `wiki/modules/daemon.md:30` overstates task-event coverage: it lists all four `auto_retry*` events as going to both channels, but `Hive::Events::EVENT_TYPES` (events.rb:14-17) only permits `auto_retry`/`auto_retry_skipped`; the exhausted path emits the task event `:auto_retry_skipped` (not `:auto_retry_exhausted`) and the failed path emits no task event at all — `auto_retry_exhausted`/`auto_retry_failed` reach the daemon log only.
- [ ] [code-simplifier] `requeue_plan_rerun` (`recoverable_error_healer.rb` ~L244-264) is a verbatim copy of `stale_agent_healer.rb:259-285`, differing only in the `trigger:` string — the plan (U6) explicitly called for extracting this into a shared helper; it was copy-pasted instead.
- [ ] [code-simplifier] `marker_match_attrs`/`marker_reason`/`marker_attrs`/`recovery_key` (`recoverable_error_healer.rb` ~L266-284) re-implement `stale_agent_healer.rb` (381-394 / 513-515 / 753-758 / 418-425) with identical behavior — move this marker-attr/match/key cluster into a shared module/base, the extraction the plan intended.
- [ ] [code-simplifier, code-reviewer] `observe_pre_clear_mtime` (`recoverable_error_healer.rb` ~L233-242) duplicates `stale_agent_healer.rb:396-416` AND drops its diagnostic: the sibling logs `marker_heal_observer_missing` when the controller lacks `observe_state_file_mtime`, while the new copy silently no-ops — consolidate into one helper so the seeded-baseline guard can't silently regress.
- [ ] [pr-test-analyzer] The integration test (`daemon_auto_retry_test.rb`) stubs all three real collaborators (FakeProbe, FakeSignal, and `AutoRetrySafety.safe_to_retry?`), so only dispatcher→healer→markers→events is wired live: keyword-seam drift between the healer's calls (`fingerprint(reason:,config:,project_root:,env:,now:)`, `probe(category)`) and the real implementations would pass unnoticed. Add at least one test driving the real probe/signal/safety through the healer.
- [ ] [pr-test-analyzer] The `@controller.running_task?` guard (skip clearing while the task is actively running) is untested — `FakeController` defaults `running: false` and no test sets it true; this data-race/double-dispatch gate could regress silently.
- [ ] [pr-test-analyzer] The `row.live_task_lock == true` early-return guard is untested — every row builder hardcodes `live_task_lock: nil`; no test proves a locked task is left parked.
- [ ] [pr-test-analyzer] The `legacy_layout_projects` skip is untested — every test passes `legacy_layout_projects: {}`; a half-migrated project being incorrectly healed is uncovered.
- [ ] [pr-test-analyzer] The `rescue StandardError → :auto_retry_failed` branch is never exercised (no test makes `heal_row` raise); the logger test only asserts the event name is accepted, not that this safety branch fires.

## Nit

- [ ] [code-reviewer, code-simplifier, type-design] `classify(reason:, attrs:, stage: nil, workflow: nil)` (`recoverable_error_classifier.rb:11`) never reads `stage:`/`workflow:`; the healer threads both through (heal_row:62-65) for nothing — drop the dead kwargs or use them.
- [ ] [type-design] `probe_result` `.compact`s nil values (`health_probe.rb:165`), so only `:name`/`:ok` are guaranteed present; `:exit`/`:ms`/`:stdout_tail`/`:stderr_tail` are absent-vs-nil ambiguous — any non-audit consumer iterating probes must treat missing keys as nil (undocumented key-drift trap).
- [ ] [type-design] The `[bool, reason]` tuple invariant in `auto_retry_safety.rb:14` is positional/unenforced, and `module_function` publishes the internal `execute_safe?`/`brainstorm_safe?`/`plan_safe?` helpers; a future helper returning a bare boolean would silently destructure `safety_reason = nil` at the call site.
- [ ] [silent-failure-hunter] `audit_skip` sets `@negative_audits[key] = true` before `emit_task_event`/`logger.event` (`recoverable_error_healer.rb` ~L168); if emission raises, the dedup flag stays set, permanently suppressing that skip rationale for the key for the life of the process, with no log signal.
- [ ] [silent-failure-hunter] `profile_version` and `doctor_rows_digest` (`health_signal.rb` ~L62) fold raw exception messages (`"#{e.class}: #{e.message}"`) into the health fingerprint — a varying error string (transient PATH/exec error) changes the fingerprint and can make `changed_or_fallback?` re-arm retries on noise rather than a genuine dependency-state change.
- [ ] [pr-test-analyzer] The `requeue_plan_rerun` rescue → `:heal_requeue_failed` (with operator remediation string) is untested; only the happy-path 3-plan requeue is asserted.
- [ ] [pr-test-analyzer] The health-probe unknown-reason branch (`unknown_reason` ok:false) and the `run_probe` top-level rescue → `health_probe_error` (`health_probe.rb:64`) are both uncovered.
- [ ] [pr-test-analyzer] `health_signal.rb` fingerprint for an unknown reason (`else → "unknown_reason"`) is never asserted; all signal tests use `:codex_auth`/`:claude_launcher`.

ivankuznetsov:

# Code Review — make-the-hive-daemon-automatically (pass 2)

Scope: `git diff origin/main..HEAD` — daemon health-probe-gated auto-retry of
recoverable terminal ERROR markers (`Hive::Daemon::RecoverableErrorHealer` +
classifier, health probe, health signal, safety guard, audit, dispatcher
wiring, config kill-switch). 31 files, ~2.3k insertions.

Overall: the implementation closely tracks the plan (U1–U9), is fail-closed at
every gate, reuses the established marker-clear / pre-clear-mtime / 3-plan
requeue primitives via the new shared `HealerSupport` mixin, and is unusually
well unit-tested (classifier, probe, signal, safety, healer, config, logger,
events, dispatcher ordering/reload all covered). `StaleAgentHealer` behavior is
preserved (extracted helpers are logically equivalent; no dangling references
to the removed `error_auto_recovery_key` / `auto_recoverable_error_match_attrs`).
The two healers act on disjoint marker reasons, so no double-handling. No
High-severity defects found.

## High
- [ ] (none)

## Medium
- [ ] Codex smoke probe runs a full agent turn unconstrained: `HealthProbe#codex_probes` shells out `codex exec --json "Reply with OK."` (lib/hive/daemon/health_probe.rb:131) in the daemon's own environment/cwd with no `--add-dir`, sandbox, or approval-mode flag — the plan's Open Question on the smoke command shape was resolved implicitly. It is gated (login-status first, changed-signal, backoff, 30s timeout, per-tick cache) so the blast radius is bounded, but confirm `codex exec` defaults are non-destructive in the daemon's working directory and that a real model turn per retry-eligible tick is an acceptable cost.
- [ ] In-process `doctor` call has no outer timeout: `HealthProbe#doctor_probe` (lib/hive/daemon/health_probe.rb:81) and `HealthSignal#doctor_rows_digest` (lib/hive/daemon/health_signal.rb:93) both invoke `Hive::Commands::Doctor#call` on the daemon hot path with no surrounding `Timeout`, relying on each doctor sub-check to carry its own (qmd and version checks currently do). A future doctor check added without an internal timeout would stall the daemon tick — unlike the shell probes, which are uniformly `Timeout`-wrapped.

## Nit
- [ ] Integration suite omits two scenarios plan U9 lists as integration coverage: AE4 (2-retry limit + 30-min backoff + permanent park across dispatcher ticks) and AE2's Claude sub-probe distinction (clears only after wrapper + tmux + version + doctor all pass). Both behaviors are exercised at the unit level (recoverable_error_healer_test.rb `test_exhausts_after_two_clears` / `test_second_retry_waits_for_backoff`; health_probe_test.rb claude cases), so this is a test-placement gap, not a behavioral gap.
- [ ] Healer in-memory state hashes never evict: `@attempts`, `@last_fingerprint`, `@last_attempt_at`, `@last_clear_at`, `@exhausted_logged`, `@negative_audits` (lib/hive/daemon/recoverable_error_healer.rb:39-45) accumulate one-or-more entries per `[project,slug,stage,reason,...]` for the life of the process (negative_audits also keyed by fingerprint+rationale). Slow unbounded growth on a long-lived daemon; reset on SIGHUP reload and consistent with `StaleAgentHealer`'s deliberate per-process budgets (KTD-4). Low impact.
- [ ] Every parked non-recoverable error row emits one `auto_retry_skipped` daemon-log line (action `unknown_reason`) per process, re-emitted after each daemon restart (lib/hive/daemon/recoverable_error_healer.rb:85). Throttled and intended per R10 (task-channel is correctly suppressed via `emit_task: false`), but adds steady-state daemon-log volume proportional to the parked-error count.

ivankuznetsov:

## High
- [ ] Project-scoped health gates are bypassed: `RecoverableErrorHealer` constructs `HealthProbe`/`HealthSignal` with daemon-global config and `project_root: nil`, so project `.hive-state/config.yml` agent overrides, doctor rows, and plugin/skill inventory are not checked before clearing a parked marker required by R2/R4/R6.
- [ ] Unknown recoverable-candidate skips are not audited to `events.jsonl`: `audit_skip(... emit_task: false)` suppresses the task-channel event for non-allowlisted diagnostics even though R10 requires dual-channel audit for throttled negative decisions, including unknown reasons.

## Medium

## Nit

ivankuznetsov:

# PR Review — pr-review-toolkit pass 02

Scope reviewed: `git diff origin/main..HEAD` (RecoverableErrorHealer feature: U1–U9). All new unit + integration tests pass individually; ordering, Doctor API, wrapper path, `tmux_status` shape, and codex-profile alignment were verified against the codebase.

## High
- [ ] [release-eng] Branch base is stale vs `origin/main`; the `stale_agent_healer.rb` refactor will collide with PR #626 on merge: `origin/main` (`7ad943e8`) added the Claude stop-hook fix-retry feature (`FIX_CLAUDE_STOP_HOOK_MESSAGE`, `fix_claude_stop_hook_failure?`, the `fix_failed → fix_claude_stop_hook` label, the retryable-reason branch) *after* this branch forked from `c96df941`. The branch never had it, so the diff shows that production recovery path as "removed", and this PR heavily rewrites the exact methods #626 touched (extracting `HealerSupport`, deleting `marker_attrs_for`/`marker_reason`/`observe_pre_clear_mtime`/`error_auto_recovery_key`/`auto_recoverable_error_match_attrs`). A 3-way merge will conflict on those methods, and a resolution favoring the branch would silently drop #626's stop-hook recovery: rebase onto current `origin/main` and re-run the StaleAgentHealer suite (incl. #626's tests) before merge.

## Medium
- [ ] [correctness] `RecoverableErrorHealer` always constructs `HealthProbe` and computes fingerprints with `project_root: nil`, so the universal `hive doctor` precondition runs without any project context on a multi-project daemon. Verified `Doctor` tolerates nil (skips `.llm-wiki`/project config rows), so this is functional, but it means project-scoped doctor checks never gate auto-retry — confirm that is the intended "global health" semantics rather than a missed wiring of the row's project root.

## Nit
- [ ] [performance] For `:claude_launcher`, a full in-process `hive doctor` runs twice per tick (once inside `HealthSignal.doctor_rows_digest` for the fingerprint, once inside `HealthProbe#doctor_probe`), plus `claude --version` shells out twice (fingerprint `profile_version` + `claude_version_probe`). Both are per-tick memoized but not shared between fingerprint and probe; acceptable but redundant work on the hot daemon loop.
- [ ] [correctness] `tick_fingerprint` passes `env: ENV` to `HealthSignal.fingerprint` while `HealthProbe` reads its injected `@env`. Both default to `ENV` today so they agree, but a future custom-env injection into the probe would make the fingerprint and the probe observe different environments — pass the probe's env through to keep them coupled.
- [ ] [maintainability] In-memory state hashes (`@negative_audits`, `@attempts`, `@last_fingerprint`, `@last_attempt_at`, `@last_clear_at`, `@exhausted_logged`) have no eviction. `@negative_audits` keys include fingerprint+rationale, so under repeated dependency churn on a long-lived daemon the map grows unbounded (the per-reason budget maps are bounded by task count, but this one is not). SIGHUP reload resets it; a slow leak otherwise.
- [ ] [test-coverage] The codex smoke probe hardcodes `["exec", "--json", "Reply with OK."]`, omitting `--add-dir`/model/sandbox flags the real execute launch uses (per `agent_profiles/codex.rb`). It correctly exercises auth + connectivity (matching the headless `exec --json` shape), but a launch failure mode beyond auth would not be caught — fine for an auth probe, worth a comment noting the deliberate narrowing (plan Open Question on smoke-shape).

ivankuznetsov:

Outcome: timeout

Agent tail:

```
The hosted tests themselves passed, but CI failed the 100% coverage gate. The uncovered lines are in the auto-retry files, so I’m switching into the debug loop for a targeted coverage fix.
```

PR #625, opened 2026-06-30
fix(review): accept clean tmux fix completions despite missed Stop hook

ivankuznetsov:
## Summary

Under tmux mode, a review-fix pass could finish cleanly — produce artifacts and even auto-commit — yet still be stamped with a terminal `REVIEW_ERROR phase=fix reason=fix_failed`. The cause was Claude's interactive Stop hook missing its `.done` / `result.json` sentinels, so `wait_for_done_signal` drained to a `:timeout` and the stage treated that as a hard failure. This stranded completed work on tasks 58/PR #622, 287/PR #623, and 288/PR #624, and forced `claude.mode: headless` as a workaround.

Hive now accepts a missed Stop-hook signal as success **only when it can prove the pass actually completed** from on-disk evidence, and records an auditable `claude_completion_fallback` event when it does. Genuine failures still terminate in `REVIEW_ERROR`.

### What changed

| Layer | Behavior |
|---|---|
| `wait_for_done_signal` (shared tmux path) | Attaches advisory `completion_evidence` (process exited, pane idle, session liveness, expected sentinel paths) when the sentinel is absent — without changing the returned `:timeout` status, so the 4-execute path is unaffected. Resolves promptly when the turn ends rather than waiting the full deadline. |
| `Hive::ClaudeCompletionFallback.suppress?` | New pure, phase-agnostic predicate. Requires a clean exit, no limit wall, readable tmux, **and** all phase facts true. Any unknown/ambiguous signal defaults to no-suppression. |
| Review-fix phase 4 | Captures a pass-start HEAD baseline, supplies phase facts (artifacts present, commit-since-baseline **or** explicit whole-pass `RESOLVED/NO-FIX:` no-change evidence, no unresolved escalation, no missing-output marker), and on suppression writes `fix-success-NN.md` + emits the WARN event instead of `mark_review_phase_failure`. |
| `claude_completion_fallback` event type | Added to the events allowlist; the emitted message carries phase, pass, pid/session, expected sentinel paths, missing-signal reason, and the artifact/commit checks. Emission failure blocks suppression so every success stays auditable. |

### Suppression decision

```mermaid
flowchart TD
    A[review-fix returns :timeout + completion_evidence] --> B{clean exit, tmux readable, session alive, no limit wall?}
    B -- no --> X[REVIEW_ERROR reason=fix_failed]
    B -- yes --> C{artifacts present AND commit-or-no-change AND no unresolved escalation AND no missing-output?}
    C -- no --> X
    C -- yes --> D[write fix-success-NN.md, emit claude_completion_fallback, proceed to SUCCESS]
```

The fix is deliberately built in the shared tmux path so other phases can opt in later, but only review-fix — the observed production failure — is wired up now.

### Root cause & recovery

The leading hypothesis (documented in `docs/notes/claude-tmux-launch-mode.md`) is that the `Stop` hook's firing semantics differ between headless `claude -p` and the interactive REPL. The fallback ships as the safety net regardless of a deterministic repro. `stop_hook_installer` tests now pin the exact sentinel/result path contract so the launcher and installer can't drift.

`docs/recipes.md` and `docs/faq.md` document recovery for the three stuck tasks — clear `REVIEW_ERROR` with `--match-attr phase=fix,reason=fix_failed` **only after** verifying artifacts and commit/no-change evidence — and confirm `claude.mode: headless` remains the recommended workaround. Local operator config is not auto-reverted.

## Test plan

- `test/unit/claude_launcher_test.rb` — evidence shape per scenario; execute-path status unchanged; cold-start idle drains to deadline with a live recorded PID
- `test/unit/claude_completion_fallback_test.rb` — predicate truth table; no single condition alone suppresses; `worktree_readable`/`missing_output_absent`/`session_alive` each block independently
- `test/unit/events_test.rb` — new event round-trips; allowlist still enforced; long message truncated
- `test/unit/stop_hook_installer_test.rb` — sentinel/result path contract (`.done` + `result.json` under task folder; `HIVE_TASK_STAGE_DIR`)
- `test/integration/run_review_test.rb` — clean-exit-with-artifacts succeeds; crash / missing-artifact / required-commit-missing / unresolved-escalation / genuine-hang still stamp `REVIEW_ERROR` (with the stop-hook message reaching the marker); limit wall routes to `limits_reached`

All targeted suites pass. The full `rake test` run had three pre-existing, unrelated failures in init binary-path environment tests (`test_current_binary_path_*`).

## Review summary

Two `ce-code-review`, two `codex-ce-code-review`, and two `pr-review-toolkit` passes ran over the diff. No High-severity defect survived to pass 2.

Resolved findings, by theme:

- **Cold-start pane-idle race** (multiple reviewers, High): the turn-ended detection could fire on poll 1 before any work began. Fixed with a work-started latch driven by `pane_idle`, dropping the unreliable persistent-REPL PID arm; the regression test now writes a live PID and asserts the cold-start idle drains to the deadline.
- **No-change over-acceptance** (High): a single `RESOLVED/NO-FIX:` line no longer satisfies the no-change branch — suppression now requires whole-pass no-change (no unapplied AUTO-FIX finding) and treats a dirty worktree as a real code change.
- **Gone/crashed tmux** (High): `session_alive: false` and crash evidence (`process_exited` on a persistent REPL) now block suppression; unknown session liveness is treated conservatively as non-suppressible.
- **Auditability** (Medium): event-emission failure blocks suppression and falls through to the recoverable `REVIEW_ERROR`; the payload carries the concrete artifact/commit evidence rather than bare booleans.
- **Cleanups** (Nit): no-change/`worktree_status` scans computed once and threaded; advisory-only fields (`exit_code`, `tmux_readable`, `missing_output_absent`, `error_message`) documented as forward-compat; wiki `log.d` fragments converted to the repo's heading form.

Three Medium findings about the shared early-return touching the 4-execute path were resolved no-fix: the early break-and-return is plan-intended (KTD2/U1), the returned status symbol is unchanged (R-A), and the cold-start guard covers the execute path too. They are recorded in `suppressed.md`. All escalation passes closed with no open user questions.

## Linked task

- Slug: `fix-review-stage-claude-stop-260629-26ed`
- Pipeline: hive `8-finalize`
- PR: https://github.com/ivankuznetsov/hive/pull/625

---

Claude Code


ivankuznetsov:

# Code Review — fix-review-stage-claude-stop (tmux completion fallback)

Scope: `git diff origin/main..HEAD` (6 commits, U1-U6). Reviewed the shared
`wait_for_done_signal` evidence path, the `ClaudeCompletionFallback` predicate,
the review-fix wiring, events allow-list, and tests. Plan is authoritative on
scope; findings below are real defects against that intent, not scope policing.

## High
- [ ] First-poll pane-idle race in `wait_for_done_signal` (`lib/hive/claude_launcher.rb:852-858`): `completion_evidence` runs on the very first loop iteration, immediately after `send_prompt`→`send_keys("Enter")` (`:289-290`) with no guard that the agent turn has actually started and no initial sleep. The post-submit pane briefly shows the idle `❯` prompt before Claude renders its working state, so `claude_ready_prompt?` can return true → `pane_idle:true` → `completion_evidence_turn_ended?` returns the `:timeout`+evidence shape on poll 1, before any fix work happens. For review-fix this evaluates the predicate against an empty worktree (no commit/no-change) → premature terminal `REVIEW_ERROR`; the failure mode the change exists to prevent. The integration tests stub `spawn_fix_agent`, so this loop is never exercised against a real pane.

## Medium
- [ ] Early-return changes the shared 4-execute path, contradicting the plan's scope boundary/KTD1/R-A (`lib/hive/claude_launcher.rb:852-858`): execute also launches with `status_mode: :exit_code_only` (`lib/hive/stages/execute.rb:277`) and now hits the new turn-ended branch. It can return `:timeout` early on pane-idle/PID-gone instead of draining to the deadline, so a `.done` written slightly after the pane goes idle becomes a new false `implementer_failed` — a behavior change to a path the plan states is "deliberately left behavior-unchanged."
- [ ] Generic-layer crash detection is effectively a no-op (`lib/hive/claude_launcher.rb:148` + `lib/hive/claude_completion_fallback.rb:41-48`): `completion_evidence` hardcodes `exit_code: nil`, and `normal_completion?` accepts `process_exited == true`. A crashed Claude (PID gone, non-zero exit) reads as clean completion (`clean_exit_code?` treats nil as 0). R3/R4's "exited 0, not crash" guard relies entirely on phase facts; the `test_process_crash_blocks_suppression` unit test passes a synthetic `exit_code: 137` the launcher never produces, giving false confidence.
- [ ] `no_change_evidence_present?` over-accepts via `.any?` (`lib/hive/stages/review.rb:383-390`): a single `- [x] RESOLVED/NO-FIX:` line in any `reviews/*-NN.md` (reviewer- or triage-authored, not the fix output) satisfies `commit_or_no_change`. When some accepted findings were no-fix but others required a real code change that never landed (and no commit exists), the predicate still passes → suppressed `REVIEW_ERROR` for a pass that did not actually complete its fixes.
- [ ] Turn-ended detection may never fire promptly in a long shared session (`lib/hive/claude_launcher.rb:157-163`): `claude_ready_prompt?` requires the `"Claude Code"` banner or `"for agents"` footer in the captured tail (`:617-618`). After triage + N reviewers the banner scrolls out of `SENTINEL_CAPTURE_BYTES`, so `pane_idle` can stay false and—absent a Claude crash—the fallback only resolves at the full `timeout_sec` (~45 min) deadline, defeating KTD2's "resolve promptly" goal.

## Nit
- [ ] `missing_output_error?` is vacuous on this path (`lib/hive/stages/review.rb:392-394`): the only `error_message` reaching the fallback is "claude stop hook did not signal completion"; missing-output errors originate from `:output_file_exists` mode, so `missing_output_absent` is always true here.
- [ ] `tmux_readable` field is always true in production (`lib/hive/claude_launcher.rb:140`): `capture_limit_tail` rescues `TmuxError` to `""` (never nil), so `!pane_tail.nil?` never reports false; dead-tmux is only caught via `session_error`. Harmless but the field is dead weight.
- [ ] Dead/redundant evidence computation (`lib/hive/claude_completion_fallback.rb:45-48`): `clean_exit_code?`'s `code == "0"` string branch is unreachable since `exit_code` is never set; and `completion_evidence` recomputes a `.lock` YAML read + `Process.kill(0)` + ready-prompt regex on every poll for up to the full timeout (minor overhead).

ivankuznetsov:

## High
- [ ] Fallback accepts mixed passes with any `RESOLVED/NO-FIX` line as no-change proof: `no_change_evidence_present?` scans every pass file for a single checked no-fix line (`lib/hive/stages/review.rb:950`), so a pass with one real `[x]` fix finding plus an unrelated no-fix disposition can suppress `claude stop hook did not signal completion` without a commit, violating the plan's commit-or-whole-pass-no-change gate.
- [ ] Fallback can suppress when the tmux session is gone: `ClaudeCompletionFallback.suppress?` blocks `session_error` and `tmux_readable: false` but never rejects `session_alive: false` (`lib/hive/claude_completion_fallback.rb:20`), even though the plan requires gone/unreadable tmux to remain a terminal `REVIEW_ERROR`.

## Medium
- [ ] `claude_completion_fallback` audit payload is missing required evidence detail: the emitted message only records booleans for artifacts and commit/no-change evidence (`lib/hive/stages/review.rb:976`), not the artifacts checked or the specific commit/no-change evidence required by R5, making fallback success hard to audit after the fact.

## Nit

ivankuznetsov:

## High
- [ ] [concurrency] Cold-start pane-idle false positive in `wait_for_done_signal` (`lib/hive/claude_launcher.rb:852-859`): the loop calls `completion_evidence_turn_ended?` on the very first poll, immediately after `send_prompt` returns (it sends Enter and returns with no confirmation the turn started — `lib/hive/tmux_runner.rb:99`). `claude_ready_prompt?`'s banner gate (`CLAUDE_READY_BANNER_MARKER = "Claude Code"`) is essentially always present, so detection hinges on the idle `❯` caret; between Enter and Claude's first streamed token the input box can momentarily still match → "turn_ended_without_stop_hook" is returned before the agent does any work, and `with_shared_session`'s `ensure` then `/quit`s + kills the session (`:248-260`). There is no "work started" latch, minimum-elapsed, or consecutive-idle guard. The launcher test (`test_wait_for_done_signal_returns_timeout_evidence_when_pane_is_idle`) enshrines this immediate-return as correct.
- [ ] [correctness] `no_change_evidence_present?` (`lib/hive/stages/review.rb:950-955`) treats the pass as "no code change needed" if *any* `- [x] RESOLVED/NO-FIX:` line exists in *any* `reviews/*-NN.md` file, using `.any?`. A mixed pass (some AUTO-FIX findings + some RESOLVED/NO-FIX findings — the common case, since triage annotates dispositions in the reviewer files) thus satisfies `commit_or_no_change` even when the AUTO-FIX findings were never applied and no commit landed. The new integration test `test_fix_agent_stop_hook_timeout_with_no_change_evidence_uses_completion_fallback` demonstrates exactly this: a `- [x] apply a fix` finding is accepted into SUCCESS with a do-nothing fix agent and no commit. This is the R-B/R-F masking risk the predicate is meant to prevent; it should require that no AUTO-FIX work remained (all findings resolved/no-fix), not merely one no-fix line.

## Medium
- [ ] [correctness] Premature `write_fix_success(ctx)` inside `handle_fix_completion_fallback` (`lib/hive/stages/review.rb:910`) runs before the post-fix path's FixGuardrail (`:637-658`). On suppression, control falls through to the normal path; if the guardrail then trips it returns `review_waiting` *without* writing the success sentinel by design (`:651-657`) — but the fallback has already written `reviews/fix-success-NN.md`. A markerless re-entry could read that stale sentinel as a completed pass and bypass the pending human guardrail approval the comment at `:644-650` explicitly defends. The write is also redundant with the legitimate `write_fix_success` at `:665`; it should be removed.
- [ ] [correctness] The `commit_or_no_change` head-diff branch is largely inert (`lib/hive/stages/review.rb:940-948`): `after_fix_head` is captured at `:580` *before* the orchestrator's post-fix `auto_commit_fix_worktree` (`:611`), and fix agents normally leave changes uncommitted for the orchestrator to commit. So for the standard dirty-worktree case `before_fix_head == after_fix_head` and the commit branch never fires — suppression then rests entirely on the loose `no_change_evidence_present?` check. The plan (KTD4/HLD) anticipated also consulting `worktree_status` ("the auto-commit landed"), which is not done here.
- [ ] [architecture] The new unconditional early-return changes the shared `wait_for_done_signal` behavior for *all* `:exit_code_only` callers, not just review-fix (confirmed routing at `wait_for_status` `:exit_code_only → wait_for_done_signal`). The 4-execute implementer path now terminates early on pane-idle/PID-exit (and is exposed to the cold-start race above) where it previously drained to the deadline. U1/KTD1/R-A's claim that execute "observes identical status outcomes" holds only for the returned `:timeout` symbol, not for *when/whether the agent is killed* — a real, untested behavior change for the un-wired path.

## Nit
- [ ] [maintainability] `completion_evidence` hardcodes `exit_code: nil` (`lib/hive/claude_launcher.rb:864`), so `ClaudeCompletionFallback.clean_exit_code?` always passes — the "process exited 0" condition from R3 is never actually validated for tmux. Either capture a real code or document the field as advisory-only to avoid a false sense of an enforced check.
- [ ] [performance] `worktree_status(worktree_path)` is invoked twice per fallback pass — once in `fix_completion_fallback_phase_facts` (`:921`) and again at `:608` — duplicating a git subprocess; the first result could be threaded through.
- [ ] [maintainability] `missing_output_error?` (`lib/hive/stages/review.rb:957`) / the `missing_output_absent` fact is effectively always true on the exit_code_only fix path, since `error_message` is fixed to "claude stop hook did not signal completion" and never matches the regex — the condition is inert here despite being listed as a guard in R3.

ivankuznetsov:

# Code Review — fix-review-stage-claude-stop (pass 2)

Scope: `git diff origin/main..HEAD` (7 commits, 18 files). Reviewed the tmux Stop-hook
completion fallback: launcher evidence capture (`claude_launcher.rb`), the generic
predicate (`claude_completion_fallback.rb`), the review-fix wiring (`stages/review.rb`),
the new event type, tests, and docs/wiki.

Overall the core logic is sound and conservatively guarded: the strict conjunction in
`ClaudeCompletionFallback.suppress?`, the cold-start `work_started` latch, the
`session_alive == false` rejection, and the "code-change evidence" requirement all push
ambiguity toward terminal `REVIEW_ERROR` (R4). No High-severity defects found. Findings
below are a test-coverage gap and minor cleanups.

Requirements R1–R9 all appear implemented; the deferred items in the plan's Scope
Boundaries (wiring the execute path / other wrappers) are correctly left out of scope.

## High

(none)

## Medium

- [ ] Truth-table coverage gap for two phase facts (`test/unit/claude_completion_fallback_test.rb`): `REQUIRED_PHASE_FACTS` gates suppression on all five phase facts, but the unit tests independently exercise only `artifacts_present`, `commit_or_no_change`, and `no_unresolved_escalation`. There is no test toggling `worktree_readable: false` or `missing_output_absent: false` (the existing `test_tmux_unreadable_blocks_suppression` exercises the *generic* `tmux_readable` evidence field, not the `worktree_readable` phase fact). This contradicts U3's stated verification ("truth-table tests cover each suppression-blocking condition independently; no condition alone can produce `suppress: true`") and leaves R3 conditions unguarded against regression. Add two cases asserting each blocks suppression and appears in `missing`.

## Nit

- [ ] Redundant filesystem scan on the no-change path (`lib/hive/stages/review.rb:973` and `:1066`): `no_change_evidence_present?(ctx)` re-reads and regex-scans every `reviews/*-NN.md` reviewer file twice when a fallback is suppressed via whole-pass-no-change — once in `fix_completion_commit_or_no_change?` (phase facts) and again in `fix_completion_commit_evidence` (audit message). The hardening fragment notes `worktree_status` was threaded through "no duplicate git subprocess"; the no-change scan deserves the same. Compute once and pass the boolean/basis through.

- [ ] Wiki log.d fragment format deviates from repo convention (`wiki/log.d/20260630T084346Z-claude-tmux-completion-fallback.md`, `wiki/log.d/20260630T085012Z-review-fix-stop-hook-recovery.md`): both use YAML frontmatter (`---\ntimestamp:\nslug:\n---`), but every existing fragment and the third new one (`20260630T120000Z-...`) use a `## [<ISO8601>] <scope> — <title>` heading with no frontmatter. The frontmatter style risks being mis-parsed by `hive wiki compile-log` / `Hive::WikiLog`. Convert the two to the `## [timestamp] scope — title` form.

- [ ] PID liveness signals are advisory but not validated against the session (`lib/hive/claude_launcher.rb:967` `recorded_claude_pid`, `:978` `process_alive?`): `recorded_claude_pid` trusts any positive integer in `.lock["claude_pid"]` without confirming it is this run's process, and `process_alive?` returns `true` on `Errno::EPERM`. A stale/reused PID could mis-report `process_exited`. Impact is contained today (the cold-start latch suppresses sealing when a stale PID is already dead and the pane is idle; a wrongly-"alive" PID only makes the predicate *more* conservative), so this is robustness hardening, not a live false-accept — but a brief comment or a session-ownership check would make the safety explicit.

- [ ] Integration coverage asymmetry for R4 hard-failure branches (`test/integration/run_review_test.rb`): the genuine-failure path is exercised end-to-end only for "no commit / no no-change evidence"; the other R4 rejections (missing `escalations-NN.md`, unresolved escalation > 0, gone tmux `session_alive: false`) are covered at the unit level but not through the full review-stage branch. Acceptable layering, but one more integration case (e.g. unresolved-escalation still lands terminal `REVIEW_ERROR`) would lock the wiring. Also, the existing failure-path test asserts marker name/reason but not that the `claude stop hook did not signal completion` message reached the marker (R9) — add that assertion for symmetry with the success-path tests.

ivankuznetsov:

## High
- [ ] PID-death-only evidence can suppress crashed tmux runs: `lib/hive/claude_completion_fallback.rb:42` treats `process_exited: true` as normal completion and `lib/hive/claude_completion_fallback.rb:46` treats `exit_code: nil` as clean, so a killed/crashed Claude process that left artifacts or a dirty worktree can bypass terminal `REVIEW_ERROR`, violating R3/R4.

## Medium
- [ ] Unknown tmux session liveness is accepted as suppressible: `lib/hive/claude_completion_fallback.rb:27` only rejects `session_alive == false` and the test at `test/unit/claude_completion_fallback_test.rb:62` locks in `nil` as success, contrary to the plan’s conservative unknown/ambiguous-evidence rule and the docs requiring a live tmux session.

## Nit

ivankuznetsov:

# pr-review-toolkit review — pass 02

Diff reviewed: `git diff origin/main..HEAD` (7 commits, tmux stop-hook completion fallback).
All shipped unit/integration tests pass locally. Findings below are the union of the
multi-persona pass; each is grounded in the diff and, where noted, reproduced empirically.

## High
- [ ] [correctness] Cold-start work-started latch is defeated in production tmux REPL mode (`lib/hive/claude_launcher.rb:826-882`): the recorded `claude_pid` is the persistent tmux REPL pane process (`record_claude_pid` → `runner.pane_pid`), which stays alive *before, during and after* every turn, so `process_exited == false` on the very first poll sets `work_started ||= ...` true immediately — exactly the pre-work prompt the guard exists to reject. With a live recorded pid and a pre-first-token idle `❯` pane on poll 1, `wait_for_done_signal` returns `turn_ended_without_stop_hook` *before Claude does any work*. For a real-change pass (reviewer files carry unapplied `[x] AUTO-FIX`) the predicate then fails `commit_or_no_change` and stamps a terminal `REVIEW_ERROR`, converting a run whose Stop hook would have signaled correctly into a spurious failure — a new failure mode strictly worse than today's behavior. Empirically reproduced: with `.lock` holding a live pid and an idle pane, the launcher returns `reason="turn_ended_without_stop_hook"` on poll 1 instead of draining to `deadline_without_stop_hook`. The guard's premise ("a live recorded process" ⇒ "turn underway") is false for a persistent REPL; only the `pane_idle == false` arm provides real protection.

## Medium
- [ ] [test-design] The cold-start regression test gives false confidence (`test/unit/claude_launcher_test.rb:1184-1201`, `test_wait_for_done_signal_ignores_cold_start_idle_before_work_starts`): it omits the `.lock` file, so `recorded_claude_pid` returns nil and `process_exited` is nil — the latch never trips and the test passes for the wrong reason. It does not exercise the production path (live REPL pid recorded by `record_claude_pid`), so it masks the High finding above. A faithful test would write a live `claude_pid` into `.lock` (as `test_wait_for_done_signal_returns_timeout_evidence_when_recorded_pid_exited` does with a dead pid) and assert the cold-start idle still drains to the deadline.
- [ ] [reliability] Suppression can reach SUCCESS with no audit trail (`lib/hive/stages/review.rb:1007-1024`): `emit_fix_completion_fallback` rescues `SystemCallError, IOError, JSON::JSONError` to nil, and `handle_fix_completion_fallback` returns `true` regardless of whether the event was emitted. If emission fails, the pass is still laundered to success with no `claude_completion_fallback` record — violating R5's "every suppression auditable" intent. Consider letting emit failure block suppression (return false / fall through to `REVIEW_ERROR`) or at minimum surfacing a warning.

## Nit
- [ ] [observability] `claude_completion_fallback` event is emitted before the post-fix guardrail runs (`lib/hive/stages/review.rb:919-928` then `:646-662`): when the FixGuardrail trips, the pass returns `:review_waiting` (pending human approval), not SUCCESS — yet a `claude_completion_fallback` WARN event was already recorded, implying the pass was accepted as complete. The sentinel is correctly withheld in that case, but the audit event's timing slightly overstates the outcome.
- [ ] [performance] `no_change_evidence_present?` re-reads every reviewer file up to ~3× per fallback (`lib/hive/stages/review.rb:956,980,1064`): once via `fix_completion_commit_or_no_change?` (phase facts) and again inside `fix_completion_commit_evidence` for the audit message, each call re-globbing `reviews/*-NN.md` and re-reading lines. Compute once and thread the result, mirroring the `post_fix_status` "compute once" change already made in this diff.
- [ ] [clarity] `limit_wall?` checks `evidence[:error_message]` (`lib/hive/claude_completion_fallback.rb:55-58`), but the launcher's `completion_evidence` hash never sets an `error_message` key — only `reason`/`session_error`. The message branch is dead on the current tmux path; harmless, but worth a comment noting it is forward-compat for callers that supply a richer evidence hash, consistent with how `clean_exit_code?`/`missing_output_error?` are documented.

ivankuznetsov:

Outcome: timeout

Agent tail:

```
Coverage is in the web/supervisor tail now. I’m waiting for the final summary and exit code.
```

PR #617, opened 2026-06-28
fix(review): correctly classify usage-limit failures in 6-review

ivankuznetsov:
## Summary

The 6-review stage misclassified Claude usage/credit-limit deaths in both directions. A real limit hit during triage or CI-fix was parked as a terminal `triage_failed` / `fix_failed` / `ci_unrunnable` marker with no `retry_after`, so the cooldown healer never retried and the task sat red forever (this stranded 5 telegram-bot PRs). In the other direction, an agent that merely *quoted* limit strings in its output, e.g. a triage agent reviewing limit-handling code, was killed mid-run because the scanner matched anywhere inside an 8 KB byte tail of scrollback (repro: task #7456). This change makes both paths classify correctly: real walls self-heal, narrated limit strings no longer kill a healthy run.

## Root cause

`AgentLimit.error_message` emits a fixed wire format (`limits reached for <agent>: <line>`). The headless agent built that message from the single matched line, so re-scanning it round-tripped. The tmux launcher instead passed the **whole pane tail** (banner, diff chrome, menu) to `error_message`, so the formatted string lost the real limit sentence and the sinks' re-scan with `limit_reached?` failed → false negative. The same loose `limit_reached?` ran over an 8 KB **byte** window of the live pane, so any quoted limit line matched → false positive.

Threading the launcher's old loose boolean straight into the sinks would have *promoted* that false positive into a confident held `limits_reached` marker. So tightening the live detector is a prerequisite for safely threading a structured signal, not an independent cleanup.

## The fix

```mermaid
flowchart TB
  pane["live tmux pane tail"] -->|live_limit_line| ll["AgentLimit.live_limit_line"]
  ll -->|matched line| wait["launcher wait result<br/>{ limit_text }"]
  wait --> res["Triage/CiFix Result.limit_text"]
  res --> sink["review sinks"]
  sink -->|limit_text OR from_limit?| mark["review_error<br/>reason=limits_reached + retry_after"]
  mark --> heal["StaleAgentHealer cooldown self-heal"]
```

[...]

PR #627, opened 2026-06-30
feat(review): classify triage/fix phase failure reasons

ivankuznetsov:
## Summary

When a 6-review **triage** or **fix** agent failed for a non-limit reason, the runner stamped one opaque marker — `reason=triage_failed` or `reason=fix_failed` — flattening the real cause (a merge conflict, a network timeout, a tool-permission denial, an agent crash) into a phase-shaped placeholder. That `reason=` is the field the daemon healer, `hive status --json`, the web diagnostic card, and the TUI status column all key off, so operators saw "the phase errored" with nothing specific to act on.

This change classifies non-limit triage/fix failures into a closed enum, so the operator-facing `reason=` names an actual failure class instead of a placeholder.

### What changed

A new `Hive::ReviewErrorReason` classifier (`lib/hive/review_error_reason.rb`) maps a captured agent `error_message` to a closed enum, structurally mirroring `Hive::AgentLimit`: ordered, first-match-wins line regexes biased toward a safe fallback. It reuses `AgentLimit.normalize` for ANSI/control-char scrubbing and emits only enum members, so the closed-enum guarantee is enforced at the classifier (and by a load-time guard) rather than by mutating the marker layer.

| `reason=` | Source | Recovery |
|---|---|---|
| `limits_reached` | `AgentLimit` (unchanged first gate) | auto-heals after cooldown |
| `merge_conflict`, `network_timeout`, `tool_permission_denied`, `agent_crashed` | new classifier | terminal (`hive markers clear`) |
| `unknown` | classifier fallback | terminal |
| `rate_limited` | reserved enum member, not emitted by v1 | n/a |

`mark_review_phase_failure` keeps the `AgentLimit.limit_reached?` check as the **first, unchanged gate** — quota / credit / session / 429 / "rate limit reached" text still maps to `limits_reached` and keeps the daemon's `retry_after` auto-heal. Only the residual non-limit branch runs the new classifier, defaulting to `unknown`. The now-redundant `terminal_reason:` keyword and its `"triage_failed"` / `"fix_failed"` call-site values are dropped; `phase=triage|fix` still carries the phase distinction those tokens redundantly encoded.

The condensed raw cause is untouched: `review_phase_error_summary` still whitespace-collapses and 300-char-caps the original `error_message` into `message=`, so even an `unknown` row stays diagnosable.

### Read surfaces unchanged

The TUI status column (`review_recovery_status`), `hive status --json`, and the web card already render `attrs["reason"]` generically, so they show the new reasons with no code change. A `tasks_pane` render assertion was added to lock that contract.

### Scope note

The classifier receives whatever text the caller passes. On today's non-limit triage/fix path that is a condensed hive wrapper string (`expected output file missing or empty: …`, `tmux_session_terminated before writing …`, `triage agent failed (timeout)`), none of which match a specific pattern — so real-world input currently falls back to `unknown`, with the raw cause preserved in `message=`. Surfacing raw agent stdout/stderr into that branch (so the specific buckets fire in production) is out of this change's deliberate "one write site" scope; the classifier is in place for when that plumbing lands. This is documented in the `classify` docstring and the wiki.

## Test plan

- `ruby -Itest test/unit/review_error_reason_test.rb` — each classified reason via a representative line, per-regex-alternative binding, first-match-wins ordering, blank / nil / ANSI input, realistic wrapper strings → `unknown`, and a guard that `classify` can only ever emit enum members.
- `ruby -Itest test/unit/daemon/stale_agent_healer_test.rb` — pins every classified reason as **not** auto-recoverable (the two-tier recovery invariant).
- `ruby -Itest test/integration/run_review_test.rb` — merge-conflict / network-timeout / permission-denied / crash / unrecognised `error_message` end-to-end through the runner; the realistic wrapper string stays terminal `unknown`.
- `ruby -Itest test/unit/tui/views/tasks_pane_test.rb` — a classified reason renders verbatim in the status column.
- Full suite + coverage gate: **6514 runs, 0 failures, 100% line coverage**.

## Review summary

Reviewed across three CE passes by multiple personas (correctness, testing, maintainability, reliability, adversarial, api-contract) plus codex and the pr-review-toolkit. **No P0/P1 findings** — the implementation is correct and contract-safe: the limit gate stays first (quota/429 still auto-heal), the closed enum is closed by construction, ReDoS is not viable (linear regexes over single stripped lines), and marker injection is neutralized downstream.

Quality findings were auto-fixed: per-regex representative-line binding (so a typo'd alternative fails a test rather than silently degrading to `unknown`), `CLASSIFIED` derived from `PATTERNS` plus a load-time closed-enum guard, a healer "never auto-recover" regression test, negative tests pinning wrapper strings to `unknown`, and comment / docstring / wiki accuracy on the input contract.

Two threads were escalated for human decision:

1. **Rebase reconciliation with #617.** This branch forked before #617 independently reworked `mark_review_phase_failure` (adding a `limit_text:` channel and a `from_limit?` gate). The two changes are orthogonal — the recommended resolution keeps both layers: #617's limit gate as the first arm, this task's `ReviewErrorReason.classify` in the residual non-limit branch. Rebase onto current `origin/main` and re-run the suite before merge.
2. **Efficacy gap (see Scope note).** The classifier's specific buckets only fire on raw agent output; today's callers pass wrapper strings, so production emits `unknown`. Merge as forward-looking, documented infra, or expand scope to surface raw stdout/stderr — a scope decision beyond the plan's single-write-site boundary.

## Linked task

Hive task `can-we-write-instead-of-260523-ee3c` (PR #627) — classify triage/fix failures into real error reasons.

---

Opus 4.8


ivankuznetsov:

## High
- [ ] Preserve the raw limit signal when triage/fix agents hit Claude's live limit menu (`lib/hive/stages/review.rb:874`): `mark_review_phase_failure` now re-scans only the summarized `error_message`; `AgentLimit.error_message(pane_tail)` can reduce a real menu to `limits reached for claude: What do you want to do?`, which does not match `limit_reached?`, so planned `limits_reached` + `retry_after` markers become terminal `unknown`.
- [ ] Restore live-pane limit detection instead of scanning the entire captured tail (`lib/hive/claude_launcher.rb:741`): replacing `live_limit_line` with `AgentLimit.limit_reached?(pane_tail)` matches old quoted/history lines such as "Stop and wait for limit to reset" even when a ready prompt appears below, reintroducing false `limits_reached` failures in healthy sessions.
- [ ] Restore `hive review --pr` and the ad-hoc review-only gates (`lib/hive/cli.rb:540`): this branch deletes `lib/hive/commands/adhoc_review.rb`, removes the `--pr` option, and removes the ad-hoc fix-off/CI-skip checks, breaking the documented existing PR review workflow outside this classifier task's scope.
- [ ] Keep CI-fix provider limits on the existing cooldown path (`lib/hive/stages/review.rb:206`): the prior `phase=ci reason=limits_reached retry_after=...` branch and `CiFix::Result#limit_text` were removed, so CI fix-agent quota failures now become manual `ci_unrunnable` errors instead of daemon-retryable holds.

## Medium
- [ ] Move the new AgentLimit regression tests back above `private` (`test/unit/agent_limit_test.rb:214`): `test_does_not_classify_ui_feature_limit_text_as_a_wall` and `test_detects_usage_credit_and_time_window_walls` are currently private methods, so Minitest will not run them.

## Nit
- [ ] Drop the compiled wiki log edit (`wiki/log.md`): project instructions require feature PRs to add `wiki/log.d/*` fragments and not edit the generated `wiki/log.md` directly.

ivankuznetsov:

# Code review — `Hive::ReviewErrorReason` classifier (pass 1)

Scope: the task's own work is the 4 commits in `03c3d342..HEAD` (16 files). The
`origin/main..HEAD` range additionally shows ~131 unrelated files because
`origin/main` has diverged far ahead of this branch's fork point — those are NOT
this task's changes and were excluded from review (see M3).

Reviewed by 6 personas (correctness, testing, maintainability, reliability,
adversarial, api-contract). The implementation is correct, contract-safe, and
reliability-sound: no P0/P1. The limit gate stays first (quota/429 still
auto-heal), the closed enum is closed by construction, ReDoS is not viable
(linear regexes over single stripped lines), and marker injection is neutralized
downstream. Findings below are quality/robustness/integration items.

## High

(none)

## Medium
- [ ] `lib/hive/review_error_reason.rb:32` — `/rebase .* conflict/i` cannot match the bare phrase "rebase conflict": the `.*` sits between two literal spaces, so it requires intervening words ("rebase X conflict") and silently returns `unknown` for the common two-word form. Low real-world impact (genuine rebase output carries stronger signals caught by the other `merge_conflict` patterns), but it's a pattern that does not do what it reads as.
- [ ] `lib/hive/review_error_reason.rb` PATTERNS — 16 of the 30 regex alternatives have no test exercising them, and the 100% line-coverage gate cannot protect them (the regexes are array-literal expressions evaluated once at `require`, so they count as covered the instant `classify` runs). A typo in any alternative degrades silently to `unknown` with no failing test — this is exactly how the `/rebase .* conflict/i` gap above went unnoticed. Add table-driven unit cases (one realistic line per alternative).
- [ ] `lib/hive/review_error_reason.rb:50,64,68` — the broad `agent_crashed`/`tool_permission_denied`/`merge_conflict` patterns (`/\bkilled\b/i`, `/panic:/i`, `/fatal error:/i`, `/permission denied/i`, `/merge conflict in /i`) match benign agent *prose that merely discusses* errors (e.g. "do not panic: this is expected", an agent quoting `permission denied` while explaining a finding), stamping a misleading `reason=`. Blast radius is cosmetic — all classified reasons are terminal, the daemon healer ignores them, and the raw cause is preserved in `message=` — but it undercuts the plan's stated "bias toward `unknown`" intent. Consider classifying only the last N lines of the tail and/or anchoring patterns to log-line shapes, per the plan's "refine against real corpus tails" note.
- [ ] Stale-base / `#617` merge divergence — this branch forked at `03c3d342`, *before* origin/main's `#617` ("fix(review): correctly classify usage-limit failures in 6-review", `262299fa`) landed. `#617` evolved the very function this task edits (`mark_review_phase_failure`): it added a `limit_text:` parameter, a `limit_failure?` helper, switched the gate `AgentLimit.limit_reached?` → `AgentLimit.from_limit?`, and added a CI-phase `limits_reached` path. This branch reverts to the old `limit_reached?` gate and has no `limit_text` threading. A naive merge will conflict in `lib/hive/stages/review.rb` (+ `run_review_test.rb`, `phase_failure_helpers_test.rb`) and risks either regressing `#617`'s improved limit detection or dropping this classifier. Rebase onto current origin/main and re-validate the classifier against `#617`'s error_message/`limit_text` threading before merge.
- [ ] Missing safety-invariant test — no test pins that the classified reasons (`merge_conflict`, `network_timeout`, `tool_permission_denied`, `agent_crashed`, `unknown`) are NOT auto-recoverable by `StaleAgentHealer#auto_recoverable_review_error?`. This invariant is the only thing keeping a misclassification (M2) cosmetic; it currently holds by inspection only, so a future allow-list edit that adds one of these strings would silently arm an agent-controllable signal with no failing test. Add a healer-level regression guard feeding each new enum value and asserting `false`.

## Nit
- [ ] `web/test/integration/tasks_test.rb:351` — straggler fixture still seeds the retired literal `reason=triage_failed` (a value the producer can no longer emit). The test still passes because it asserts on the `phase`/marker, not the reason, so leaving it is plan-permissible (the plan said to update only assertions that test the value). Worth updating to `merge_conflict`/`unknown` for vocabulary consistency so it doesn't model a non-producible marker shape. (flagged by maintainability + api-contract)
- [ ] `lib/hive/review_error_reason.rb:5-20` — no test enforces the closed-enum invariant against drift in `PATTERNS`. `test_classified_outputs_are_closed_over_enum` checks `CLASSIFIED + ["unknown"] ⊆ REASONS`, but nothing asserts `PATTERNS.map(&:first).sort == CLASSIFIED.sort`, so a future PATTERNS entry whose reason is outside the enum could make `classify` return an undocumented value undetected. Also, `REASONS` is a producer-side contract with no schema/consumer pin (`schemas/hive-status.v4.json` leaves `attrs` free-form), so a misspelled written reason surfaces only in production. Add a self-enforcing test.
- [ ] `lib/hive/review_error_reason.rb:5-20` — the three-constant split (`REASONS` = full marker vocabulary incl. `limits_reached`/`rate_limited` which `classify` never emits; `CLASSIFIED` = the classifier's actual output set; `PATTERNS` = the regex tables) is not self-evident from the names. Add a one-line comment on each constant, in particular marking `rate_limited`/`limits_reached` as "stamped by AgentLimit, never by classify" so a reader scanning `REASONS` doesn't have to grep to learn which values this module produces.

ivankuznetsov:

# PR review (pr-review-toolkit) — `Hive::ReviewErrorReason` classifier (pass 1)

Scope: the task's own work, `git diff 03c3d342..HEAD` (16 files). The
`origin/main..HEAD` range shows ~131 unrelated files only because `origin/main`
has diverged past this branch's fork point — those are not this task's changes
and were excluded.

Six personas ran (code-reviewer, pr-test-analyzer, comment-analyzer,
silent-failure-hunter, type-design-analyzer, code-simplifier). All six independently
verified the core behavior is correct: the closed enum cannot leak today (fuzzed,
5000 inputs, zero leaks), the `AgentLimit` limit gate stays first (quota/429 still
auto-heal), the raw cause is preserved in `message=`, no downstream consumer keyed
off the old `triage_failed`/`fix_failed` literals, and no transient failure is newly
stranded. type-design ratings: encapsulation 8/10, invariant-expression 7/10,
usefulness 8/10, **enforcement 4/10** (see Medium). No Critical/High findings.

## High

(none)

## Medium
- [ ] [tests] `lib/hive/review_error_reason.rb` PATTERNS — 15 of 30 regex alternatives have zero test demonstrating they match (e.g. `/\beacces\b/i`, `/could not resolve host/i`, `/network is unreachable/i`, `/\bsigsegv\b/i`, `/\bsigkill\b/i`, `/fatal error:/i`, `/not allowed to (?:use|run|call)/i`, `/rebase .* conflict/i`). The 100% line-coverage gate cannot catch this — the regexes are array-literal expressions evaluated once at `require`, so they read as covered after a single `classify` call. A one-character typo in any untested alternative silently reclassifies real failures to `unknown` with a green build. Add a data-driven table test (`{representative-string => expected_reason}`) covering all 30 alternatives.
- [ ] [type-design] `lib/hive/review_error_reason.rb:5-26` — the "closed by construction" invariant is upheld only by manual discipline across three parallel lists, not structurally. `CLASSIFIED` is a hand-maintained duplicate of `PATTERNS.map(&:first)`, and the guard test `test_classified_outputs_are_closed_over_enum` only checks `CLASSIFIED + ["unknown"] ⊆ REASONS` — it never inspects `PATTERNS`. So a future `PATTERNS` entry with a new/typo'd reason key (e.g. `"disk_full"`, `"tool_perm_denied"`) would be returned by `classify`, fall outside `CLASSIFIED`/`REASONS`, and ship green. Fix: derive `CLASSIFIED = PATTERNS.map(&:first).freeze` (makes `PATTERNS.map(&:first) ⊆ CLASSIFIED` true by construction) and/or add `assert_equal CLASSIFIED.sort, PATTERNS.map(&:first).sort` to the closure test.

## Nit
- [ ] [comments][type-design][silent-failure] `lib/hive/review_error_reason.rb:7` — `rate_limited` is an orphan member of `REASONS`: not in `CLASSIFIED`, no `PATTERNS` entry, never returned by `classify`, and unreferenced elsewhere in `lib/` (limit text writes `limits_reached` via the `AgentLimit` gate). A maintainer can't tell if it's dead, aspirational, or emitted elsewhere — a latent dead-branch trap, and a contributor could wire `classify` to emit it, regressing a self-healing `limits_reached` hold into a terminal reason. Either drop it or add a one-line comment marking `limits_reached`/`rate_limited` as "owned by the AgentLimit gate, never produced here; listed so `REASONS` is the complete `reason=` vocabulary." (flagged by 3 personas)
- [ ] [comments][type-design] `lib/hive/review_error_reason.rb:5-20` — the `REASONS` (full vocabulary) vs `CLASSIFIED` (classifier's emitted subset) vs `PATTERNS` (regex tables) distinction carries no comment; the file is the authoritative reason list the wiki points readers to. Add a one-line comment on each constant to prevent future drift (e.g. someone adding a `PATTERNS` entry but forgetting `CLASSIFIED`).
- [ ] [code-review][silent-failure] `lib/hive/review_error_reason.rb:58,68` — the `agent_crashed` patterns `/\bkilled\b/i` and `/fatal error:/i` are broad enough to match benign agent prose that merely mentions those words ("killed the flaky test", quoted compiler output), stamping a misleading `reason=agent_crashed`. Impact is cosmetic only — all classified reasons are equally terminal, the healer ignores them, and the raw cause rides in `message=` — so flag only if operators triage by the `reason=` class. Consider anchoring to log-line shapes or classifying only the tail lines.
- [ ] [tests] `lib/hive/review_error_reason.rb:84-87` — per-line matching semantics (a signal whose words straddle a newline is not matched) is a deliberate design property but is pinned by no test; a future refactor to whole-string `match?` would change behavior with a green suite. Add one assertion documenting the intended line-scoped behavior.
- [ ] [simplify] `lib/hive/review_error_reason.rb:81-84` — `classify` walks the full normalized string once for the `text.strip.empty?` guard, then strips every line again in the scan. The guard could fold into the line scan (`lines = normalize(...).each_line.map(&:strip).reject(&:empty?)`; empty `lines` falls through to `"unknown"`) to drop one full-string pass. Minor and defensible as-is — it mirrors `AgentLimit#limit_reached?`'s `return false if normalized.empty?` shape; verify the 100% line gate if changed.
- [ ] [type-design] `lib/hive/review_error_reason.rb:5-20` — `REASONS`/`CLASSIFIED`/`PATTERNS` are public constants but no `lib/` consumer reads them (only tests do); they could be `private_constant` with a `classified?(reason)` query method to avoid exposing the table shape. Optional — keeping them public matches the `AgentLimit` sibling's house style.

ivankuznetsov:

# Code Review — classify triage/fix failures into real error reasons (pass 2)

Scope reviewed: the branch's own delta (`git diff 03c3d342..HEAD`, the merge-base),
which is the classify-review-error work described by the plan — `lib/hive/review_error_reason.rb`,
`lib/hive/stages/review.rb` (`mark_review_phase_failure`), and the test/wiki updates.
The raw `git diff origin/main..HEAD` is dominated by 24 unrelated origin/main commits
(a large deletion refactor) that this branch never touched; those deletions are not
this task's work and are excluded.

Overall the authored implementation is high quality and faithful to the plan:
the classifier mirrors `AgentLimit` (ordered line-based regex, false-negative bias),
the limit gate stays first (Open Question A resolved as recommended — `rate_limited`
reserved, 429/quota still auto-heal), old `triage_failed`/`fix_failed` literals are
fully removed, and coverage is thorough (per-regex-alternative representative lines,
a `CLASSIFIED`↔`PATTERNS` drift guard, a newline-straddle pin, a healer "never
auto-recover" safety invariant, and triage+fix end-to-end classification). The new
unit test, healer test, and tasks_pane test all pass locally. The findings below are
integration/mergeability concerns against current `origin/main`, plus one nit.

## High
- [ ] Branch is 24 commits behind `origin/main` and does not merge cleanly — `git merge-tree origin/main HEAD` reports content conflicts in 5 files (`lib/hive/stages/review.rb`, `test/integration/run_review_test.rb`, `test/unit/daemon/stale_agent_healer_test.rb`, `wiki/gaps.md`, `wiki/stages/review.md`): needs a rebase onto `origin/main` before it can land.
- [ ] `origin/main` independently reworked the *same* `mark_review_phase_failure` this branch rewrites — main now threads a separate `limit_text:` channel through `limit_failure?(limit_text:, error_message:)` (`limit_text` OR `AgentLimit.from_limit?`), whereas this branch still gates on the older `AgentLimit.limit_reached?(error_message)` only: a naive merge that keeps this branch's version drops main's `limit_text` limit-detection path, so a triage/fix provider-limit wall captured only in the raw `limit_text` stream (absent from `error_message`) would misclassify as `unknown` and go terminal instead of self-healing — the exact regression main's refactor fixed.
- [ ] Rebase must re-apply the classifier *inside* main's limit-gated `else` branch, not replace it — keep main's `limit_failure?(limit_text:, error_message:)` as the first arm and swap only the terminal `reason: terminal_reason` for `reason: Hive::ReviewErrorReason.classify(error_message)`, and re-add the `limit_text:` kwarg the two call sites (`review.rb:429`/`:565`) must forward, so both improvements (limit self-heal + reason classification) survive.

## Medium
- [ ] The plan's "current shape" of `mark_review_phase_failure` (limit gate = `AgentLimit.limit_reached?`, single `else`) reflects the stale merge-base, not current `origin/main` (`limit_failure?` + `from_limit?` + `limit_text`): re-verify the classifier wiring against main's actual helper after rebase so the review-stage behavior described in the updated `wiki/stages/review.md` ("provider-limit text → `limits_reached`") still holds for the `limit_text`-only detection case.

## Nit
- [ ] `agent_crashed` pattern `/\bkilled\b/i` is broad — a hive-enforced deadline/OOM kill can surface bare "Killed" in captured output and would classify as `agent_crashed` rather than the plan's intended `unknown` for a clean phase-deadline timeout; it's plan-sanctioned and mitigated (first-match ordering, `message=` preserves the raw cause), but worth a corpus check that real timeout tails don't carry "killed".

ivankuznetsov:

# PR Review Toolkit — multi-persona review (pass 2)

Task: "classify triage/fix failures into real error reasons." Personas run: code-reviewer,
pr-test-analyzer, comment-analyzer, silent-failure-hunter, type-design-analyzer, code-simplifier.

Scope reviewed: the authored delta `git diff 03c3d342..HEAD` (the merge-base). The raw
`origin/main..HEAD` diff is polluted by 24 unrelated origin/main deletion-refactor commits this
branch never touched; those are excluded.

Consensus across personas: the implementation is correct, well-tested, and idiomatic (a faithful
mirror of `Hive::AgentLimit`). The limit gate stays first, `message=` is preserved, classified
reasons are provably never auto-recoverable, and the drift guards close the regex coverage blind
spot. code-simplifier found nothing to simplify. Findings below are the union of all persona
findings; most are Medium/Nit polish plus one integration (merge) blocker.

## High
- [ ] [code-reviewer] Merge-conflict signature reconciliation on `mark_review_phase_failure`: this branch removed the `terminal_reason:` kwarg from the method and both call sites (`lib/hive/stages/review.rb:429`, `:565`, `:873`), while origin/main independently added a `limit_text:` param + `limit_failure?` helper to the same method — a naive merge must reconcile both so neither the `ReviewErrorReason.classify` call nor origin/main's limit-text self-heal channel is dropped. Needs rebase before merge.

## Medium
- [ ] [comment-analyzer, pr-test-analyzer] `rate_limited` is phantom/dead vocabulary: the `REASONS` header comment claims `rate_limited` is "stamped by the AgentLimit gate," but nothing in `lib/` ever writes `reason=rate_limited` (the gate only ever stamps `limits_reached`), and the closure test only checks `CLASSIFIED + [unknown] ⊆ REASONS` so the unused member passes silently — drop it or relabel it a reserved-but-unemitted slot (`lib/hive/review_error_reason.rb:5-18`).
- [ ] [comment-analyzer] "the one authoritative reason list" overstates `REASONS`: `review_error` markers elsewhere also carry `ci_unrunnable`, `reviewer_partial_failure`, `fix_status_check_failed`, `fix_tampered`, and the `phase=resume` reasons — scope the comment to "the reasons produced by `mark_review_phase_failure`" so a maintainer isn't misled into thinking it enumerates all `review_error` reasons (`lib/hive/review_error_reason.rb:5-9`).
- [ ] [silent-failure-hunter] No benign-line filter (unlike `AgentLimit::BENIGN_PATTERNS`): `permission denied`, `fatal error:`, and `\bkilled\b` can match incidental words in an agent's final output (e.g. `Permission denied (publickey)`, an agent describing a "fatal error:" it found, "task was killed"), so the `reason=` shown in TUI/status/web can lead an operator with a specific but wrong diagnosis — mitigated (terminal either way, raw cause preserved in `message=`), so it's a routing/UX risk, not a silent failure (`lib/hive/review_error_reason.rb:44-65`).
- [ ] [type-design-analyzer] `CLASSIFIED` is a hand-maintained duplicate of `PATTERNS.map(&:first).uniq` whose correctness rests entirely on a drift-guard test — derive it (`CLASSIFIED = PATTERNS.map(&:first).uniq.freeze`) to make the invariant true by construction, retire the guard test, and eliminate the drift class at zero behavior cost (`lib/hive/review_error_reason.rb:23-28`).

## Nit
- [ ] [type-design-analyzer] The central closed-enum invariant (`CLASSIFIED ∪ {unknown} ⊆ REASONS`) is enforced only by tests — add a one-line load-time guard that raises if `(CLASSIFIED - REASONS)` is non-empty so the highest-cost drift fails at require time, not just under a comprehensive suite run (`lib/hive/review_error_reason.rb`).
- [ ] [type-design-analyzer] `PATTERNS` is exposed at module scope but is a pure internal that only tests read — consider `private_constant :PATTERNS` (feasible today since the per-alternative tests already go through `classify`) to stop a future consumer hardening the regex table into contract (`lib/hive/review_error_reason.rb:32`).
- [ ] [silent-failure-hunter] `agent_crashed` for an OS OOM/`SIGKILL` (`/\bkilled\b/i`, `core dumped`) mislabels a resource-exhaustion/infra condition as an agent code bug, potentially biasing the operator away from "the box was starved" — label-precision only; terminal + `message=` preserved (`lib/hive/review_error_reason.rb:61`).
- [ ] [pr-test-analyzer] After the `triage_failed`/`fix_failed` fixtures were retired, no TUI/web/dispatcher render test exercises those display paths with `reason=unknown` — the value most likely to appear in production; behavior-neutral (reason is opaque display payload) so no action strictly required, flagged for awareness (`test/unit/tui/bubble_model_test.rb`, `test/unit/web/dispatcher_test.rb`).

ivankuznetsov:

## High
- [ ] Massive out-of-scope reversion in `git diff origin/main..HEAD`: the plan calls for a small classifier/wiring change, but the branch deletes or rewrites shipped commands, daemon/web helpers, schemas, docs, and tests (for example `lib/hive/commands/adhoc_review.rb`, `lib/hive/daemon/recoverable_error_healer.rb`, and `lib/hive/web/app_bundle.rb`), so merging would remove unrelated production behavior.
- [ ] Ad-hoc review fix-off gates were removed in `lib/hive/stages/review.rb:190` and `lib/hive/stages/review.rb:497`: ad-hoc PR reviews with the default `review.adhoc.fix: false` can now run CI/fix agents and mutate a borrowed PR worktree.
- [ ] CI provider-limit failures no longer self-heal at `lib/hive/stages/review.rb:206`: every `CiFix` error is stamped `reason=ci_unrunnable`, so quota/rate-limit walls lose the previous `reason=limits_reached retry_after=...` cooldown path.
- [ ] Triage/fix limit classification dropped raw `limit_text`/canonical limit support at `lib/hive/stages/review.rb:429`, `lib/hive/stages/review.rb:565`, and `lib/hive/stages/review.rb:873`: lossy agent errors that used to be recognized by the limit gate can now fall through to terminal classified reasons, violating the plan's "limit gate first and unchanged" requirement.
- [ ] Claude tmux limit detection now scans whole pane tails with `AgentLimit.limit_reached?` at `lib/hive/claude_launcher.rb:741` and `lib/hive/claude_launcher.rb:825`: deleting the live-menu filter lets quoted or stale "Stop and wait for limit to reset" text falsely stamp `limits_reached` after the agent has moved on.

## Medium
- [ ] `test/unit/agent_limit_test.rb:214` moves `private` above intended `test_*` methods: Minitest will not run the UI-limit and real-wall assertions at lines 228 and 240, leaving the changed classifier behavior unpinned.

## Nit

ivankuznetsov:

# Code Review — pass 3 — `can-we-write-instead-of-260523-ee3c`

Scope reviewed: the feature commits `git diff origin/main...HEAD` (merge-base `03c3d342`) —
`lib/hive/review_error_reason.rb` (new classifier), its wiring into
`Hive::Stages::Review#mark_review_phase_failure`, unit/integration tests, and wiki docs.
Intent (per plan): classify non-limit triage/fix review-phase agent failures into a closed
enum (`merge_conflict`, `network_timeout`, `tool_permission_denied`, `agent_crashed`,
falling back to `unknown`) instead of the opaque `triage_failed`/`fix_failed` token, with the
`AgentLimit` limit-gate kept first so quota/429 still auto-heal.

Reviewers: correctness, adversarial, testing, maintainability (parallel personas) + orchestrator
verification of the data-flow into `classify`. The design is correct and well-tested; the enum is
closed by construction, the two-tier recovery model holds (verified against
`auto_recoverable_review_error?` — all classified reasons fall through to terminal, and the new
`stale_agent_healer_test.rb` guard pins this off `CLASSIFIED`), and the wiki accurately reflects the
new vocabulary. The findings below are led by one efficacy gap against the plan's core goal.

## High
- [ ] Classifier will emit `unknown` for essentially every real non-limit triage/fix failure — `ReviewErrorReason.classify` is fed `error_message`, but on the non-limit triage/fix path that value is always a hive-generated wrapper string (`tmux_session_terminated before writing expected output file: …`, `expected output file missing or empty: …`, `claude reported :error via result.json`, `claude stop hook did not signal completion`, `exit_code=N`, or the `"triage agent failed (…)"` fallback — see `claude_launcher.rb:742-856`, `agent.rb:437-462`, `triage.rb:148`), none of which contain the merge-conflict/network/permission/crash text the regexes target. Raw agent pane/stderr output reaches `error_message` only on the limit path (`AgentLimit.error_message`), which the limit gate consumes first (`review.rb:874`). So the plan's headline goal — a specific machine-readable `reason=` in the field for Q3/Q8 — is not realized and is not deferred anywhere; net production effect is renaming the `triage_failed`/`fix_failed` fallback to `unknown`, and the daemon-healer/`hive status --json`/TUI `reason=` the plan set out to de-genericize stays generic. Not a crash (fallback is safe, `message=` still carries the wrapper cause) but the classifier's four specific reasons fire only on the injected test strings. Fix: surface the captured pane/stderr tail into `error_message` on the non-limit triage/fix branch (or tune patterns to the actual wrapper vocabulary), and add an integration case that drives a realistic non-limit failure through the real spawn and asserts a specific reason — decide explicitly whether to merge this as forward-looking infrastructure that emits `unknown` today.

## Medium
- [ ] `test/unit/review_error_reason_test.rb` `test_every_pattern_alternative_has_a_representative_line` guards only count parity (`alternative_count == REPRESENTATIVE_LINES.size`), not the one-to-one regex→line correspondence its own comment promises — a future edit that double-covers one alternative and orphans another passes green, letting a per-regex typo (the exact failure this table exists to catch, since the 100%-line gate reads a dead alternative as covered) ship silently. Fix: assert each individual regex in `PATTERNS` matches ≥1 representative line, so the property is enforced structurally rather than by count.
- [ ] Branch is stale relative to `origin/main`: merge-base is `03c3d342` while `origin/main` has advanced far past it, so the literal review target `git diff origin/main..HEAD` shows ~24k deletions of unrelated features that exist on main but not this branch. The feature diff itself (three-dot `origin/main...HEAD`, reviewed here) is clean, but merging the branch without rebasing would revert large amounts of main. Fix: rebase onto current `origin/main` before merge and re-run the suite/coverage gate.

## Nit
- [ ] `agent_crashed` (`/\bkilled\b/i`, `/fatal error:/i`, `/panic:/i`, `review_error_reason.rb:64-73`) and `merge_conflict` (`/needs merge/i`, `:34`) alternatives are broad enough to false-positive on benign agent narration or ordinary `git status` chrome, contradicting the file's own stated false-negative bias (flagged by both correctness and adversarial reviewers). Cosmetic today — all classified reasons are terminal and the raw cause is preserved in `message=` — but would mislabel once raw output is surfaced (see High). Fix: anchor these to a process/exit/signal context rather than a bare mid-sentence keyword.
- [ ] No negative test that realistic hive wrapper strings / benign narration classify to `unknown` — the suite proves patterns FIRE but never that they don't over-fire on prose; notably the healer's own `tmux_session_terminated before writing expected output file` signature (a real `error_message` that reaches `classify`) is untested against the classifier. Fix: add a small negative-case table asserting these land on `unknown`.
- [ ] `rate_limited` is a reserved-but-never-emitted `REASONS` member (`review_error_reason.rb:17`) — accepted on read, never written by `lib/`. It is well-documented and pinned by `test_rate_limit_language_is_reserved_for_agent_limit_gate`, but is a mild clarity trap; if no read-side consumer keys off it, consider dropping it from the enum in a later pass.
- [ ] Pre-existing / out-of-scope: shared `AgentLimit.normalize` strips `\r` and ANSI CSI as empty (`agent_limit.rb:175`), so a keyword split by a color reset or CR rejoins into a single matchable line — the per-line non-match invariant is pinned only for `\n` (`test_signal_split_across_newline_is_not_matched`). Harmless for `classify` (terminal either way) and lives in the unchanged `AgentLimit` module; noted only for awareness since `classify` inherits it.

ivankuznetsov:

# PR Review (pr-review-toolkit) — pass 3 — `can-we-write-instead-of-260523-ee3c`

Multi-persona review of the feature commits `git diff origin/main...HEAD` (merge-base
`03c3d342`; the two-dot `origin/main..HEAD` is polluted by a stale branch base — the branch
must be rebased before merge). Personas run: code-reviewer, silent-failure-hunter,
pr-test-analyzer, comment-analyzer, type-design-analyzer, code-simplifier.

Intent: classify non-limit triage/fix review-phase agent failures into a closed enum
(`merge_conflict`, `network_timeout`, `tool_permission_denied`, `agent_crashed`, else
`unknown`) via new `Hive::ReviewErrorReason.classify`, replacing the opaque
`triage_failed`/`fix_failed` reason, with the `AgentLimit` limit-gate kept first.

Union of findings below. The mechanism is correct, safe, and well-tested (closed enum by
construction, load-time guard, two-tier recovery verified, `message=` preserved on every
branch, no swallowed errors); the lead finding is an efficacy gap all four investigating
personas reached independently.

## High
- [ ] [code-reviewer, silent-failure-hunter, comment-analyzer, type-design] Classifier is starved of the failure signal — `classify` returns `unknown` for essentially every real triage/fix failure: on the non-limit path `error_message` is always a hive-generated wrapper string (`tmux_session_terminated before writing expected output file: …`, `tmux_pane_unreadable: …`, `expected output file missing or empty: …`, `claude reported :error via result.json`, `claude stop hook did not signal completion`, `exit_code=N`, or the `"triage agent failed (…)"` fallback — `claude_launcher.rb:753/774/783/843/852`, `agent.rb:441/448/461`, `triage.rb:148`), none of which match the merge/network/permission/crash regexes; raw agent pane output reaches `error_message` only via `AgentLimit.error_message` on the limit path, which the gate consumes first (`review.rb:874`). The four specific reasons fire only on the stubbed integration inputs (each persona verified empirically, confidence ~88). So the plan's core goal — a specific machine-readable `reason=` in the field for the healer/`hive status --json`/TUI (Q3/Q8) — is unrealized and not deferred; net production effect is renaming the fallback token to `unknown`. Not a data-loss bug (`message=` still carries the raw wrapper cause) but a correctness/efficacy gap. Fix: surface the captured pane/stderr tail into the non-limit `error_message` (mirror `AgentLimit.error_message`) so the classifier receives the signal — or explicitly scope this down and merge as inert forward-looking infra with the docs corrected (see Medium doc finding).

## Medium
- [ ] [pr-test-analyzer] `test_every_pattern_alternative_has_a_representative_line` (`test/unit/review_error_reason_test.rb:100`) checks only count parity (`alternative_count == REPRESENTATIVE_LINES.size`), not per-alternative binding — an add-a-regex-without-a-line + duplicate-cover-a-sibling edit passes green while a per-regex typo ships (empirically confirmed by the persona; this is the exact coverage-gate blind spot the table was written to close). Fix: assert each individual regex matches ≥1 representative line that no earlier-priority regex matches, or index `REPRESENTATIVE_LINES` by `[reason, alternative_index]`.
- [ ] [pr-test-analyzer] The two new integration cases (`run_review_test.rb` triage/fix "classified_from_captured_output") inject crafted `error_message` via stubbed `Triage.run!` / `spawn_fix_agent` — they prove the `mark_review_phase_failure → classify → marker` wiring but largely duplicate the unit suite and never drive a non-injected spawn failure through classification; the only realistic-input test is `test_triage_non_limit_error_stays_terminal_unknown` (feeds `"triage agent failed (timeout)"`, asserts `unknown`), which is load-bearing and should be recognized as such. Fix: keep one wiring positive per phase, and add real-path negative coverage rather than more synthetic positives.
- [ ] [comment-analyzer, silent-failure-hunter] Code comments and wiki overstate what the classifier delivers in the field: `review_error_reason.rb:26-27/94-96` and `wiki/{stages/review.md:93,112, state-model.md:104, modules/markers.md:59}` imply the specific reason buckets fire on real causes, but production input yields `unknown` (see High). A maintainer may build TUI affordances, healer arms, or metrics around buckets that never appear. Fix: state the input contract in the `classify` docstring (it classifies whatever text the caller passes; callers currently pass wrapper strings) and note in the wiki that specific buckets require raw agent output the current triage/fix plumbing does not forward.

## Nit
- [ ] [type-design] `"unknown"` membership in `REASONS` is guaranteed only by the hand-written literal (`review_error_reason.rb:23`) plus `test_classified_outputs_are_closed_over_enum`, NOT by the load-time `raise` (which validates only the derived `CLASSIFIED`) — the most-emitted value has the weakest structural guarantee. Fix: `FALLBACK = "unknown"` and extend the guard to `((CLASSIFIED + [FALLBACK]) - REASONS).any?`.
- [ ] [code-simplifier] `classify`'s `return "unknown" if text.strip.empty?` (`review_error_reason.rb:99`) is a redundant early return — blank/whitespace input yields `lines == []`, the `PATTERNS` loop short-circuits, and the trailing `"unknown"` returns the identical result. Drop it (single clear win; keeps one fallback site).
- [ ] [pr-test-analyzer] Add a unit test pinning realistic wrapper strings (`"expected output file missing or empty: …"`, `"exit_code=1"`, `"tmux_session_terminated before writing expected output file: …"`) to `unknown`, so a future regex broadening (e.g. adding `/terminated/i` to `agent_crashed`) fails loudly instead of silently reclassifying an infra timeout as a crash.
- [ ] [comment-analyzer] The `CLASSIFIED` comment (`review_error_reason.rb:81-85`) says the lockstep holds "by construction … no guard test to keep it honest," but the adjacent CLASSIFIED⊆REASONS invariant IS guarded by the load-time `raise` (:90) and `test_classified_outputs_are_closed_over_enum` — a reader could delete those thinking they're redundant. Fix: tighten to "no hand-maintained duplicate reason list to drift."
- [ ] [type-design] Reserved `rate_limited` (`review_error_reason.rb:17`) is permitted by the enum but produced by nothing in `lib/` — documented and tested, but a minor "illegal-states-representable" smell (read vocabulary broader than write vocabulary). Consider dropping it if no read-side consumer (status/JSON/TUI/healer) keys off it.
- [ ] [code-simplifier] Trim the redundant data-shape half of the `PATTERNS` comment (`review_error_reason.rb:26`) — "ordered [reason, [regexes...]] rows" duplicates the literal directly below; keep the load-bearing priority sentence.
- [ ] [code-simplifier] Cosmetic asymmetry in `mark_review_phase_failure`: `error_message.to_s` is passed to `limit_reached?` (`review.rb:874`) but the raw `error_message` to `classify` (:881). Both are safe (`classify`→`AgentLimit.normalize`→`.to_s.scrub` handles nil/false from the fix call site's `fix_result && fix_result[:error_message]`); optional to align for readability.

ivankuznetsov:
## Architecture review (user + best-practices lens)

Verdict: **correct shape, well-guarded, honestly documented** — the closed enum is closed by construction (`CLASSIFIED` derived from `PATTERNS`, load-time guard, emitted-subset test), the `AgentLimit` gate stays first so quota/429 keeps the cooldown auto-heal, and dropping the redundant `terminal_reason:` kwarg (its information already lives in `phase=`) is good API subtraction. One small fix is stacked as #646.

### Fixed in #646

- The reserved-but-unemitted `rate_limited` enum member. Its stated justification ("accepted on read") has no consumer — nothing reads `REASONS`; the healer keys on the literal `limits_reached`. A dead member in a closed enum invites the exact misrouting the two-tier recovery design prevents: future 429 handling emitted as `rate_limited` would sound like a limit but land terminal, bypassing the cooldown auto-heal. Details in the #646 comment.

### The real architectural question (already escalated, agreeing out loud)

The PR's scope note is admirably honest: today's triage/fix callers pass condensed wrapper strings, so **every production failure classifies as `unknown`** — the four specific buckets are dead code until raw agent stdout/stderr is plumbed into `error_message`. From the user's perspective, v1 changes `reason=triage_failed` to `reason=unknown`, which is arguably a *worse* label for the same information (`triage_failed` at least said which phase; that now lives only in `phase=`, which the TUI/web do render, so no net loss — but no net gain either).

I'd recommend not letting this sit: either (a) land the raw-output plumbing as the immediate follow-up task so the buckets fire, or (b) if that plumbing is far off, consider whether the classifier should live in the follow-up PR alongside it. Merging infrastructure ahead of its signal is defensible (the write-site, enum, healer invariants, and render contracts are all locked by tests now), but only if (a) actually happens.

### Considered, no change requested

- **Classifier placement** (`lib/hive/review_error_reason.rb`, top-level like `AgentLimit`) is right — the healer and status surfaces are cross-cutting consumers of the vocabulary even though only review writes it.
- **Structural mirroring of `AgentLimit`** (normalize → line-scan → ordered first-match tables) without extracting a shared classifier base: correct call. Two uses with different result shapes (boolean vs enum) don't justify the abstraction.
- The per-regex representative-line test binding (a typo'd alternative fails a test instead of silently degrading to `unknown`) is a pattern worth reusing elsewhere.

Stacked refactor: #646.

PR #628, opened 2026-06-30
Hive patrol: git grep can smuggle exec options after a consumed separator

ivankuznetsov:
## Summary

`bin/hive-babysitter-stub-git` is the dry-run git stub the babysitter uses so agents can run read-only git commands without reaching real git. It decided whether a command was a safe read by scanning the tokens before the `--` pathspec separator. This PR closes two ways to smuggle an *executing* option past that scan:

- **`git grep` consumed-separator smuggling** (original finding): the scan stopped at the first `--` via `take_while`, but `git grep`'s value-taking options (`-e`, `--file`, …) consume the following token — including a `--`. So `git grep -e -- --open-files-in-pager=/tmp/pager` or `git grep -e -- --textconv` fed the `--` to `-e` as the pattern and left the exec-bearing option for real git to parse, slipping a pager/textconv command past the stub. The fix replaces `take_while` with a subcommand-aware scan (`command_option_scan`) that advances over value-taking options the way git does, so `--` is treated as the separator only when it was not consumed as an option value.

- **log/show/rev-list `--pretty`/`--format` over-consumption** (review auto-fix): the new scanner must *not* treat `--pretty`/`--format` as consuming the next token. Git registers them stuck-only (`PARSE_OPT_OPTARG`): a bare `--pretty` uses the default format and the following argv element is parsed on its own, so `git log --pretty --output=/tmp/x` makes git write the file. Skipping that token would slip a trailing `--output`/`-o` past the write-file guard. `command_option_consumes_next?` returns `false` for these, and the scan still records the next token so the existing `%G` signature over-block stays intact.

Both paths fail closed: ambiguous parses are logged and skipped rather than reaching real git.

## Test plan

- `bundle exec rake test` — unit suite passes (`test/unit/babysitter/dry_run_env_test.rb`).
- `bundle exec rubocop` — lint passes.
- New regressions in `dry_run_env_test.rb` assert the stub skips, and never execs real git for:
  - `git grep -e -- --open-files-in-pager=…` and `git grep -e -- --textconv`
  - `git log --pretty --output=…`, `git show --pretty --output …`, `git rev-list --format --output=… HEAD`
  - and that the dry-run skip log records each refused command.

## Review summary

Two codex-native review passes. Pass 01 raised a single High finding — the log/show/rev-list `--pretty --output` over-consumption above — which was auto-fixed in commit `17ec602a` with regressions. Pass 02 found no correctness, security, or maintainability regressions. No findings were suppressed and nothing was escalated to the user.

## Linked task

Hive patrol finding `command-bin-hive-babysitter-stub-git` (slug `patrol-command-bin-hive-babysitter-stub-git-16fdbcdb`).

- Category: `security` · Severity: `high` · Confidence: `high`
- Fingerprint: `16fdbcdbf9d89209d3be460748e265533320b7e01baf1b50a3b87ea0ff91a7ed`


ivankuznetsov:
### Reviewer: codex-native-review - Pass 01

## High
- [ ] Do not skip log/show options after --pretty — /home/asterio/Dev/hive.worktrees/.patrol/hive-patrol-command-bin-hive-babysitter-stub-git-16fdbcdb/bin/hive-babysitter-stub-git:138-139 When a dry-run command uses `git log --pretty --output=/tmp/file` (or `git show`), git does not consume the next argv as a format value; it parses `--output` as the write-to-file option. This new consumption skips the existing output guard, so the stub execs real git and creates/writes the file instead of logging a s…

## Medium
No findings.

## Nit
No findings.

PR #629, opened 2026-06-30
feat(bot): add Telegram pairing and CLI approval flow

ivankuznetsov:
## Summary

A stranger who DMs the hive Telegram bot used to be dropped silently: the only way to grant access was to hand-edit `bot.chat_id_allowlist` in `config.yml` and restart. This PR adds a self-service pairing flow modeled on OpenClaw. An unknown chat presses **/start**, the bot replies with their Telegram user id and a one-time pairing code, and the owner approves from the CLI:

```
hive pairing approve telegram <CODE>
```

On approval the chat_id is appended to the allowlist through the existing locked atomic writer, the running bot is signalled to reload, and the user receives a proactive "✅ Approved" DM. The bot can now also bootstrap with an empty allowlist when pairing is enabled, so a fresh install is reachable before any chat is approved.

Scope is deliberately narrow: DM-only, Telegram-only, single global allowlist. `deny`/`revoke`, a web UI, multi-platform abstraction, and per-project allowlists are explicit non-goals — `telegram` is a fixed literal, not a dispatch dimension.

## Flow

```mermaid
flowchart TB
    S[Stranger DM: /start] --> R[Router: unauthorized + pairing_enabled]
    R --> PS[(PairingStore<br/>.bot.pairings.json)]
    R --> Reply[Reply: user id + code + approve command]
    O[Owner: hive pairing approve telegram CODE] --> PS
    O --> CFG[(config.yml<br/>chat_id_allowlist)]
    O --> Q[(PairingApprovalQueue)]
    O -. SIGHUP .-> BOT[Running bot]
    BOT -->|reaper drain| Q
    Q --> DM[Approved DM to chat_id]
```

The bot mints codes; the tokenless CLI consumes them, mutates config, and hands the "approved" DM back to the running bot to send. The two processes coordinate only through files.

## Design decisions

- **Throttle reuses the existing once-per-chat gate.** The router replies to an unauthorized `/start` at most once per bot lifetime (subsequent ones fall through to the silent `:noop`), while the on-disk store keeps the *code* stable across replies and restarts. The two invariants are tested separately so they can't drift.
- **A dedicated `PairingApprovalQueue`, not an extension of `DispatchResultQueue`.** It reuses the same seam — `QueueDirectory` (0700), atomic write-then-rename, versioned schema, reaper drain — but with its own `hive-pairing-approval` schema, keeping the dispatch-result schema honest.
- **The approval drain does *not* re-gate on the in-memory allowlist.** Unlike `drain_dispatch_results`, it must not drop the notice on an allowlist miss: the config write and the bot's SIGHUP reload are asynchronous, so the chat may not be live in memory yet. The drain trusts the owner-only 0700 queue plus a 1h expiry bound instead, so the reload race can't silently swallow the "approved" DM.
- **Bootstrap relaxation is scoped to emptiness.** `validate_bot_runtime!` permits an empty allowlist only when `pairing_enabled` is true; the token guard and the Integer-array shape check (`validate_bot_allowlist!`) are unchanged and still reject a missing token or a malformed allowlist.
- **Codes are chat_id-bound, 8 chars A–Z, 24h expiry.** Repeated `/start` returns the same pending code; cross-process access to `.bot.pairings.json` is guarded by an exclusive `flock` plus atomic write-then-rename.

## Agent surface

`hive pairing list` and `hive pairing approve telegram <CODE>` both accept `--json`, emitting `hive-pairing-list.v1` / `hive-pairing-approve.v1` envelopes. Exit codes: `0` success, `1` unknown/expired code (plus other runtime failures, disambiguated by the `--json` `error_kind`), `64` invalid arguments, `78` bad global bot config. Approval is idempotent (re-approving an already-allowlisted chat adds no duplicate) and degrades cleanly when the bot is down (config still persists; the notice waits, bounded by expiry).

## Test plan

- `test/integration/bot/pairing_flow_test.rb` exercises the end-to-end path with real file I/O and only the Telegram HTTP transport doubled: empty-allowlist boot → unauthorized `/start` reply → throttle → CLI approve → allowlist mutation on disk → approval-queue drain → "approved" DM → authorized message accepted after reload, plus the expired-code and unknown-code rejection paths.
- Unit tests cover the store lifecycle (`pairing_store_test.rb`), router classification (`router_test.rb`), the supervisor drain and its send-cap/prune ordering (`supervisor_test.rb`), the CLI command and its failure branches (`commands/pairing_test.rb`), the approval queue (`pairing_approval_queue_test.rb`), the pid-file guard (`pid_file_test.rb`), and the digest line (`digest/run_test.rb`).
- `rubocop` clean, `brakeman` clean (0 warnings), full suite green, and `rake coverage` at 100% line coverage.
- A true stranger-account E2E needs a second Telegram identity and token and is left to manual operator validation.
- Visual demo (`hive pairing` CLI/TUI) captured against an isolated offline state home; recordings live under the task's `media/`, but Screenote was not connected so no shareable URLs are available.

## Review summary

Three CE review passes reached `review_complete` (pass 3). Reviews were scoped to the branch's own merge-base diff (`03c3d34..HEAD`), since `origin/main..HEAD` is polluted by unrelated post-merge-base deletions.

Resolved across the passes (auto-fixes, all verified present):

- The once-per-chat throttle was re-gated on a dedicated *pairing-reply-sent* flag rather than the shared unauthorized-log gate, so a non-`/start` message before `/start` no longer suppresses the code.
- `approve` now consumes the code **last** (after the allowlist write and notice queue succeed) so a failed config write can't strand a consumed code.
- `PairingStore#pending` no longer writes on a pure read, and the digest count is guarded, so an I/O-faulted state dir can't crash the daily digest or `hive pairing list`.
- Code input is stripped and upcased on resolve; the drain got a per-tick send cap; the bot-reload signal is guarded against non-positive pids (no SIGHUP broadcast); transiently-unreadable approval notices are no longer deleted; and dead methods were removed.

**Open item (owner decision pending):** reviewers flagged, across all three passes, that DM-only is not actively *enforced* — the pairing path gates on `chat_id` + `/start` without a `chat_id == from_id` check, so a `/start` in a group the bot has joined would mint a code bound to the group's chat_id and let that group be approved. The plan states R1 "DM-only"; the brainstorm also states "zero group-chat handling." The escalation (`### A1.`) was left unanswered in all three passes, so the one-line DM guard was not applied. This is the one known gap a human reviewer should weigh before merge; the suggested fix is a `chat_id == from_id` guard in `Router#classify` before the pairing branch.

## Linked task

- Hive task **12119** — `i-need-the-same-pairing-260626-9dcb` ("Telegram Pairing Confirmation").
- Originated from a request to replicate OpenClaw's pairing UX: provide a bot token, press **/start**, receive a Telegram user id + pairing code, then `hive pairing approve telegram <CODE>` from the CLI.

---


ivankuznetsov:

## High
- [ ] Unauthorized pairing is not DM-only (`lib/hive/bot/router.rb:121`): group `/start` updates have a distinct `from_id`, but the pairing branch mints and approves by `chat_id` alone, so a group chat can be added to the allowlist despite R1/R15.
- [ ] The branch reverts unrelated released functionality (`lib/hive.rb:1`, deleted `lib/hive/commands/answer_digest.rb`): the diff downgrades Hive from 0.3.2 to 0.3.1 and removes answer-digest command/scheduler/schema files outside the pairing plan.

## Medium
- [ ] Bot JSON usage errors lost their envelope (`bin/hive:19`): removing the `bot` entry from `JSON_USAGE_ERROR_CONTRACTS` makes `hive bot status extra --json` exit 64 with empty stdout instead of the documented `hive-bot-status` error payload.
- [ ] `hv` can hang again without GNU `timeout` (`bin/hv:30`): the macOS/no-timeout fallback now probes candidates directly, reintroducing the blocking `--version` path that origin/main guarded with a watchdog.

## Nit
- [ ] `PairingStore#mint_or_get` does not persist pruning on reuse (`lib/hive/bot/pairing_store.rb:25`): it computes pruned entries but returns an existing code before `write_entries`, leaving expired sibling codes on disk until a later read.

ivankuznetsov:

# Code Review — Telegram pairing / approval flow (pass 1)

Reviewed the branch's own changes (`git diff 03c3d34..HEAD`), not `origin/main..HEAD`
— the branch is based on an old merge-base, so the `origin/main` diff is polluted with
thousands of unrelated deletions (features main added after the merge-base that this
branch simply lacks). Those were excluded. Verified: `rubocop` clean, `brakeman` clean
(0 warnings), all focused unit + integration tests green, schema-files test green.

## High
- [ ] Unauthorized `/start` can return **no** pairing code, violating R2/R4 (`lib/hive/bot/router.rb:124-130`): the throttle keys on `@unauthorized_logged`, which `authorized?` stamps on **every** rejected update (`router.rb:203`), not on "a pairing reply was already sent". So if a stranger sends any non-`/start` message first — or had any unauthorized contact earlier in the same bot lifetime — `already_unauthorized` is already true when `/start` arrives and it routes to silent `:unauthorized` instead of `:unauthorized_pairing`. The plan's own U3 prose says "the first `/start` replies"; the implementation only guarantees that when `/start` is the chat's *first ever* contact. Gate on a pairing-reply-sent flag instead.

## Medium
- [ ] `approve` consumes the code before the fallible allowlist write (`lib/hive/commands/pairing.rb:84-88`): `resolve_and_consume` deletes the pending entry (`pairing_store.rb:64`) at line 85, *then* `append_allowlist` writes config at line 86. If `update_global_config!` raises (read-only/full dir, malformed-on-disk config → `ConfigError`), the code is gone, the chat is never allowlisted, and no SIGHUP fires — so the in-memory throttle stays latched and the printed "ask the user to run /start again" is misleading (the next `/start` is throttled). Append-then-consume, or compensate the consume.
- [ ] `PairingStore#pending` is a query with an unguarded write side effect, widening the digest's failure surface (`lib/hive/bot/pairing_store.rb:43-49,116-128`): `pending` always calls `write_entries` (tmp+rename+fsync, no rescue), and `Digest.run` reads it via `pairing_store.pending.size` (`digest.rb:56`) under only `rescue ModelError` (`digest.rb:72`), while `Commands::Digest#call` rescues only `Hive::Error`/`ArgumentError`. An unwritable/full state dir (`Errno::EROFS`/`ENOSPC`/`EACCES`) now crashes the entire daily digest (and `hive pairing list`) — a path previously hardened to degrade gracefully. Write only when pruning changed something (mirror `resolve_and_consume`'s conditional write) and/or guard the digest read.

## Nit
- [ ] `resolve_and_consume` strips but does not upcase the code (`lib/hive/bot/pairing_store.rb:52`); a manually-retyped lowercase code reports "was not found" rather than matching the uppercase-minted key — and no test pins the strip/case/empty-string normalization.
- [ ] `drain_pairing_approvals` has no per-tick send cap (`lib/hive/bot/supervisor.rb:382-390`), unlike the sibling `drain_dispatch_results` (`DISPATCH_RESULT_SEND_CAP`) it mirrors; a backlog or post-reconnect burst could exceed Telegram's per-chat rate limit.
- [ ] Every SIGHUP reload (including each `pairing approve`) rebuilds the Router and clears `@unauthorized_logged` (`lib/hive/bot/router.rb:88`), so the documented "once per chat per bot lifetime" code throttle actually resets on every reload — weaker than stated (bounded only by 24h code reuse, so not an exhaustion vector).
- [ ] `pending` rewrites + fsyncs the backing file on every read, so a digest-only user with no bot configured gets `.bot.pairings.json` + `.bot.pairings.json.lock` created/rewritten on each `digest`/`pairing list` run (`lib/hive/bot/pairing_store.rb:46`) — surprising I/O from a count/list query.
- [ ] `approve --json` success `code` field echoes the raw CLI arg rather than the normalized stored code (`lib/hive/commands/pairing.rb:90,194`); a whitespace-padded-but-matching input would emit a `code` that fails the schema's `^[A-Z]{8}$` pattern.
- [ ] Test gaps for new branches: no test covers a non-`/start` message then `/start` (the High path), the consume-then-config-write-failure window (`pairing.rb:84-88`), the `PairingApprovalQueue.expired?` off-by-one boundary (age == `EXPIRY_SEC`), or the silent prune-on-miss arm of `resolve_and_consume` (`pairing_store.rb:60`).

ivankuznetsov:

# PR Review (pr-review-toolkit, multi-persona) — Telegram pairing / approval flow (pass 1)

Union of six personas: `[code]` code-reviewer, `[errors]` silent-failure-hunter,
`[tests]` pr-test-analyzer, `[type-design]` type-design-analyzer, `[comments]`
comment-analyzer, `[simplify]` code-simplifier.

Scoped to the branch's own changes (`git diff 03c3d34..HEAD`); `origin/main..HEAD`
is polluted by ~16k lines of unrelated deletions (features main added after the
old merge-base) and was excluded. Gates verified green by the personas: `rubocop`
clean, `brakeman` 0 warnings, `rake coverage` 100%, pairing tests pass. The
concurrency test was empirically confirmed load-bearing (removing the flock fails
it 30/30), so it is **not** flagged.

## High
- [ ] [errors] `PairingStore#pending` performs an unguarded write on a read path (`lib/hive/bot/pairing_store.rb:43-49,116-128`): on a read-only/full state dir the `SystemCallError` (EROFS/ENOSPC/EACCES) escapes `Digest.run` (rescues only `ModelError`, `digest.rb:56,72`) and `hive pairing list` (`Commands::Pairing#call` rescues only `Hive::Error` subtypes) — a previously read-only digest computation now crashes with a raw backtrace and no `--json` envelope.
- [ ] [errors] An unauthorized `/start` gives the user total silence if `mint_or_get`'s write fails (`lib/hive/bot/router.rb:344-348` → `pairing_store.rb:38`): the `SystemCallError` unwinds to the poll loop's per-update `rescue StandardError` (`supervisor.rb:251`) which only logs `:fatal` — no code, no "try again" reply.

## Medium
- [ ] [code] `approve` consumes the code before the fallible allowlist write (`lib/hive/commands/pairing.rb:85-86`): `resolve_and_consume` deletes the pending entry, then if `append_allowlist`→`update_global_config!` raises (ENOSPC/read-only) the code is lost, the chat is never allowlisted, no SIGHUP fires, and the once-per-lifetime throttle blocks the next `/start` — making the printed "run /start again" misleading.
- [ ] [errors] `approve`'s `PairingApprovalQueue.write!` and its `directory_for` `mkdir_p` have no rescue (`lib/hive/commands/pairing.rb:88`, `pairing_approval_queue.rb:24-48`): a `SystemCallError` after the code is consumed + chat allowlisted crashes the command with no envelope, queues no approval DM, and a retry reports "code not found".
- [ ] [errors] `PairingStore#load_entries` rescues `Errno::EACCES` → `{}` (`lib/hive/bot/pairing_store.rb:112-113`), silently masking a read-permission error as "no pending pairings" in `pairing list` and the digest — no log, no error surfaced.
- [ ] [errors] `signal_bot_reload` collapses "no/dead bot", "SIGHUP failed (EPERM)", and "unreadable pid-file" into one `reloaded: false` (`lib/hive/commands/pairing.rb:144-153`) that is merely omitted from output, so an operator can't distinguish a stale-config running bot from a clean skip; the `SystemCallError` rescue is also broader than the intended `Errno`.
- [ ] [type-design] `resolve_and_consume` returns a `Integer | :expired | :unknown` union and `resolve_code!`'s `case` folds every non-Integer/non-`:expired` value into the `else` "not found" arm (`pairing_store.rb:51-69`, `pairing.rb:118-129`) — non-exhaustive, so a future sentinel / `nil` / String silently maps to "code not found" rather than failing loudly.
- [ ] [tests] The two data-loss paths are untested: digest-crash-on-`pending`-write (the digest test injects a non-writing double) and the consume-then-config-write-failure window (only the *pre*-consume validation failure is covered) (`test/unit/digest/run_test.rb`, `test/unit/commands/pairing_test.rb:223`).
- [ ] [tests] The non-`/start`-then-`/start` ordering — which suppresses the pairing reply for the chat's lifetime because `authorized?` sets the gate on the first (junk) message — is untested; existing throttle/silence tests never chain a junk message into a `/start` from the same chat (`test/unit/bot/router_test.rb:122,140`).

## Nit
- [ ] [code] `PairingStore#pending` rewrites + fsyncs the file even when nothing was pruned (`lib/hive/bot/pairing_store.rb:43-49`); `resolve_and_consume` already guards its write with a length check (`:60`) and `pending` could too — every digest tick fsyncs the lock+file for zero expirations.
- [ ] [code] A successful first `/start` emits three log events including a misleading `:update_rejected_unauthorized` — `authorized?` logs it (`router.rb:200-203`) before the pairing branch emits `:pairing_code_issued` (`router.rb:344-348`); the `:update_received` skip only excludes plain `:unauthorized`.
- [ ] [type-design] `Notice#path` is populated but never read by the consumer (`pairing_approval_queue.rb:18`); removal re-globs + re-parses the dir by `notice_id` (`:89-101`) instead of using the path it already holds — either drop the field or use it for O(1) removal.
- [ ] [type-design] `Entry`/`Notice` Struct constructors enforce no invariants (code `[A-Z]{8}`, Integer `chat_id`, valid time live only in the parse/load filters, `pairing_store.rb:15`, `pairing_approval_queue.rb:18`); a `from_hash` factory would co-locate the invariant with the type.
- [ ] [type-design] `mint_or_get` writes whatever `chat_id` it is given with no type check (`pairing_store.rb:25-41`) while `load_entries` strictly drops non-Integer `chat_id` — a bad write would fail invisibly on the next read (asymmetric strictness).
- [ ] [tests] `PairingApprovalQueue.expired?` strict-`>` boundary (age == `EXPIRY_SEC` vs +1), `resolve_and_consume` normalization (strip is load-bearing; lowercase → `:unknown`; empty → `:unknown`), and the remaining digest pluralization crossings (singular on non-empty, plural on empty, non-empty+0 guard) are all unasserted (`pairing_approval_queue.rb:64-71`, `pairing_store.rb:51-53`, `digest/run_test.rb:203,220`).
- [ ] [comments] `wiki/operating.md:463-464` says approval "signals the running bot to reload when a bot PID file exists" — overstates: a stale/dead PID yields `reloaded: false` and no SIGHUP; sibling docs correctly say "live PID".
- [ ] [simplify] `pid_alive?` / `pid_file_payload` / the pid→live→HUP triad are duplicated between `Commands::Pairing` (`pairing.rb:144-175`) and `Commands::Bot` (`bot.rb:367-392`) — the plan flagged this; extract a shared liveness/pid-read helper while keeping each caller's corruption policy.
- [ ] [simplify] `approve` reads + parses the global config twice — `validate_current_bot_config!` discards its load (`pairing.rb:114-115`) and `bot_config` loads again (`:155-156`); have validation reuse the memoized `bot_config`.
- [ ] [simplify] `pairing_enabled?`'s `... ? true : false` is redundant since config validation already guarantees a boolean (`router.rb:356-358`); the bare `fetch` suffices.

ivankuznetsov:

## High
- [ ] Unauthorized pairing is still not DM-only (`lib/hive/bot/router.rb:125`): the router mints on any unauthorized `/start` using `chat_id` alone, and a group-style update with `chat_id != from_id` receives a code bound to the group chat id, violating R1/R15.

## Medium
- [ ] Expired pairings can still make read-only digest/list paths crash (`lib/hive/bot/pairing_store.rb:65`): `pending` now avoids no-op rewrites, but when it prunes an expired entry it still writes during a read, so `Digest.run`'s nonessential `pairing_store.pending.size` can raise `Errno::EACCES`/`ENOSPC` and fail the whole digest.

## Nit

ivankuznetsov:

# Code Review — Telegram pairing / approval flow (pass 2)

Reviewed the branch's own changes via the merge-base diff
(`git diff 03c3d34..HEAD`), not `origin/main..HEAD` — the branch sits on an old
merge-base, so the two-dot diff is polluted with thousands of unrelated deletions
(features main added after the merge-base). Pass-1's eight auto-fixes are verified
present in the current tree (reply-sent throttle flag, consume-last ordering,
conditional `pending` write, code normalization, drain send-cap, added tests).
Reviewers: correctness, security, adversarial, reliability, testing, cli-readiness,
maintainability. The DM-only High is the unresolved pass-1 escalation (Q1), re-confirmed
here with a concrete exploit path; the digest-crash Medium is corroborated by
correctness + reliability + the codex pass-2 review.

## High
- [ ] DM-only is not enforced — unauthorized `/start` mints, replies, and allowlists by `chat_id` alone with no `chat_id == from_id` guard (`lib/hive/bot/router.rb:132,361`): a `/start` in any group the bot has joined posts the requester's Telegram id + pairing code **publicly into that group** and mints a code bound to the group's (negative) `chat_id`; an owner running `hive pairing approve telegram <CODE>` then appends that **group** `chat_id` to `bot.chat_id_allowlist`, authorizing every member of the group to drive hive (`/run`, `/approve`, `/idea`…). This contradicts plan R1 ("DM-only") and the brainstorm's own "no code path consumes [groups]" invariant — which held before this feature (group `/start` was a silent `noop`). `Telegram::Update` already carries `from_id` (in a DM `chat_id == from_id`; in a group they differ), so the guard is one line. **OPEN ESCALATION from pass 1 (Q1) — owner answer still pending.**

## Medium
- [ ] `PairingStore#pending` performs a write during a read and that write can crash the daily digest (`lib/hive/bot/pairing_store.rb:65`, `lib/hive/digest.rb:56`): when an expired entry exists, `pending` calls `write_entries` (tmp+rename+fsync, no `SystemCallError` rescue); on a read-only/full state dir that raises `Errno::EROFS/ENOSPC/EACCES`, which is neither `ModelError` (the only rescue in `Digest.run`, digest.rb:72) nor `Hive::Error`/`ArgumentError` (`Commands::Digest#call`), so a previously-hardened-to-degrade path aborts the whole digest over a purely informational count. The same unguarded `@store.pending` in `hive pairing list` (`lib/hive/commands/pairing.rb:63`) emits a raw backtrace instead of a clean envelope. Make `pending` a pure read (prune only on write paths) or guard both call sites with `rescue SystemCallError, IOError`.
- [ ] The new `pairing` command is absent from `JSON_USAGE_ERROR_CONTRACTS` (`bin/hive:19`), so a Thor-layer usage error (e.g. an unknown flag on `hive pairing list --json`) escapes without the `hive-pairing-*` envelope the command's `long_desc` promises for exit 64 — breaking the `--json` machine-parseable-error guarantee for an agent-facing surface. Add `"pairing" => { schema: "hive-pairing-approve", error_kind: "usage" }`, mirroring the existing `"digest"` entry. (`bin/hive` is not in this branch's diff; apply on rebase onto current main.)

## Nit
- [ ] `write_approval_notice!` rescues `SystemCallError` but not `IOError` (`lib/hive/commands/pairing.rb:150`): an `IOError` from `write!`'s `file.write/flush/fsync` would bypass the clean `ApprovalError` envelope and the consume-last retry contract. Near-unreachable in practice, but widen to `rescue SystemCallError, IOError` for parity with the documented behavior.
- [ ] `PairingStore#resolve_and_consume` has no production callers (`lib/hive/bot/pairing_store.rb:109`) — `approve` deliberately uses `resolve` + `consume`; the method is exercised only by its own tests. Remove per YAGNI (drops its tests from the coverage surface too), or note the intended near-term caller in the comment.
- [ ] Duplicate/concurrent `hive pairing approve telegram <CODE>` double-sends the "✅ Approved" DM (`lib/hive/commands/pairing.rb:92-96`): the store lock is released between `resolve` (92) and `consume` (96), so two overlapping invocations both resolve the still-pending code and both `write!` a notice, and the reaper sends two DMs. Single-operator + harmless payload, so low impact; dedupe the approval queue by `chat_id` at `write!` time (or claim the code atomically) if hardening.
- [ ] `ApprovalError` exit codes are undifferentiated (`lib/hive/commands/pairing.rb:17`): `notice_write_failed` (retryable) and `internal_error` (not) both inherit exit `1`, identical to unknown/expired-code — yet the `long_desc` documents `1` only for unknown/expired, so an agent cannot branch retry behavior on the exit code. Map software/transient kinds to distinct codes (e.g. 70/75) and document them.
- [ ] Test/coverage hardening (several new branches under-asserted): `test_pending_does_not_rewrite_when_nothing_is_pruned` spies on the private `write_entries` (brittle on rename — assert on file mtime instead) (`test/unit/bot/pairing_store_test.rb:148`); `test_approve_is_idempotent_for_already_allowlisted_chat` never asserts the code was consumed or a notice queued (`test/unit/commands/pairing_test.rb:211`); `test_approve_notice_write_failure_…` never asserts the allowlist was persisted in the partial-success window (`pairing_test.rb:428`); no test covers `approve telegram <CODE> <extra>` (the `extra.empty?` false arm, pairing.rb:75) nor `Digest.run` when `pairing_store.pending` raises `SystemCallError`.

ivankuznetsov:

# PR Review Toolkit — Telegram pairing / approval flow (pass 2)

Multi-persona review (code, tests, errors, comments, type-design, simplify) over the
**merge-base diff** (`git diff 03c3d34..HEAD`) — `origin/main..HEAD` is polluted by an
old merge-base. Pass-1's eight auto-fixes are verified present. Findings below are the
union across personas, deduped, with the surfacing persona prefixed. Two persona claims
were corrected during synthesis and dropped: (a) the bot pid file stores `started_at` as
an ISO **string** (bot.rb:81), so `PidFile.read` does **not** break the wired reload path;
(b) the approval-drain's missing allowlist re-gate is the plan's **deliberate Q-D decision**
("send trusting 0700-dir provenance"), so it is not flagged as a defect — only the comment
that oversells "mirroring" is kept.

## High
- [ ] [comments] DM-only is asserted but not enforced (`lib/hive/cli.rb:1204` long_desc "an unknown Telegram **DM** sends /start"; `lib/hive/bot/router.rb:125-134,361`): the router gates pairing on `chat_id` + `start_command?` only, and `Telegram::Update` carries no chat-type, so a `/start` from a non-allowlisted group mints a code (posted publicly) and lets a group `chat_id` be approved into `bot.chat_id_allowlist` — contradicting plan R1; add a `chat_id == from_id` DM guard (re-confirms the open pass-1 escalation Q1, still pending an owner answer).

## Medium
- [ ] [errors/code/tests] `Digest.run` can crash the entire daily digest on an I/O-faulted state dir (`lib/hive/digest.rb:56`): `pairing_store.pending.size` runs outside the only rescue (`ModelError`, digest.rb:72), and `PairingStore#pending` opens `<file>.lock` with `CREAT` (pairing_store.rb:132) and may `write_entries` on prune (pairing_store.rb:65) — raising `Errno::EROFS/ENOSPC/EACCES` on a read-only/full dir; `Commands::Digest#call` only rescues `Hive::Error`, so a previously-hardened path now aborts (and the daemon hot-loops it). Guard the count (`… rescue 0`) or make `pending` non-raising on its own write failure. (silent-failure-hunter: Critical.)
- [ ] [errors] `@store.consume` is unguarded after the side effects already landed (`lib/hive/commands/pairing.rb:96`): on a read-only/full dir `consume → write_entries` raises `SystemCallError`, which escapes `call`'s rescue (only InvalidTaskPath/ApprovalError/ConfigError) as a raw backtrace **after** allowlist+reload+notice succeeded — the chat is functionally approved but the operator sees a crash, no success envelope, and retries into `unknown_code`. Wrap consume; the cleanup failure must not mask a completed approval.
- [ ] [errors] `signal_bot_reload` collapses "signal failed" into a silent `false` (`lib/hive/commands/pairing.rb:171-180`): a **live** bot whose SIGHUP fails for a real reason (`Errno::EPERM`, transient `IOError`) is indistinguishable from "bot not running", so the operator gets no warning and never learns the live bot kept serving the **stale** allowlist (the freshly-approved chat stays locked out until restart). Separate genuine absence from a failed signal and surface the latter.
- [ ] [errors] `pid_file_payload` corruption handling diverges from its sibling (`lib/hive/commands/pairing.rb:186-190`): it rescues `Psych::Exception → {}` (treats a corrupt `.bot.pid` as "bot down" and silently skips the reload), directly contradicting `commands/bot.rb:376-381` which now `warn`s + raises "refusing to assume bot state" on the same corruption. Mirror bot.rb so a corrupt pid file surfaces instead of silently skipping live reload.
- [ ] [errors] `write_approval_notice!` rescues `SystemCallError` but not `IOError` (`lib/hive/commands/pairing.rb:150`): `PairingApprovalQueue.write!`'s `file.write/flush/fsync` (pairing_approval_queue.rb:39-44) can raise `IOError`, which escapes the clean `ApprovalError` envelope as a raw backtrace after the chat was allowlisted. Broaden to `rescue SystemCallError, IOError`.
- [ ] [comments] `long_desc` exit-code table is misleading (`lib/hive/cli.rb:1206-1210`): it documents exit `1` as only "unknown/expired code", but `notice_write_failed` (reachable) and `internal_error` (defensive) raise `ApprovalError` with no `exit_code` override and so also exit `1` — an agent scripting on the exit code mis-classifies a notice-write failure as a bad code. Reword to note `1` is a catch-all and to branch on the `--json` `error_kind`.
- [ ] [type-design] `PidFile`'s two readers have silently divergent YAML contracts (`lib/hive/pid_file.rb:30-35` vs `:74-86`): module `PidFile.read` uses `YAML.safe_load` **without** `permitted_classes: [Time]`, while instance `read_pid_file_payload` permits `Time`; a daemon pid file embeds a `Time`, so `PidFile.read` pointed at it raises `Psych::DisallowedClass`. The wired pairing path is safe (bot pid file uses a string), but the near-identical names + incompatible parsers are a latent trap — give `PidFile.read` the same `permitted_classes: [Time]`.
- [ ] [type-design/simplify] `PairingStore#resolve_and_consume` is dead public API (`lib/hive/bot/pairing_store.rb:109-113`): no production caller (approve deliberately uses `resolve`+`consume`), exercised only by its own tests, and non-atomic (two separate `with_lock`s) despite its "atomic" name. Remove it (and `prune_expired!`, also test-only, pairing_store.rb:115) to shrink the surface and the coverage burden.
- [ ] [type-design] the `resolve`/`consume` two-phase contract is unenforced by the type (`lib/hive/bot/pairing_store.rb:75-104`): `consume(code:)` will delete a code the caller never `resolve`d (or resolved as `:expired`), and the "fallible work then consume" invariant lives only in `approve`'s prose. Return a small lease/handle from `resolve` whose `#consume` is the only happy-path entry point, or at minimum document the protocol on `consume` itself.

## Nit
- [ ] [code] Expired-code dead-end (`lib/hive/bot/router.rb:132`): `@pairing_reply_sent` is a lifetime latch that never resets, so a legitimate user whose code expired (24h) cannot obtain a fresh one via `/start` within the same bot lifetime — the second `/start` routes to silent `:unauthorized` although `mint_or_get` would mint a new code; stuck until restart. Matches the documented R5 throttle, but consider gating on whether a *non-expired* code is still pending for the chat.
- [ ] [code/simplify] `signal_bot_reload`'s rescue list is both over- and under-broad (`lib/hive/commands/pairing.rb:178`): `Psych::Exception`/`IOError` are already swallowed by `pid_file_payload → {}` (unreachable here), while a non-integer pid (YAML `pid: "4242"`) makes `Process.kill(0, "4242")` raise an **un-rescued** `TypeError` after the allowlist write. Drop the dead arms and add `TypeError` (or coerce via `Integer(pid, exception: false)`).
- [ ] [errors] `drain_pairing_approvals` can re-send the "✅ Approved" DM every tick (`lib/hive/bot/supervisor.rb:393-396`): the `remove_pairing_approval` return is ignored and `remove_known_path` rescues only `Errno::ENOENT`, so a successful send + failed unlink (`EACCES`) leaves the notice and re-sends until expiry (1h). Check the remove result, log a distinct event, and rescue `EACCES` to a clean `false`.
- [ ] [errors] `load_entries` masks a corrupt pairings file as "no pending" (`lib/hive/bot/pairing_store.rb:155-157`): `hive pairing list` prints "No pending pairing requests." for a merely garbled/unreadable file, so the owner can't distinguish empty from broken. Log a single `warn`/event on the rescue while keeping the `{}` in-memory fallback.
- [ ] [comments] The `(R3)` tag is reused for two distinct latches (`lib/hive/bot/router.rb:365` pairing-reply latch vs `:208/:240` unauthorized-log throttle): both claim "R3", inviting a reader to conflate the mechanisms, and the out-of-repo R-labels will rot. Cite the specific sub-requirement or the guaranteed behavior instead.
- [ ] [comments] "Per-tick send cap **mirroring** drain_dispatch_results" oversells the parallel (`lib/hive/bot/supervisor.rb:390`): unlike the sibling, the pairing drain has no overflow-summary and no allowlist re-check (the latter is the deliberate Q-D design), so a reader trusting "mirroring" assumes parity that isn't there. Tighten to "mirroring the send cap (no overflow-summary; over-cap notices stay queued)".
- [ ] [tests] No test for `approve telegram <CODE> <extra>` (`lib/hive/commands/pairing.rb:75-79`): the `extra.empty?` false arm is uncovered, so a refactor dropping `*extra` would silently accept trailing garbage. Add a test asserting `InvalidTaskPath` (usage) with no allowlist mutation / notice write.
- [ ] [tests] No test for the two newly-flagged failure branches: `Digest.run` when `pairing_store.pending` raises `SystemCallError` (digest-crash contract), and `drain_pairing_approvals` when the send succeeds but `remove` returns false/raises (duplicate-DM contract). Add both to pin behavior the 100%-line gate doesn't exercise.
- [ ] [type-design] `resolve` returns a `Integer | :expired | :unknown` symbol union with no enforced exhaustiveness (`lib/hive/bot/pairing_store.rb:75-90`); the hand-written `else` "programming error" arm in `resolve_code!` (pairing.rb:139-145) is the tell. Optional: result objects / sentinel constants if the pattern recurs.
- [ ] [type-design] `PairingApprovalQueue::Notice#path` is a producer-populated fast-delete hint threaded back through `remove(..., path:)` (`lib/hive/bot/pairing_approval_queue.rb:18,85-91`) — a mild leaky abstraction (a mismatched id/path is only caught silently by `remove_known_path`'s re-verify). Consider `remove(notice)` so the pair can't be mismatched by construction.
- [ ] [type-design] `Entry`/`Notice` are mutable `Struct`s returned from read APIs (`lib/hive/bot/pairing_store.rb:15`, `pairing_approval_queue.rb:18`); a consumer can mutate what looks like a point-in-time snapshot. Freeze on construction or use `Data.define` for immutable value semantics.
- [ ] [simplify] Inline `validate_current_bot_config!` (`lib/hive/commands/pairing.rb:85,122-127`): its only effect is forcing the memoized `bot_config` load (which `signal_bot_reload` re-triggers anyway), and the name overpromises "validation" beyond what `load_global_bot` already does. Replace the call with a direct `bot_config` and drop the method.

ivankuznetsov:

## High
- [ ] Unauthorized pairing is still not DM-only (`lib/hive/bot/router.rb:132`, `lib/hive/bot/telegram.rb:17`): the router mints on any unauthorized `/start` using `chat_id` alone, so a group-style update where `chat_id != from_id` can receive a public pairing code and approval would append the group chat id, violating plan R1/R15.

## Medium

## Nit

ivankuznetsov:

# Code Review — Telegram pairing / approval flow (pass 3)

Reviewed the branch's own changes via the merge-base diff
(`git diff 03c3d34..HEAD`), not `origin/main..HEAD` — the branch sits on an old
merge-base, so the two-dot diff is polluted by unrelated deletions (features main
added after the merge-base; those were correctly suppressed in pass 1). Pass-1 and
pass-2 fixes are verified present (reply-sent latch, consume-last ordering, digest
count guard, drain send-cap + failed-unlink event, code normalization, struct
freeze, exit-code doc, dead-method removal). Reviewers: correctness (opus),
adversarial/security (opus), testing (sonnet); grounded by rubocop (clean on all 6
changed Ruby files) and the codebase's own pid-guard convention.

Focus of this pass was **regressions in the pass-2 hardening commits** and the one
persistent open escalation. Two genuinely NEW defects surfaced (both corroborated
by ≥2 reviewers); the DM-only High re-confirms with a fresh exploit path and
matches codex-ce-code-review-03. Deliberate plan scope boundaries (spam/flood
acceptable per Risks/A6; once-per-lifetime throttle per R5; concurrent-approve
duplicate DM per single-operator A7) were checked and left unflagged.

## High
- [ ] ESCALATE: DM-only is still not enforced — unauthorized `/start` mints/replies/allowlists by `chat_id` alone with no `chat_id == from_id` guard (`lib/hive/bot/router.rb:125,132-133,363`; `Telegram::Update` carries `from_id` at `lib/hive/bot/telegram.rb:17`): a `/start` (or `/start@botname`, which matches `start_command?`'s `\A/start\b`) in any group the bot has joined posts the requester's id + pairing code **publicly into the group** and mints a code bound to the group's negative `chat_id`; `hive pairing approve telegram <CODE>` then appends that **group** `chat_id` to `bot.chat_id_allowlist`, authorizing every group member to drive hive (`/run`, `/approve`, `/idea`). Contradicts plan R1 ("DM-only") — a plan requirement, not a deferred non-goal — and the brainstorm's "no code path consumes [groups]" invariant that held before this feature. One-line fix (`chat_id == from_id` guard in `classify` before the pairing branch). **OPEN ESCALATION from passes 1 & 2 (Q1) — owner `A1.` still blank; re-raised independently by correctness, adversarial, and codex-03.** 

## Medium
- [ ] `signal_bot_reload` can broadcast SIGHUP to the whole process group (or all user processes) on a corrupt/hand-edited pid file (`lib/hive/commands/pairing.rb:200`): the guard is `return false unless pid && pid_alive?(pid)` with no `pid > 0` check, so a pid file holding `pid: 0` / `pid: -1` / `pid: "0"` coerces to a truthy `0`/negative, `PidFile.alive?(0)` → `Process.kill(0, 0)` does **not** raise `ESRCH` (it probes the caller's process group, `lib/hive/pid_file.rb:17`), and `@process.kill("HUP", 0)` then signals every process in the group (`-1` → every process the user can signal) — after the allowlist write, on a normal `hive pairing approve`. Every other pid consumer guards this exact case (`pid_file.rb:46`, `babysit.rb:348,373`, `daemon.rb:350,439` all use `pid && pid > 0`); this NEW method's own comment claims to defend hand-edited pid files (the `pid: "4242"` string case) but omits `pid > 0`. Corroborated by correctness + testing. Fix: `return false unless pid && pid > 0 && pid_alive?(pid)`. 
- [ ] Transient read errors delete a valid approval notice before the "✅ Approved" DM is sent (`lib/hive/bot/pairing_approval_queue.rb:150`, `lib/hive/bot/supervisor.rb:381`): `parse_file` folds `Errno::EACCES`/`IOError` into the same `:malformed_json` sentinel as genuine `JSON::ParserError`, and `drain_pairing_approvals`'s `bad_handler` unconditionally `FileUtils.rm_f`s it — so a structurally-valid notice that hits a transient FS fault (perm-locked/NFS-hiccup dir) is permanently deleted; the chat is already allowlisted (approval persists) but the confirmation DM is lost with no retry. Real-world impact is low (informational DM only; `write!`'s tmp+atomic-rename makes the trigger rare and non-attacker-controlled), but it is a silent drop of owner-authored state that contradicts this same PR's "surface, don't silently drop" hardening of `PairingStore#load_entries` (now `warn`s). Corroborated by adversarial + testing. Fix: split transient I/O errors out of the corruption arm so the drain leaves the notice for a later tick instead of deleting it. 

## Nit
- [ ] No test exercises `signal_bot_reload` with a non-positive pid (`test/unit/commands/pairing_test.rb`): every case pins `pid: 4242` or forces a nil pid, so the missing `pid > 0` guard (Medium above) has no catching test — the `FakeProcess` records all `kill` calls, so a `{ "pid" => 0 }` fixture asserting `process.kills` never targets a non-positive pid would pin it. 
- [ ] `drain_pairing_approvals`'s send-cap + expired-prune interaction is untested (`test/unit/bot/supervisor_test.rb`): one test writes only fresh over-cap notices, another only a single expired notice — none mixes expired (oldest, pruned first) + fresh where fresh exceeds `PAIRING_APPROVAL_SEND_CAP`. The code is correct (prune before slice), but a regression that capped before pruning would waste cap slots on stale notices and starve fresh approvals for a tick with no failing test.

ivankuznetsov:

# PR Review Toolkit — Telegram pairing / approval flow (pass 3)

Multi-persona review (code, silent-failure, tests, type-design, comments, simplify)
over the **merge-base diff** (`git diff 03c3d34..HEAD`) — `origin/main..HEAD` is
polluted by an old merge-base. Pass-1/pass-2 fixes verified present; rubocop clean on
all changed Ruby files. Findings are the union across personas, deduped, prefixed by
the surfacing persona. Three items corroborate orchestrator-held findings (DM-only,
`pid>0`, transient-notice-delete); the rest are NEW this pass. Deliberate plan scope
boundaries (spam/flood per Risks/A6, once-per-lifetime throttle R5, concurrent-approve
dup DM A7, Q-D no-allowlist-re-gate, telegram fixed literal, single `pairing_enabled`
boolean) were checked and left unflagged.

## High
- [ ] [silent-failure] The once-per-chat pairing-reply latch is committed before the reply is delivered (`lib/hive/bot/router.rb:363-370` sets `@pairing_reply_sent[chat_id]` right after `mint_or_get`, but the reply is only *returned* and later sent by `supervisor.rb` `execute_result` → `safe_send_message`, which rescues `StandardError`, logs `:send_failure`, and returns `nil` with no feedback to the router): any swallowed Telegram send failure (transient network/`IOError`, HTTP 429 rate-limit, 403 "bot blocked", chat-not-found race) leaves the latch set and `:pairing_code_issued` logged, so the user never receives their code AND every later `/start` is silently `:unauthorized` → `noop` until a SIGHUP reload/restart clears the latch — while `pairing list` + the digest still say the code was "issued". Distinct from the suppressed mint-*write*-failure item (that raises pre-latch as `:fatal`; this one succeeds, latches, then loses delivery). Fix: set the latch only after a confirmed send (or clear it when the send returns `nil`), or document the latch as delivery-independent by design. 
- [ ] [code, comment, tests] DM-only is not enforced — the router mints/replies/allowlists on `chat_id` + `/start` alone with no `chat_id == from_id` guard (`lib/hive/bot/router.rb:125,132-133,363`; `Telegram::Update` carries `from_id` at `telegram.rb:17`, never parses `chat.type`): a `/start` (or `/start@botname`) in any group the bot joined posts the pairing code **publicly** into the group and lets the group's negative `chat_id` be approved into `bot.chat_id_allowlist`, authorizing every group member to drive hive. Contradicts plan R1 ("DM-only") — a requirement, not a deferred non-goal. Independently re-surfaced by code, comment, and test personas (and codex-03). One-line `chat_id == from_id` DM guard. **OPEN ESCALATION from passes 1 & 2 (Q1) — owner `A1.` still blank.** 

## Medium
- [ ] [silent-failure] The expired-notice branch of `drain_pairing_approvals` discards the removal result (`lib/hive/bot/supervisor.rb:378-386`): `remove_pairing_approval(notice)` is called and its `false` return ignored, so an expired notice whose unlink keeps failing (the clean `false` the queue now returns on `EACCES`, or a mismatched/corrupt file) is silently re-evaluated-and-re-dropped every reaper tick (1 Hz) with zero observability — the exact "quietly wedged" class the successful-send branch was just hardened against (b8465749 added `:pairing_approval_notice_unremovable`). Mirror that fix: check the result and log the event in the expired branch too. 
- [ ] [silent-failure, code, tests] Transient read errors delete a valid approval notice before the "✅ Approved" DM is sent (`lib/hive/bot/pairing_approval_queue.rb:150` folds `Errno::EACCES`/`IOError` into the same `:malformed_json` sentinel as genuine `JSON::ParserError`; `drain_pairing_approvals`'s `bad_handler` at `supervisor.rb:381` unconditionally `FileUtils.rm_f`s it): a structurally-valid notice hit by a transient FS fault is permanently deleted — the chat is already allowlisted (approval persists) but the confirmation DM is lost with no retry. Low real-world likelihood (atomic tmp+rename makes it rare, informational-only impact) but a silent drop of owner state that contradicts this PR's own `load_entries` hardening. Fix: separate transient I/O from corruption so the drain leaves the notice for a later tick. 
- [ ] [code, tests] `signal_bot_reload` lacks the codebase-wide `pid > 0` guard (`lib/hive/commands/pairing.rb:200`): `Integer("0")` is truthy, so a corrupt/hand-edited `pid: 0`/`-1`/`"0"` passes `pid && pid_alive?(pid)` (`PidFile.alive?(0)` → `Process.kill(0, 0)` does not raise `ESRCH`), and `@process.kill("HUP", 0)` then broadcasts SIGHUP to the whole process group (`-1` → all user processes) — after the allowlist write, on a normal `hive pairing approve`. Every sibling guards this (`pid_file.rb:46`, `babysit.rb:348,373`, `daemon.rb:350,439`); this new method's own comment claims to defend hand-edited pid files but omits `pid > 0`. Fix: `return false unless pid && pid > 0 && pid_alive?(pid)`. 
- [ ] [comment] The "DM" wording overstates the enforced trigger (`lib/hive/cli.rb:1149` long_desc: "when an unknown Telegram **DM** sends /start"; `templates/hive_config.yml.erb:14`: "unknown **DM** /start requests get pairing codes"): the code applies no chat-type / `chat_id == from_id` check, so any non-allowlisted chat (including a group) triggers a code. This is the doc-accuracy angle of the DM-only High — even if the code guard is added separately, drop/qualify "DM" so the docs don't describe behavior the code doesn't guarantee ("an unknown Telegram chat not on the allowlist"). 

## Nit
- [ ] [simplify] Dead method `PairingApprovalQueue.prune_expired` has no production caller (`lib/hive/bot/pairing_approval_queue.rb:189-199`): the bot's `drain_pairing_approvals` prunes inline (`pending` + `expired?` + `remove(path:)`) and, unlike the sibling `DispatchResultQueue.prune_expired`, nothing schedules a prune tick for this queue — only its own unit test references it. Same dead-code class the prior sweep removed from `PairingStore` (`resolve_and_consume`/`prune_expired!`), missed here. Remove it + its test, or accept it as sibling-parity dead weight. 
- [ ] [type-design] `PidFile` now has a split personality (`lib/hive/pid_file.rb`): the new stateless `self.read`/`self.alive?` duplicate the mixin's instance `read_pid_file_payload`/`pid_alive?`, and the parse logic has **already drifted** (the instance version accepts a legacy bare-integer file; `self.read` does not). `read_live_pid` still calls the instance `pid_alive?` while `Commands::Bot#pid_alive?` now delegates to `PidFile.alive?`, so a future EPERM/ESRCH change must touch both or they diverge. Fix: have the instance methods delegate to the module functions to collapse the duplicated logic. 
- [ ] [tests] No test drives `signal_bot_reload` with a non-positive pid (`test/unit/commands/pairing_test.rb` — every case uses `4242` or a nil-yielding file): a `{ "pid" => 0 }` fixture asserting `reloaded == false` and that the fake process received no `HUP` would catch the missing `pid > 0` guard (fails today). 
- [ ] [tests] The `drain_pairing_approvals` send-cap × expired-prune interaction is untested (`test/unit/bot/supervisor_test.rb:2336-2369` — one test is all-fresh over-cap, one is a single stale): the load-bearing property is that expired notices sorted ahead of fresh ones are pruned *before* `fresh.first(CAP)`, so a cap-before-prune regression would burn cap slots on stale notices and starve fresh approvals with no failing test. Add a mixed expired(oldest)+fresh over-cap drain asserting exactly CAP fresh DMs and the expired ones pruned without a send. 
- [ ] [tests] The transient-read-error path in `PairingApprovalQueue.parse_file` is untested (`test/unit/daemon/pairing_approval_queue_test.rb` only covers genuine `{not json` corruption): force `File.read` to raise `Errno::EACCES` on a well-formed notice and assert it is not silently discarded/deleted — pins the Medium data-loss branch the line gate can't see. 
- [ ] [tests] The operator-facing "live bot, SIGHUP failed — restart the bot" warn is executed but never asserted (`lib/hive/commands/pairing.rb:213-214`; `test/unit/commands/pairing_test.rb:271` drives the branch but only checks `reloaded == false`): a refactor dropping the `warn` would keep the line covered and every test green while silently stranding operators on a stale allowlist. Wrap in `capture_io`/`assert_output` and assert the message reaches stderr. 
- [ ] [comment] The long_desc "requests a live bot reload when a bot PID is present" understates the liveness requirement (`lib/hive/cli.rb:1150`): `signal_bot_reload` requires the pid to be parseable AND alive — a present-but-dead pid triggers no reload. Reword to "when a live bot PID is present". 
- [ ] [comment] The `PairingStore#pending` comment implies skipping the write alone fixed the read-only/full-dir crash (`lib/hive/bot/pairing_store.rb:60-66`), but `pending` still enters `with_lock`, which opens `<path>.lock` with `RDWR|CREAT` and can itself raise on such a dir (the callers correctly still rescue it). Add a half-sentence noting the lock open remains a raise point. 
- [ ] [type-design] `Notice`'s `chat_id`/`created_at` invariants are enforced only in `parse_file`; the struct constructor and `write!` accept anything (`lib/hive/bot/pairing_approval_queue.rb:129-134,140`). Same asymmetric-strictness class as the already-suppressed `mint_or_get`/`load_entries` decision (every `write!` caller passes an Integer in practice). Flagged only to confirm the same no-fix ruling applies here. 
- [ ] [code] `Digest.append_pairing_notice` appends raw, un-escaped text to a message the `Renderer` otherwise escapes before sending with `parse_mode: markdown_v2` (`lib/hive/digest.rb`): the current static string (backtick code span + em-dash) is valid MarkdownV2 so there is no bug today, but editing it to include a reserved char (`.`, `-`, `!`, `+`, `=`) outside the code span would 400 the whole digest send. Latent fragility only.

ivankuznetsov:
## Architecture review (user + best-practices lens)

Verdict: **this is how a bot feature should be built** — after three CE review passes there was genuinely little left at the architecture level. One structural fix is stacked as #645; the rest of this comment records what holds up and a few small observations that don't warrant code churn.

### What holds up well

- **The two-process contract is exactly right.** The bot mints codes; the tokenless CLI consumes them, mutates config through the existing locked writer, and hands the DM back to the running bot via a file queue. No shared memory, no HTTP between them, SIGHUP only as a hint — every coordination point is a file with a schema.
- **`resolve` (peek) → fallible work → `consume` (delete last)** in `Commands::Pairing#approve` is a textbook two-phase protocol, and the store documents it as a contract ("approve is the reference implementation"). The consume-failure policy (warn, let the success envelope through, let the code expire) is the right call and the comment explains *why*.
- **Failure-policy discipline.** Every rescue is scoped and justified: `:io_error` vs `:malformed_json` in the queue so a transient read fault can't feed a deleting drain; the unremovable-notice log event so a wedged notice is attributable; the pid > 0 guard so a hand-edited `pid: 0` can't SIGHUP-broadcast. The `PidFile.alive?`/`read` extraction (module-level stateless vs instance ownership-aware, with the non-wrapper rationale documented) is a nice piece of API design in its own right.
- **Scope discipline.** DM-only, Telegram-only, `telegram` as a fixed literal with `deny`/`revoke`/web-UI as explicit non-goals — and the bootstrap relaxation scoped to *emptiness* only when `pairing_enabled`.

### Fixed in #645

- The feature adds atomic-write implementations **six and seven** to a codebase that already has five (`Markers`, `Events`, `Init`, `ServiceInstaller::Base`, review suppression). #645 introduces `Hive::AtomicFile.write` and points the two new writers (`PairingStore#write_entries`, `PairingApprovalQueue.write!`) at it — identical on-disk behavior. Migrating the legacy five is left as a follow-up pass.

### Observations, no change requested

- **Error-envelope schema for pre-dispatch usage errors.** `bin/hive`'s `JSON_USAGE_ERROR_CONTRACTS` maps all `pairing` argv-shape errors to `hive-pairing-approve`, while an in-command unknown subcommand emits under `hive-pairing-list` when `@subcommand` isn't "approve". Automation branching on `schema` for usage errors could see either. Harmless today (both carry `error_kind`), just worth knowing.
- **`Digest` now depends on `Bot::PairingStore`** (with injection, and the count is fail-soft). Acceptable coupling for a one-line notice; if digest grows more bot-derived lines, an interface for "digest contributors" would be the moment to invert it.
- **Router pairing logic stays inline** (~35 lines, two latches). I considered a `Handlers::PairingHandler` for symmetry with Slash/Callback handlers and rejected it: one intent, no conversation state — extraction would be ceremony. Worth revisiting only if deny/revoke ever land.

Stacked refactor: #645.

PR #643, opened 2026-07-01
docs(openclaw): add task watch recipe to hive skill

ivankuznetsov:
## Summary

Hive operators can now wait on selected in-flight tasks and be notified *only* when a task's state actually changes — without ever touching Hive state. A new `## Watch Selected Tasks` recipe in the hive skill (`openclaw/skills/hive/SKILL.md`) documents a read-only, copy-pasteable `bash`/`jq` loop over `hive status --json`; there is no new CLI command and no library change.

The recipe builds a composite state key from the meaningful fields (`stage`, `action`, `marker`, `attrs.phase`, `attrs.pass`, `claude_pid_alive`, held `retry_after`, `pr_url`) and deliberately excludes noisy churn (`mtime`, `folder_mtime`, `age_seconds`) so it reports transitions, not heartbeat. It surfaces PR-URL and `7-artifacts`/`8-finalize` stage transitions, exits early when every tracked slug is terminal, and otherwise prints a final per-slug snapshot on timeout.

The recipe guarantees:

- **Read-only.** Only `hive status --json` reads plus a local `mktemp`/`rm -f` snapshot file. `Ctrl-C is safe`; the loop **never kills Hive agents**, **never clears markers**, and **never advances stages**.
- **Selectable + safe defaults.** Positional args pick slugs; with none, it tracks all non-`9-done` tasks. `HIVE_WATCH_INTERVAL` (default 15s) and `HIVE_WATCH_TIMEOUT` (default 1800s) are env-overridable. A `trap` removes the temp snapshot on exit or interrupt.
- **Orphan cleanup hint.** Left-running loops can be found with `pgrep -af 'hive status --json'` or the shell `jobs` table.

## Test plan

- `LANG=C.UTF-8 ruby -Itest test/unit/openclaw_skills_test.rb` — green (new + existing methods). Added `test_umbrella_skill_documents_watch_selected_tasks`, which pins the section's literal anchors (heading, env knobs, `retry_after`, `pr_url`, stage names, and the three safety promises) so the guidance can't silently drift. Existing tests — including the `version: 0.1.1` frontmatter assertion — are untouched.
- `git diff --check` clean; the documented bash block passes `bash -n`.

## Review summary

Two `ce-code-review` passes, two `codex` passes, and two PR-review-toolkit passes ran. No High/Medium findings survived; all were auto-fixed or resolved:

- **Baseline noise (High/Medium):** the first poll seeded pre-existing state as fresh events — now the baseline snapshot is seeded silently so unchanged slugs print nothing.
- **Empty TSV field corruption (High):** `IFS=$'\t' read` collapsed empty interior fields (e.g. an empty `marker`), corrupting stage/action/marker/pr_url extraction — fixed with a delimiter that survives empty fields.
- **Held-quota with null `retry_after` (Medium):** keying only on `.held.retry_after` missed a `held:{reason:"quota",retry_after:null}` transition — the key now includes held presence/reason.
- **Hidden-field changes (Medium):** summary lines now surface the changed hidden field so a held/`retry_after` or `claude_pid_alive` flip is reported meaningfully rather than as a no-op line.
- **Nits:** documented the `bash` 4+/`mapfile` and `awk`/`date`/`mktemp` dependencies, added an early "nothing to watch" exit, distinguished INT/TERM exit codes.

The change is docs/skill-only: no `lib/` changes, no CLI command, `version: 0.1.1` preserved, existing tests untouched. The snippet is read-only (`hive status --json` reads plus `mktemp`/`rm -f` on temp files).

## Linked task

- Slug: `update-the-openclaw-hive-skill-260630-b0d3`
- PR: https://github.com/ivankuznetsov/hive/pull/643

---

Claude Code


ivankuznetsov:

# Code Review — update-the-openclaw-hive-skill-260630-b0d3 (pass 1)

Scope: `git diff origin/main..HEAD` — `openclaw/skills/hive/SKILL.md` (+120), `test/unit/openclaw_skills_test.rb` (+24). Documentation/skill change only, matching the plan's scope boundaries. All 15 pinned literals from the Requirements Trace are present in the SKILL body and asserted by the new test; `version: 0.1.1` and existing test methods are untouched. No `lib/` or CLI changes. The snippet is read-only (only `hive status --json` reads plus local `mktemp`/`rm -f`/`mv` on temp files); the `trap` cleanup and all safety promises are in place. Findings below concern output-quality behavior of the documented snippet, not scope violations.

## High
(none)

## Medium
- [ ] Hidden-field state changes render an uninformative identical summary line (SKILL.md:283-284): the composite `key` (R6) includes `attrs.phase`, `attrs.pass`, `claude_pid_alive`, and `held.retry_after`, but `print_change` only shows `stage/action/marker`. When one of the hidden fields changes while stage/action/marker do not, the loop prints `slug: 3-develop/develop/none → 3-develop/develop/none` — a no-op-looking line. This directly undercuts the plan's acceptance scenario "a task enters held quota with `retry_after` ⇒ the held/retry_after change is reported once": a line is emitted but conveys nothing about the hold or the `retry_after`. Same problem when `claude_pid_alive` flips (agent dies). Consider appending the changed hidden fields (e.g. `held/retry_after`, `pid_alive`) to the summary line.
- [ ] First poll reports pre-existing state as fresh events (SKILL.md:280-293): `$snap` starts empty (`: >"$snap"`), so on iteration 1 every tracked slug has an empty `old_key` and fires `print_change`; additionally, a task that already has a `pr_url` prints "pr_url became available", and a task already in `7-artifacts`/`8-finalize` prints "entered <stage>". These events did not occur during the watch — they predate it. This contradicts the acceptance example "one summary line for the changed slug, none for the unchanged one" (at startup both slugs print) and produces misleading transition notices. Consider seeding the baseline snapshot silently on the first iteration before reporting.

## Nit
- [ ] Prerequisites understate tool dependencies (SKILL.md:201): the prose states the recipe "assumes `bash` and `jq`", but the snippet also relies on `mapfile` (bash 4+), `awk`, `date +%s`, and `mktemp`. All are standard on Linux but `mapfile` notably excludes bash 3.x (default macOS `/bin/bash`). Worth a one-line note so the copy-paste recipe is unambiguous about its runtime.
- [ ] Empty tracked set spins silently until timeout (SKILL.md:306): when no slugs are passed and no active (non-`9-done`) tasks exist, `slugs` is empty, the terminal-exit guard `[ "${#slugs[@]}" -gt 0 ]` prevents early exit, and the loop polls with no output until `HIVE_WATCH_TIMEOUT` elapses. A brief note or an early "nothing to watch" exit would spare the operator a 30-minute silent wait.

ivankuznetsov:

## High
- [ ] Initial poll reports every tracked task as changed: `openclaw/skills/hive/SKILL.md:280` leaves `old_key` empty for unseen slugs, so lines 283-290 emit `unknown -> <current>`, `pr_url became available`, and `entered 7-artifacts/8-finalize` on the baseline snapshot instead of notifying only when state actually changes.

## Medium

## Nit

ivankuznetsov:

# PR Review — update-the-openclaw-hive-skill-260630-b0d3

Scope: docs/skill change only (`openclaw/skills/hive/SKILL.md` + one new test method
in `test/unit/openclaw_skills_test.rb`). Reviewed `git diff origin/main..HEAD`.
Verified grounding facts against `lib/hive/commands/status.rb`. All pinned literals
(R1–R18) present; `version: 0.1.1` untouched; no `lib/` or CLI changes; existing
tests unmodified. Snippet is read-only (only `hive status --json` reads +
`mktemp`/`rm -f` on temp files). Findings below are minor.

## High

_(none)_

## Medium
- [ ] [portability] `mapfile -t slugs < <(...)` is a bash 4+ builtin, but macOS ships `/bin/bash` 3.2 by default and the skill's Homebrew install targets `os: [darwin]` (SKILL.md:19); a mac operator running default bash gets `mapfile: command not found` on the no-args path: the stated prerequisites ("assumes `bash` and `jq`") don't warn about the bash-4 requirement, so the "canonical, copy-pasteable" snippet silently fails for a plausible target audience.

## Nit
- [ ] [documentation] Prerequisites list only `bash` and `jq`, but the snippet also depends on `awk`, `date`, and `mktemp` (all external commands): worth naming them (or at least noting the bash-4 `mapfile` dependency) so copy-paste portability expectations are accurate.
- [ ] [correctness] On the first poll the baseline snapshot is empty, so every tracked slug's key differs from empty and `print_change` emits a `slug: unknown → <stage>...` line for all slugs: this contradicts the prose/acceptance framing that "unchanged slugs print nothing" (a fresh watcher is momentarily noisy). Behaviorally defensible as an initial-state print, but undocumented — a one-line note would prevent surprise.
- [ ] [documentation] The plan's grounding stated `hive status --json` returns top-level `{ "tasks": [...] }`, but the real envelope is `{ "projects": [ { "tasks": [...] } ] }` (`status.rb:140-155`); the snippet correctly falls back via `tasks_jq='(.tasks // ([.projects[]?.tasks[]?]))'`, so behavior is right, but the `.tasks //` primary branch is dead against the current CLI (no single-project shape is ever emitted) — harmless defensive code, flagged only so the inaccurate grounding isn't mistaken for intent.

ivankuznetsov:

# Code Review — update-the-openclaw-hive-skill-260630-b0d3 (pass 2)

Scope: docs/skill change only — `openclaw/skills/hive/SKILL.md` (+142) and one new
test method in `test/unit/openclaw_skills_test.rb` (+24). No `lib/` changes, no
version bump (`0.1.1` preserved), no other test methods touched. Matches the plan's
Scope Boundaries.

Verification performed:
- Ran `ruby -Itest test/unit/openclaw_skills_test.rb`: 10 runs, 282 assertions, green;
  all 15 pinned literals asserted by the new `test_umbrella_skill_documents_watch_selected_tasks`.
- Extracted the canonical bash snippet and executed it end-to-end against a mock
  `hive status --json`, exercising: no-arg fallback (filters `9-done`), explicit slug
  args, `3-develop → 7-artifacts` transition, `pr_url` null→non-null, hidden-field
  (`phase`/`pass`) diff, early terminal exit (all slugs terminal → exit 0 before
  timeout), timeout final-snapshot branch, and missing-slug handling. All branches
  behaved per Requirements Trace R5–R14.
- Confirmed the snippet is read-only: only `mktemp`/`rm -f` on temp files and
  `hive status --json` reads; no state-mutating `hive` verb, no kill of `claude_pid`.
- Note (not a finding): the plan's grounding claimed a top-level `.tasks` array, but
  the real envelope is `{"projects":[{"tasks":[...]}]}` (`lib/hive/commands/status.rb:140`).
  The snippet correctly handles this via `(.tasks // ([.projects[]?.tasks[]?]))` — a
  correctness improvement over the plan's stated grounding.

## High

_None._

## Medium

_None._

## Nit
- [ ] Prose bash-version note is imprecise about where `mapfile` is used (`SKILL.md:201`): it says "the no-args path uses `mapfile`", but `mapfile` is also used unconditionally in the always-executed change-detection loop (`SKILL.md:298-299`). The umbrella claim "assumes `bash` 4+" is correct, so operators are warned; however, the parenthetical could mislead a macOS bash-3.2 user into thinking explicit slug args avoid the bash-4 requirement, when they do not. Reword to "the recipe uses `mapfile`" or drop the "no-args path" qualifier.
- [ ] `trap 'cleanup; exit 130' INT TERM` uses exit code 130 for both signals (`SKILL.md`, trap line): 130 is the `128+SIGINT` convention; a `TERM` exit is conventionally 143. Harmless and common, but strictly the code doesn't distinguish the two signals.

ivankuznetsov:

## High

## Medium
- [ ] Held state changes can be missed when retry_after is null: `openclaw/skills/hive/SKILL.md:237` keys only on `.held.retry_after`, but Hive status explicitly allows `held.retry_after` to be null, so a task can enter `held: {reason: "quota", retry_after: null}` without changing the composite key or notifying the watcher.

## Nit

ivankuznetsov:

# PR Review (pass 2) — update-the-openclaw-hive-skill-260630-b0d3

Scope: docs/skill change only — `openclaw/skills/hive/SKILL.md` (+142) and one new
test method in `test/unit/openclaw_skills_test.rb` (+24). No `lib/` changes, no
version bump (`0.1.1` preserved), no other test methods touched — matches the plan's
Scope Boundaries. All 15 pinned literals present; the new
`test_umbrella_skill_documents_watch_selected_tasks` asserts them. Snippet is
read-only (`hive status --json` reads + `mktemp`/`rm -f` on temp files only).

Multi-persona findings (union) below. The headline is a bash TSV-parsing defect that
prior pass-2 reviewers missed because their mocks populated every field; it only
manifests when an interior field is empty, which is the common case (`marker` is
`""` for any task without an active marker, per `lib/hive/commands/status.rb:283`).

## High
- [ ] [correctness] `IFS=$'\t' read` collapses empty TSV fields, corrupting stage/action/marker/pr_url extraction (`SKILL.md:289`, `:293`, `:331`): tab is IFS-whitespace, so `read` with `IFS=$'\t'` treats runs of tabs as a single delimiter and trims trailing tabs (verified: `printf 'a\t\t\tb' | IFS=$'\t' read w x y z` yields `w=a x=b`, not `w=a …z=b`). `state_fields` emits `[.stage,.action,.marker,.pr_url] | @tsv`, and `marker` is `""` for any task with no active marker (`status.rb:283`), so a realistic row `7-artifacts/ready/""/<pr>` parses to `marker=<pr_url>, pr=""` — reproduced end-to-end against the snippet's own functions. This defeats R11 (pr_url-became-available is **missed** because `new_pr` is empty), R12 (stage-transition line prints without the PR URL), and R7 (summary label shows the URL as the marker); a `8-finalize` row with an empty `action` and `marker=complete` also fails the terminal check (R13), so the loop runs to timeout instead of exiting 0. Fix requires a delimiter that is not IFS-whitespace (e.g. join with a sentinel/`\x1f` and split on that, or read each field with a separate `jq` call) — plain `IFS=$'\t' read` cannot preserve empty TSV fields.

## Medium
- [ ] [correctness] Held-quota transition is silently missed when `retry_after` is null (`SKILL.md:237`): the composite key uses only `.held.retry_after // ""`, but `held` is `{reason:"quota", provider:…, retry_after:<iso8601|null>}` and `retry_after` is explicitly allowed to be null (`lib/hive/agent_limit.rb`). A task entering `held:{reason:"quota",retry_after:null}` produces the same key as not-held, so the watcher never reports it — contradicting the plan's acceptance scenario "a task enters held quota ⇒ the held/retry_after change is reported once." Keying on held presence/`reason` (not just `retry_after`) would close the gap.

## Nit
- [ ] [documentation] Prose "the no-args path uses `mapfile`" is misleading (`SKILL.md:201`): `mapfile` is also used unconditionally in the always-executed change-detection loop (`SKILL.md:298-299`), so passing explicit slug args does **not** avoid the bash-4 requirement. Reword to "the recipe uses `mapfile`" so a macOS bash-3.2 operator isn't led to think explicit args are a workaround.
- [ ] [portability] `trap 'cleanup; exit 130' INT TERM` uses exit 130 (`128+SIGINT`) for `TERM` as well (`SKILL.md:262`); a `TERM` exit is conventionally 143. Harmless, but the two signals are not distinguished.

PR #644, opened 2026-07-01
refactor: architecture pass over the local web install/run mode (#622)

ivankuznetsov:
Post-merge architecture review of #622 (*first-class local (non-Docker) web install and run mode*), from the user and best-practices perspective, followed by five behavior-preserving refactors. Each change is documented in its own PR comment below, with the finding it resolves and the reasoning.

## What this PR does

| # | Change | Finding it resolves |
|---|--------|---------------------|
| 1 | Extract `Hive::Daemon::StatusReport` | Web tier constructed a CLI command object for data access; `public :status_payload` escape hatch |
| 2 | Derive `target_path` in `ServiceInstaller::Base`; migrate the bot installer to the shared renderers | `target_path` triplicated with divergent unsupported-platform contracts; bot installer skipped in #622's own dedupe |
| 3 | Explicit subcommand dispatch in `Commands::Web` | `start` without `--detach` mutated `@subcommand` and re-entered `#call` |
| 4 | Rails auth bypass delegates to `Hive::Web::Loopback` | The module's "single source of truth" claim was false |
| 5 | `hive setup`: one `#phase` runner; remove dead `--yes` | 4× duplicated rescue boilerplate; advertised flag nothing read |

One deliberate contract change (change 2): on an **unsupported platform**, `Web::ServiceInstaller#target_path` now returns `nil` (like the daemon/bot installers) instead of raising `Hive::InvalidTaskPath`, so `hive web status` degrades to "not installed" the way `hive daemon status` does. Everything else is behavior-preserving.

## Verification

- `rake test`: failure set byte-identical to the `main` baseline on this host (the shared failures are host-environmental, subprocess-spawn related; targeted unit suites for daemon/web/setup/installers all green).
- RuboCop: 14 changed files, no offenses.

Review summary with the findings **not** changed here (flagged for maintainer judgment) is posted on #622.


ivankuznetsov:
## Change 1 — `Hive::Daemon::StatusReport` (8ce3c6ec)

**Finding.** The `hive-daemon-status` envelope was produced by a cluster of private methods inside `Commands::Daemon`, and the web dashboard consumed it by constructing a CLI command object purely for data access:

```ruby
Hive::Commands::Daemon.new("status", json: true).status_payload
```

with `public :status_payload` after the `private` marker — the code itself admitting the seam was in the wrong place ("Public despite living in the private section…"). A Rails controller should not need to know how to build a Thor-command wrapper (including a meaningless `json: true`) to read daemon health.

**Change.** The producer is now `Hive::Daemon::StatusReport` (`lib/hive/daemon/status_report.rb`): `#running_state` (PID-file liveness), `#payload` (full envelope), `#safe_payload` (the web contract — never raises, degrades to a minimal not-running hash). `BINARY_DRIFT_STATES` / `BINARY_DRIFT_ACTIONABLE` move with it, so the schema, the `_daemon` view guard, and the docs all read the actual producer. `Commands::Daemon` keeps a memoized `#status_report` and shrinks by ~150 lines; `StatusController` calls `StatusReport.new.safe_payload`.

**Why it's better for the user.** The CLI and the web UI now provably render the same envelope from the same object, and the next consumer (bot push? TUI footer?) has a first-class API instead of a hack to copy.

ivankuznetsov:
## Change 2 — one unit-path rule, one renderer for the installer family (149503f6)

**Finding (a).** #622 extracted `render_systemd_from` / `render_launchd_from` into `ServiceInstaller::Base` — good — but migrated only the daemon and web installers. The bot installer kept byte-for-byte copies of the pre-extraction render bodies, so the family had a shared helper *and* a fork of it on day one.

**Finding (b).** `target_path` was implemented three times, each a platform-case over the same two derivable patterns (`~/.config/systemd/user/<name>.service`, `~/Library/LaunchAgents/local.<name>.plist`). Worse, the three siblings disagreed about unsupported hosts: daemon/bot return `nil` (what `Base#install!` and `#service_state` are written to handle), while the web installer raised `Hive::InvalidTaskPath` — so `hive web status` on FreeBSD/WSL-without-systemd would crash mid-envelope where `hive daemon status` degrades to `"not installed"`.

**Change.** `Base#target_path` derives from `service_name`/`launchd_label`; all three overrides deleted; the bot renders through the shared helpers (its templates have no `HIVE_BIN` line and no bare `WorkingDirectory`, so those substitutions are documented no-ops). One contract everywhere: `nil` on unsupported hosts.

**Deliberate behavior change.** `Web::ServiceInstaller#target_path` no longer raises on unsupported platforms — `test_target_path_raises_on_unsupported_platform` became `test_target_path_nil_on_unsupported_platform`, asserting the degraded `service_state` instead. The raise defended against `nil` "silently flowing into systemctl argv", but `install!` returns `:unsupported` before ever touching `target_path`, so the guard mostly turned a readable status into a crash.

ivankuznetsov:
## Change 3 — explicit dispatch in `Commands::Web` (419bda03)

**Finding.** `#call` branched on `@subcommand ? service_command : <180-line foreground body>`, and `start` without `--detach` re-entered `#call` after mutating state:

```ruby
when "start"
  if @detach
    start_service
  else
    @subcommand = nil
    call
  end
```

Control flow that depends on mutating your own constructor arguments is the kind of thing that reads fine today and bites the next person who adds a subcommand or memoizes anything keyed on `@subcommand`.

**Change.** One case expression routes every subcommand (`nil` → `run_foreground`, `install` → bootstrap+install, `start` → `@detach ? start_service : run_foreground`, …), and the foreground boot body lives in `#run_foreground`. The lazy `require` of the service installer moved into the three service methods, so bare `hive web` still boots without loading installer code. Behavior unchanged — the routing tests pass as-is; only the test comment describing the recursion hack was updated.

ivankuznetsov:
## Change 4 — make the loopback single source of truth true (e14c2fe9)

**Finding.** `lib/hive/web/loopback.rb` documents itself as:

> Single source of truth for "is this address the loopback interface?". Shared by the CLI bind-policy check … **and the Rails loopback-peer bypass (`ApplicationController#local_loopback_request?`)** so the two ad-hoc string checks can't drift apart.

But the controller never called it — it kept its own inline check (`remote == "::1" || remote.start_with?("127.")`). This is the *auth bypass* check; the one place where the "can't drift apart" guarantee actually matters was the one place not using the module. If someone later teaches `Loopback.address?` about IPv4-mapped `::ffff:127.0.0.1`, the CLI would set `HIVEBOX_LOCAL_LOOPBACK=1` for binds the Rails side then refuses (or vice versa).

**Change.** `local_loopback_request?` now delegates to `Hive::Web::Loopback.address?(request.remote_ip)`; the module is required from `web/config/initializers/hive.rb` alongside the other gem classes the controllers use. Accepted set is identical (`::1`, `127.0.0.0/8`; plus `"localhost"`, which `remote_ip` never yields), so behavior is unchanged — the claim in the module doc is just no longer aspirational.

ivankuznetsov:
## Change 5 — `hive setup`: one phase runner, no dead flags (003b008d)

**Finding (a).** Four of the five provisioning steps carried the same `rescue StandardError => e; add_phase(name, false, "message" => …)` tail, each with its own copy of the AE5 comment explaining why. A fifth phase added later that *forgets* the rescue silently reintroduces the abort-before-envelope bug the comments warn about.

**Finding (b).** `hive setup --yes` ("accept non-interactive defaults") was advertised in `hive help setup` and documented in the wiki — but `@yes` was assigned and never read. Setup has no interactive path at all. A flag that changes nothing trains users to cargo-cult it (and would make a *future* interactive default ambiguous: does `--yes` mean the old no-op or the new prompt-skipping?).

**Change.** A single `#phase(name)` runner owns the rescue policy — blocks return `[ok, data]`, any raise records a failed phase so the `--json` envelope always emits and the exit code stays honest. Each step is now just its own logic. `--yes` is removed from the CLI, the command, and `wiki/commands/setup.md`; #622 has not shipped in any release tag, so no released caller can be passing it.

ivankuznetsov:

Outcome: timeout

Agent tail:

```
The root suite is still active; it is likely in the slower subprocess/e2e-adjacent integration tests now.
```

PR #645, opened 2026-07-01
refactor(pairing): Hive::AtomicFile — stop the atomic-write pattern at five copies

ivankuznetsov:
Architecture-review follow-up for #629, stacked on its branch (see the review comment there). One structural change, documented in a PR comment below; everything else in #629 held up under review.

- New `lib/hive/atomic_file.rb`: `Hive::AtomicFile.write(path, content, mode:, fsync:)` — tempfile in the same directory, optional fsync, rename, guaranteed tempfile cleanup.
- `PairingStore#write_entries` and `PairingApprovalQueue.write!` (the PR's two new state writers) now use it; identical on-disk behavior (same modes, same fsync discipline, same tmp-in-same-dir naming shape).
- Unit tests cover mode/parent-dir/return-value, the no-fsync path, and failed-rename leaving the target intact with no orphaned tempfile.

Legacy copies (`Markers`, `Events`, `Init`, `ServiceInstaller::Base`, review suppression) are deliberately NOT migrated here — that's a separate mechanical pass; this PR only keeps the pairing feature from adding copies six and seven.

Verification: pairing unit + integration suites green (219 runs), RuboCop clean on touched files.


ivankuznetsov:
## Change — `Hive::AtomicFile` (0e0d2819, amended)

**Finding.** `PairingStore#write_entries` and `PairingApprovalQueue.write!` each hand-roll tempfile → write → flush → fsync → rename → `ensure` cleanup. The codebase already has at least five private implementations of exactly this (`Markers#write_atomic`, `Events#write_atomic`, `Commands::Init#atomic_write`, `ServiceInstaller::Base#atomic_write`, `review/suppression.rb#write_atomic`) — each with its own subtle choices about fsync, tmp naming, and cleanup. Every new copy is a chance to get one of those choices wrong (and reviewers must re-verify the whole dance every time).

**Change.** `Hive::AtomicFile.write(path, content, mode: 0o644, fsync: true)` is the one obvious way; both new pairing writers now call it. On-disk behavior is identical: the store keeps `mode: 0o600` and fsync; the queue keeps 0o644 (its directory is already 0700 via `QueueDirectory`) and fsync; the tmp name keeps the `.name.tmp.pid.hex` same-directory shape, so nothing about crash-safety or the queue's glob patterns changes.

**Scope.** The five legacy copies are left alone deliberately — migrating them is a mechanical, separately-reviewable pass, and doing it inside a feature PR's follow-up would bury the pairing diff. The helper's doc comment points future state files at it.

PR #646, opened 2026-07-01
refactor(review): drop the reserved rate_limited enum member

ivankuznetsov:
Architecture-review follow-up for #627, stacked on its branch (review comment posted there). One change, documented in a PR comment below.

`Hive::ReviewErrorReason::REASONS` loses its reserved-but-unemitted `rate_limited` member. Nothing reads `REASONS` anywhere in `lib/`/`web/` — the healer keys on the literal `limits_reached`, `notification_builders` keeps its own manual-only lists, and the wiki's enum table already omits it — so the "accepted on read" reservation guarded nothing while inviting misuse (emitting it would *sound* like a limit but bypass the `limits_reached` cooldown auto-heal).

Verification: review-error-reason, phase-failure-helper, and stale-agent-healer suites green (112 runs); RuboCop clean.

### DIFF
--- a/lib/hive/review_error_reason.rb
+++ b/lib/hive/review_error_reason.rb
@@ -3,19 +3,16 @@ require "hive/agent_limit"
 module Hive
   module ReviewErrorReason
     # The `reason=` vocabulary `mark_review_phase_failure` can stamp onto a
-    # review_error marker, plus one reserved slot. Its limit arm writes
-    # `limits_reached`; `classify` emits the CLASSIFIED subset below (plus the
-    # always-available `unknown` fallback). `rate_limited` is reserved-but-
-    # unemitted: it is accepted on read, but nothing in lib/ ever writes
-    # `reason=rate_limited` (the gate only stamps `limits_reached`) — it is
-    # kept so the enum tolerates the value without tripping the drift guards.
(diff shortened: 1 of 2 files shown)


ivankuznetsov:
## Change — remove `rate_limited` from REASONS

**Finding.** The enum shipped with a member no code can ever produce, justified in its own comment as "accepted on read … kept so the enum tolerates the value without tripping the drift guards". Grepping `lib/` and `web/`: no reader consults `REASONS` at all. The daemon healer branches on the literal string `limits_reached`; the bot's manual-only gating uses its own lists; the TUI/web render `attrs["reason"]` generically. So the reservation had no consumer to tolerate anything — it was vocabulary without grammar.

**Why it matters.** A dead member in a *closed* enum is an attractive nuisance: a future author wiring real 429 handling would reasonably emit `reason=rate_limited` (it's right there in the enum), and the failure would silently land on the terminal manual-only path instead of the `limits_reached` cooldown auto-heal — the exact misrouting the two-tier recovery design exists to prevent. YAGNI applies doubly to closed enums.

**Change.** `rate_limited` removed; the docstring now states the rule that keeps this from recurring: every member is written by exactly one live code path, and members are added only together with their emitter. The wiki's enum table in `wiki/stages/review.md` already matched this shape, so docs and code now agree.

PR #656, opened 2026-07-02
feat(refactor-patrol): add hive refactor-patrol discovery command

ivankuznetsov:
## Summary

`hive refactor-patrol` is a new on-demand command that surfaces high-leverage refactoring work as ranked, evidence-backed *refactor theses* for a human to review before any code is written. Where `hive patrol` reports bugs, `refactor-patrol` answers a different question: *"which architecture problem, if fixed, buys the most leverage — and what would that refactor cost?"*

It is a **sibling** to patrol, not a flag on it. The unit of output is a leverage-ranked thesis — *feature X has architecture problem Y causing cost Z; refactor A reduces the cost while preserving behavior* — scored by a transparent weighted blend of measurable signals, never a magic number.

**v1 is discovery/reporting only.** It maps features, asks an agent for schema-shaped theses per candidate, scores them, flags anything that breaks risk-scope caps or admissibility, and emits a ranked report. It does **not** auto-fix, touch worktrees, open PRs, or enqueue into `6-review` — those land later behind per-candidate approval.

### What's new

You can now run:

```
hive refactor-patrol PROJECT [--json] [--dry-run] \
  [--feature X | --entrypoint E | --path P] [--changed-since REF]
```

and get a ranked list of refactor theses, each carrying cited evidence, an explicit feature boundary, a leverage-score breakdown, declared risk/budget, and behavior-preservation guidance.

### How it works

The command mirrors patrol's structure — Thor command → orchestrator → `Hive::RefactorPatrol::*` components → durable state under `.hive-state/refactor_patrol/` → structured `hive-refactor-patrol.v1` output — while keeping its schema, scoring, memory, and reporting fully independent so it can evolve without distorting ordinary patrol.

| Stage | Component | Responsibility |
|---|---|---|
| Map | `Hive::Patrol::Mapper` (reused) | Discover feature slices; no patrol change required |
| Score | `leverage.rb` | Blend churn, fan-in/blast-radius, complexity/size, cross-boundary coupling into a score with a per-signal breakdown |
| Review | `reviewer.rb` + ERB prompt | Ask the agent for a schema-shaped thesis anchored on measured signals; validate + gate admissibility in code |
| Guard | `caps.rb` | Filter-or-flag cap violations; detect public-API and cross-feature impact (advisory in v1) |
| Dedup | `fingerprint.rb`, `collisions.rb` | Own fingerprint/dismissal memory; read patrol state read-only to suppress/flag in-flight collisions |
| Report | `reporter.rb` | Ranked text + `hive-refactor-patrol.v1` JSON envelope |

### Design decisions

- **Leverage is explainable, not a number.** Every thesis carries `expected_leverage.breakdown`; a breakdown-less thesis fails schema validation. Weights are config-tunable without code changes, and optional signals (bug density, coverage gaps) default to weight 0 and are never required.
- **Admissibility is enforced in code, not trusted from the agent.** Evidence must cite a concrete file path *plus* ≥1 measurable signal. "Feels messy" theses are marked `admissible: false` and flagged — retained and surfaced, never silently dropped.
- **Caps filter or flag, never silently drop.** Cap-exceeding candidates land in `flagged_theses` with a reason. The return shape carries a `blocked` field (always `false` in v1) so the later hard-block mode can reuse it.
- **Independent memory, read-only awareness.** refactor-patrol owns its fingerprint/dismissal store but reads patrol's `fingerprints.json` read-only to suppress already-seen theses and flag collisions with open patrol work.
- **Partial-scan safety.** A single feature's agent error records a review error and does not advance `last_scanned_sha`, mirroring patrol's incomplete-scan invariant.

### Out of scope (v1)

Any code mutation, PR opening, secret scanning, `6-review` enqueue, hard-blocking guardrail enforcement, and daemon/`continuous` cadence are explicitly deferred to follow-up work behind per-candidate approval.

## Test plan

`rake` is unavailable in this environment, so suites were run directly:

- Focused refactor-patrol suite (config, leverage, fingerprint, caps, reviewer, collisions, integration): `407 runs, 1818 assertions, 0 failures`
- CLI tests: `33 runs, 160 assertions, 0 failures`

The acceptance integration test asserts the v1 contract end-to-end with a mocked agent: ranked evidence-backed theses with explicit boundaries, cap-exceeding candidates flagged (never silently emitted), behavior-preservation guidance on accepted theses, inadmissible theses flagged, re-run suppression of seen theses, and **no** auto-fix/PR/`6-review` side effects.

## Review summary

Two full code-review passes ran over the branch (Claude + Codex + PR-review-toolkit reviewers) with all findings triaged and resolved:

- **Dry-run contract (High).** `--dry-run` no longer persists any durable artifacts: reviewer thesis writes, run dirs, agent logs, and the reused `Mapper`'s feature writes are all suppressed under dry-run, so a preview run touches nothing under `.hive-state/refactor_patrol/`.
- **Leverage robustness & speed (High/Medium).** Non-UTF8/binary tracked files no longer crash scoring (encoding scrubbed on read); tracked-file content and `fan_in` are memoized on the shared instance and honor exclude globs, removing the O(features × files) double scan.
- **Ranking integrity (Medium).** Ranking sorts by the deterministic computed leverage score; the agent's `expected_leverage.score` is advisory only, preserving the "no magic number" guarantee.
- **Behavior-preservation guidance (Medium).** Every admissible thesis is guaranteed validation guidance (commands or `characterization_first`), including test-rich slices with empty commands.
- **Flag-not-drop guarantee (Medium/High).** Evidence-less or agent-declared-risk theses are retained and flagged rather than dropped as `schema_invalid`; agent-declared public-API/cross-feature risk is honored when caps disallow it.
- **Scope-hint precision (Medium/Nit).** `--feature` uses exact id match and `--path` uses directory/exact-file boundaries (no mid-filename prefix matches); `--changed-since` degrades to the scoped set on git failure; the duplicate `git diff` is memoized.
- **Config validation (Nit).** An all-zero leverage-weight set is rejected so it can't silently produce empty breakdowns.

One Nit was suppressed by design: loose basename substring matching in leverage heuristics is explicitly accepted under plan Risk #1 (cheap/noisy v1 signals, tunable via weights later). No open escalations; all passes recorded fix-success.

## Linked task

- Plan: `hive refactor-patrol` v1 (discovery/reporting) — sibling to `hive patrol`
- Task slug: `add-a-refactor-patrol-architecture-260702-9048`
- PR: https://github.com/ivankuznetsov/hive/pull/656

---

Claude Code


ivankuznetsov:

# Code Review — `hive refactor-patrol` (v1 discovery/reporting)

Scope: `git diff origin/main..HEAD` (27 files, ~2768 insertions). All 31 refactor_patrol unit + integration tests pass. Findings below are corroborated across the orchestrator read plus two independent reviewers (correctness, plan-conformance).

## High
- [ ] `--dry-run` still persists state via the real `Reviewer` (`lib/hive/refactor_patrol/reviewer.rb:109`, `:57`, `:211`): `normalize_thesis` calls `@state.write_thesis`, `review_feature` calls `@state.run_dir("review")` (which `ensure!`s `features/theses/runs` dirs), and `record_feature_error` writes run logs — all unconditionally. The command's `@dry_run` guard only wraps `persist`/`update_scan_state` (`lib/hive/commands/refactor_patrol.rb:85`), so a real run with `--dry-run` writes `theses/*.json` and creates `.hive-state/refactor_patrol/` — violating plan U8/DoD "`--dry-run` does not write `state.json`/`theses/`". The integration dry-run test passes only because it injects a `FakeReviewer` that never persists; the real path is uncovered.
- [ ] Leverage scoring crashes on any non-UTF8/binary tracked file (`lib/hive/refactor_patrol/leverage.rb:84`, `:92`, `:121`): `read`'s `rescue ArgumentError` is dead for this case — `File.read` succeeds on binary bytes, then `String#downcase`/`#match?` in `fan_in`/`complexity`/`coupling` raise `ArgumentError: invalid byte sequence in UTF-8` outside the rescue. `Leverage#score` is called with no per-feature rescue (`refactor_patrol.rb:68`), so one tracked image/latin-1 file aborts the entire run — contradicting R5/U4's required "no crash / graceful degradation" posture (churn has a rescue; the other three do not). Fix belongs in `read` (scrub encoding on the returned string).

## Medium
- [ ] Ranked score is agent-overridable, weakening R5's "no magic number" guarantee (`lib/hive/refactor_patrol/reviewer.rb:131`): `"score" => score.fetch("score", leverage["score"] || 0).to_f` lets the agent's `expected_leverage.score` (and `breakdown`, line 132) override the deterministic churn/fan_in/complexity/coupling blend, and `Reporter#score` (`reporter.rb:85`) sorts by that agent-supplied value. R5 wants ranking explained by measured signals; nothing enforces it. Prefer ranking by the computed `leverage["score"]` and treating the agent value as advisory.
- [ ] R8 behavior-preservation guidance is skipped for test-rich slices (`lib/hive/refactor_patrol/reviewer.rb:170`): `characterization_first` is forced only when `!has_tests && commands.empty?`. A test-rich slice where the agent returns empty `commands` and `characterization_first:false` yields an admissible thesis with no validation guidance at all, violating R8's "each thesis names validation commands OR characterization_first." Require guidance for every admissible thesis (e.g., inject the configured `test` command or flag when commands are empty).
- [ ] Redundant, unbounded full-repo content scanning in leverage (`lib/hive/refactor_patrol/leverage.rb:104`, `:125`): `coupling` recomputes `fan_in(feature)` even though `collect` already computed it (two full passes reading every tracked file per feature), and `tracked_files` honors neither the `include`/`exclude` config nor size limits, so it reads content of every tracked file including vendored/large/binary assets. On non-trivial repos this is O(features × tracked_files) file reads and can be very slow. Reuse the computed `fan_in` and apply the configured exclude globs.

## Nit
- [ ] Every thesis is written to disk twice (`lib/hive/refactor_patrol/reviewer.rb:109` then `lib/hive/commands/refactor_patrol.rb:107` in `persist`): the reviewer's write happens before the command mutates the thesis (caps flags, collision), so the first write is immediately stale and overwritten — wasted I/O. Drop the reviewer's `write_thesis` and persist only post-mutation (this also resolves the dry-run leak above).
- [ ] `--path` matching is over-broad (`lib/hive/commands/refactor_patrol.rb:131`): the trailing `path.start_with?(normalized)` clause makes the exact-match and `"#{normalized}/"` clauses redundant and matches mid-filename prefixes (`--path lib/se` matches `lib/search.rb`). Drop the third clause so `--path` means a directory/exact-file boundary as documented.
- [ ] `--feature` uses substring `include?` (`lib/hive/commands/refactor_patrol.rb:126`): `feature.id.include?(@feature_hint)` over-selects (`--feature check` matches `checkout` and `recheck`). Plan R11 says `--feature X` restricts to feature X; prefer exact id match (or document the substring behavior).
- [ ] Text-output path builds a throwaway `Reporter` with a synthetic cfg (`lib/hive/commands/refactor_patrol.rb:175`): `Reporter.new({ "refactor_patrol" => { "max_theses_per_run" => ... } })` instead of reusing the reporter instance already built in `run_cycle`. Harmless today but will mislead if `text` later reads other cfg keys.

ivankuznetsov:

## High
- [ ] `--dry-run` still persists refactor-patrol artifacts: `RefactorPatrol#run_cycle` calls the real reviewer before the dry-run gate, and `Reviewer#review_feature`/`normalize_thesis` create `.hive-state/refactor_patrol/runs/*` plus write thesis JSON, violating the v1 dry-run/no-persistence contract. (`lib/hive/commands/refactor_patrol.rb:71`, `lib/hive/refactor_patrol/reviewer.rb:57`, `lib/hive/refactor_patrol/reviewer.rb:109`)
- [ ] Accepted theses can omit behavior-preservation guidance: `enforce_behavior_guidance!` only fills `characterization_first` when a slice has no tests, so a tested slice with empty/unknown validation commands remains admissible and can count as accepted despite R8/A3 requiring commands or characterization-first guidance. (`lib/hive/refactor_patrol/reviewer.rb:165`)

## Medium
- [ ] `--path` scope over-selects sibling paths: the filter accepts any owned file starting with the raw hint, so `--path lib/foo` also reviews `lib/foobar.rb`, violating the “under that path” scope semantics. (`lib/hive/commands/refactor_patrol.rb:131`)

## Nit

ivankuznetsov:

## High
- [ ] [correctness] `--dry-run` still persists durable state: the real `Hive::RefactorPatrol::Reviewer#normalize_thesis` calls `@state.write_thesis(thesis)` unconditionally (reviewer.rb:109), so a real dry run writes `.hive-state/refactor_patrol/theses/*.json` and creates `runs/` dirs — violating U8/R15's "`--dry-run` does not write `state.json`/`theses/`" contract (only `persist`/`update_scan_state` are `unless @dry_run`-guarded in refactor_patrol.rb:85). The integration `test_dry_run_does_not_write_state_or_theses` passes only because it injects a `FakeReviewer` that never writes, so the shipped defect is hidden behind a green test.

## Medium
- [ ] [correctness] `--path` scope hint over-matches sibling paths: `apply_scope_hints` ORs `path.start_with?(normalized)` with no trailing `/` (refactor_patrol.rb:131), so `--path lib/search` also selects features owning `lib/search_helper.rb` or `lib/search2.rb`, breaking the "owned files under this path" semantics of U8/R11 (the exact and `"#{normalized}/"` clauses are already correct; the bare prefix clause is the bug).
- [ ] [correctness] Inadmissible thesis with no `file` in evidence is dropped, not flagged: the thesis schema requires `evidence[].file`, so an evidence item lacking a path fails `@schemer.valid?`, is recorded as a `schema_invalid` review error, and returns `nil` (reviewer.rb:103-106) — diverging from U5's scenario "evidence has no file path → `admissible: false` … still returned (flagged, not dropped)". Side effect: any such agent output marks the scan incomplete and pins `last_scanned_sha`, forcing perpetual re-scan.

## Nit
- [ ] [maintainability] Reviewer's inline `@state.write_thesis` (reviewer.rb:109) is redundant — `Commands::RefactorPatrol#persist` re-writes each thesis after caps/collision mutations (refactor_patrol.rb:107), which is the write that actually captures flags/collisions. Removing the early write eliminates duplicate I/O and is the fix for the dry-run leak above.
- [ ] [perf] `Leverage#coupling` re-invokes `fan_in` (leverage.rb:104) and `fan_in` reads every tracked file per feature (leverage.rb:81-86), so full-repo content reads run ~2×(feature count); memoize `tracked_files`/`fan_in` to bound cost on large repos (plan risk #7).
- [ ] [security] `Reviewer::TemplateBindings` computes and passes `user_supplied_tag` (reviewer.rb:80) but `templates/refactor_patrol_review_prompt.md.erb` never references it, so repo-derived paths/content are injected into the agent prompt without the intended data-fence. Matches patrol's existing template posture (not a regression), but the unused binding suggests a fence was intended and forgotten.
- [ ] [correctness] `Caps#candidate_paths` includes `feature_boundary.entrypoints` (caps.rb:98); if an entrypoint is not also listed in `owned_files`, `detect_cross_feature` will false-positive a `cross_feature_impact` flag against the feature's own entrypoint.

ivankuznetsov:

# Code review — refactor-patrol v1 (pass 2)

Scope: `git diff origin/main..HEAD` (28 files, ~2935 insertions). Intent: add discovery/reporting-only `hive refactor-patrol` command per plan. All 291 refactor_patrol + config + integration tests pass locally. No High-severity defects and no plan-required-but-missing units found (U1–U9 all implemented). Findings below are runtime/quality issues.

## High
- (none)

## Medium
- [ ] Dry-run leaks durable agent logs into `.hive-state/refactor_patrol/logs`: `Reviewer#run_agent` (lib/hive/refactor_patrol/reviewer.rb:247) sets `log_dir: File.join(@state.root, "logs")` and `Hive::Agent#ensure_log_dir` (lib/hive/agent.rb:481) `mkdir_p`s it. In a real `--dry-run` the reviewer scratches its output in a tmp dir but the agent still creates and writes logs under the durable state tree, violating the stated invariant ("A dry run must not create durable artifacts under .hive-state/refactor_patrol/", refactor_patrol.rb:61) that the command advertises as a pure preview (tests mock the agent so this isn't caught).
- [ ] Leverage scanning is O(features × tracked_files) in file reads: `fan_in` and `coupling` (lib/hive/refactor_patrol/leverage.rb:77,98) iterate all tracked files and call `read(path)` with no content memoization, and the `Leverage` instance is shared across every mapped feature — so each of N features re-reads the full tracked-file set. On a large repo this is a real per-run cost the earlier "harden/speed up" commit did not remove.

## Nit
- [ ] `references_path?` recomputes `content.downcase` on the full owned-file content once per candidate inside the `files.count { ... }` loop (leverage.rb:119-123), redundantly lowercasing the same string T times per owned file; hoist the downcased content out of the loop.
- [ ] `changed_files` runs `git diff --name-only` twice per invocation — once in `apply_scope_hints` (refactor_patrol.rb:140) and again in `run_cycle` (refactor_patrol.rb:66); compute once and reuse.
- [ ] Loose substring matching in `reference_needles`/`references_path?` uses bare basenames (e.g. entrypoint `cli.rb` -> needle `cli`) that match any file whose content contains the substring, inflating fan_in/coupling; acknowledged by plan Risk #1 but worth tightening (word-boundary or require the path stem, not the bare basename).
- [ ] All-zero leverage weights are accepted by config validation (only `>= 0` is enforced) but produce an empty `breakdown`, which fails the thesis schema's `expected_leverage.breakdown` `minProperties: 1`; every thesis is then silently recorded as a `schema_invalid` review error with no operator-facing explanation. Consider validating that at least one weight is positive.
- [ ] `normalize_thesis` builds `Thesis.from_h(hash)` twice (reviewer.rb:106 and :115) — once only to stamp the fingerprint; reuse the first instance.
- [ ] `detect_dependency_bumps` appends manifest names to `cross_feature_details` and sets the `dependency_bump` flag but never sets `cross_feature_impact: true`, so a reader seeing populated `cross_feature_details` with `cross_feature_impact: false` is mildly inconsistent (caps.rb:85-97).

ivankuznetsov:

## High

## Medium
- [ ] `--dry-run` still writes refactor-patrol feature state: `RefactorPatrol#run_cycle` skips `state.ensure!` for dry runs, but immediately calls the real patrol mapper with the same `StateStore`; `Hive::Patrol::Mapper#call` always calls `@state.ensure!` and writes each mapped feature, so `hive refactor-patrol PROJECT --dry-run` creates `.hive-state/refactor_patrol/features/*.json` despite the v1 dry-run contract to avoid persistence. See `lib/hive/commands/refactor_patrol.rb:60` and `lib/hive/patrol/mapper.rb:18`.
- [ ] Agent-declared public-API or cross-feature risk can be accepted as clean: `Caps#apply` only adds report flags for heuristic path detections, so a schema-valid thesis with `risk.public_api_impact: true` or `risk.cross_feature_impact: true` but no path matching the heuristic keeps an empty `risk.flags`; `Reporter#accepted_count` then counts it as an accepted thesis even though the default caps disallow those risks and the plan requires cap/guardrail violations to be flagged, never silently clean. See `lib/hive/refactor_patrol/caps.rb:34` and `lib/hive/refactor_patrol/reporter.rb:74`.

## Nit

ivankuznetsov:

## High

## Medium
- [ ] [correctness] Empty-evidence thesis is silently dropped, not flagged: an agent thesis with `evidence: []` marks `admissible: false` in `enforce_admissibility!` (reviewer.rb:198) but then fails the thesis schema's `evidence.minItems: 1` (hive-refactor-patrol-thesis.v1.json:34), so `normalize_thesis` records a `schema_invalid` review error and returns `nil` (reviewer.rb:110-113) — the candidate is dropped and, worse, the review error pins `last_scanned_sha` (update_scan_state) forcing perpetual re-scan. This is the zero-evidence variant of the pass-1 High "never silently dropped" contract gap (R10/U5/DoD); the missing-file/no-signal cases are covered because they still carry ≥1 evidence item, but a truly evidence-less "feels messy" thesis escapes the flag-not-drop guarantee.

## Nit
- [ ] [correctness] `Caps#public_api_content?` (caps.rb:110-112) flags any `.rb` path whose name merely contains the substring `"cli"`, so `lib/hive/client.rb` (or any `*cli*.rb`) false-positives `public_api_impact` even though it is not a Thor command surface; tighten to the CLI file/`bin/` heuristic already covered by `PUBLIC_API_PATTERNS`.
- [ ] [robustness] `changed_files` (refactor_patrol.rb:145-153) returns `[]` both for "no changes" and for a git failure; when `--changed-since` is combined with a scope hint, `apply_scope_hints` (refactor_patrol.rb:140-141) then filters the scoped set down to empty on any git error, yielding zero theses instead of degrading to the scoped set — inconsistent with the graceful-degradation posture used for churn/leverage.
- [ ] [perf] `changed_files(project_root)` runs `git diff` twice per invocation — once in `run_cycle` (refactor_patrol.rb:66) and again inside `apply_scope_hints` (refactor_patrol.rb:140); memoize the result to avoid the duplicate subprocess.
- [ ] [maintainability] `normalize_thesis` constructs `Thesis.from_h(hash)` twice — once only to stamp the fingerprint (reviewer.rb:106) and again as the return value (reviewer.rb:115); compute the fingerprint from the hash and build the struct once.
- [ ] [correctness] Config permits all leverage weights to be `0` (validation only enforces `>= 0`, config.rb), but an all-zero weight set makes `Leverage#score` emit an empty `breakdown`, which fails the thesis schema's `expected_leverage.breakdown.minProperties: 1` and silently drops every thesis as `schema_invalid`; the shipped defaults are safe, but a zeroed-weights config produces empty output with a pinned SHA and no diagnostic.

ivankuznetsov:
### Dogfood polish pass — evidence contract

Root cause of the 0-accepted-theses dogfood run: the review prompt showed only an empty `{"theses": []}` skeleton, so the agent invented its own (reasonable) evidence spelling — `files` array + `claim` prose, named `signal` without a `value`, `refactor` instead of `proposed_refactor`. The admissibility gate correctly refused it, but the content was genuinely file-anchored: the guardrail rejected the spelling, not the substance.

Fixed in 65398911:

- **Prompt teaches the contract by example** — full worked thesis with the exact evidence item shape (`file` singular + measured `value`), plus explicit rules.
- **Reviewer normalizes honestly recoverable drift** — expands `files`/`paths` arrays per path, anchors file-less evidence to owned files *only when the evidence text literally names them*, backfills a named measurable signal's `value` from the feature's own measured leverage, maps `refactor`/`characterization_notes`/object-`feature` aliases. Path-less evidence stays flagged inadmissible.
- **Tests** replay the captured dogfood shape end-to-end (accepted + schema-valid) and cover text-anchoring; the real run's `theses.json` now yields **3/3 accepted** theses instead of 0.

Focused suites: 317 runs, 1437 assertions, 0 failures. RuboCop clean.

ivankuznetsov:
### Architecture pass — step 1/2: extract `ThesisNormalizer` (ee2b4f72)

**Why.** `Reviewer` had grown to 354 lines mixing four roles: per-feature orchestration, raw-thesis normalization (the alias/evidence repair from the dogfood fix), admissibility+guidance policy, and agent plumbing. Dogfooding showed normalization/policy is the fastest-evolving part, so it now lives in a pure `ThesisNormalizer` (no I/O, no agent coupling) returning a `Thesis` or an `Invalid` carrying schema errors. `Reviewer` (192 lines) keeps orchestration + error recording. Behavior unchanged.

**Live test.** Real-agent `--dry-run --feature command-bin-hive-eval` against the dogfood project: both theses came back **admissible** with per-file evidence and backfilled signal values (the pre-fix run produced 0 admissible). `theses: 0` accepted in the envelope is caps policy, not an evidence failure — the slice is entirely `bin/`-owned, so every thesis is flagged `public_api_impact` under the default caps and surfaces in `ranked`/`flagged_theses` with reasons. Worth deciding later whether `bin/`-heavy slices deserve a caps carve-out.

**Autoreview.** 8 finder angles + verification: no confirmed correctness findings. One confirmed cleanup — ~60 lines of fixtures duplicated across the two thesis test files — fixed via a shared `thesis_fixtures.rb` module. Altitude findings (fingerprint stamping inside the normalizer) refuted: the thesis schema requires `fingerprint`, so stamping must precede validation, which is the normalizer's seam.

Suites: 66 runs / 247 assertions green; RuboCop clean. Next: slim the Reviewer's agent plumbing and decompose `Commands::RefactorPatrol#run_cycle`.

ivankuznetsov:
### Architecture pass — step 2/2: extract `ReviewAgentRunner`, stage `run_cycle` (3daa5e4d)

**Reviewer is now pure orchestration (~130 lines).** The agent spawn + usage recording moved to `ReviewAgentRunner`; the `agent_runner` call protocol is the only seam between them, so tests and future runners swap in without touching orchestration. The awkward const-swapping agent tests now target the runner directly.

**`run_cycle` reads as a pipeline.** `resolve_project!` → `scoped_features` → `score_features` → `guard_theses` → `build_payload`, all following the module-wide `(project_root, cfg, state)` argument order.

**Live test.** Real-agent `--dry-run` through the new runner path: 2 theses parsed, both admissible with file-backed evidence, caps flags applied (`not_single_feature`, `exceeds_max_diff_lines`, `public_api_impact` — this agent run declared bigger scopes), dry-run contract intact, exit 0.

**Autoreview.** 3 verification angles on the diff: zero correctness findings (both the removed-behavior audit and line-by-line scan confirmed exact behavior preservation, including `caps.apply` ordering and `scanned_sha` timing); one argument-order inconsistency confirmed and fixed.

**Closing architecture verdict — happy with it.** Each unit now has one reason to change: `ThesisNormalizer` (what agent output we accept + honest repair), `Reviewer` (per-feature orchestration + error recording), `ReviewAgentRunner` (how agents run + usage), `Commands::RefactorPatrol` (pipeline), and the existing small units (`Leverage`, `Caps`, `Collisions`, `Fingerprint`, `Reporter`, `StateStore`) were already well-factored. Suites: 353 runs / 1608 assertions green, RuboCop clean, wiki fragment added. No further refactoring planned.

ivankuznetsov:
### Public-API semantics fix (05e90381): touching surface files ≠ changing the API

**The principle:** refactoring preserves the public contract by definition — if a proposal changes the public API, it is not a refactor. The caps heuristic conflated *working inside files that host public surface* with *changing the contract*: `candidate_paths` included the feature's whole owned boundary and `public_api_content?` matched any `bin/` path, so every thesis in a `bin/`-owned slice was flagged `public_api_impact` and the slice could never yield an accepted thesis (the dogfood runs' `accepted: 0`).

**Semantics now:**
- Path-heuristic matches (`bin/`, `cli.rb`, `schemas/`, routes, docs) → non-blocking `touches_public_api_surface` **advisory** in the new `risk.advisories` array, with `public_api_details` still naming the files for the human.
- Agent-**declared** `risk.public_api_impact: true` (an admitted contract change) → still flags and blocks acceptance; flag-not-drop keeps it visible as out-of-contract.
- The review prompt states the definition so agents self-select: contract-changing proposals must be declared or dropped.
- `ranked` envelope items carry `advisories` (schema-required); text report appends `advisories=` when present.

**Live result on the same dogfood slice: `accepted: 0 → 2`**, both theses ranked clean with the advisory visible. Suites 356 runs / 1633 assertions green, RuboCop clean, autoreview pass clean.

Note: this commit also carries `docs/dogfood-reports/2026-07-04-…-dogfood.md` — the `/ce-dogfood` report found in the worktree; shout if it wasn't meant to ship with the branch.

PR #680, opened 2026-07-08
Hive patrol: hive-eval deletes report paths before validation finishes

ivankuznetsov:
## Summary

`bin/hive-eval` deleted the selected `--report` path too early: it unlinked the
report immediately after option parsing, before rejecting unexpected positional
arguments or a missing scenario. A malformed invocation whose `--report` pointed
at an existing **non-report** file could therefore delete that file even though
no eval run ever happened, and the parse-error cleanup unlinked the captured
report path unconditionally.

This PR finishes argv and scenario validation before any destructive cleanup and
scopes cleanup to files hive-eval actually owns:

- **`hive_eval_report?(path)`** — treats a path as removable only when it is a
  regular file whose JSON parses to a Hash with `schema == "hive-eval-report"`.
  The `File.file?` guard is deliberate: cleanup now inspects arbitrary
  caller-supplied paths, and a bare `File.read` on a FIFO or blocking device
  would wait for EOF and stall the usage-error message.
- **`remove_report_if_stale(path, default_report)`** — removes a path only when
  it is the default report path or passes `hive_eval_report?`, so arbitrary
  caller files are preserved.
- The unconditional `FileUtils.rm_f(report_path)` now runs **after** all usage
  validation (parse errors, unexpected positionals, scenario resolution, and the
  scenario-exists check), immediately before the eval run. Every usage-error
  exit path routes through `remove_report_if_stale` instead.

Net effect: stale eval JSON from a previous run is still cleared before a real
run, but a usage error can no longer clobber a caller's unrelated file.

## Test plan

- `bundle exec rake test` — passing.
- `bundle exec rubocop` — passing.
- New regression `test_cli_usage_errors_preserve_existing_non_report_files` in
  `test/eval/support/reporter_test.rb` drives `bin/hive-eval` as a real
  subprocess across three usage-error cases — missing option value (`--scenario`
  with no argument), unexpected positional argument, and a missing scenario —
  and asserts each exits `64`, prints the expected message, and leaves the
  caller's pre-existing non-report file byte-for-byte intact.

## Review summary

- Two `codex-native-review` passes. Pass 01 raised one Medium finding
  (auto-fixed): guard the stale-report check against blocking paths by requiring
  `File.file?` before `File.read`, so a FIFO/blocking `--report` target cannot
  stall the usage-error path. Pass 02 reported no findings across all
  severities.
- No High findings, no user escalations, no active suppressions.

## Linked task

- Source: hive patrol finding `command-bin-hive-eval-5`
- Fingerprint: `3fb28c39ecc86baf8c009c16221da55c0282699eda0c7933887db95365b6b65d`
- Task slug: `patrol-command-bin-hive-eval-3fb28c39`


ivankuznetsov:
### Reviewer: codex-native-review - Pass 01

## High
No findings.

## Medium
- [ ] Avoid blocking when checking stale report paths — /home/asterio/Dev/hive.worktrees/.patrol/hive-patrol-command-bin-hive-eval-3fb28c39/bin/hive-eval:88-88 When `--report` points at a FIFO or blocking device and a later usage error occurs (for example a missing scenario), this `File.read` waits for EOF before printing the error; previously the cleanup path returned immediately. Since the cleanup now inspects arbitrary caller paths, guard it to regular files or use a bounded/nonblocking read before…

## Nit
No findings.

PR #706, opened 2026-07-10
feat(config): route model and effort per built-in stage

ivankuznetsov:
## Summary

Operators can now pick the model and reasoning effort per Hive stage in one config block, instead of every agent inheriting a single Claude default. A `models:` map routes each built-in identity — `plan`, `execute`, the review roles, digest, patrol, diagnosis, rebase, and the rest — and the selection is profile-neutral: the same config can run plan/execute on Codex `gpt-5.6-sol` at `xhigh` while fanning review out across Claude Opus and Codex in a single pass. Projects that set nothing keep their previous Claude/descriptor/CLI behavior byte-for-byte.

The setting is deliberately decoupled from Claude. `Hive::AgentProfile` now owns model-control capability and native argv rendering, so Claude emits `--model`/`--effort`, Codex emits `--model gpt-5.6-sol` plus its one-run `model_reasoning_effort` override, and a control aimed at a profile that cannot render it fails at config load rather than silently leaking a Claude-only flag to Codex.

## Resolution model

Each field resolves independently — an exact stage can take its model from itself and its effort from the coarse parent. Only field *absence* continues down the chain:

```mermaid
flowchart TB
  A[Named agent spawn] --> B{Exact models entry has field?}
  B -->|yes| F[Effective model / effort]
  B -->|no| C{Coarse parent has field?}
  C -->|yes| F
  C -->|no| D{Workflow / reviewer supplied field?}
  D -->|yes| F
  D -->|no| E[Profile / Claude global fallback]
  E --> F
  F --> G[AgentProfile validates + renders native argv]
```

The stage vocabulary is a single closed constant (`MODEL_STAGE_PARENTS`) shared by callers, validation, and docs, so a typo cannot silently fall through to a different model.

## What changed

| Area | Change |
|---|---|
| Config | Source-aware `models:` parsing, resolution, and validation. Project config owns pipeline/runtime identities; global config owns `digest` and rejects the inert project-side key. |
| AgentProfile | Native model/effort capability + argv rendering for Claude and Codex; Pi advertises none until a native mapping is added. Legacy raw `cli_flags` stay isolated from profile-rendered controls. |
| Core + review stages | Every core, review, council, and workflow spawn threads its exact identity and `cfg:`. Shared Claude tmux reviewer sessions now group by effective model/effort as well as permission scope, since launch flags are fixed for the session lifetime. |
| Auxiliary runtimes | Digest, patrol, diagnosis, babysitter, and rebase route through the shared resolver while keeping their existing marker/usage/timeout/process-group lifecycles unchanged. |
| Docs + templates | Copyable `models:` examples in the project and global templates, plus wiki/ADR updates recording the profile-neutral architecture and the Codex config source. |

## Design decisions

- **Field-by-field, not entry-by-entry** — a stage may inherit effort from `review` while overriding only its model. Sentinel values (`inherit`, `default`) keep their existing profile-specific meaning; the resolver never assigns cross-profile meaning to strings.
- **Coarse review is a common override, not heterogeneous routing** — mixed reviewer lists still carry profile-specific values on reviewer entries. `models.review` is a deliberate shared override, and Hive validates control *capability* offline, not provider model catalogs.
- **Digest is globally owned** — `Hive::Digest.run` loads the global config, so `models.digest` lives there; project config rejects it rather than accepting an override that would never fire.
- **Backward compatibility is contractual** — `claude_cli_flags(cfg, model:, effort:)` and `spawn_agent` gain optional `stage:` rather than requiring it; no-stage calls ignore the map and stay byte-compatible, and the public custom-`AgentProfile` constructor is preserved.

## Test plan

No paid provider request is required for acceptance — a deterministic mixed-agent integration smoke (`test/integration/mixed_agent_model_routing_test.rb`) drives the real config loader, stage/profile selection, and command builders through fake Claude/Codex binaries that record argv. It asserts negative space as well as presence: no Claude-only flag in Codex logs, no Codex config override in Claude logs, and no cross-reviewer value reuse. Live billable smoke against real accounts is deferred to follow-up.

Full gate on the supported Ruby matrix:

- Tests: 7,437 runs, 76,228 assertions, 100% line coverage
- RuboCop: 797 files, no offenses
- Brakeman: passed
- bundler-audit: no vulnerabilities
- Worktree: clean

## Review summary

Two review passes ran (claude-ce, codex-ce, and pr-review-toolkit personas). Every finding was auto-fixed or resolved from the plan/brainstorm/prior decisions — none required a user decision.

- **Pass 1** — 13 AUTO-FIX, 4 RESOLVED/NO-FIX. Fixed a deterministic test failure (budget warning wrote `config-warnings.log` before the capability check raised), dropped a dead `respond_to?(:render_model_controls)` guard for consistency with the digest path and the 100% gate, guarded the nil-cfg/no-stage `spawn_agent` path to stay argv byte-compatible, moved Codex `--model`/`-c` controls before the `review` subcommand (Codex 0.144.0 rejects post-subcommand globals), reordered config validation so a bad `plan.agent` raises `ConfigError` not `NoMethodError`, and corrected the copyable Codex/digest template examples to select `agent: codex`.
- **Pass 2** — 9 AUTO-FIX, 3 RESOLVED/NO-FIX. Added load-time `claude.effort` vocabulary validation (R6), a positional test pinning Codex `exec`-path control ordering, a shared-constant for Claude's render-nothing effort sentinels, a both-or-neither renderer/values constructor guard, and a reciprocal drift guard for the project/global stage partition, plus argv-seam routing tests for the review roles.

The codex-ce reviewer hit its usage limit partway through pass 2 (recorded in `errors-02.md`); the remaining personas covered the diff. No findings were escalated to the user.

## Linked task

- Hive task `1846` — `per-stage-claude-model-config-260709-35f7` ("Per Stage Claude Models").
- Origin: idea captured 2026-07-09 while driving hive-bench v3 with fable-plans/opus-executes, where the global-only `claude.model` forced a config flip between invocations for mixed-model pipelines. This PR generalizes that request into profile-neutral per-stage routing.

---

Claude Code


ivankuznetsov:

## High

- [ ] Native Codex review places `--model` after the `review` subcommand (`lib/hive/reviewers/codex_review.rb:475`): real Codex 0.144.0 rejects `codex review --model ...`, so any routed reviewer model makes this required review surface exit before running.
- [ ] Legacy cfg-less spawns now gain `--model default` (`lib/hive/stages/base.rb:536`): unconditional resolution from `cfg || {}` changes the public no-config/no-stage argv despite the plan's byte-compatibility requirement.
- [ ] Coarse review validation checks profiles that cannot inherit the field (`lib/hive/config.rb:1823`): it rejects valid configs for disabled roles and for children whose exact fine-stage value shadows the incompatible coarse value, contradicting field-by-field reachability.
- [ ] The advertised Codex plan example still selects Claude (`docs/architecture.md:115`, `templates/project_config.yml.erb:38`): copying or uncommenting the example sends `gpt-5.6-sol`/`xhigh` to Claude because `models:` does not change the selected agent profile.

## Medium

- [ ] Claude effort fallback is narrowed to a hard-coded enum (`lib/hive/agent_profiles/claude.rb:39`): values outside today's list used to pass through and are still documented as open, but now raise before Claude can interpret a newly supported tier.
- [ ] Council precedence documentation remains stale (`wiki/modules/workflows.md:39`, `wiki/stages/council.md`): it still says descriptor model/effort overrides project config and omits the new `review_reviewers` route and pre-fan-out capability validation.

## Nit

- [ ] Model capability errors omit the config source (`lib/hive/config.rb:1802`): invalid effort or unsupported-profile messages name the key but not the project/global config path required by the validation contract.
- [ ] The Agent wiki omits the new model-control channel (`wiki/modules/agent.md:15`): its constructor shape lacks `model_control_flags:` and still describes legacy Claude-only `cli_flags` as the model/effort mechanism.
- [ ] The Config wiki's API inventory is incomplete (`wiki/modules/config.md:211`): it documents only the old `claude_cli_flags(cfg)` signature and omits the generic resolver/rendering entry points and `models` validation surface.

ivankuznetsov:

# Code Review — per-stage model/effort routing (pass 1)

Scope: `git diff origin/main..HEAD` (72 files, ~1433/-124). Implements U1-U6 of the
per-stage `models:` routing plan. Focused unit + integration suites run green
locally (config, agent_profile(s), spawn_agent, run_reviewers, digest/categorizer,
diagnosis_agent, codex_review, descriptor_parser, mixed_agent_model_routing).
Every named spawn site the plan enumerates (core stages, review substages,
council, reviewers, digest, patrol, diagnosis, babysitter, rebase) threads a
stage identity; shared-tmux reviewer grouping correctly includes effective
model/effort; the legacy Claude-only `cli_flags` channel is kept isolated from
the new `model_control_flags` channel. No plan-required-but-missing gap found.

## High

_None._

## Medium

- [ ] Global `claude.effort` vocabulary silently narrowed and validated at spawn instead of load (`lib/hive/agent_profiles/claude.rb:39` `effort_values`, via `lib/hive/config.rb:781` `claude_cli_flags` → `render_model_controls`): a previously-open value (e.g. a future/undocumented Claude tier) passes `Config.load` but then raises `Hive::ConfigError` on the first Claude spawn mid-run. Verified: `claude.effort: "turbo"` → `LOAD OK` then `RENDER RAISED: Hive::ConfigError ... must be one of ["default","inherit","low","medium","high","xhigh","max"]`. This contradicts the plan's explicit intent (U1 "leave existing global/native Claude value semantics untouched"; risk table "keep old global Claude effort behavior backward compatible"; R6 "invalid efforts fail during configuration or descriptor loading" — i.e. at load, not spawn). The still-present comment at `lib/hive/config.rb:778-780` ("any other tier Claude Code accepts passes through verbatim ... the config key is open") is now stale/wrong. Either validate the global `claude` block at load time too, or exempt the global block from the closed vocabulary to preserve passthrough.

## Nit

- [ ] Dead capability guards `profile.respond_to?(:render_model_controls)` at `lib/hive/diagnosis_agent.rb:88`, `lib/hive/patrol/reviewer.rb:162`, `lib/hive/patrol/fixer.rb:151`: `render_model_controls` is defined on the base `AgentProfile`, so the branch is always true and the `: []` fallback is unreachable. `Config.agent_control_flags` already renders `[]` for profiles without a renderer, so the guard adds nothing and is inconsistent with `lib/hive/digest/categorizer.rb:288` which calls it unguarded.
- [ ] `spawn_agent` with `cfg: nil` on a Claude profile now emits `--model default` where it previously emitted no model flag (`lib/hive/stages/base.rb:540-545`): the old code gated cfg-derived flags on `cfg` being truthy, so nil-cfg and `{}`-cfg diverged; now both resolve to `--model default`. No production caller passes nil cfg (all thread `cfg:`), so impact is limited to callers/tests relying on the prior nil-cfg silence — worth a note, not a blocker.

ivankuznetsov:

# PR Review — per-stage model/effort routing

Scope: `git diff origin/main..HEAD`. Grok changes (`lib/hive/agent_profiles/grok.rb`,
`test/unit/grok_preflight_test.rb`, `wiki/log.d/…grok-auth-path-preflight.md`) are
stale-base diff noise — origin/main #713 landed after this branch's merge-base (#712)
and no commit here touches grok — so they were excluded from review.

Note on false positives dropped after verification: the "invalid effort not rejected at
config-load when a stage's `agent` is unset" claim was disproven — `merge_defaults`
populates each stage's default agent before `validate_models!`, so a bogus effort under
`models.brainstorm` (agent unset) raises `Hive::ConfigError` at `Config.load`, identical
to the explicit-agent case. Verified by reproduction against the real loader.

## High

## Medium
- [ ] [comments] `lib/hive/config.rb:30-32` and `:783-785` — Claude `effort` comments still claim open pass-through ("any other value passes through verbatim … change without notice" / "the config key is open"): `agent_profiles/claude.rb:39` now pins a closed `effort_values` list and `validate_effort_control!` raises on anything outside it, so a newly-released CC tier would be rejected until the array is updated — the comment describes behavior this change removed.
- [ ] [comments] Template model examples route a Codex model to the default Claude profile: `templates/project_config.yml.erb:36-49` ("runs planning and implementation on Codex SOL") and `templates/hive_config.yml.erb:45-48` set `model: gpt-5.6-sol` but never set the stage/digest `agent:` to codex; copied verbatim, `plan.agent`/`digest.agent` stay `claude`, model names aren't validated offline, so config loads and then hands `--model gpt-5.6-sol` to Claude at runtime — the example is incomplete versus its stated effect.
- [ ] [type-design] `lib/hive/config.rb:1822` `model_profiles_for_stage` is a second hand-maintained `case` over the stage vocabulary that `MODEL_STAGE_PARENTS` already owns; a stage added to the constant but missed here falls through `else => []` and silently skips profile validation with no load-time signal — the plan's own risk table calls for one closed source of identities.
- [ ] [tests] `test/unit/claude_cli_flags_test.rb` / `config_test.rb` prove exact-beats-coarse only for `model` (`review_fix.model` > `review.model`); no test asserts an exact `effort` overriding a coarse `effort` on the same field, so a resolver that reversed exact/coarse priority for effort specifically would pass the suite (R3/R9 precedence edge).
- [ ] [tests] The Claude-only blank→`inherit` coercion in `agent_control_flags` (`lib/hive/config.rb:~817`) has no direct test; all cli-flags/integration cases use non-blank values, so removing that transform (which would let a blank-resolved control reach `render_model_controls` and raise) is undetected by current tests.
- [ ] [tests] `test_model_profile_targets_cover_every_static_stage_family` (`config_test.rb:97`) omits `plan`, `open_pr`, the `execute` family, and `digest`; the digest branch's three-way fallback (`digest.agent || patrol.agent || "claude"`) and those `case` arms are unverified, so a `cfg.dig` typo there would skip validation undetected — this is the test gap that would catch the drift in the type-design finding above.

## Nit
- [ ] [errors] Dead defensive guard `profile.respond_to?(:render_model_controls) ? … : []` in `lib/hive/diagnosis_agent.rb:88`, `lib/hive/patrol/fixer.rb:151`, `lib/hive/patrol/reviewer.rb:158` — every `AgentProfile` defines the method, so the `[]` branch is unreachable in production and silently masks a stubbed/misconfigured profile; inconsistent with `digest/categorizer.rb:288` which calls unconditionally.
- [ ] [type-design] `lib/hive/agent_profiles/claude.rb:36-39` `effort_values` flattens sentinels (`default`, `inherit`) and real tiers into one array; the "renders nothing" semantics live only in the renderer lambda, so the value vocabulary and render behavior must be kept in sync by hand.
- [ ] [comments] `templates/hive_config.yml.erb:40` says digest `defaults to patrol.agent, then claude` yet the adjacent example (`:45-48`) assumes a Codex agent — add a one-line "assumes a Codex `digest.agent`" note to prevent copy-paste misroute (effort `xhigh` is valid on both profiles; only the model name is agent-specific).
- [ ] [tests] Weak-coverage sub-cases: same-model/differing-`effort` reviewer specs are not proven to split shared Claude tmux sessions (`run_reviewers_test.rb` covers differing-model only), and the Symbol-value type path in `validate_models!` and the "supports model but not effort" branch of `validate_effort_control!` are untested.
- [ ] [code] Branch is behind `origin/main` (merge-base #712); rebasing onto current `origin/main` before merge drops the grok #713 reversion noise from the diff and keeps the review/merge clean.

ivankuznetsov:

# Code Review — per-stage model & effort routing (pass 2)

Scope: `git diff origin/main..HEAD` (merge-base `acf1a74f`), ~1587 exec lines of Ruby.
Verdict: Ready with fixes. The change faithfully implements plan units U1–U6 (config schema/resolver/validation, profile-native rendering, stage threading, auxiliary agents, docs, mixed-agent smoke) with strong tests including negative-space assertions. No shipping-breaking defect found; the items below are correctness/back-compat refinements and coverage gaps.

## High
_(none — no plan-required item is missing and no normal-path crash/data-loss was found.)_

## Medium
- [ ] Claude `effort` closed-vocabulary validation fires at spawn/render time, not config load (`lib/hive/config.rb` `claude_cli_flags`/`agent_control_flags` -> `render_model_controls`; `lib/hive/agent_profiles/claude.rb:effort_values`): there is no load-time validator for the top-level `claude.effort`/`claude.model` block (only `models:` entries and reviewer entries are load-validated), so a `claude.effort` outside `%w[default inherit low medium high xhigh max]` passes `Config.load` and then raises `ConfigError` mid-pipeline at the first Claude spawn — and it also narrows the previously-open pass-through (old `claude_cli_flags` emitted `--effort <any tier>` verbatim). This is in direct tension with the plan's own mitigation ("validate the effective target set **at load time**; keep old global Claude effort behavior backward compatible") and the CHANGELOG's "Existing configs need no migration." Fix: validate the top-level `claude` block controls in `validate!` at load, or document the narrowing.
- [ ] Codex `effort_values` includes `none` and `xhigh` (`lib/hive/agent_profiles/codex.rb:39`) which diverge from the Codex `model_reasoning_effort` vocabulary the ADR itself cites (`minimal|low|medium|high`) with no justifying comment — unlike the Claude list, which carries a hand-maintained-sync note: a value Hive accepts (`effort: xhigh`, the exact value used across all new docs/templates/tests) could be rejected or ignored by the real Codex CLI at spawn, and the whole test suite masks it because `test/fixtures/fake-codex` accepts any value. `xhigh` is plan-mandated so this is a verify-and-annotate item, not a behavior change: confirm the accepted set against the pinned Codex (>=0.144.0) and add a comment pinning the source, mirroring the Claude sentinel comment.
- [ ] `command:` reviewers silently drop declared `model:`/`effort:` (`lib/hive/workflows/descriptor_parser.rb:221` and `lib/hive/config.rb` `validate_council_model_controls!` `next if reviewer.command`): `parse_reviewer` accepts both `command` and `model`/`effort` on the same reviewer, but both validators skip command reviewers, so the controls are neither validated nor rendered — a silent drop that contradicts this feature's stated "validate instead of drop" principle. Fix: when a `command:` reviewer (or revise) also sets `model`/`effort`, raise a descriptor error rather than skipping.
- [ ] R7 no-context byte-compat guard is not exercised through `spawn_agent` (`lib/hive/stages/base.rb` `if cfg || !model.nil? || !effort.nil?`; test should live in `test/unit/spawn_agent_test.rb`): the byte-identical goldens construct `Hive::Agent.new` directly and default `model_control_flags` to `[]`, so no test drives `Stages::Base.spawn_agent(task, cfg: nil, model: nil, effort: nil)` and asserts the control argv stays empty. Flipping that guard to unconditional would silently leak `--model default` onto every legacy no-context spawn with no test going red.

## Nit
- [ ] Shared-reviewer tmux grouping keys on the raw resolved controls, pre the empty-string->`inherit` transform (`lib/hive/stages/review.rb:1620-1626` `shared_reviewer_groups` vs the transform in `agent_control_flags`): two Claude reviewers whose effective values differ only as `""` vs `inherit` (both render to no `--effort`) split into separate sessions unnecessarily. It never causes model bleed (equal raw controls always render equal flags), and the trigger is narrow (an empty-string `claude.*` control). Optional: key the group on the rendered flags for exact parity with launch.
- [ ] An unknown *string* agent name inside a `models.<stage>` entry raises the generic "unknown agent profile" error from `validate_models!` before the dedicated `validate_role_agent_names!`/`validate_reviewers!` run (`lib/hive/config.rb` `model_profiles_for_stage` `lookup`): still a typed `ConfigError`, but a less specific message than the ordering was designed to produce. Rescue `UnknownAgent` and `next`, or order the role validators ahead of `validate_models!`.
- [ ] Council `stage:` identity threading and the `revise` branch are unasserted (`lib/hive/stages/council/reviewer.rb` `stage: :review_reviewers`, `lib/hive/stages/council/revise.rb` `known_model_stage?(...) ? name : nil`, and the revise arm of `validate_council_model_controls!`): unlike rebase/babysitter/plan/execute/reviewers, no test asserts the identity handed to the resolver for the two council spawn sites, and the pre-fanout validation test only covers the reviewer loop, not the revise agent.
- [ ] Codex `exec` (non-review) control ordering is unpinned (`lib/hive/agent.rb` `build_cmd` places `@model_control_flags` after `exec`, the mirror of `codex_review.rb` placing them before `review`): tests assert the exec-path flags are present but never assert position relative to `exec`, while the review-path ordering IS pinned. Not a bug today (real Codex tolerates post-`exec` controls) but the two hand-encoded opposite orderings could drift.

ivankuznetsov:

## High
- [ ] Validate project/global `claude.effort` during config loading (`lib/hive/config.rb:1728`): an unsupported tier currently passes `Config.validate!` and raises only when a Claude spawn renders flags, violating R6 and the documented load-time contract.
- [ ] Resolve nil built-in agent names exactly as runtime does (`lib/hive/config.rb:1895`): `model_profiles_for_stage` compacts `plan.agent: null` away while `stage_profile` falls back to Claude, so a Claude-invalid `models.plan.effort` loads successfully and fails at spawn.
- [ ] Validate descriptor controls against the project-selected profile (`lib/hive/workflows/descriptor_parser.rb:213`): omitted descriptor agents are hardcoded to Claude in parser/council preflight, but runtime can select `cfg[stage].agent`, causing valid Codex controls to be rejected or validated against the wrong CLI.

## Medium
- [ ] Make the project-template Codex example safe to uncomment (`templates/project_config.yml.erb:45`): its duplicate `plan`/`execute` keys are followed by the live mappings at lines 94-104, whose last-key-wins values override Codex and can route `gpt-5.6-sol` back to Claude.
- [ ] Update the workflow and council wiki contracts (`wiki/modules/workflows.md:39`): they still say descriptor model/effort overrides project config and omit `models.review_reviewers` precedence plus profile-aware council preflight, contradicting the changed runtime and mandatory wiki policy.
- [ ] Cover diagnosis routing through the real config-to-command seam (`test/unit/diagnosis_agent_test.rb:445`): the test injects `@model_control_flags` directly, so U4/R9's required proof that `models.diagnose` overrides `models.execute` while preserving plain output is absent.

## Nit
- [ ] Record the deferred live-routing uncertainty in `wiki/gaps.md`: the feature leaves this mandatory file untouched, and its existing Claude-global flag gap does not cover per-stage mixed-provider routing or Codex reasoning overrides.
- [ ] Move the model-routing changelog note out of the already released `0.3.7` section (`CHANGELOG.md:5`): the feature commits postdate tag `v0.3.7` and current `origin/main` is `v0.4.0`, so the entry currently claims availability in a version that never contained it.

ivankuznetsov:

# PR Review (Pass 2) — per-stage model/effort routing

Scope: `git diff origin/main..HEAD`. Union of five personas (code, silent-failure,
tests, type-design, comments). Grok changes (`agent_profiles/grok.rb`,
`grok_preflight_test.rb`, grok wiki log) are stale-base diff noise (branch behind
`origin/main`) and were excluded. Pass-1 fixes verified present and not re-reported:
closed-effort comment scoping, template Codex `agent:` examples, `model_profiles_for_stage`
exhaustiveness guard, and the added precedence/coercion/session-split tests.

Overall: correctness posture is strong — the design consistently fails loud (nil renderer
raises "unsupported" rather than no-op; unknown stage raises; `model_profiles_for_stage`
`else` raises on drift; `validate_models!` rejects nil/blank/wrong-type/unknown-field at
load). `code-reviewer` and `comment-analyzer` found no High/Medium defects. Remaining items
are maintainability and test-coverage gaps.

## High

## Medium
- [ ] [tests] No config-load test rejects a real profile's out-of-vocab effort (`test/unit/config_test.rb` validate_models! cases): `validate_effort_control!`'s vocab guard (`agent_profile.rb:234-237`) is exercised only against a synthetic profile with `effort_values: %w[low high]` (`agent_profile_test.rb:215`), so a drift in Codex/Claude `effort_values` (e.g. Codex silently accepting `max`) would ship untested against R6/R9.
- [ ] [tests] Descriptor per-stage `model`/`effort` never asserted to reach a spawned subprocess argv: `descriptor_parser_test.rb:74-99` proves round-trip and unsupported-control rejection, but `stages/agent.rb:50-51` threads `stage.model`/`stage.effort` into `spawn_agent` with no test proving those become native child flags — the one R9 config-to-subprocess path covered only at the parse boundary.
- [ ] [type-design] Claude effort sentinel set duplicated across two hand-synced literals (`agent_profiles/claude.rb:37` and `:44`): the `effort_renderer` render-nothing branch (`%w[default inherit]`) and `effort_values` (`%w[default inherit low medium high xhigh max]`) encode the same sentinel knowledge; the comment concedes they "must stay in sync by hand" — a sentinel added to one but not the other silently mis-renders (e.g. `auto` in values without the branch emits `--effort auto`).
- [ ] [type-design] Illegal AgentProfile capability combinations are representable at construction (`agent_profile.rb:100-134`): a profile may pass `effort_values:` without `effort_renderer:` (vocabulary silently unused) or `effort_renderer:` without `effort_values:` (silently open vocabulary), yet the constructor validates `prompt_style`/`status_detection_mode` against closed sets but adds no cross-field guard for the new renderer/values trio.
- [ ] [type-design] Project/global stage partition lacks a reciprocal drift guard (`config.rb` `PROJECT_MODEL_STAGES`/`GLOBAL_DIGEST_MODEL_STAGES` ~:639-640): unlike `model_profiles_for_stage` (now guarded), the split is encoded as a literal `- ["digest"]` subtraction, so any future global-only stage added to `MODEL_STAGE_PARENTS` would be silently misclassified into project scope; derive the partition from a per-entry attribute rather than a name subtraction.

## Nit
- [ ] [silent-failure] Runtime resolver relies on load-time validation for its non-blank invariant (`config.rb:837-843`): `resolve_model_control_field` treats an explicit `models.<stage>.<field>:` (nil in YAML) as "present, use nil" → renders no flag; unreachable because `validate_models!` (`config.rb:1828-1837`) raises on nil/blank first, but worth a one-line note documenting the dependency.
- [ ] [silent-failure] `validate_entry_model_controls!` does not skip `kind: "linter"` reviewer entries (`config.rb:2252-2260`), unlike `reviewer_agent_names` (`config.rb:1908-1910`): a linter entry with a stray `model:`/`effort:` is validated against `entry["agent"] || "claude"` — stricter (fails loud) not looser, but a minor asymmetry between the two reviewer-iteration helpers.
- [ ] [tests] `review_ci`/`review_triage`/`review_fix`/`review_browser` have no argv- or `model_control_flags`-level routing test (`ci_fix_test.rb`, `triage_test.rb` carry zero model/effort assertions): covered only at the resolver level (`config_test.rb:97-135`), unlike patrol_review/patrol_fix which each have an exact-beats-coarse seam test.
- [ ] [tests] No explicit assertion that a Codex spawn does NOT coerce a blank-resolved control to `inherit` (the coercion at `config.rb:812-816` is Claude-gated): `claude_cli_flags_test.rb:78` pins the Claude side but the Codex-raises-instead negative is only covered indirectly.
- [ ] [type-design] Model/effort controls are validated twice — at config load and again in `render_model_controls` (`agent_profile.rb:240-246`, `config.rb:1842`) — with no shared normalized result threaded through, so `.strip` normalization is recomputed and discarded at render; benign duplication.
- [ ] [type-design] `construction_kwargs` mirrors the `initialize` kwarg list 1:1 by hand (`agent_profile.rb:359-382`): the three new fields are correctly present, but a future field forgotten here would be silently dropped by `with_overrides` — standing maintenance hazard, not a live bug.

ivankuznetsov:
Closing after fresh-main verification. This head still reproducibly fails StagesBaseUsageTest#test_profile_without_model_capability_fails_instead_of_dropping_controls because spawn_agent writes config-warnings.log before model-capability validation raises. More importantly, current main now contains the durable implementation-identity/profile contract from #773 across the same AgentProfile, launcher, config, stage, patrol, and test surfaces; merging main into this branch produces broad semantic conflicts rather than a mechanical refresh. The remaining review findings around profile selection and load-time validation are therefore being rejected with this stale implementation. If per-stage routing beyond implementation-owning stages is still wanted, it should be rebuilt as a fresh-main follow-up on the #773 identity APIs.

PR #759, opened 2026-07-16
feat(routing): add provider and model circuit routing

ivankuznetsov:
## Summary

Hive now avoids retry storms when a provider account or model becomes unavailable. Every provider-backed dispatch passes through one deterministic, capability-aware router backed by durable global circuits and attempt leases: healthy work selects the first eligible entry in its declared pool, while open circuits, concurrency caps, strict pins, quality floors, context needs, tools, and permissions produce either a safe fallback or an explicit wait reason.

- Provider identity is separate from the CLI adapter, so account-wide failures and model-scoped failures have the correct blast radius. Adapter-owned normalizers classify quota, rate, session, credit, authentication, permission, billing, context, transient, and unknown outcomes without exposing raw provider text to circuit policy; unknown and transient outcomes do not open shared circuits.
- Timed circuits admit exactly one durable real half-open probe. Administrative failures remain open until explicitly cleared, corrupt durable state fails closed, and successful recovery affects new dispatches without preempting fallback work already in flight.
- Provider concurrency and workflow-recovery deduplication use process-shared durable leases. Generic resumable workflows, including the built-in `bench` adapter, resume only pending/provider-retryable children from durable checkpoints while preserving completed artifacts and terminal failures; conversational state is never transferred across providers.
- Operators can inspect human or JSON circuit state, retry timing, probe ownership, and observed/configured concurrency with `hive circuits`, then perform a reasoned provider/model clear when administrative recovery is confirmed.

Existing configuration remains compatible: without `routing.pool`, the resolved `agent`/`model`/`effort` is synthesized as a pool of one with the prior dispatch behavior.

Session-settled decisions carried from planning: provider-account/model health boundaries, global durable circuit truth, strict ordered selection and pins, adapter-owned normalization, one real half-open probe, durable attempt leases, checkpoint-only cross-provider recovery, and recovery through the current routing policy (all user-directed).

## Test plan

- Full suite passed: **7,670 runs, 117,813 assertions, 0 failures, 0 errors**.
- Line coverage passed at **100.00% (33,866/33,866)**; RuboCop inspected **838 files with no offenses**.
- Deterministic fake-clock replays cover Claude session-limit fallback/probe recovery and Grok credit exhaustion with restart-safe `bench` recovery, immutable completed artifacts, terminal-child preservation, and exact attempt counts proving no retry storm.
- Multi-process coverage verifies circuit-store atomicity, provider caps, lease heartbeats/reclamation, single-winner half-open probes, and workflow/checkpoint recovery deduplication across restarts.
- Negative coverage preserves strict-pin and quality-floor waits, indefinite administrative opens, task-local context mismatch, bounded unknown/transient retries, and legacy pool-of-one behavior. The suite is hermetic and requires no paid-provider calls.

## Review summary

Review completed in pass 01 with no reviewer findings or escalations; triage required no user questions. The fix pass completed successfully. Browser review was skipped because this change has no web surface.

## Linked task

Hive pipeline task **9770**: `add-provider-and-model-circuit-260716-0c3b` (Capability-Aware Provider Circuit Breakers).

---

Claude Code


ivankuznetsov:
Fresh review against current `main` found this implementation is not merge-safe despite its old green checks.

Material blockers:

1. **Half-open probes can wedge forever.** `ProviderRouting::Circuit.claim_probe` persists only `attempt_id`, owner text, and `claimed_at`; `Store#availability_for` treats `half_open` as unavailable, but there is no probe expiry or owner-liveness reclamation. If the worker dies after `claim_probe` and before `probe_succeeded` / `probe_failed` / `abandon_probe`, that provider/model stays permanently half-open until manual clear.
2. **Context fallback is not actually resumed.** `Stages::Base#finish_routed_attempt` writes `ERROR reason=routing_context_mismatch` plus the exclusion fields, and `routing_exclusions` can read them on a later run. But no healer/requeue path recognizes `routing_context_mismatch`, so the task parks at the error marker instead of automatically selecting the next eligible provider as the PR promises.
3. The branch is now deeply stale against the durable-attempt/implementation-identity architecture on `main`: a fresh integration has semantic conflicts across the agent/profile implementations, stage routing, daemon dispatcher/healer, workflow descriptors, paths/schema registry, and their tests (more than twenty files).

The claimed no-findings review has no reviewer artifact/comments, and the old CI predates these current contracts. Closing rather than guessing through the newer ownership model. A fresh-main follow-up needs generation-bound probe claims with expiry/liveness recovery, a durable retry transition for context exclusions, and concurrency/lease tests against the current attempt store.

PR #676, opened 2026-07-06
Hive patrol: GNU timeout probe can still hang hv

ivankuznetsov:
## Summary

Hardens `bin/hv` — the fallback launcher that probes candidate `hive` binaries when Apache Hive shadows our binary on `PATH` — so a misbehaving candidate can no longer hang the probe past the documented 5-second cap.

- **Route every probe through the watchdog/sweep path** (`6f52fc7`): the GNU `timeout` command-substitution shortcut is removed. Previously a bad candidate could fork a child that inherited stdout; when the wrapper timed out, the child kept the command-substitution pipe open and `hv` waited on it well past the cap, orphaning the child. All probes now go through the temp-file + process-group watchdog/sweep path, so a forked stdout-inheriting helper can no longer hold the caller's pipe open.
- **Reject a candidate that prints a valid semver then hangs** (`62e0809`): with the always-on watchdog, a candidate whose `hive --version` printed a bare semver and then hung — trapping `TERM` and exiting 0 — was accepted, because `wait` recorded status 0 and the semver matched. `probe_version` now writes a sentinel the moment the watchdog's arming delay elapses and forces a non-zero status (124, mirroring GNU `timeout`) after teardown whenever the sentinel is present, so the hung candidate is rejected and `hv` falls through to the next one. The fast path is untouched — a probe that finishes first tears the watchdog down before the sentinel is written.

## Test plan

- `bundle exec rake test` — passed
- `bundle exec rubocop` — passed

Two regression tests added to `test/unit/hv_test.rb` (both pass locally):

- `test_candidate_that_prints_semver_then_hangs_is_rejected_not_exec_d` — with `timeout` present, the first candidate prints `1.2.3` then hangs and traps `TERM` with `exit 0`; `hv` must fall through to the healthy candidate and leave no child behind.
- `test_timeout_present_probe_tree_uses_watchdog_and_does_not_leak_stdout_pipe` — a candidate that forks a stdout-inheriting, `TERM`-ignoring child must not hold the probe's pipe open past the cap.

## Review summary

Two automated review passes. Pass 1 auto-fixed one Medium finding (the semver-then-hang acceptance regression, addressed by `62e0809` and its regression test). Pass 2 (`codex-native-review-02`) reported no High/Medium/Nit findings, and no escalations required user input. Stage exited `review_complete pass=2`.

## Linked task

Hive patrol finding `command-bin-hv-1` (category `bug`, severity `medium`, confidence `high`).

[...]

PR #781, opened 2026-07-17
Hive patrol: GNU timeout does not bound probes that leave stdout-owning children

ivankuznetsov:
## Hive Patrol Finding

Slice: `command-bin-hive-e2e`
Category: `bug`
Severity: `medium`
Confidence: `high`
Fingerprint: `cabfe0afe74c5755f5c8309eca632dd5fbf26371b7371fb80fcd4107e89772fb`

On the normal GNU-timeout path, `hv` wraps only the candidate process. If `<candidate> --version` forks a child that inherits stdout and the direct process exits, `timeout` returns without killing that child, while the surrounding command substitution continues waiting for the inherited pipe to close. A candidate can therefore hang `hv` indefinitely and leave an orphan despite the documented five-second bound; the process-group watchdog used when `timeout` is absent does not have this flaw.

## Evidence

- bin/hv:35 "$timeout_bin" 5 "$1" --version </dev/null 2>/dev/null
- bin/hv:117 if ! version="$(probe_version "$candidate")" || [[ ! "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then

## Applied Fix

Use the process-group/watchdog probe implementation on all platforms, or explicitly launch each probe in a dedicated group and terminate/reap that whole group after the direct child exits as well as on timeout. Add a regression test with a candidate that forks a stdout-inheriting sleeper and exits immediately while the system `timeout` command is available.

## Validation

## Diffstat

```text
 bin/hv                                             | 15 ++----
 test/unit/hv_test.rb                               | 60 +++++++++++++++++++++-
 .../20260717T201408Z-hv-unified-probe-watchdog.md  | 16 ++++++
 wiki/operating.md                                  |  4 +-
 wiki/testing.md                                    |  2 +-
 5 files changed, 82 insertions(+), 15 deletions(-)

```


ivankuznetsov:
Closing as stale/superseded after verifying against freshly fetched `origin/main@3c1c7512`.

The PR was generated from parent `069bba06`, not current main. Current main already contains the same process-group/watchdog fix from #676, plus a timeout-fired sentinel that preserves GNU `timeout`'s non-zero overrun semantics. It also already includes `test_timeout_present_probe_tree_uses_watchdog_and_does_not_leak_stdout_pipe` for the exact stdout-owning child case.

Fresh-main verification: `bundle exec ruby -Itest test/unit/hv_test.rb` → 14 runs, 53 assertions, 0 failures, 0 errors, 0 skips. Applying this stale patch would duplicate coverage and discard newer hardening.

PR #972, opened 2026-08-10
perf(attempts): bound durable storage history

ivankuznetsov:
## Summary

- Make daemon cost depend on live and unresolved attempts instead of total attempt history. `Store#scan` now sees only hot records; permanent schema-v3 proof and decision indexes remain available through point lookups.
- Finalize completed work through durable, restart-safe acknowledgements before removing it from the hot set. Closed raw logs move to a digest-sharded cold archive and expire after three days, or earlier after canonical task archival.
- Cut existing stores forward from the physical v2 layout to v3 before admission. The cutover is fenced, crash-resumable, corpus-verified, and intentionally has no dual reader or reverse hydration path.

Session-settled decisions carried from planning: separate live scheduler state from permanent history, and rotate raw logs after three days or earlier canonical archival.

### Performance evidence

| Workload | Before | Final-head proof |
|---|---:|---:|
| Ordinary reconcile with historical attempts | 2,677 records caused repeated full scans and about 59 MiB retained RSS growth | 30,000 cold proofs cause one hot scan and zero cold proof opens |
| Hourly cold-log maintenance | Historical directory cardinality could become the work unit | 30,000 cold logs examine at most 512 IDs per run through a persisted 256-shard cursor |
| Admission during one daemon tick | Capacity and recovery could rescan attempt history | One reconciler snapshot is reused and updated under the admission lock |

## Test plan

- [x] Focused final-head storage, migration, command, and schema tests: 296 runs / 1,568 assertions.
- [x] Component-boundary tests: 46 runs / 678 assertions.
- [x] Review-regression tests for restart idempotency, duplicate successors, maintenance isolation, migration custody, expired replay, and bounded cold sweeps.
- [x] RuboCop passed for every changed Ruby file; `git diff --check` passed.
- [x] Broad local storage-head checkpoint: 12,336 runs / 157,997 assertions, zero failures or errors, and 12 expected skips.\n- [x] Storage-head coverage gate: 100.00% line coverage (79,880 / 79,880).\n- [x] Final test-isolation head: exact leaking order reproduced red-to-green; two affected files pass at 17 runs / 202 assertions.
- [x] Hosted CI on exact final head: all 23 checks passed, including the complete suite, 100% coverage gate, packaging, platform, and dedicated merge gates.

## Notes for reviewer

- Permanent proof remains the existing immutable attempt schema v3. This PR changes physical ownership and lookup paths, not the record contract.
- Invalid records, unresolved losses, and incomplete downstream obligations remain hot and reserve capacity fail-closed.
- Startup remains migration-aware so existing user stores are cut over automatically. After the fence is published, runtime storage reads only v3; older binaries are prevented from opening the split store.
- Raw-log retention is age- and task-archive-bounded. This deliberately does not add the speculative byte ceiling from an earlier plan draft.
- Foreground dispatch does not fail when opportunistic maintenance fails. `attempt_storage` reports the bounded migration or maintenance failure for operator action.

## Post-deploy monitoring and validation

- During the next dogfood cycle, watch `hive status --operational --json` for `attempt_storage`, `maintenance_failed`, and `migration_failed`. Also confirm task advancement, daemon RSS, and tick latency.
- Healthy means layout generation 3 is complete, invalid hot count is zero, the hot count reflects only actionable work, hourly maintenance completes, and no storage warning interrupts dispatch.
- On failure, stop the candidate daemon and preserve the v2 fence, v3 root, cutover checkpoint, proofs, and logs. Do not downgrade or reconstruct the old layout; diagnose the preserved state before retrying.

---


ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 13, 3am (Europe/London)
```

PR #983, opened 2026-08-11
feat(routing): add provider health and route selection

ivankuznetsov:
## Summary

- Adds opt-in provider-account routing. A routed attempt selects the first compatible, healthy, unsaturated route in configured order; stages without a routing pool keep the existing provider path.
- Adds durable, generation-fenced provider and model health circuits. Trusted adapter evidence can open a scoped circuit, and one later admitted attempt can make a half-open probe claim.
- Keeps retry authority in `RecoveryCoordinator`. A failed route becomes one ordinary recovery request, waits for the durable health acknowledgement, and can select the next eligible route without a second charge or provider-specific scheduler.
- Adds `hive circuits` inspection and reason-required scoped block, unblock, and reset controls. Human status and JSON expose sanitized route decisions, health, capacity, probe ownership, and next-action ownership.

## Test plan

- [x] Broad local checkpoint - 12,635 runs, 159,926 assertions, 0 failures, 0 errors, 12 declared skips
- [x] Changed Ruby lint - 157 files, no offenses
- [x] Real CLI E2E gate - 22 executed, 22 passed, 0 failed; 3 pending incident contracts remained visible and unexecuted
- [x] Hosted incident timing budget - 4.400s plan-only, 5.557s provider-limit, 2.906s repository-routing; 12.863s aggregate
- [x] E2E harness library - 238 runs, 1,462 assertions, 0 failures, 0 errors
- [x] Standalone `agent-cli-runtime` - 15 runs, 101 assertions, 0 failures
- [x] Focused profile/routing - 103 runs, 520 assertions, 0 failures
- [x] Focused provider health - 47 runs, 283 assertions, 0 failures; direct compaction and corruption regressions also green
- [x] Focused queue/public schemas - 275 runs, 1,376 assertions, 0 failures
- [x] Hosted CI coverage and dedicated merge gates green - 100% line coverage; 14 CI jobs and 8 install-smoke jobs passed

## Notes for reviewer

The routing policy is frozen per task generation, while the selected route is immutable per durable attempt. This allows a recovery-owned successor to move from account A to B without conflating account identity with the existing adapter/profile provider field.

Provider accounts use CLI-owned subscription-session directories. Hive does not add API-key provisioning. Provider-specific session-directory and ambient-override compatibility metadata lives in the extracted `agent-cli-runtime` package and is parity-tested against Hive's temporary internal profile copy. Named bindings clear recognized ambient overrides so the selected subscription session remains authoritative; the default binding preserves the caller's normal session context.

Health state accepts only normalized, allowlisted adapter transport or diagnostic signals. Raw prompts, output, credentials, and provider messages stay in protected attempt artifacts; shared state, audit records, and operational output contain only bounded safe metadata. Current shared transport evidence is enabled only for the reviewed Claude subscription event shape; Codex, Grok, and Pi remain task-local until real captures establish a stable safe contract.

Provider concurrency is derived from live durable attempts. The implementation adds no second lease store, retry loop, deadline, queue, or successor owner. An all-route capacity observation is scheduler-owned and causes no attempt, probe, marker, retry charge, or deadline mutation.

The final Compound Engineering review found no remaining actionable issue after fixes. It covered fail-closed journal genesis and probe replay, full provider/model exclusion receipts, bounded idempotency compaction, default/named session-alias rejection, and subscription-binding isolation.


ivankuznetsov:

## High

- [ ] Ground circuit-opening evidence in verified adapter channels (`test/fixtures/provider_errors/inventory.yml:10`): every real capture remains `real_capture_pending` and task-local while runtime trusts synthetic envelopes that no current CLI is proven to emit, so real provider failures cannot reliably enter the required fallback lifecycle.
- [ ] Admit one enclosing route for the review stage (`lib/hive/config.rb:2179`): configuration discovers only nested review-role pools while durable `6-review` admission resolves `review.routing`, leaving nested pools unused and top-level review routing unable to activate provider accounts.
- [ ] Handle foreground initial no-route results through RecoveryCoordinator (`lib/hive/attempts/entrypoint.rb:52`): the attempt-less result bypasses the required markerless recovery request, then command dispatch calls attached-result methods and bot dispatch dereferences `result.attempt` before deleting its queued request.
- [ ] Publish a recoverable review provider-failure marker (`lib/hive/stages/review.rb:757`): `ProviderRouteFailed` is re-raised by the earlier `AgentError` rescue before any `REVIEW_ERROR reason=provider_route_failed` transition, leaving `REVIEW_WORKING` and preventing RecoveryCoordinator ownership.
- [ ] Isolate provider-health consumer failures per attempt (`lib/hive/attempts/reconciler.rb:60`): one corrupt or unavailable scope raises out of the record loop, aborts the host-wide reconciliation sweep, and remains unacknowledged so the same outage repeats on every daemon tick.
- [ ] Retire a live probe when an operator mutation fences it (`lib/hive/provider_health/store.rb:785`): block and unblock advance generation but retain `probe`, so the old completion cannot clear ownership and the route remains `half_open_probe_owned` even after unblock.
- [ ] Enforce the journal event cap before writing (`lib/hive/provider_health/store.rb:618`): the writer can append event 4097 because it checks only bytes, but the reader rejects more than 4096 events and converts an ordinary long-lived scope into unavailable corruption.
- [ ] Preserve the verified journal prefix during corrupt-state reset (`lib/hive/provider_health/store.rb:471`): a parseable final JSON record without a newline raises with `Circuit.closed` before replaying earlier records, so reset loses the last verified generation and manual block.
- [ ] Clear all ambient Claude credentials for named bindings (`lib/hive/agent_profiles/launch_bindings.rb:16`): named accounts unset `ANTHROPIC_API_KEY` but inherit the alternate `CLAUDE_API_KEY`, allowing a route to use the wrong billing context instead of its configured directory.
- [ ] Reject effective launch-binding aliases (`lib/hive/provider_routing/configuration.rb:102`): validation compares symbolic binding names only, so distinct names or symlinked paths can resolve to one credential directory and defeat account isolation, concurrency, and fallback.
- [ ] Apply the shared secret detector to operator reasons (`lib/hive/provider_health/audit.rb:97`): canonical credentials such as `sk-ant-...` pass the narrow label regex and are then persisted and rendered in audit receipts.
- [ ] Validate route metadata against the selected profile's hard limits (`lib/hive/provider_routing/configuration.rb:212`): tools and permissions are checked only against global enums, so an impossible declaration can satisfy hard requirements and select an adapter that cannot enforce it.
- [ ] Complete the required AE2-AE8 and legacy incident proof (`test/e2e/scenarios/incident_provider_limit_retry.yml:2`): the only provider-routing E2E covers AE1, with no durable half-open, restart/race, strict-pin, capacity, operator, corruption, or real no-pool `limits_reached` characterization required by R13-R15.

## Medium

- [ ] Convert preselection reconciliation failure into a typed no-route result (`lib/hive/attempts/dispatcher.rb:230`): `health.reconcile!` can raise `Unavailable` before route evaluation and bypass the durable `health_state_unavailable` decision and recovery receipt.
- [ ] Wire configured account cooldowns into production health stores (`lib/hive/provider_health/store.rb:46`): every production constructor omits `cooldown_resolver`, so accepted per-account and per-class `cooldown_sec` settings are inert and no-hint circuits always use 300 seconds.
- [ ] Refresh the shipped agent and dispatch-request contracts (`skills/hive/references/status-and-watch.md:6`): canonical guidance still advertises operational-status v3, omits `hive circuits`, and current-facing wiki pages still claim dispatch-request v4 although runtime now emits v4 status and v5 requests.

## Nit

ivankuznetsov:

# Code Review - Provider Health and Route Selection (add-provider-health-and-route-260722-765f)

Scope: `git diff origin/main..HEAD` (185 files, ~16.5k Ruby exec lines + 7 schemas). Base `92df2e3`, head `a8a8eb4`.
Reviewers: correctness, security, adversarial (deep/session-model) + reliability, api-contract, data-migration, testing, maintainability, project-standards, agent-native, learnings.

Note on the deep passes: the security, correctness, and adversarial (cross-attack) reviews each returned **zero** P0/P1 findings after tracing determinism, generation-CAS/idempotency, crash reconciliation, the health-before-redispatch barrier, single-recovery/charge dedup, sanitization (R3), credential binding, operator auth (R11), and lock order (admission -> generation -> health). The core routing/health engine is heavily and consistently guarded. The High findings below are all **error-path / fail-closed-isolation** and **migration-tolerance** defects on the edges, not logic bugs in the engine. All High/Medium findings were verified by orchestrator direct reading of the cited code paths (the P0 migration finding was reproduced live by the data-migration reviewer); an independent per-finding validator wave was not run (budget) - see Coverage.

## High
- [ ] Migration hard-aborts daemon startup on one un-convertible legacy record (`lib/hive/recovery/migration.rb:529`): `corpus_summary` raises an unrescued `Migration::Error` for a v3 record that `migrate_hot_records!` intentionally left unconverted (e.g. a receipt identifier passing v3 schema but exceeding the new 128-byte v4 bound), aborting `Migration.ensure!` / `Store.open_default` permanently and defeating the documented invalid-record tolerance; retries fail identically since the file is already renamed into `attempts/v4`. Live-reproduced by the reviewer. Fix: count the leftover legacy-version record as invalid (as `migrate_hot_records!`/`migration_scan` do) instead of raising. Trigger is narrow today (generation.rb emits 64-char hex) but it is a startup-blocking fail-open-to-crash.
- [ ] Uncaught provider-health error in the terminal-ack path stalls reconciliation daemon-wide (`lib/hive/attempts/reconciler.rb:59`): `acknowledge_provider_health` -> `AttemptObserver#observe` -> `complete_probe`/`apply_evidence` can raise `Unavailable`/`ManagedDirectory::UnsafeError`/`InvalidMutation`, but the per-record loop rescues **only** `CompareAndSwapFailed` (`reconciler.rb:72`) and `acknowledge_provider_health` (`finalization_maintenance.rb:89`) has no internal guard, so one corrupt/unavailable health scope on a single explicit-routed terminal record aborts `reconcile()` and fails the whole tick (`daemon/dispatcher.rb:262` "fail closed for this tick") for every task including legacy no-pool tasks - and since the terminal record is immutable and revisited each tick, the stall is permanent. Violates the plan's fail-closed-per-scope and legacy-bypass guarantees. Fix: rescue `ProviderHealth::Error`/`UnsafeError` inside `acknowledge_provider_health` and return false.
- [ ] Provider-health store mutation entry points are not fail-closed like the read paths (`lib/hive/provider_health/store.rb:431`): `with_mutation_lock` rescues only `JournalCorruption`, and `with_route_admission`/`apply_evidence`/`complete_probe`/`reconcile!`/`operator_mutation` lack the `Unavailable, ManagedDirectory::UnsafeError` outer rescue that `inspect_scope`/`evaluate_route` already have; the admission loop rescues only `StaleGeneration` (`attempts/dispatcher.rb`), so `health.reconcile!` / `with_route_admission` on a corrupt scope escape admission and (per reviewer trace) crash/stall the daemon dispatch loop rather than excluding only the enclosing scope. Fix: extend the read-path fail-closed rescue pattern to the mutating entry points. (Same root cause as the finding above; distinct code site.)

## Medium
- [ ] Operator block/unblock/reset don't guard against an unresolved probe intent before bumping generation (`lib/hive/provider_health/store.rb:754`): after a crash inside the multi-scope claim window (intent+attempt persisted, claims not yet finalized), an operator action advances the circuit generation, then `reconcile_intent` hits `raise Unavailable, "...cannot be reconciled"` (`store.rb:1082`) for the still-live attempt instead of gracefully retiring the stale claim - which the plan (U7/AE7: "block/unblock/reset during a live probe ... retires the stale claim exactly once") explicitly requires, and which feeds the daemon stall above. Fix: add the `unresolved_intent?` guard to `operator_mutation`, or make `reconcile_intent` treat an advanced generation on a live attempt as a benign rollback.
- [ ] `wiki/modules/daemon.md` contradicts itself on the dispatch-request schema version: this diff updated some prose to v5 (lines 50, 636, 662, 669-670) but left v4 in others (lines 65, 422, 763-764, 886), so the wiki now says both v4 and v5 are "current". CLAUDE.md treats the wiki as source of truth and the plan (U6/U8) requires updating `daemon.md`. Fix: make every dispatch-request-version reference consistent with the v5 writer/reader reality.
- [ ] `hive circuits --provider/--model` narrows `accounts` but not `decisions` (`lib/hive/provider_routing/operational_projection.rb:109`): `decision_projection` calls `routing_decisions(limit: DECISION_LIMIT)` ignoring the `accounts:` filter that `account_projection` honors, there is no `--project` filter despite `identity.project` in the schema, and no filter-echo - so an agent running `hive circuits inspect --provider X` sees a scoped `accounts` array beside an unscoped `decisions` array with nothing signaling the mismatch (R12). Fix: scope `decision_projection` by the filtered accounts (and/or emit the applied filter in the payload).
- [ ] Malformed/version-skewed routing disposition is dropped with no degraded signal (`lib/hive/operational_status.rb:597`): `routing_payload` returns `nil` on any key/shape mismatch and, unlike `attempt_storage`/scheduler-join failures, appends no entry to the top-level `issues` array, so a routing-blocked explicit-pool task becomes indistinguishable from a no-pool task in `hive status --json` - an agent cannot tell explainability data existed but was discarded (R12 "one durable operational projection"). Fix: record a sanitized issue when a non-empty-but-malformed routing block is dropped.
- [ ] `config/component-boundaries.yml:34` omits `Hive::ProviderHealth::InvalidScope` from provider-health `public_contract.errors`, though this same diff defines it (`lib/hive/provider_health.rb:12`) and rescues it across the facade (`lib/hive/commands/circuits.rb:63`); the plan requires the boundary catalog to be accurate and tested rather than drift. Fix: add `Hive::ProviderHealth::InvalidScope` to the errors list.
- [ ] Circuit->JSON summary shape is hand-duplicated across three producers with diverging `state` semantics (`lib/hive/provider_routing/candidate.rb:44`, `operational_projection.rb` `inspection_projection`, and the `dispatch_request_queue.rb` admission-observation validator), and no test ties the producer shape to the admission validator's `ADMISSION_CIRCUIT_KEYS` (corroborated by api-contract). Fix: add one canonical sanitized-circuit serializer owned by ProviderHealth and have all three consume it.
- [ ] AE4 strict-pin exclusion matrix covers only 1 of 5 plan-required states (`test/integration/provider_routing_replay_test.rb:63` asserts only `circuit_cooldown`+`hard_pin_mismatch`): missing `manual_block`, `requirements_incompatible`, `half_open_probe_owned`, and `provider_concurrency_saturated` on the pinned route, each of which AE4 requires to return `no_eligible_provider_route` with its exact reason. Fix: parametrize the test over all five pinned-route exclusion states.
- [ ] No explicit provider-vs-non-provider one-request/one-charge comparison (`test/unit/daemon/recovery_coordinator_test.rb:713`): the Verification Contract requires comparing exactly-one recovery request/charge against an equivalent non-provider recoverable failure under identical fixtures, but the current proof is implicit. Fix: add a side-by-side test asserting both flows yield one pending request, `retry_count==1`, and identical due time.

## Nit
- [ ] `Hive::ProviderHealth::Reconciler` (`lib/hive/provider_health/reconciler.rb`) is an unused wrapper class - autoloaded (`provider_health.rb:363`) and unit-tested but never called in production; the real restart reconciliation runs via `health.reconcile!` directly at `attempts/dispatcher.rb:230,308`. Verified it is NOT an unwired restart seam. Delete the class + autoload + its test, or wire it in like `Attempts::Reconciler`.
- [ ] `config/component-boundaries.yml` catalog drift for the new wiring: `attempts.allowed_hive_dependencies` (~line 530) omits the new provider-health/routing requires added in `finalization_maintenance.rb`, `evidence_channel.rb`, `decision_index.rb`; provider-health `hive_consumers` (~line 70) omits `evidence_channel.rb`; provider-routing-policy `hive_consumers` (~line 145) omits `decision_index.rb`, `dispatch_request_queue.rb`, `recovery_coordinator.rb`. Tolerated only because these components are still `candidate` status (no `validate_requires!`). Fix: reconcile the catalog with the new requires.
- [ ] Exclusion `reason` field validation strictness is inconsistent across sibling schemas (`schemas/hive-dispatch-request.v5.json:139` uses a closed 8-value enum; `hive-attempt.v4.json`, `hive-circuits.v1.json`, `hive-operational-status.v4.json` use loose `Identifier` bounds) and `circuit_open` is dead enum vocabulary never emitted (the store emits `circuit_cooldown` for an open circuit). Fix: extract one shared `$defs` reason definition + add a `schema_files_test.rb` drift-guard tying `ADMISSION_EXCLUSIONS` to the enum.
- [ ] `Hive::ProviderRouting.digest` (`lib/hive/provider_routing.rb:168`) always `canonical_json`s its input, silently diverging from `Hive::ProviderHealth.digest`'s String short-circuit; latent (no current caller passes a raw String, confirmed by two reviewers) but an easy trap for the next caller. Fix: share one digest/canonical-JSON helper so the identically-named methods cannot diverge.
- [ ] Failure-class/provenance vocabularies are duplicated verbatim between `lib/hive/attempts/record.rb:25` and `lib/hive/provider_health.rb:22` with nothing keeping them in sync (no drift-guard test). Fix: single source of truth or an equality assertion test.
- [ ] Read-only inspection mutates durable state under the host-global health lock: `inspect_scope`/`evaluate_route`/`OperationalProjection#to_h` trigger `publish_projection` and torn-suffix-trim `atomic_write` (`store.rb:490,568`) and acquire `mutation.lock` per account+model, so `hive circuits`/status rendering can rewrite `journal.jsonl`/`current.json` and contend with admission on a nominal read. Fix: pass `publish_projection:false`/`recover_torn:false` on pure-report paths.
- [ ] Operator projection is a torn cross-scope read (`lib/hive/provider_routing/operational_projection.rb:138`): account and model circuits are sampled under separate `inspect_scope` lock acquisitions, so a concurrent probe claim between them can render an account/model generation pair that never coexisted (status-only, cosmetic). Fix: snapshot all enclosing scopes for an account under one lock hold.
- [ ] `Route` and `Pin` disagree on the canonical field name (`account` vs `provider`) (`lib/hive/provider_routing/route.rb:29`). Fix: standardize on one primary name with the other as an alias.

## Coverage / notes
- Verdict: Ready with fixes. No defect was found in the routing/health decision engine itself; the three High findings are all error-path fail-closed-isolation (2x) and migration-tolerance (1x) defects that let a single corrupt/unavailable health scope, or one un-convertible legacy record, stall reconciliation or block daemon startup for all tasks - directly contradicting the plan's fail-closed-per-scope and legacy-bypass guarantees. Fix those before merge.
- Requirements completeness: U1-U8 are structurally present (`hive circuits` registered at `cli.rb:1149`; `recovery_authority_test.rb:135` forbids the rejected PR #759 lease/recovery namespaces; `incident_provider_limit_retry.yml` activated with `pending: false`). No plan-required requirement was found missing/undeferred.
- Validation method: security/correctness/adversarial deep passes returned no P0/P1. All High/Medium findings verified by orchestrator direct reading of the cited code (P0 migration finding independently reproduced live by the data-migration reviewer). An independent per-finding validator-subagent wave was not run due to budget; treat the High findings as CONFIRMED-by-direct-read and the crash-vs-permanent-stall distinction on findings 2-3 as the one open detail (impact is High either way).
- Notable residual risks (not findings): evidence-poisoning defense depends on no adapter surfacing model/tool content as a top-level `provider_transport` frame (verify per adapter); a `succeeded`+`provider_evidence` record would probe-close then evidence-open the same circuit, prevented only by the single Supervisor producer, not at the observer boundary; `Attempts::Reconciler` calls `acknowledge_provider_health` unconditionally (not gated on `prepare`'s result).
- Notable testing gaps (not findings): no fault-injection across every `perform_probe_intent` crash phase then `reconcile!` asserting one-or-zero probe owner / no double capacity count; concurrency races use deterministic single-shot hooks rather than real threads at the dispatcher/health boundary; no end-to-end secret-canary test proving planted stdout/stderr/message secrets never reach circuit/journal/audit/CLI/JSON; no test that an oversized/multiline/secret-like `--reason` is rejected with no journal append/generation bump.

ivankuznetsov:

## High

- [ ] Real adapter evidence captures are still missing (`test/fixtures/provider_errors/inventory.yml:10`): all four trusted normalizers and the fallback incident rely on invented contract envelopes while every declared real capture remains `real_capture_pending`, so the U1/U5/R15 provenance proof is not delivered.
- [ ] Health-store construction bypasses fail-closed route admission (`lib/hive/attempts/dispatcher.rb:229`): an unsafe or unavailable store can raise before `reconcile_provider_health` and before a durable `health_state_unavailable` decision is recorded.
- [ ] One corrupt probe-intent file disables every explicit route without a supported repair (`lib/hive/provider_health/store.rb:1010`): `load_intents` fails the whole host-global scan, but inspection and reset expose repair tokens only for scoped journals.
- [ ] Unavailable named launch bindings remain eligible and repeatedly win selection (`lib/hive/provider_routing/configuration.rb:128`): configuration swallows binding-resolution failure, then launch fails after route commitment without health evidence, so later attempts select the same unusable preferred account instead of falling back.
- [ ] Named Pi accounts can collapse onto ambient provider credentials (`lib/hive/agent_profiles/launch_bindings.rb:19`): Pi clears no API-key environment variables and ordinary child spawns inherit the ambient environment, so distinct `PI_CODING_AGENT_DIR` bindings can execute as the same external account.
- [ ] Nested review-role routing pools are accepted but never control the durable review attempt (`lib/hive/config.rb:2190`): configuration validates `review.ci`, `review.fix`, `review.triage`, browser, and reviewer pools even though `6-review` admission reads only `review.routing` and its admitted route overrides every nested spawn.
- [ ] Crashed live v3 attempts permanently block the v4 cutover (`lib/hive/recovery/migration.rb:312`): migration refuses persisted launching/running records before the v4 reconciler can prove them lost, contradicting the required in-flight v3 migration and restart behavior.
- [ ] The claimed AE2-AE8 release matrix bypasses durable admission (`test/e2e/lib/provider_routing_incident_matrix.rb:47`): it calls health, router, and recovery objects with in-memory attempt maps instead of exercising `Dispatcher`, durable attempt ownership, `Supervisor`, real races, and crash/restart boundaries required by R15.
- [ ] The required `circuit_open` exclusion is absent (`lib/hive/provider_health/store.rb:128`): every automatic open state emits `circuit_cooldown`, and the changed schemas and tests omit the distinct stable code promised by the authoritative decision vocabulary.
- [ ] Torn final journals are neither transparently recovered nor atomically reset (`lib/hive/provider_health/store.rb:338`): route evaluation reports a malformed torn tail unavailable, while token reset truncates it first and then rejects the now-stale token, producing an unaudited failed-but-mutating repair.
- [ ] A valid full health journal has no recovery path (`lib/hive/provider_health/store.rb:625`): at 4,096 events or 2 MiB the scope remains inspectable but every evidence, probe, block, unblock, and reset append fails permanently because no compaction or epoch rotation exists.

## Medium

- [ ] Operator generation CAS is checked before probe-intent reconciliation mutates the generation (`lib/hive/provider_health/store.rb:784`): reconciliation can accept a probe claim and then append the operator action without rechecking, allowing two accepted transitions from one observed generation.
- [ ] Hyphenated OpenAI project keys pass reason validation (`lib/hive/provider_health/audit.rb:98`): the shared `sk-` pattern does not match `sk-proj-...`, so credential material can be persisted and rendered in operator audit receipts despite R11.
- [ ] Capacity-saturated routing is projected as idle (`lib/hive/operational_status.rb:541`): the daemon records `capacity_saturated` as `attempt_capacity`, but the status classifier has no such branch and falls back to a false ready/idle state.
- [ ] Human circuit inspection omits a required corruption-reset token field (`lib/hive/commands/circuits.rb:280`): it prints the epoch and fingerprint but not `last_verified_generation`, so the default inspection output cannot construct the documented reset command.
- [ ] `hive circuits --json` loses its JSON contract on Thor usage errors (`bin/hive:91`): `circuits` is absent from `JSON_USAGE_ERROR_CONTRACTS`, so malformed options or extra positionals emit non-`hive-circuits` output before command dispatch.
- [ ] Operational status still advertises attempt layout generation 3 (`schemas/hive-operational-status.v4.json:184`): runtime state moved to `attempts/v4`, but the status producer/schema and public path helpers still identify or resolve the fenced v3 layout.
- [ ] The component catalog hides provider-health and routing-policy state dependencies (`config/component-boundaries.yml:59`): provider health declares no dependency while importing `Attempts::OutputReference`, and routing policy claims no mutation authority while its owned `PolicyStore` writes through Attempts storage, defeating the required DAG and authority checks.
- [ ] `ProviderRouting::PolicyStore` is declared public but missing from its entrypoint (`lib/hive/provider_routing.rb:122`): a clean `require "hive/provider_routing"` cannot resolve the catalogued constant unless an internal Attempts require happened first.
- [ ] Admission reparses health state once per route under the host-global admission lock (`lib/hive/attempts/dispatcher.rb:614`): pools may contain 1,024 routes, and each candidate separately acquires the health lock, rescans all intents, and replays both journals, creating multiplicative host-wide admission stalls.

## Nit

PR #987, opened 2026-08-12
fix(recovery): normalize marker attribute encodings

ivankuznetsov:
## Summary

Recovery no longer stalls when a provider error contains non-ASCII text. Hive now treats byte-identical valid UTF-8 marker attributes as the same value before and after durable JSON storage, while malformed bytes still fail closed.

The shared `Hive::Recovery` compatibility boundary now owns this normalization for recovery requests, observation tokens, marker clearing, and operational scheduler joins. This removes the encoding-only `generation_conflict` that blocked retries and prevents the same task from appearing stale in operational status.

## Validation

- Recovery coordinator: 42 runs, 247 assertions, 0 failures
- Operational status and marker focused suites: 89 runs, 350 assertions, 0 failures
- Broad suite on the same production diff: 12,649 runs, 160,060 assertions, 0 failures
- RuboCop: no offenses in changed Ruby and test files
- Invalid UTF-8 regression: request persistence fails closed and leaves the marker intact

---

### DIFF
--- a/lib/hive/daemon/recovery_coordinator.rb
+++ b/lib/hive/daemon/recovery_coordinator.rb
@@ -89,8 +89,7 @@ module Hive
         def observation_token(row)
           attrs = row_value(row, :marker_attrs)
           attrs = row_value(row, :attrs) unless attrs.is_a?(Hash)
-          attrs = attrs.is_a?(Hash) ? attrs.to_h.transform_keys(&:to_s) : {}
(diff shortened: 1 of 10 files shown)


ivankuznetsov:
Integrated into and shipped through #989. The exact #989 head passed all required CI, and the merged main tree at 1d7fba8927d0d8fc8808dda9a5db485c01e127f7 is byte-for-byte identical to that reviewed tree. Closing this source PR because main enforces linear history and required the integration PR to be squash-merged.

PR #998, opened 2026-08-12
feat(agent-runtime): add first-class OpenCode support

ivankuznetsov:
## Summary

- Make OpenCode a fully opt-in Hive and `agent-cli-runtime` profile, with exact `provider/model` routing, offline capability checks, and an invocation-owned isolated configuration overlay.
- Keep process control with the caller while adding strict JSONL and sanitized-session normalization, typed outcomes, requested-versus-observed model identity, nullable usage, and cleanup on every Hive lifecycle path.
- Make OpenCode available across existing role and skill capability checks without changing defaults, councils, or fallbacks; document the 0.2.0 component candidate and validate its standalone mirror.

## Test plan

- [x] Component suite: 80 runs, 975 assertions
- [x] Hive suite: 12,761 runs, 161,068 assertions
- [x] OpenCode 1.18.17 offline smoke: 17 assertions, with no model request
- [x] Mirror projection: 9 runs, 189 assertions
- [x] Built `agent-cli-runtime-0.2.0.gem` candidate (SHA-256 `512a04179b39ab79b42d4be269969d4193777946a087bb12dac1a0a9adcf1a2d`)
- [x] Authenticated live smoke explicitly skipped because no live route was supplied

## Notes for reviewer

- OpenCode remains opt-in. Existing default profile selection, reviewer councils, and fallback order are unchanged.
- Session-settled decisions carried from planning: caller-owned process lifecycle and explicit isolated configuration (user-directed); exact nested route attribution and sanitized export correlation (evidence-driven); fail-closed permission policy and no release action (user-directed).
- This is a pre-release 0.2.0 candidate only. No tag, publication, deployment, or release workflow was run.


ivankuznetsov:

## High

- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/overlay.rb:261` — Compile edit permissions in OpenCode's worktree-relative path form: the generated absolute allow patterns never match `edit`/`write` requests, so the trailing deny rule blocks every `workspace-write` edit.
- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/overlay.rb:334` — Reject or sanitize per-agent permissions from the selected config: OpenCode merges `agent.*.permission` after the generated top-level deny-first policy, allowing caller config to re-enable shell, edits, or external writes.
- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/probe.rb:168` — Clear the ambient environment for probe subprocesses: `Open3.popen3` inherits unselected credentials and configuration, so readiness can pass using sources that the hermetic prepared invocation will not receive.
- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/values.rb:386` — Reserve overlay-owned environment keys before forwarding credential variables: a caller can name `OPENCODE_CONFIG`, XDG homes, or disable flags as credentials and overwrite the confinement environment used by run and export.
- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/probe.rb:95` — Bind authentication evidence to the requested provider: any nonempty allowlisted key or staged auth file currently satisfies the probe even when it cannot authenticate the requested `provider/model`, violating offline exact-route readiness.
- [ ] `lib/hive/stages/base.rb:466` — Preserve the scope of qualified edit permissions: `Edit(/project/docs/**)` is collapsed to the containing `/project` root, widening write access from the declared subtree to the entire project.
- [ ] `lib/hive/agent.rb:739` — Carry the inherited implementation route into prepared OpenCode requests: downstream stages with identity-only launch arguments lose the required nested model and fail preparation or select an unrelated overlay default.
- [ ] `lib/hive/implementation_identity/store.rb:158` — Give each repeated review CI-fix observation a distinct idempotency identity: multiple fixes in one stage/attempt reuse the same journal key, so the second differing observation conflicts and aborts the retry loop.
- [ ] `lib/hive/usage_db.rb:248` — Propagate usage availability through aggregation and presentation: nullable OpenCode metrics are stored as numeric zero plus availability flags, but aggregate queries discard those flags and report unknown usage as measured zero.
- [ ] `components/agent-cli-runtime/test/opencode_offline_smoke_test.rb:9` — Make the installed OpenCode offline smoke a mandatory pre-release gate: the test silently skips when the CLI is absent and no changed task or workflow converts that skip into the required failure evidence.
- [ ] `components/agent-cli-runtime/bin/verify-candidate:86` — Exercise OpenCode probe, prepare, normalize, and cleanup from the installed candidate and mirror projection: the current consumer smoke checks only registry/version and a missing-Codex probe, leaving the required packaged lifecycle unverified.
- [ ] `lib/hive/agent_skills/adapters/opencode.rb:28` — Validate Compound Engineering readiness inside the prepared invocation environment: setup/readiness inspects ambient plugin caches, while runtime redirects config/data/cache to fresh roots containing only a remote plugin spec, so a role can pass readiness yet lack its skill at spawn.

## Medium

- [ ] `lib/hive/agent_profile.rb:751` — Use one credential-placeholder grammar across Hive and the component: Hive accepts `${KEY}` while the overlay accepts only `{env:KEY}`, making every placeholder invalid at one of the two validation layers.
- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/result_parser.rb:257` — Bind session identity even for the first unknown additive event: an initial unknown event from session A is ignored for correlation and a later required event from session B is incorrectly accepted.
- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/result_parser.rb:24` — Narrow nonzero-exit classification to recognized configuration diagnostics: matching generic words such as `provider` or `model` mislabels operational provider failures and rate limits as `configuration_failure`.
- [ ] `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/overlay.rb:294` — Permit resolved, ownership-safe invocation paths beneath symlinked system ancestors: rejecting any symlink in the ancestor chain makes OpenCode unusable on otherwise valid platforms whose temporary directory path is a symlink.

## Nit

ivankuznetsov:

# Code Review — First-Class OpenCode Support (pass 1)

Scope: `git diff origin/main..HEAD` (7 commits, U1–U7). Reviewers: correctness,
security, reliability, testing, maintainability + orchestrator fast pass, all
against the local worktree. Verdict: **Ready with fixes** — plan requirements
appear fully implemented (opt-in defaults preserved: `DEFAULT_GLOBAL_AGENTS` is
still `[claude codex]`; `0.1.0`→`0.2.0` version/changelog/gemspec consistent;
route-aware probe, strict parser, deny-first permissions, requested/actual
identity, nullable usage, cleanup-in-`ensure`, and skill readiness all present).
No P0/P1 / High defects found. The items below are P2 (Medium) and P3 (Nit).

## High

_None._

## Medium

- [ ] Env-placeholder secret validators disagree across the ABI (`lib/hive/agent_profile.rb:751` accepts only `${VAR}`, `components/agent-cli-runtime/lib/agent_cli_runtime/opencode/overlay.rb:9` accepts only `{env:VAR}`): both run on the same inline `agents.opencode.config`, so `{env:VAR}` (OpenCode's real syntax) is rejected by Hive and `${VAR}` is rejected by the component — no placeholder satisfies both, making inline env-referenced provider credentials unusable. Fails closed (no leak) but blocks a documented path; flagged independently by correctness, security, and maintainability. Fix: align Hive's regex to `{env:VAR}` and add a parity assertion.
- [ ] workspace-write edit policy can leave a declared read-only root editable when it is nested under a writable root (`overlay.rb:249-263`): the writable-root `#{root}/**` => `allow` is inserted *after* the nested read-only `#{root}/**` => `deny`, and OpenCode applies the last matching rule, so the parent allow shadows the nested deny — contradicting the code's own comment ("additional read roots remain explicitly denied"). The top-level `"*" => "deny"` keeps the AE1 "no writes outside declared roots" guarantee intact, so blast radius is read-only carve-outs *inside* an already-writable tree; still a real deviation from stated intent. Fix: re-emit nested read-only denies after the writable allows (deny-last for exceptions) and add a nested-root test.
- [ ] `prepared.cleanup!` is called unguarded in the `spawn_opencode_and_wait` `ensure` block (`lib/hive/agent.rb:679`): `Overlay::Cleanup#call` deliberately raises `UnsafePathError` on a replaced root and `FileUtils.remove_entry_secure` can raise `Errno`, so a cleanup failure propagates from the `ensure` — masking any in-flight body exception and turning a fully-built successful `result` into a raise. Violates the stated "cleanup must not raise past the lifecycle" contract. Fix: rescue around `cleanup!` and record the failure (e.g. `result[:cleanup_error]`) instead of propagating.
- [ ] `web/Gemfile.lock` still pins `agent-cli-runtime (~> 0.1.1)` / `(0.1.1)` after `hive.gemspec` bumped the dependency to `~> 0.2.0` (`web/Gemfile.lock:5,101`; file not touched by the diff): web bootstraps `hive-cli` by path (`web/Gemfile:79`), which transitively requires `~> 0.2.0`, so a frozen/deployment bundle under `web/` would fail resolution. Conditional on a frozen bundle and may be handled by release tooling (owner: release), but the lock is inconsistent as of this PR. Fix: regenerate `web/Gemfile.lock`.

## Nit

- [ ] `final_message_truncated` is inferred as `outcome.final_message&.bytesize == MAX_FINAL_MESSAGE_BYTES` (`lib/hive/agent.rb:650`): because `ResultParser.bounded_string` truncates via `byteslice(0, MAX).scrub("?")`, a multibyte-cut truncation can shrink below MAX (false negative, and this flag gates `runtime_policy.rb:69` host-output validation) while an exactly-MAX untruncated message reports true (false positive). Fix: have `ParsedRun`/`NormalizedOutcome` carry an explicit truncated flag from the parser.
- [ ] Sanitized-export inspection timeout path leaks a zombie (`lib/hive/agent.rb:854-865`): the timeout branch does `kill_group` + `sleep_grace_then_kill` then `break` with no blocking reap, so an export child that survives SIGTERM through the 3s grace is SIGKILLed but never waited on, leaving a zombie in the long-lived hive process. The main run path (`wait_for_opencode_process:814`) correctly follows with `Process.wait2(pid)`. Fix: add the same blocking reap (rescue `ECHILD`).
- [ ] CE plugin pin `compound-engineering-v3.21.4` is duplicated as a magic constant `SkillCheck::OpenCode::PINNED_COMPOUND_ENGINEERING_PLUGIN` (`lib/hive/skill_check.rb:424`, used for install-hint text/tests) separate from the authoritative `config/agent-skills.yml:52`, with no test tying them — a one-sided version bump silently leaves the install hint stale. Fix: derive the constant from the manifest or add an equality test.
- [ ] The new `:shadowed` SkillCheck status and the OpenCode/Grok delegate modules are undocumented in the `SkillCheck` module header (`lib/hive/skill_check.rb:16`), which still enumerates only `:present`/`:missing`/`:not_applicable` and Claude/Codex/Pi. `:shadowed` is correctly surfaced as a failure by `stages/base.rb#format_verified_skill_invocation`; only the contract docstring is stale. Fix: update the header.
- [ ] The probe's fail-closed guard for a missing export `--sanitize` capability (`components/agent-cli-runtime/lib/agent_cli_runtime/opencode/probe.rb:81-84`) has no test — the only capability-removal fixture strips `--variant` from run-help (`opencode_preparation_test.rb`). Since an unsanitized export could leak secrets, this security guard should be exercised. Fix: add a `:missing_export_capability` fixture mode asserting `prepare!` raises `UnsupportedCapability` and removes the invocation root.
- [ ] The TUI token-stats `opencode` agent row is unasserted (`test/unit/tui/views/token_stats_test.rb`): `token_stats.rb` added `:opencode` to `AGENTS`, but the render test only asserts `claude`/`grok` appear, so dropping `:opencode` would pass. The plan's "nullable usage in TUI" surface for OpenCode is covered in the DB and status JSON but not the TUI row. Fix: `assert_includes out, "opencode"` plus a nil-vs-zero rendering case.
- [ ] Untested behaviors worth backfilling (all present in code, none blocking): a signal-terminated-but-not-cancelled run normalizing to a failure kind (`success?` relies on unguarded `signal.nil?`); the Hive-level cancellation lifecycle (only timeout is exercised for the no-inspection-after-failure guarantee); `validate_additive_session!` (unknown event with a mismatched `sessionID` → malformed); and cleanup idempotency/refusal against a *replaced* invocation root plus a 0600/removed assertion on the staged `auth.json`.

ivankuznetsov:
Resolved both review passes in `a380b3e71`.

Security and confinement:

- OpenCode edit grants are now worktree-relative, qualified `Edit(path/**)` scopes stay narrow, and nested read-only roots are re-denied after writable parents.
- Selected `agent.*.permission` overrides are removed; overlay-owned environment keys are reserved; probe children explicitly clear ambient configuration and credentials.
- Credential evidence is bound to the requested provider for environment keys and staged auth files, and ownership-safe invocation roots beneath symlinked system ancestors are canonicalized before validation.

Lifecycle and evidence:

- Inherited routes survive identity-only launches; repeated OpenCode observations have distinct session-scoped idempotency keys; nullable usage remains unknown through aggregation and TUI presentation.
- Additive events bind session identity, operational provider failures no longer look like configuration failures, and truncation is an explicit parser field.
- Export children are reaped after forced termination; cleanup failures are reported without masking a successful/body result; Hive-level timeout, TERM cancellation, replaced-root refusal, auth-file mode/removal, and additive-event mismatch paths have regression coverage.
- Compound Engineering readiness is rechecked against the prepared OpenCode configuration/environment, with shadow detection and a manifest-pin parity test.

Package and release proof:

- The exact installed candidate and independently reconstructed mirror now exercise OpenCode probe, prepare, normalize, and cleanup.
- The component release workflow installs OpenCode 1.18.16 and requires the guarded offline smoke instead of accepting a missing-CLI skip; missing export `--sanitize` capability is explicitly tested.
- No release was authorized, so the premature 0.2.0 selection was removed. The feature remains unreleased, Hive/web stay resolvable on the published 0.1.x dependency, and version/tag/publication remain a separate operator action.

Validation:

- Component suite: 91 runs, 1,022 assertions, 0 failures/errors; only the opt-in authenticated live-model smoke skipped.
- Focused Hive/component/release/mirror suites: green, including installed-candidate and mirror lifecycle checks.
- Repository checkpoint: 13,516 runs / 192,018 assertions reached two stale integration-fixture errors (generic credential name and absolute worktree-rule expectation); both fixtures were corrected and rerun together green (2 runs / 36 assertions).
- RuboCop: 29 changed Ruby files, no offenses. Root and web dependency checks both pass.

All reported High, Medium, and Nit items are addressed; exact-head CI is running on `a380b3e71061f2025cd3abb676d6acefa5da01de`.

ivankuznetsov:
Exact-head follow-up: the first CI run exposed a source-Web dependency mismatch, not a Rails behavior failure. Web was loading the published 0.1.1 component while Hive source registered the new OpenCode profile.

Fixed in `b19172e80` by making the component an explicit Web path dependency:

- source checkouts default to `../components/agent-cli-runtime`;
- managed bundle install, asset build, and server launch export `HIVE_AGENT_CLI_RUNTIME_ROOT` for the exact installed component gem, parallel to `HIVE_CLI_ROOT`;
- the Web lock remains on version 0.1.1, so this still does not select or publish a new release.

Local proof after the fix:

- Rails integration: 313 runs / 1,786 assertions, green;
- Playwright system: 62 runs / 628 assertions, green;
- committed packaged-Web bootstrap (`git archive HEAD:web`): 1 run / 763 assertions, green;
- focused AppBundle/Web command contracts: 88 runs / 326 assertions, green.

CI has restarted on exact head `b19172e80dfa20183389ecf8aeaefe2b02e28e47`.

ivankuznetsov:
Coverage-gate follow-up is fixed in `32317020c`.

The failed aggregate identified 87 uncovered OpenCode paths even though every leaf test job was green. This head adds fail-closed regression coverage across profile/config validation, plugin setup and inventory, prepared-skill resolution, implementation observations, managed permissions, Web component roots, process-group races, cancellation, inspection timeouts, and defensive cleanup.

Two small contract hardenings fell out of that proof: signal cancellation and IO cleanup now use directly testable helpers, and an absent OpenCode plugin package is reported as `nil` (matching every other inventory) rather than `false`.

Local exact gate on this head:

- component: 91 runs / 1,022 assertions / 0 failures or errors / 1 authenticated-live skip;
- repository: 13,534 runs / 192,185 assertions / 0 failures or errors / 12 intentional skips;
- line coverage: **100.00% (94,724/94,724)**;
- RuboCop: 98 changed Ruby files, no offenses.

Exact-head CI has restarted on `32317020cb7319bc7972002c51fa89bf9175323e`.

PR #1000, opened 2026-08-13
fix(installer): harden managed QMD installation

### LINKED ISSUE #209: QMD install: broken better-sqlite3 native module goes undetected
From ce-code-review of PR #192 (finding P2, reviewers: reliability, adversarial, cli-readiness; confidence 100; validated).

**Problem.** In `install.sh` `install_qmd()`:
- `install.sh:279` runs `npm rebuild --global --prefix "$qmd_home" better-sqlite3 >/dev/null 2>&1 || true`, fully silencing rebuild failures (missing node-gyp/compiler/python leave the old broken ABI build in place).
- `install.sh:286` health-gates only on `qmd --version`, which proves the process starts but does NOT load the `better-sqlite3` native addon.

So a `NODE_MODULE_VERSION`/ABI mismatch is reported as a healthy install and only fails later at `qmd update`/wiki-refresh time, silently. The same silent rebuild appears in the `install.md` repair snippet (`install.md:84`).

**Suggested fix.** Surface the `npm rebuild` exit status / stderr instead of swallowing it, and add a health probe that actually exercises better-sqlite3 (open the DB) rather than relying on `--version`. Emit a warning + the `npm rebuild better-sqlite3` repair hint when the native module fails to load.

### LINKED ISSUE #210: QMD install: macOS BSD readlink -f breaks symlink refresh on hive update
From ce-code-review of PR #192 (finding P2, reviewers: adversarial, reliability; confidence 100; validated).

**Problem.** In `install.sh` `install_qmd()` (~lines 291-300), the existing-symlink conflict check computes the two sides asymmetrically:
- `existing_qmd_link="$(readlink -f "$qmd_link" 2>/dev/null || true)"` -> empty string when `readlink -f` is unsupported
- `managed_qmd_link="$(readlink -f "$qmd_bin" 2>/dev/null || echo "$qmd_bin")"` -> real path fallback

`readlink -f` is unavailable on stock macOS pre-Sonoma. There the comparison `[[ "" != "/real/path" ]]` is always true, so install.sh prints a spurious "existing qmd ... leaving it unchanged" warning AND skips `ln -sfn`, leaving the `qmd` symlink stale (never refreshed to a rebuilt binary) on `hive update`.

**Suggested fix.** Use a portable canonicalizer with the SAME fallback on both sides so a resolver failure compares like-to-like (e.g. fall back to the literal path on both sides, or use `cd -P`/`pwd -P`).

### LINKED ISSUE #211: QMD install: pin @tobilu/qmd and validate HIVE_QMD_NPM_PACKAGE (supply chain)
From ce-code-review of PR #192 (finding P2, reviewers: security, learnings; confidence 75; validated).

**Problem.** The hive `.gem` is cosign-verified + SHA256 fail-closed (`install.sh:372-405`). The managed QMD step (`install.sh:271`) then runs `npm install --global --prefix ... @tobilu/qmd` with the package UNPINNED (no version, no integrity, no `--ignore-scripts`) and default postinstall + native build enabled — entirely outside that trust boundary. A compromised/typosquatted/hijacked upstream `@tobilu/qmd` gets arbitrary code execution as the installing user on `install.sh` / `hive update`. `HIVE_QMD_NPM_PACKAGE` (`install.sh:10`) is also not shape-validated like `HIVE_VERSION`/`HIVE_REPO_OWNER`.

This is the standard risk of any `npm install` on a local single-user tool (hence P2, not higher), with `HIVE_INSTALL_QMD=0` as an opt-out — but it should be a conscious, documented decision.

**Suggested fix.** Pin a specific `@tobilu/qmd` version (ideally with an integrity hash); document that this step downloads/executes third-party npm code outside Hive's signature/checksum boundary; shape-validate `HIVE_QMD_NPM_PACKAGE` (package-name allowlist) for parity with the other env inputs.

### LINKED ISSUE #212: QMD discovery: normalize --prefix to absolute before writing install-prefix sidecar
From ce-code-review of PR #192 (finding P3, reviewer: adversarial; confidence 75; validated).

**Problem.** `install.sh` (~line 471) writes the raw `--prefix`/`HIVE_PREFIX` value into the `install-prefix` sidecar without normalizing to an absolute path. The readers — `doctor.rb#find_qmd` (`File.join(prefix, ...)`, ~line 159) and the generated bash `find_qmd` (string concat, `scripts.rb` ~line 51) — do not expand `~` or resolve relative paths. So a `--prefix=~/foo` or relative-path install produces an `install-prefix` whose derived candidate path never resolves, and qmd becomes silently undiscoverable later (the `-x`/`File.executable?` guard just fails).

**Suggested fix.** Normalize `$PREFIX` to absolute before writing `install-prefix`, and/or `File.expand_path` it on read in `doctor.rb#find_qmd` and the scripts `find_qmd`, matching `InstallChannel.read_prefix`.

ivankuznetsov:
## Summary

- QMD installs now use one exact downloaded tarball whose local sha512 SRI must match Hive's checked-in pin. The verified tarball is installed and native-health-checked in a sibling staging tree before activation.
- Failed downloads, installs, rebuilds, probes, timeouts, or activation attempts keep the previous managed QMD usable. The installer also preserves operator-owned launchers and reports optional QMD failures without misreporting them as healthy.
- Prefix sidecars now persist absolute paths for relative and `~/...` inputs, and QMD link ownership no longer depends on GNU `readlink -f` on macOS.

Fixes #209
Fixes #210
Fixes #211
Fixes #212

## Test plan

- [x] `test/unit/install_script_test.rb`: 35 runs, 218 assertions, 0 failures
- [x] `test/unit/install_channel_test.rb`: 13 runs, 27 assertions, 0 failures
- [x] `test/unit/commands/update_test.rb`: 21 runs, 98 assertions, 0 failures
- [x] `test/unit/wiki_log_test.rb`: 7 runs, 27 assertions, 0 failures
- [x] `bundle exec rake test`: 12,742 runs, 160,827 assertions, 0 failures, 0 errors, 12 skips
- [x] RuboCop on `test/unit/install_script_test.rb`: no offenses
- [x] `bash -n install.sh` and `git diff --check`
- [x] Real `@tobilu/qmd@2.5.3` smoke: checked-in SRI matched the downloaded tarball; local `.tgz` install, native rebuild, `qmd 2.5.3`, and in-memory better-sqlite3 query passed
- [x] Current-base head `cec237bf8789ff937f313e109c37de49fd4e8596` contains base `0b16e4f72ac82d9529929180532a74ab96d41fd2`; installer/update/wiki focus passed 79 runs / 384 assertions, plus RuboCop, `bash -n`, and diff/clean-tree checks
- [x] Hosted CI on the current head is fully green. The first Hive web job hit the golden-path task-resolution race after the task was already present in `5-open-pr`; the exact local E2E reproduced once and then passed (1 run / 14 assertions), the same-base `main` E2E passed, and the single targeted full-job rerun `94342026946` passed
- [ ] ShellCheck was unavailable locally; hosted installer CI owns this check

## Notes for reviewer

QMD remains optional, but invalid enabled-QMD configuration fails before network work. Runtime failures warn and preserve the last healthy managed tree. Each optional npm, QMD, or Node subprocess has a configurable 600-second default deadline.

The review specifically challenged the two-resolution `npm view` then `npm install` design. This version removes that gap by hashing the downloaded tarball and installing only that same local file.

---


ivankuznetsov:
Merge-verification follow-up for exact head `604586ee24cb93c55ed79c35465a7bd673baefd1`:

- Closed the remaining dependency-boundary review finding: caller-supplied QMD integrity must now equal the release pin; `npm ci` runs with `--ignore-scripts`; `better-sqlite3` is built from its integrity-checked source by locked `node-gyp` 11.5.0, using local Node headers with offline mode.
- Real native proof: the locked closure installed 341 packages with lifecycle scripts disabled, the explicit source build completed on Node 26.2.0, and an in-memory SQLite query returned `1`.
- Installer unit suite: 42 runs / 261 assertions, 0 failures. Aggregate issue-fix suite: 560 runs / 3,822 assertions, 0 failures.
- The exact-head Install smoke workflow's eight cells are green, including `install.sh fixture regressions`. The temporary `bundler-audit` failure is the repo-wide sqlite3 2.9.5 advisory, addressed independently by #1106 before this PR is merged.

PR #996, opened 2026-08-12
fix(recovery): dispatch pinned historical workflows

ivankuznetsov:
## Summary

- classify managed recovery from the task's immutable pinned workflow descriptor
- resolve durable recovery requests through their canonical task folder and revalidate project/stage identity
- replace the queue-safe slug with the validated canonical folder only at worker dispatch

## Why

PR #995 fixed status-side resolution of historical managed workflow folders. Dogfooding then exposed the next compatibility boundary: the retained Architecture `4-review` task resolved successfully but retry policy inspected the currently installed Architecture descriptor, which no longer contains that stage, and returned `no retry verb for stage 4-review` every scheduler tick.

This remains the existing universal recovery lifecycle; there is no schema or migration change.

## Verification

- `test/unit/daemon/recovery_coordinator_test.rb`: 44 runs, 250 assertions
- `test/unit/attempts/dispatcher_test.rb`: 34 runs, 161 assertions
- `test/unit/bot/handlers/recovery_result_builder_test.rb`: 21 runs, 49 assertions
- `test/integration/status_legacy_layout_test.rb`: 10 runs, 29 assertions
- live read-only proof on retained Architecture task: pinned `architecture` descriptor, `4-review`, retry verb `run`, canonical folder target
- broad `rake test`: 12,725 runs / 171,143 assertions; its two environment-sensitive failures passed separately under the expected Bundler process (packaging 10/48 and gem scripts 2/23)

PR #1001, opened 2026-08-13
fix(workflows): migrate retained tasks on update

ivankuznetsov:
## Summary

- Hive installs and `hive update` now run the newly installed binary's `hive migrate --all` across the current user's registered projects before daemon dispatch resumes. A project failure stops the update with a recovery command and defers daemon restart.
- Retained managed-workflow tasks now move to the selected generation by stable semantic stage name. Folder moves, state-artifact renames, immutable pin rewrites, and workflow pointer activation share one guarded state commit with rollback.
- Runtime accepts only the selected workflow generation. Migration removes stale nonterminal deliveries, consume-time identity checks reject deliveries that race the cutover, and workflow removal refuses while retained tasks exist.

## Test plan

- [x] Focused migration, workflow lifecycle, task admission, durable queue, dispatcher, installer, schema, Rails, and fleet-restart tests pass.
- [x] Multi-task failure tests prove folder, artifact, metadata, and selection rollback.
- [x] Exact default local suite passed through Ruby because the worktree has no `rake` executable: 12,738 runs, 160,835 assertions, 0 failures, 0 errors, 12 skips. CI-only gates remain hosted checks.
- [ ] Hosted CI and dedicated merge gates pass on this exact head.

## Notes for reviewer

Supersedes #996. That PR restored historical package dispatch at runtime; this replacement keeps old-descriptor loading inside the one-way migration boundary and removes the runtime fallback.


ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your session limit · resets 5:40am (Europe/London)
```

PR #1133, opened 2026-08-16
fix(patrol): recover interrupted fix attempts

ivankuznetsov:
## Summary

Patrol no longer crashes the Hive daemon when a killed fix agent leaves a local `attempt` effect uncertain. A recovered cycle now adopts the exact persisted patch when one exists. Otherwise it records an interrupted attempt only after proving the deterministic Patrol checkout has no work to preserve.

Ambiguous, dirty, broken-registration, or uniquely committed checkouts remain recovery-active for operator inspection. The scheduler also keeps occurrences with nonterminal effects blocked instead of calling strict finalization and raising through the daemon tick.

## Safety boundaries

- Patch JSON is subordinate to the outer attempt effect, which removes the nested state-effect crash window.
- Recovery matches the occurrence and 64-character finding fingerprint exactly, then scans patch and finding state once per recovery pass.
- Checkout retirement uses non-force worktree removal and an expected-head ref lease. Work arriving after proof is preserved.
- The Patrol v3 output schema includes the new `interrupted_fix_attempt` terminal reason.

## Validation

- 392 focused/schema runs, 2,267 assertions, 0 failures or errors.
- The broad checkpoint completed 13,561 runs and 192,265 assertions. It exposed the schema omission fixed in this head. The remaining independent failures were four browser-bundle tests caused by the host's broken `npm` mise shim; a sparse-file timeout passed on immediate isolated rerun.
- Five changed production Ruby files pass syntax checks, and `git diff --check` is clean.

---


ivankuznetsov:
## Live dogfood evidence

- Functional replay on `0ee91c561` settled the exact interrupted ordinary-Patrol attempt from `occ-c06418b9947be24631b647ac5f530498b986c1b31f5962f7451afe3199fa772f` as `failed` with patch `patch-ae56fe72ae85-f1243a` and reason `interrupted_fix_attempt`.
- The disposable Patrol worktree and exact branch were retired; no nonterminal effects remain in that occurrence.
- Current PR head `d73927f1c4d2455d2685ae09d14ad3843faddade` is installed locally. The daemon reports binary drift `none`, completed repeated scheduler ticks, and has `NRestarts=0`.
- The second commit routes the same cleanup proof through validated Git operations; focused GitOps/Fixer tests and the exact hosted Brakeman command are green locally.

Residual dogfood findings are independent of this fix: Hive review launches currently hit exhausted Fable credits, and Architecture Patrol job `pr-1015-8181522c6a7dcff1` hits the pre-existing occurrence effect limit.

ivankuznetsov (commented):

The safety design is the point here and it's done well: proof-before-retirement, non-force removal, expected-head ref lease via `update-ref -d <oldvalue>`, fail-closed preservation of anything ambiguous, scheduler blocking instead of raising through the tick. The `write_patch` de-nesting (dropping the per-patch state effect) is itself a good simplification. Tests are adversarially thorough.

Simplification observations, none blocking:

1. **`interrupted_attempt_proof!` returns `[head, head]`** on the registered path — base and head are always identical by construction (the proof *is* that HEAD == branch head == ancestor of default). Returning a tuple suggests they can differ; returning a single `head` (or `nil` for the no-branch case) and letting the caller derive `base_sha:`/`head_sha:` would make the invariant visible in the signature.

2. **Two failure-wrapping helpers with near-identical rescue/re-raise scaffolding** (`interrupted_attempt_proof!` and `retire_interrupted_attempt!` each have the `rescue PublicationRecoveryError; raise; rescue StandardError => wrap` pattern plus `preserve_interrupted_worktree!`). Fine at this size; if a third appears, extract a `fail_closed_preservation(message)` wrapper.

3. **`reconcile_attempts` batch API** is justified by the once-per-recovery-pass scan requirement, but note only one caller needs the batch form today. If recovery stays single-fingerprint in practice, the hash-building/transform_values machinery could collapse back to the simple filter_map — revisit if no second consumer materializes.

4. **Scheduler blocked-event emission** writes directly into `@events` inline; every other terminal path in `finalize_pending` appears to go through helpers. Cosmetic consistency only.

The one thing I'd genuinely push on: **architecture patrol symmetry**. The PR description scopes this to ordinary patrol; `RefactorPatrol::Fixer` has the same killed-process exposure. If symmetric treatment is planned, say so in the PR (or gaps.md) so reviewers know the asymmetry is deliberate.

ivankuznetsov:

Outcome: failure

Agent tail:

```
```

PR #1135, opened 2026-08-16
fix(web): keep waiting questions answerable

ivankuznetsov:
## Summary

A task waiting on bound questions no longer disables its answer controls only because legacy execution checkpoint, attempt, or resource evidence is incomplete. Answer readiness now follows fresh native status plus exact question bindings, while execution actions remain fail-closed. Draft restoration is scoped to the opaque question binding, so a new question round cannot inherit an answer from the previous round. This also corrects the browser-test oracle that had encoded the disabled state as expected behavior, and makes the setup-orchestrator test load the command it stubs instead of depending on test order.

## Validation

- Builder regressions: 15 runs, 78 assertions, no failures
- Web integration: 94 runs, 556 assertions, no failures
- Focused browser regressions: 2 runs, 16 assertions, no failures
- Focused setup orchestrator: 45 runs, 208 assertions, no failures
- Hosted coverage-shard failure reproduced locally with its exact seed; the repaired shard passed with 1,718 runs and 9,064 assertions
- JavaScript syntax check: passed
- Broad checkpoint: 13,554 runs and 192,226 assertions; four transient failures came from the unchanged browser-bundle test losing access to `npm` during the combined run. Its isolated rerun passed with 7 runs and 41 assertions.

---

### DIFF
--- a/lib/hive/task_workspace/builder.rb
+++ b/lib/hive/task_workspace/builder.rb
@@ -33,8 +33,7 @@ module Hive
                      cursor_codec:, limits: Limits.new, attempt_store: nil,
                      projection_store: nil, projection_read: nil,
                      event_reader: nil, journal_reader: nil, usage_reader: Hive::UsageDb,
-                     current_context_observation: nil, questions_count: nil,
-                     questions: nil,
+                     current_context_observation: nil, questions: nil,
                      daemon_enabled: true, archive: false,
                      clock: -> { Time.now.utc })
         @task = task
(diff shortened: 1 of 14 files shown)


ivankuznetsov:

Outcome: failure

Agent tail:

```
That failure is environmental — the test shells out to `bundle exec rake` in a subprocess that can't see my hand-set gem path (`Bundler::GemNotFound`), unrelated to my change. Let me confirm the byte-balance assertion specifically passes.
```

PR #1148, opened 2026-08-17
fix(agents): make Pi workflows autonomous and review-safe

ivankuznetsov:
## Summary

Make Pi-backed Hive workflows detect real provider outcomes and recover end to end without manual retries or stage advancement, while mandatory plan review and controller-owned artifacts remain fail-closed.

This PR closes the failure chain exposed by live Webmail dogfooding:

- Pi provider refusals, upstream stream failures, quota responses, and output-limit stops are typed even when the CLI exits zero.
- Provider-reported usage and cost survive normalization, and large prompts travel over native non-TTY stdin instead of a single argv element.
- Plan authors and reviewers run in disposable Git worktrees with repository, search, shell, and network access; review identity includes the selected routes and changing projection.
- The daemon owns bounded retries, incomplete verification, exact unchanged markerless or dirty-work recovery, and deterministic-failure parking.
- Protected artifacts use one bounded aggregate manifest and exact snapshot/restore custody across agent spawns.

The `agent-cli-runtime` component metadata is prepared at 0.2.3. This PR does not tag, publish, release, or deploy Hive.

## Design decisions

- **Type provider failures once at the runtime boundary.** Profile-owned extractors classify structured Pi events and recovered text; Hive consumes the typed result instead of re-parsing provider syntax.
- **Prefer terminal completion evidence.** A completed response wins over a recoverable mid-stream provider error, while a zero-exit refusal without completion remains a failure.
- **Give reviewers a real checkout.** Plan review and Codex planner revisions share one detached disposable-worktree implementation, so they can inspect code and wiki context without mutating the operator checkout.
- **Keep review fail-closed but self-healing.** Route capability is part of review identity, incomplete attestations retry within a bounded round budget, and capability-only gaps do not become permanent judgments about a plan.
- **Use one custody boundary.** Artifact Firewall validates one aggregate inventory of protected anchors, required outputs, and writable roots with a hard 4,096-entry bound.
- **Use one retry ladder.** Fresh request identity and progress evidence control retry pacing; three identical deterministic failures park durably until a freshness-bound `workflow.retry` explicitly releases them.
- **Make exceptions exact and auditable.** Guardrail exceptions are `(pattern, SHA-256)` operator waivers that emit events; the legacy broad bypass key is inert.

## Operational effect

| Before | After |
| --- | --- |
| Pi could exit zero after a provider refusal | Hive records a typed provider failure and schedules the correct retry |
| A large plan could fail before launch with `E2BIG` | The prompt is supplied on stdin with bounded argv |
| A repaired reviewer route replayed stale blocked coverage | The route fingerprints a fresh review and capability gaps retry |
| Reviewers saw only a copied plan file | They inspect the full repository from disposable Git worktrees |
| Repeated identical failures consumed hourly attempts indefinitely | The workflow parks after three stable failures and stays parked |
| Custody limits applied independently to multiple collections | One aggregate manifest enforces the real total bound |

## Test plan

- [x] Exact-head hosted CI: 32/32 checks green at `33066da85b9c3f7f16f1be153fcf1aa325f773fe`.
- [x] Hosted aggregate Ruby coverage: 100%.
- [x] Hosted `rake test (Ruby 3.4)`, RuboCop, Brakeman, Bundler audit, install smoke, packaged web, TUI, multi-agent, babysitter security, and real-CLI gates green.
- [x] Focused post-merge-conflict regressions: 116 runs / 607 assertions, green.
- [x] Final coverage regressions cover durable deterministic parking and disposable-worktree cleanup failures.
- [x] `git diff --check` clean.

## Notes for reviewer

- The last architecture and simplification review rounds are fully addressed, including the seven deferred cleanup items; the cleanup pass is net-negative in production/test surface.
- A local aggregate prerequisite can fail when the installed OpenCode inventory does not expose its configured smoke-test model. The exact-head hosted component and integration gates are green.
- The repository wiki and append-only `wiki/log.d` fragments document the provider, plan-review, custody, attempt, and daemon recovery contracts.


ivankuznetsov:

Outcome: timeout

Agent tail:

```
Pushed as a fast-forward. Let me monitor CI.
```

ivankuznetsov:

Outcome: timeout

Agent tail:

```
A new commit from the PR author landed on top of mine (my commit is intact, no conflict). Let me check which SHA the green run covered and whether the new commit has CI.
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your session limit · resets 10:10pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 20, 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets 1pm (Europe/London)
```

ivankuznetsov:
# Architectural review — actionable fix list

Consolidated from a five-subsystem architectural review of this branch. Ordered by priority: **P0** confirmed defects, **P1** simplifications that delete mechanism, **P2** the autonomy/spin-detection model, **P3** smaller cleanups. Executing agent: work top-down; each item has a fix shape and acceptance criteria. The **Do NOT change** list at the bottom marks load-bearing designs that must survive the fixes. Line anchors are as of head `8db8ea594`; prefer the named methods/constants if lines drift.

---

## P0 — Confirmed defects

### P0-1. The motivating 402 payload still misclassifies as generic `provider_error`
`Hive::Agent#detected_limit_text` (lib/hive/agent.rb): `typed_provider_limit` (402/429 via `provider_limit_status?`) opens the gate, but the method then **re-gates on `Hive::AgentLimit.limit_reached?` wording**. The verbatim incident text `Prompt tokens limit exceeded: 25770 > 8471` matches no `LIMIT_PATTERNS` entry (plural "tokens" defeats `token[\s-]?limit`; nothing covers 402). Result: no `limits_reached`, no `retry_after` — a hard credit ceiling is retried on the ordinary ladder. This contradicts three in-tree claims (the "status code is authoritative" comment in agent.rb, wiki/log.d/20260817T095455Z, and the stages/artifacts.rb cooldown comment — `role_agent_limit?` never reads `status_code` either).

**Fix shape:** when `provider_error` carries a limit status, classify `limits_reached` on status alone; wording remains a fallback for untyped streams only. Full version is P1-2.
**Acceptance:** end-to-end test — zero-exit run whose stream carries the verbatim `402: {"message":"Prompt tokens limit exceeded: 25770 > 8471",...}` payload produces `reason=limits_reached` with `retry_after`, in both execute and artifacts stages.

### P0-2. `provider_error` outranks completion evidence in `handle_exit`
In `state_file_marker` mode, `handle_provider_error` runs before `handle_exit_state_file_marker`, and the stream scan latches the **first** error event for the whole run (`provider_error ||=`). A zero-exit run that wrote a valid terminal marker but emitted one transient error event mid-stream (e.g. a CLI that internally retried and recovered) is overwritten as ERROR — for **every** provider, not just Pi. This is a cross-provider behavior change the 0.2.1 changelog explicitly denies.

**Fix shape:** order the `provider_error` branch after marker/completed-output evidence; completion evidence beats a mid-run error event.
**Acceptance:** test — zero exit + valid terminal marker + earlier stream `error` event ⇒ marker trusted, run succeeds.

### P0-3. ce_doc_review's hand-rolled custody batching skips remaining batches on a validation raise
lib/hive/plan_review/adapters/ce_doc_review.rb hand-rolls `manifests.map{capture}` + `zip(snapshots).reverse_each.map{validate_and_restore}`. If `validate_and_restore` raises for one batch, the map aborts: remaining batches are neither validated nor restored, and the adapter-level `rescue ArtifactFirewall::Error` reports failure while silently skipping custody checks on the rest of history. `ProtectedAnchorCustodySet` handles exactly this via per-guard `ensure` — but plan review never adopted it.

**Fix shape:** superseded by P1-1 (delete batching entirely). If P1-1 is rejected, retrofit this adapter onto `ProtectedAnchorCustodySet` (core manifest via a plain `AgentCustody` wrapping the set).

### P0-4. Patrol cooperative shutdown is unreachable during agent runs
`Hive::Agent#run!` replaces the INT/TERM handlers with `kill_group(pgid)` for the duration of every child run (lib/hive/agent.rb, `old_int = trap("INT") { kill_group(pgid) }` ~:369) and restores the originals afterwards — without chaining. Patrol's reviewer and fixer spend most wall time inside `Hive::Agent`, so a daemon SIGTERM arriving mid-agent kills the in-flight child but never sets `Patrol::Shutdown.@requested`; patrol spawns the **next** agent and still runs until SIGKILL. This mostly defeats dbe535277.

**Fix shape:** chain to the previous handler from the replacement (or set the shared shutdown flag inside it). Also: `Shutdown.install_trap!` overwrites prior handlers unconditionally — make it chain too.
**Acceptance:** test — TERM delivered while the agent-run handler is installed ⇒ `Shutdown.requested?` is true after the run returns.

### P0-5. Widened reviewer-loop rescue grants repair rounds to custody violations
stages/artifacts.rb widened the reviewer-loop rescue from `RoleOutputError` to its superclass `StoreError`. Attempt 1's "reviewer modified protected task state" custody violation now buys a fresh reviewer run, and if attempt 2 is clean the violation is masked. A custody violation is not repairable by one more reviewer turn — the protected state was already touched.

**Fix shape:** rescue only verdict-shape errors (`RoleOutputError` + `Contract.verdicts!` violations); custody violations stay terminal.

### P0-6. Guardrail false negative from the receiver allowlist
For `password: params[:x] || "realsecret99"` the scanner's single hit is `password: params[:x]` (value token stops at whitespace), which `dynamic_password_reference?` now classifies as dynamic and skips — the literal fallback on the same line goes unflagged, where pre-PR the line tripped.

**Fix shape:** superseded by P1-3 (delete the allowlist). If kept, only skip when the whole RHS is a lookup (unquoted **and** containing no string literal).

---

## P1 — Radical simplifications (delete mechanism, keep behavior)

### P1-1. Delete `ProtectedAnchorCustodySet`; parameterize the manifest bound
Nothing about manifests is persisted — they are ephemeral in-memory objects — and all 16 batches' snapshots are held simultaneously during the spawn, so 16×128 nested guards are byte-for-byte the same peak memory as one 2048-entry manifest. The landed design is "raise the effective bound to 2048, with ceremony."

Replace the set class, `CombinedReport`, execute's `each_slice` partitioning, and ce_doc_review's hand-rolled loop with `Manifest.new(..., max_entries:)` — default 128, hard cap (e.g. 4096) so the bound stays fail-closed and non-negotiable except where explicitly widened. ~150–200 lines of mechanism → ~5, one code path instead of two divergent ones, and **P0-3 is fixed by deletion**.

Also: file the actual driver in wiki/gaps.md — receipt/record growth is unbounded with no retention policy, so any fixed bound is a deferred wall (today it would raise `InvalidManifest` from the set constructor at execute.rb outside the begin/rescue, reproducing the original pre-spawn failure). Solve once with receipt retention/compaction, not custody topology.

### P1-2. Classify provider failures once, at the extractor
"Is this a limit?" is currently derived in three-and-a-half places: the component's `status_code_from` regex, agent.rb's `provider_limit_status?` **plus** the legacy wording re-gate in `detected_limit_text`, and `role_agent_limit?` in stages/artifacts.rb re-matching wording a fourth time. The `EXTRACTED_FAILURE_KINDS` enum exists but has exactly one member.

**Fix shape:** populate typed kinds at the extractor where the provider shape is known — `:provider_limit` (402/429/quota), `:rate_limited`, `:model_output_limit`, `:provider_error` — and make `handle_exit` a dispatch on kind. Delete the wording re-gate for typed events, delete the artifacts-stage wording re-match (dispatch on the marker's typed reason instead). `AgentLimit` wording patterns survive only for live-TUI wall scraping where no typed event exists. **Fixes P0-1 structurally.**

Release hygiene: cut **agent-cli-runtime 0.2.3** for this. 0.2.2's dated changelog section was retroactively amended twice (`model_output_limit`, `piped_stdin` both appended after the 2026-08-18 date) — do not amend it a third time.

### P1-3. Replace guardrail allowlists + retroactive bypass with fingerprint-scoped waivers
Delete `DYNAMIC_PASSWORD_REFERENCES` (fix_guardrail.rb), `OBVIOUS_TEST_PASSWORD_VALUES` (secret_patterns.rb — it grew twice within six hours of landing; that's the treadmill signature), and the retroactive-release paths of `review.fix.guardrail.bypass` (stages/review.rb checkbox-free advance + task_action.rb reclassification of parked rows).

Replace with a per-project waiver list keyed on the `[pattern_name, SHA256(match)]` fingerprints auto_commit already computes: waiving is one auditable, revocable human action per unique finding, ever. Emit a loud event whenever a parked finding is released by waiver rather than checkbox. Optional single generalization replacing all 14 receivers: unquoted RHS containing `[`, `.`, or `(` **and no string literal** ⇒ lookup, skip. **Fixes P0-6 by deletion.** (Also resolve the truthiness inconsistency: `run!` skips on any truthy bypass, release paths require `== true` — standardize on `== true`.)

### P1-4. Disposable git worktree as the reviewer cwd
workspace_scope.rb is a stack of per-provider carve-outs: Codex needs `--skip-git-repo-check` (cwd isn't a checkout), Claude's workspace-scoped `Edit` is bypassable via unscoped `Bash`, Pi runs fully unrestricted, and the templates' "do not edit project source" is advisory for two of four providers. A disposable `git worktree` as cwd collapses most of it: Codex sees a real checkout (flag deleted), "don't touch the live tree" becomes structural for everyone, and only the redacted plan copy remains as per-review setup. Cost is ~hundreds of ms against 30+ minute reviewer runs.

### P1-5. One-shot migration for the `dirty_worktree` legacy bridge
`attributed_dirty_execute_waiting?` (stale_agent_healer.rb) rewrites a marker state new code can never produce (`derived_wait_reason` no longer returns it; execute writes `:error` directly) — it is migration code resident in the perpetual tick loop. Move it to a one-shot `hive migrate` step (or repair the known affected task) and delete the healer branch; at minimum attach an explicit sunset condition. Otherwise every marker-vocabulary change accretes a permanent sibling.

---

## P2 — Autonomy model: spin detection instead of caps or human gates

Design principles: maximal agent autonomy, no hard limits; minimize human interruptions; escalate real issues; never run in rounds. Mechanism: **evidence of change, not counts**. A retry is justified exactly when something might differ this time; provably-identical repetition at the ladder ceiling is the definition of running in rounds.

### P2-1. Failure fingerprinting in the recovery request
Fingerprint each terminal failure: `reason + provider + status_code + digest(normalized message)`, optionally + digest of produced output/worktree diff. Track the consecutive-identical count in the durable recovery request, next to the existing `retry_count` (dispatch_request_queue).

- Fingerprint changed between attempts ⇒ retry freely on the ladder, forever. (The DigitalOcean stream-failure recovery this PR proved stays fully autonomous.)
- K consecutive identical fingerprints at the ladder ceiling ⇒ the daemon has proven its own actions have no effect: park with `reason=deterministic_failure` + fingerprint + attempt history, stop charging daily slots, continue other tasks.
- Keep the no-exempt-class retry policy exactly as landed — parking is **per-instance on evidence**, never per-class.

Note: the primitive already exists in this branch — `repair_failed_terminal_marker` uses exact-unchanged byte matching to justify *acting*; this uses the same unchanged-detection to justify *stopping*.

### P2-2. Probe-before-dispatch in the plan-review capability loop
`reset_for_capability` (plan_review/orchestrator.rb) has no cap, no backoff: an unlaunchable mandatory reviewer redispatches every daemon tick forever, each cycle appending route rows to current.json and writing 4 immutable attempt files; `capability_only_block?` also treats exhausted transients (all-timeout) as non-verdicts, so an always-timing-out reviewer burns 1800s + budget per attempt indefinitely with no terminal `blocked` ever surfacing.

**Fix shape:** re-run the **cheap capability probe** each cycle; dispatch the expensive reviewer attempt only when the probe outcome changes. Stable-failed probe after K cycles ⇒ park the review as `blocked reason=reviewer_unlaunchable` with the diagnosis — capability failures (install skill, fix gem, credential) are human-only fixes, so this is escalation, not nagging. Record the rule in the wiki: **unbounded retries are fine when the retry is cheap; expensive retries need an evidence gate, not a count gate.**

### P2-3. Tiered escalation semantics — never a blocking question
- **Tier 1** silent retry: fingerprints varying. Zero human involvement.
- **Tier 2** retry + `degraded` operational-status row: fingerprints repeating, below ceiling. Discoverable in `hive status`/watch, not interruptive.
- **Tier 3** park + at most one notification: proven-deterministic failures (P2-1), reviewer-unlaunchable (P2-2), contract violations, security findings (P1-3 waiver events).

Escalation = durable parked marker with a real diagnosis + one status row + one notification. The daemon never waits on an answer; the human acts whenever.

### P2-4. Same treatment for hive-babysitter (follow-up; separate PR fine)
This PR's own thread carries 30+ identical hourly "hit your weekly limit · resets 1pm" give-up comments over two days — the anti-pattern live: known reset time, hourly rounds anyway, one external side effect per round, no dedup. Fingerprint give-up reasons, dedup comments, honor known reset times with one scheduled retry.

### P2-5. 402 ⇒ `limits_reached` is autonomy work
Framing for whoever implements P0-1/P1-2: a quota wall with a posted reset is the environmental case with a known end — cooldown-until-reset **is** the autonomous behavior; generic ladder retries against a credit ceiling are rounds.

---

## P3 — Smaller cleanups
- Finish "one retry ladder": provider-limit reprobes and lost-attempt heal pacing (`attempt_loss_retry_due?`) still use the flat `AgentLimit.retry_cooldown_sec` hour.
- Plan-review identity fingerprint: split operational tuning (`attempts` timeout/retry — this PR's own 900→1800s bump re-keys every in-flight review on upgrade) **out**; add `reviewers`/`adapter` config **in** (changing the review skill must invalidate cached verdicts; a timeout bump must not).
- Replace the byte-hash-of-current.json progress token and the `argv[1] == "plan-review-run"` carve-out in the generic attempts dispatcher with a semantic `(review_id, version)` token supplied by the owner (plan_review), or a general "commands may contribute a progress token" interface.
- `remove_terminal_recoveries` has no stage filter — admitting a new stage's recovery deletes the prior stage's terminal docs, silently resetting ladder history on stage re-entry.
- `StaleAgentHealer#retry_assessment` passes no `retry_count` (rung 0), causing a per-tick task-lock + full queue-directory parse per parked error row and early "scheduler due" display.
- attempt_observer: `rescue Errno::ENOENT → :pending` can leave a deleted task folder perpetually pending — map to not_applicable.
- Occurrence journal: transition stranded regenerable effects to a terminal state (e.g. `abandoned`) instead of silent `delete_if` — keeps the audit trail the journal exists for.
- Council: extract the duplicated pre-spawn/post-round budget check; make the terminal marker's round number consistent between the two paths.

---

## Do NOT change (load-bearing, correct as landed)
- Retry ladder + stage-scoped durable backoff + fresh launch timestamps — they compose correctly; no storm paths.
- `repair_failed_terminal_marker` — general, generation-guarded, fail-closed.
- Review identity keying (plan digest + task generation + policy/routes fingerprint, linked lineages) and never caching capability failures as verdicts.
- Detection-and-restore custody doctrine — do not prune receipts from custody; do not pretend confinement.
- No-exempt-class auto-retry — replace class-parking with evidence-parking (P2-1); do not reintroduce exemptions.
- `piped_stdin` transport, usage/`provider_reported_cost` extraction, effort sentinel, open-pr task-lock fix, bot lock-holder attribution — clean as landed.

---
🤖 Architectural review generated with [Claude Code](https://claude.com/claude-code) — a re-review will follow once fixes land.

ivankuznetsov:
Addressed in `c1c465a33`.

### P0 and P1

- **P0-1 / P1-2:** provider failures are classified once in agent-cli-runtime as `provider_limit`, `rate_limited`, `model_output_limit`, or `provider_error`. HTTP 402 and 429 are authoritative. Hive trusts typed reasons; wording is now only the untyped legacy fallback. Exact zero-exit 402 regressions cover Agent, Execute, and Artifacts.
- **P0-2:** valid terminal marker or required-output evidence now wins over an earlier transient provider error from a CLI that recovered internally.
- **P0-3 / P1-1:** deleted custody batching, `ProtectedAnchorCustodySet`, and `CombinedReport`. Execute and plan review each use one exact-inventory manifest with a 128 default and 4096 hard ceiling. The actual unbounded receipt-retention driver is recorded in `wiki/gaps.md`.
- **P0-4:** Agent INT and TERM handlers chain the previously installed handler; Patrol shutdown does the same. TERM during an agent run now sets cooperative shutdown state.
- **P0-5:** Artifacts retries only verdict-shape failures. Custody violations remain terminal.
- **P0-6 / P1-3:** deleted receiver/value/path allowlists and retroactive global bypass release. Guardrail waivers are exact `(pattern, SHA-256(full detector match))` entries and every applied waiver emits a task event. Mixed lookup plus literal RHS remains detected.
- **P1-4:** plan reviewers run in detached disposable Git worktrees. They can inspect repository code/wiki, use shell and network, and write only inside the disposable checkout. The live checkout is structurally separate; controller artifacts remain under the firewall. Codex no longer needs `--skip-git-repo-check`.
- **P1-5:** the legacy attributed `dirty_worktree` execute-wait rewrite moved from every daemon tick into one-shot `hive migrate`.
- **Release hygiene:** agent-cli-runtime `0.2.3` is prepared with version, changelog, root/web locks, package, mirror, and consumer assertions. It is not tagged or published by this PR.

### P2 autonomy model

- **P2-1:** durable recovery requests carry the requested failure fingerprint, consecutive-identical count, and bounded attempt history. Changed evidence retries freely. Three identical failures at the ladder ceiling park once as `deterministic_failure` without consuming further dispatches.
- **P2-2:** plan-review capability recovery keys only on cheap `unsupported` probes. Stable failed probes park as `reviewer_unlaunchable`; timeouts remain ordinary expensive-attempt outcomes rather than being mistaken for capability failures.
- **P2-3:** receipts and operational status expose `silent_retry`, `degraded`, and `parked` escalation tiers plus fingerprint/count evidence. The daemon never blocks waiting for an answer.
- **P2-4:** intentionally deferred to the separate babysitter PR allowed by the review. No babysitter behavior changed here.
- **P2-5:** typed 402 outcomes become `limits_reached` with reset-aware cooldown instead of generic ladder churn.

### P3

- Attempt-loss retry pacing now uses the shared recovery ladder.
- Plan-review identity includes adapter/reviewer routes and excludes timeout/retry tuning.
- Generic attempt commands can contribute semantic progress; plan review contributes `(review_id, version)` instead of hashing `current.json` or relying on an argv carve-out.
- Terminal recovery cleanup is stage-scoped.
- Stale-healer assessment loads durable retry count.
- Deleted task folders observe as `not_applicable`.
- Stranded regenerable occurrence effects remain as terminal `abandoned` audit cells.
- Council max-round exits share one helper and consistent terminal round numbering.

### Preserved load-bearing behavior

The retry ladder, stage-scoped durable backoff, generation-guarded terminal-marker repair, linked review identity, detection-and-restore custody, no-exempt-class retry policy, Pi stdin transport, usage/cost extraction, effort sentinel, open-PR task locking, and bot lock attribution remain intact.

### Additional dogfood result

The live Ox reviewer completed a free 27,351-token repository review through Pi, but used an invalid closed finding lifecycle after the prompt asked for a closed outcome. The prompt now says terminal adapter outcome and requires every new finding to start at `lifecycle: open`; Hive alone owns later transitions. The rendered primary/adversarial/verification prompt regression is included.

### Verification

- `bundle exec rake test`: **13,839 runs, 193,668 assertions, 0 failures, 0 errors, 12 skips**
- agent-cli-runtime suite in the same checkpoint: **106 runs, 1,096 assertions, 0 failures, 0 errors, 2 skips**
- post-checkpoint prompt regression: **25 runs, 140 assertions, 0 failures, 0 errors**
- RuboCop: **58 changed Ruby files, 0 offenses**
- `git diff --check`: clean

ivankuznetsov:
# Re-review at `15f595d6a` — checklist verified, 2 blocking items remain

Verification of the fix round against the prior review comment (all 32 CI checks green; every claim below checked in code, not from the wiki fragment).

## Verified fixed ✅

Every P0, P1, and P3 item landed and is covered by tests:

- **P0-1/P1-2**: typed failure kinds (`provider_limit`/`rate_limited`/`model_output_limit`/`provider_error`) populate once at the runtime boundary (`failure_kind_from`); the wording re-gate is gone from `detected_limit_text`; the verbatim 402 payload has an end-to-end test asserting `limits_reached` + `retry_at`; artifacts dispatches on the typed reason; **0.2.3** cut with its own dated changelog section.
- **P0-2**: `recovered_provider_error?` — completion evidence beats an earlier stream error event, with test.
- **P0-3/P1-1**: `ProtectedAnchorCustodySet`/`CombinedReport` deleted repo-wide; one manifest with `max_entries:` (default 128, fail-closed `HARD_MAX_ENTRIES = 4096`); ce_doc_review's skip-remaining-batches hazard structurally eliminated; receipt-growth wall filed in wiki/gaps.md.
- **P0-4**: chained signal traps in both `Agent` spawn paths and `Patrol::Shutdown`.
- **P0-5**: reviewer-loop rescue narrowed to `RoleOutputError`; custody violations terminal again, with test.
- **P0-6/P1-3**: allowlists and the retroactive bypass fully deleted (legacy `bypass: true` pinned inert by three tests); waivers are exact `(pattern, SHA-256(full match))` tuples in operator-owned config, malformed waivers fail closed, applied waivers emit task events; `password: params[:x] || "literal"` now trips.
- **P1-4**: reviewers run in detached disposable git worktrees with cleanup in `ensure`; `--skip-git-repo-check` deleted; live checkout structurally separate. (But see R2-2.)
- **P1-5**: dirty-worktree bridge moved to idempotent one-shot `hive migrate`; healer branch deleted.
- **P2-2**: capability probe is genuinely cheap (skill inventory, no spawn); expensive dispatch only on probe change; 3 stable probes park terminal `blocked reason=reviewer_unlaunchable`; all-timeout now bounded by `max_transient` into recoverable `blocked`.
- **P3**: attempt-loss pacing on the ladder; stage-scoped terminal cleanup and retry counts; healer rung-0 busy-poll gone; `ENOENT → :not_applicable`; identity fingerprint drops `attempts`, adds `reviewers`+`adapter` (note: the key-set change re-keys all in-flight reviews on upgrade — expected, mention in release notes); semantic `(review_id, version)` progress token via `CommandProgress`; stranded regenerable effects → terminal `abandoned` with validator support; council max-round paths unified.

New commits (generation-mismatch deferral, brainstorm resume, planner-revision custody isolation/recovery, verification follow-up continuation) are sound overall.

## R2 — remaining items

### R2-1 (blocking): deterministic-failure parking is written but not enforced
The park is real for exactly one tick. The parked request stays in the pending queue, and every daemon tick `process_dispatch_request_iteration` calls `RecoveryCoordinator#resume` on it unconditionally; `resume`'s admitted branch never consults `blocked_reason` and transitions admitted→cleared with `"blocked_reason" => nil` (recovery_coordinator.rb `resume`, dispatcher.rb dispatch iteration). Net behavior: park → next-tick un-park → dispatch → fail → count+1 → re-park → … — the hourly retry storm continues forever and **daily slots keep being charged**; the "parked" tier is visible only sub-tick. Only 3 references to `deterministic_failure` exist repo-wide (the write, the tier mapping, one test) — nothing enforces it.

**Fix:** `resume` (or the dispatch iteration before it) must refuse admitted-phase requests with `blocked_reason == "deterministic_failure"` unless the transition is an explicit operator action. Add the missing operator un-park verb (`task_action`/CLI) so the remediation text ("change the failing input … before retrying") is actually actionable — today the only exit is manual file surgery.
**Acceptance:** test that a parked request survives N dispatch ticks unchanged (no dispatch, no slot charge, tier stays `parked`); test that the operator verb un-parks and re-enters the ladder; test that a changed fingerprint after un-park resets the count.

### R2-2 (blocking): probable codex planner-revision launch regression
`PlannerRevision` still runs in a plain non-git `Dir.mktmpdir` (planner_revision.rb `workspace`), shares `WorkspaceScope.launch_kwargs`, and the codex `--skip-git-repo-check` that previously covered this path was deleted (the removed comment itself said codex refuses non-checkout dirs). A codex-identity planner revision should now fail at launch → retry exhaustion → blocked. No test covers codex + planner_revision.

**Fix:** give planner revision the same disposable-worktree treatment as reviewers (preferred — one mechanism), or restore the skip flag for the planner_revision role only.
**Acceptance:** unit test asserting codex planner-revision launch kwargs work against a non-checkout cwd, or that the cwd is a worktree checkout.

### R2-3: verification revision cap doesn't gate the re-entry path
`MAX_VERIFICATION_REVISION_ROUNDS = 3` gates only the auto-continue transition. A capped-blocked record still has accepted findings + a terminal success verification route, so any subsequent `advance!` re-enters the accepted-findings path in `resume_review`, dispatches planner revision #4 + verification (two expensive spawns per external nudge), and re-blocks at rounds 4, 5, … Not self-driving (blocked is operator-owned) but the guard is leaky and untested.
**Fix:** consult the round count in the re-entry path too; acceptance: round-4 `advance!` on a capped record is a no-op receipt, no spawns.

### R2-4: status-schema drift for the new blocker/route fields
The `reviewer_unlaunchable` blocker carries `fingerprint`/`diagnostic` and reset routes carry `capability_probe_fingerprint`/`capability_probe_count`/`verification_followup` — none declared in the strict (`additionalProperties: false`) `PlanReviewBlocker`/`PlanReviewRoute` schemas, so a parked or follow-up review's status output violates the published contract (the commit titled "declare plan review route diagnostics" declared only one of the new fields). Add the fields to both closed schemas + a shape test that emits a parked blocker through the schema.

### R2-5: `Store#progress_token` crash paths
`File.binread(path, len)` returns nil for an empty file → `NoMethodError`; valid non-object JSON (`null`, array) raises through `projection["review_id"]`; oversized file raises `InvalidRecord` — all uncaught at the dispatcher call site, in a method whose rescue list shows corruption tolerance was intended. Degrade to the unreadable-token path instead.

### R2-6 (minor): guardrail residual on unquoted dotted values
The new structural heuristic skips any fully-unquoted RHS containing `.`/`[`/`(`, so `export DB_PASSWORD=s3cr3t.pass42` no longer flags (old prefix list caught it; `.env` file edits still trip via `dotenv_edit`). Consider requiring a known lookup shape (identifier before `.`/`[`/`(`) or at least documenting the trade.

### Notes, no action required
- `>4096` anchors in execute raises `InvalidManifest` pre-spawn but unhandled (CLI backtrace, no `:error` marker) — documented as deliberate in wiki/gaps.md; plan review rescues the same breach into `terminal_failure`. Asymmetry worth remembering when the receipt-retention gap is picked up.
- Worktree leak on SIGKILL between add and cleanup (prune is opportunistic); spurious double "cleanup failed" warn on add-failure; stale "batches" test name; stale comment above `capability_only_block?` describing the old any-verdict semantics.
- `hive migrate` dirty-wait rewrite takes the project lock but not the per-task lock the healer used — narrow race for an operator-initiated one-shot.
- Deferred non-recovery durable commands behind a long-lived different-generation owner still expire at 600s (pre-existing policy; strictly better than attaching to the wrong generation).
- Finding-lifecycle enforcement is prompt-only; a new finding emitted as `approved`/`answered` enters `accepted_findings` without a decision (pre-existing, but worth a parser-side clamp to `open` for unknown fingerprints as a follow-up).

---
🤖 Re-review generated with [Claude Code](https://claude.com/claude-code) — R2-1 and R2-2 are the merge blockers; the rest can ride this PR or a fast-follow.

ivankuznetsov:
## Simplification review (follow-up)

Baseline: no prior inline threads; the architectural round recorded in `wiki/log.d/20260821T113000Z-pr1148-architectural-review-fixes.md` (single custody manifest, disposable reviewer worktrees, exact fingerprint waivers, stage-scoped fingerprints + retry ladder, one-shot legacy migration) is correctly implemented — no regressions against those decisions. Findings below are simplification-only, ranked.

### High impact
1. **Dead template branches + contradictory pi anchor recompute** — `plan_review/adapters/ce_doc_review.rb:370-371` hardcodes `host_computed_anchors: false` / `host_output_mode: false`, so the `<% if %>` blocks in all three `templates/plan_review_*.md.erb` are unreachable; meanwhile `normalize_pi_result` (`ce_doc_review.rb:377-403`) host-recomputes anchor digests for pi anyway, duplicating the prompt contract. Delete both.
2. **Dead `WorkspaceScope.supported?`** — `workspace_scope.rb:33` is now `= true`, zero production callers, kept alive by two tautological tests (`workspace_scope_test.rb`). Delete method + tests.
3. **Duplicated secret scan in auto_commit** — `stages/auto_commit.rb:288` hand-walks `SecretPatterns::PATTERNS` even though this same PR added `sha256` to `SecretPatterns.scan`. Collapses to `SecretPatterns.scan(text).map { |h| [h[:name], h[:sha256]] }`. Also delete the dead `path:` kwarg threaded through 3 methods there.
4. **Double rescue boundary on error extraction** — `agent_runtime.rb:260-268` rescues around extraction that `components/agent-cli-runtime/.../profile.rb:226-231` already rescues. Keep one (the component layer); drop the Hive-level rescue + its duplicate test.
5. **Byte-identical recovery-reset logic in two files** — `orchestrator.rb:448-456` vs `decision_service.rb:377-386`; extract one shared helper.

### Medium
6. **Council budget checked three times** — `council.rb:53,93,168`; one top-of-loop `>= max_rounds` suffices (also fixes a resume at exactly `max_rounds` paying an extra full reviewer round).
7. **Firewall `entry_limit` isn't actually total** — `artifact_firewall.rb:140-151` caps anchors/outputs/roots independently (3× limit possible). Normalize first, check aggregate once.
8. **Dead production method** `Dispatcher#brainstorm_answers_pending?` (`daemon/dispatcher.rb:1393`) — only tests reach it via `send`; point them at `brainstorm_answer_state`.
9. **Triplicated queue scan loop** — `dispatch_request_queue.rb:584-620`; one private `each_matching_request` iterator collapses ~15 lines and makes ENOENT handling uniform.
10. **`resume_review` publish shape repeated 6×** — `orchestrator.rb:188-193,224-227,385-393,431-439,822-825,838-844`; extract a `publish_transition` helper.
11. **Over-defensive nil/rescue layers** — healer's inline ladder fallback (`stale_agent_healer.rb:305-313`), unreachable `value()` rescue (`recovery_coordinator.rb:1129`), nil-attempt guard in `superseding_loss?` (`attempts/dispatcher.rb:486`) + its test, dead UTF-8 fallback branch in `plan_text` (`orchestrator.rb:124`), dead emptiness re-check (`plan.rb:68`).

### Low / test bloat
12. Launcher kwargs that exist only to be rejected (`claude_launcher.rb:240-253`); `EXTRACTED_FAILURE_KINDS` registry guarding a single literal construction site (`values.rb:84-100`); duplicated pi-refusal fixtures across component + Hive seams; twin provider-error-beats tests in `agent_test.rb`; PermanentProofStore sidecar-read duplication (`permanent_proof_store.rb:45-55` vs `102-116`); waiver pattern name printed twice per finding line (`review.rb:2252`); `resolve_waivers` Hash-as-set.

### Verdict
Directionally sound and honors every prior architectural decision; no further structural complaints on recovery/fingerprint/custody. But 39 commits of accretion left dead branches, double-guarded boundaries, and copy-pasted helpers that a focused cleanup pass (~200–300 lines net deletion) should remove before merge. Items 1–5 are worth requesting changes over; 6–12 nice-to-have.

ivankuznetsov:
Addressed in `5b1039a0c`.

> R2-1 (blocking): deterministic-failure parking is written but not enforced

All six round-two items are fixed:

- Deterministic parks now survive repeated daemon ticks without a child spawn or daily-slot charge. The existing freshness-bound `workflow.retry` action is the explicit operator unpark after remediation, and changed evidence starts a fresh failure series.
- Reviewer and original-planner revision agents now share one detached disposable Git-worktree implementation, so Codex revisions run in a real checkout.
- The three-round verification cap is enforced on external `advance!` re-entry; a capped record is a no-op and launches nothing.
- Both strict status schemas now include emitted blocker fingerprints/diagnostics and capability/reset route fields. The regression projects a real operational-status payload through the schema.
- Empty, non-object, invalid-identity, and oversized plan-review projections degrade to a stable unreadable progress token.
- Prefixed password assignments with dotted literal values are detected; only syntactic bracket/call lookup shapes receive the runtime-reference exemption.

> Items 1–5 are worth requesting changes over

All five requested high-impact simplifications also landed:

1. Deleted the unreachable host-output/host-anchor template branches and Pi anchor rewriting; every provider now follows one exact-file/anchored-evidence contract.
2. Deleted the tautological `WorkspaceScope.supported?` and its tests.
3. Deleted AutoCommit's duplicate regex walker and unused `path:` threading; it now consumes `SecretPatterns.scan` `(name, sha256)` evidence.
4. Deleted Hive's duplicate provider-error rescue and duplicate test; the component profile remains the single non-raising boundary.
5. Consolidated recovery-reset route construction into `Hive::PlanReview.recovery_reset_route` for orchestration and operator decisions.

The seven explicitly non-blocking simplification notes are left for a focused follow-up rather than expanding this already broad correctness round.

Verification:

- Expanded focused regressions: **812 runs, 3,672 assertions, 0 failures/errors**
- Component error-extractor boundary: **9 runs, 30 assertions, 0 failures/errors**
- Default Hive corpus (run without the environment-dependent component prerequisite): **13,857 runs, 193,809 assertions, 0 failures/errors, 12 skips**
- RuboCop: **26 changed Ruby files, no offenses**
- `git diff --check`: clean

The aggregate `bundle exec rake test` prerequisite still stops on this machine's installed OpenCode inventory not exposing `opencode/deepseek-v4-flash-free`; that is the previously observed local environment smoke, not a failure in the changed Hive corpus.

GitGuardian initially identified the dotted-password regression's literal fake fixture. Because it scans PR history, I replaced that fixture with runtime-joined fragments and amended only this fix-round commit under an exact force-with-lease; no secret-shaped literal remains in the branch. Fresh CI is now running on `5b1039a0c`.

ivankuznetsov:

Outcome: failure

Agent tail:

```
```

ivankuznetsov:
Finished the remaining simplification pass in d3ec4491f.

All seven previously deferred medium/low items are now addressed:

- Council has one pre-spawn round-budget boundary, including exact-limit resume behavior.
- Artifact Firewall enforces one aggregate max_entries bound across anchors, outputs, and writable roots; long review histories and growing execute receipts size that complete inventory.
- The dead brainstorm predicate wrapper is deleted; callers consume brainstorm_answer_state directly.
- Dispatch-request scans share one ENOENT-tolerant iterator.
- Plan-review transitions share one publish_transition helper.
- Defensive residue was removed or narrowed across attempt dispatch, plan decoding, planner emptiness, retry pacing, and recovery row lookup. The broad suite exposed the typed Struct edge, so value lookup now preserves generic indexables without restoring the blanket rescue.
- Low-level duplication is gone: shared Pi refusal fixture, table-driven completion precedence, one proof-sidecar reader, Set-backed waivers, OpenCode-only Claude kwargs removed, and the single-site failure-kind registry deleted.

The cleanup is net -66 lines and includes the required wiki/log fragment.

Verification:

- Default Hive corpus: 13,852 runs / 193,790 assertions / 1 failure / 0 errors / 12 skips on the first post-cleanup checkpoint. The sole failure was the recovery row-lookup contract identified above.
- Exact post-fix recovery checkpoint: 51 runs / 310 assertions, green.
- Exact post-fix daemon auto-retry integration: 4 runs / 13 assertions, green.
- Aggregate-limit regressions: attempt loss, plan review, and execute focused files total 74 runs / 331 assertions, green.
- RuboCop: 27 changed Ruby files, no offenses.
- git diff --check: clean.

Fresh exact-head CI is now running.

PR #1149, opened 2026-08-17
feat(patrol): kill silent validation runs at an idle-output deadline

ivankuznetsov:
## Why

Patrol validation commands are bounded only by the wall-clock cap (`timeout_sec.patrol`). On the dogfood box every fix-lane preflight from Aug 15–17 died at exit 124: the full-suite run needs ~27 min idle and 60+ min under daemon load, so the cap either kills healthy runs (too low) or, once raised, lets a genuinely wedged child hold the patrol cycle — and its locks — for hours (too high). One knob cannot be both the runaway backstop and the wedge detector.

## What

`Hive::Patrol::Validator` now enforces two independent deadlines per command run:

- `timeout_sec.patrol` — unchanged wall-clock backstop (total runtime/cost).
- `timeout_sec.patrol_idle` — new, optional: reader threads stamp a mutex-guarded monotonic pulse on every stdout/stderr chunk; the wait loop kills the process group as soon as the child has been silent for the idle window. A progressing suite prints as it runs, so a generous wall clock plus a short idle window fails hangs fast without punishing slow-but-live validation.

`CommandResult` records which deadline fired in a new `timeout_reason` field (`"wall_clock"` / `"idle_output"`, nil otherwise). Exit code stays 124 and `passed?` semantics are unchanged, so fixer fail-closed classification (`normal_nonzero_command_result?`, preflight refusal) is untouched. Both `Hive::Patrol::Fixer` and `Hive::RefactorPatrol::Fixer` plumb the knob. Unset or non-positive disables the idle deadline — the historical behavior — so existing configs see no change.

Deliberately not extended to agent children (stage/reviewer timeouts): an agent's stream is legitimately silent during long tool calls (e.g. the fix agent running the test suite inside one Bash call), and dead-runner detection already exists via attempts-v4 supervisor heartbeats. A follow-up applies the same pattern to the one other piped command runner, `review.ci.command`.

## Testing

- `ruby -Itest -Ilib test/unit/patrol/validator_test.rb` — 19 runs (6 new: idle kill fires before wall clock, stdout/stderr both defer the idle deadline, wall clock still bounds a chatty child, non-positive disables, reason reporting), 0 failures.
- `test/unit/patrol/fixer_test.rb` (67 runs) and `test/unit/refactor_patrol/fixer_test.rb` (63 runs) — 0 failures.
- Scoped coverage run: `lib/hive/patrol/validator.rb` uncovered lines `[]`.


ivankuznetsov (commented):

Intent is crisp (split wall-clock backstop from wedge detection) and execution is solid: fail-safe default (unset/non-positive = disabled), exit-124/passed? compatibility preserved, monotonic pulse correctly mutex-guarded, wait slice clamped away from busy-wait. Tests cover both deadlines and both output streams. Approve-level quality.

One simplification request:

- **`OutputPulse` is duplicated verbatim in #1150** (`lib/hive/stages/review/ci_fix.rb` gets its own copy with a slightly different API — `idle_for` vs `last`). After both PRs merge there are two near-identical thread-coordination classes in the tree. Whichever lands second should extract one shared helper (e.g. `Hive::Support::OutputPulse`) and use it in both call sites. Thread-safety primitives are exactly the kind of code you want to exist once.

Minor, non-blocking:

- `await_completion` recomputes `idle_deadline = pulse.last + @idle_timeout_sec` every iteration; computing it against a fixed start would be equivalent since `pulse.last` only moves forward — current form is fine, just noting it's more general than needed.
- `WAIT_SLICE_SEC = 0.05` floor means sub-50ms configured timeouts overshoot slightly; irrelevant in practice, no change needed.

PR #1150, opened 2026-08-17
feat(review): kill silent CI runs at an idle-output deadline

ivankuznetsov:
## Why

Companion to #1149 (patrol validation idle deadline). The review stage's per-attempt CI subprocess (`review.ci.command`) is bounded only by `timeout_sec.review_ci`, so one knob has to be both the runaway backstop and the wedge detector: sized for a slow-but-healthy suite it lets a hung `bin/ci` hold the review stage for the whole cap; sized to fail hangs fast it kills healthy runs.

## What

`run_ci_once` accepts an optional idle-output deadline plumbed from a new `timeout_sec.review_ci_idle` knob:

- The pipe reader stamps a mutex-guarded monotonic pulse per drained chunk — including capped-and-dropped bytes past `max_log_bytes`, which are still proof of life.
- The wait loop TERM/KILLs the process group as soon as combined stdout+stderr have been silent for the idle window, returning a `CommandError` with an `idle-output timeout` message through the existing `:error` path — same downstream handling as a wall-clock kill.
- Unset or non-positive disables the idle deadline; existing configs see no behavior change.

Semantics mirror `timeout_sec.patrol_idle` (#1149): generous wall clock as the cost/runaway backstop, short idle window as the wedge detector.

## Testing

- `ruby -Itest -Ilib test/unit/stages/review/ci_fix_test.rb` — 32 runs (3 new: idle kill fires long before the wall clock, progressing output defers the idle deadline through a green run, non-positive disables), 0 failures.
- Scoped coverage: every new line in `lib/hive/stages/review/ci_fix.rb` covered by this file (the single remaining uncovered line, 461, is pre-existing tmux spawn code covered by other suites).


ivankuznetsov (commented):

Right follow-up to #1149 and the right semantics (capped-and-dropped bytes still count as proof of life is a nice detail). Two simplification asks:

1. **Deduplicate `OutputPulse` with #1149.** This PR adds a second near-copy of the mutex-guarded monotonic pulse (`idle_for` here, `last` there) inside `CiFix`. Extract one shared class (e.g. `Hive::Support::OutputPulse` with both `touch`/`last`/`idle_for`) and use it from both `lib/hive/patrol/validator.rb` and this file. Hand-rolled thread coordination should exist exactly once in the codebase.

2. **Consider reusing the wait-loop shape.** #1149's `await_completion` computes a bounded sleep from both deadlines; this PR's loop checks wall clock, then idle, then `WNOHANG`. Both work, but two subtly different deadline loops for the same concept invites drift (e.g. one gets a fix the other doesn't). A shared "run child with wall + idle deadline" helper is probably overkill given different spawn/read patterns — but at minimum keep the pulse shared so the semantics can't diverge.

Non-blocking: the error message interpolates `idle_timeout_sec` (the raw config value) rather than the normalized float; harmless, just be aware a string config value would render oddly.

PR #1167, opened 2026-08-21
fix(ci): make CI faster to diagnose without weaker gates

ivankuznetsov:
## Summary

- Makes CI faster to diagnose without skipping or weakening any required proof. The six-way coverage lane still enforces exact 100% line coverage, and every existing functional gate still runs.
- Gives failed runs portable, shell-safe single-test repros, visible missing-evidence warnings, complete uncovered-line and shard-map artifacts, bounded jobs, and clear aggregation summaries.
- Keeps measurement separate from policy. The nightly lane accepts only a complete, fully loaded, positive-count seed matrix; its timing and flake outputs cannot alter required PR CI until a later reviewed change consumes real hosted evidence.

## Test plan

- [x] `bundle exec rake test` broad checkpoint: 13,275 runs, 273,957 assertions, 0 failures, 0 errors, 10 opt-in skips. One local OpenCode catalog route required the supported current-model override; the release workflow remains pinned to its deterministic CLI/catalog version.
- [x] Changed-area checkpoint: 66 runs, 453 assertions, 0 failures, 0 errors, 0 skips.
- [x] The exact formerly failing shard is green through the plain Ruby fallback: 1,754 runs, 8,368 assertions, 0 failures, 0 errors, 6 skips. The fallback now unbundles inherited Bundler activation before loading tests.
- [x] The failure-evidence integration fixture isolates `GITHUB_WORKSPACE`, so its temporary output root cannot change production evidence routing.
- [x] RuboCop: 18 touched Ruby files, no offenses.
- [x] Workflow YAML parses cleanly; workflow contracts cover required ownership, timeout headroom, evidence uploads, stable e2e artifact layout, fixed seed identity, and exact issue deduplication.
- [x] Independent adversarial review completed; all seven actionable findings were fixed and regression-tested.
- [x] Hosted CI on exact head `b4c80be40`: 33/33 checks completed successfully. GitHub reports `MERGEABLE`, `CLEAN`, and current base `a1194a8174e`.

## Notes for reviewer

The architecture has one deliberate boundary:

| Lane | Owns now | Does not own yet |
|---|---|---|
| Required PR CI | Functional tests, deterministic six-way byte partition, exact coverage, stable e2e proofs | Runtime partitioning, path-based proof skips, retries |
| Nightly measurement | Three fixed seeds, suite-load and test-count completeness, flake candidates, raw timings | Automatic gate or partition changes |
| Advisory performance | Absolute TUI latency signal | Merge eligibility |

The earlier docs-only bypass, empty quarantine layer, and timing partition with no reviewed input were removed. Future speed policy must cite complete hosted evidence from the nightly producer. Browser QA was skipped because this diff changes no product route or rendered page.


ivankuznetsov:

Outcome: failure

Agent tail:

```
```

ivankuznetsov (commented):

The goals (agent-legible CI, flake budget, fast local loop) are right and nothing weakens a gate. But this is five PRs wearing one trench coat, and several components ship before their inputs exist. Required first:

0. **CI is red** (`coverage` and `rake test` both failing on head). Diagnose that before anything below matters.

Simplification / sequencing asks:

1. **Split it.** At minimum: (a) failure evidence + timeouts, (b) shard partitioning, (c) docs-only short-circuit + path filters, (d) local fast loop, (e) nightly sweep. Each is independently revertible today only as one 27-file revert.

2. **Runtime shard partitioning is dead code on arrival.** `test/support/shard_timings.json` is deliberately not checked in, so every run takes the byte-size fallback until the nightly sweep produces timings — which itself needs a manual `workflow_dispatch` to exercise. Land the sweep + a real timings file first; add the LPT partitioner when it has input. Shipping both halves together means the primary new mechanism is untested in production until weeks after merge.

3. **Empty quarantine list ships machinery with zero users**, including the `NEVER_MATCH` sentinel hack to defeat minitest-retry's "empty list = retry everything" semantics. Simpler: don't call `HiveFlakeQuarantine.activate!` at all while `QUARANTINED_TESTS.empty?` — the sentinel disappears, and activation happens the day the first evidence-backed entry lands. The test-only override API can keep working via explicit activation.

4. **`HiveChangedCoverage` has three mapping layers** (override table → mirrored convention → basename glob fallback). The basename fallback especially is guesswork-as-mapping: a wrong basename match silently runs an unrelated test file and reports exact coverage for a source it may not exercise... except the enforcement step catches never-loaded sources, so the fallback's main effect is wasted runtime plus warning noise. Drop it initially: mirror convention + override table + loud fail-open warning covers the real cases.

5. **Micro:** `"#{ENV.fetch('GITHUB_STEP_SUMMARY')}"` in `failure_evidence.rb` — redundant interpolation around fetch.

6. **Gemfile.lock collision** with #1147 (brakeman bump, just merged) — rebase will need a lockfile regen.

The eventually-bounds relaxation of the TUI/e2e latency assertions is defensible given the dedicated perf gates, but each relaxed assertion should name its owning gate in a comment — two of them do, verify none lost their owner.

ivankuznetsov:

Outcome: failure

Agent tail:

```
```

ivankuznetsov:
## hive-babysitter: CI diagnosis — coverage gate, not a flake

`rake test (Ruby 3.4)` is a pure aggregator; it is red only because `coverage (Ruby 3.4)` is red. There is nothing to fix in it.

**The real failure:** `line coverage 96.05% (92702/96515) below minimum 100.00%` — 3813 uncovered lines across 108 files.

`main` at the base HEAD is green (`100.00% (97302/97302)`, 1730 process results), so this is PR-caused. Note this is the *first* run on the branch where all six shards passed and the merge gate actually executed — the previous run died in a shard, so the number was never observable before.

### What is not the problem

- **No test files were dropped.** The shard map union is 811 files = 815 glob − 4 gate tests. `HiveShardPartition` assigns every file, and with no `shard_timings.json` checked in it falls back to the historical byte split.
- **No test failed.** All six shards are green; shard 3 alone reports `1566 runs, 0 failures, 0 errors, 0 skips`.

### Where the coverage went

The loss is concentrated in one shard. Per-shard collected process results:

| shard | files | process results | artifact |
|---|---|---|---|
| 0 | 202 | 216 | 42 MB |
| 1 | 203 | 310 | 10 MB |
| 2 | 102 | 305 | 8.3 MB |
| **3** | **101** | **101** | **3.3 MB** |
| 4 | 101 | 457 | 12 MB |
| 5 | 102 | 230 | 1.2 MB |

Shard 3 produced exactly one dump per test file — i.e. **no forked/spawned child coverage flushes landed at all**, while every other shard averages 2–4.5 per file. Mapping the uncovered sources back to their owning tests, ~2787 of the 3813 lines belong to tests in shard 3.

### Why shard 3

The byte-size partition is input-unstable: this PR adds ~1000 lines of new test files, which reshuffles it. Recomputing the same partition against `main` versus this branch:

| test file | main | this PR |
|---|---|---|
| `refactor_patrol/action_runner_test.rb` | shard 3 | shard 3 |
| `commands/watch_test.rb` | shard 2 | **shard 3** |
| `modules/target_executor_test.rb` | shard 5 | **shard 3** |
| `daemon/dispatch_request_queue_test.rb` | shard 0 | **shard 3** |
| `commands/uninstall_test.rb` | shard 5 | **shard 3** |
| `tui/state_source_test.rb` | shard 0 | **shard 3** |

The reshuffle **concentrated the subprocess-heavy tests into shard 3**. These exercise their sources almost entirely through child processes — `action_runner_test.rb` alone drives `ActionRunner` via `Open3.capture3`, `fork`/`exit! 91`, and a deliberate `Process.kill("KILL", ...)`. Their coverage depends on the `HiveCoverageExitFlush` `exit!` hook in `test/hive_coverage_boot.rb` landing a sparse dump. Shard 3's profile matches that load: 127,589 assertions versus ~9–18k elsewhere, and 244s runtime.

This is the exact failure mode `coverage:prepare_shard` already warns about:

> redundant preloads can make coverage-aware `exit!` flushes exceed bounded subprocess deadlines under runner contention

Children killed by their own bounded deadlines skip the flush, so the tests still pass while the coverage silently vanishes.

### Suggested direction (author's call)

The PR already contains the right fix — the runtime-measured partition — but **`test/support/shard_timings.json` is not checked in**, so `HiveShardPartition.partition` takes the byte-size fallback and inherits its instability. Landing a timings file (or otherwise keeping subprocess-heavy tests from piling into one shard) is the first thing I would test. A byte-size partition will keep re-rolling this dice on every PR that adds test files.

### Two unrelated things I noticed

1. `test/test_helper.rb` does `Minitest.extensions << HiveFailureEvidence`. Minitest's `init_plugins` iterates `extensions` and calls `Minitest.plugin_#{name}_init`, so pushing a **module** yields `plugin_HiveFailureEvidence_init`, which `Minitest` does not respond to — the hook is silently skipped. The module also defines `self.minitest_plugin_init`, which is not the convention Minitest calls. The failure-evidence reporter is very likely never registering.
2. The shard checkout moved from `fetch-depth: 0` to `1`, with only the baseline-catalog shard unshallowing. That is unrelated to this failure (shard 3 *is* the unshallowing shard) but it is a behavior change for the other five shards worth a second look.

### Action taken

**None — branch left alone deliberately.** The PR is `MERGEABLE` and only 8 commits behind, so there is no conflict pressure, and merging `main` would reshuffle the byte-size partition again: that would move or mask this failure without fixing it, and would destroy the reproducibility of the evidence above. Better to land a partition fix first, then update.

<sub>Evidence: run `32489414668`, `coverage-evidence` artifact (`uncovered.json`, `shards.json`).</sub>

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your session limit · resets 12:30am (Europe/London)
```

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your session limit · resets 12:30am (Europe/London)
```

ivankuznetsov:
> CI is red (`coverage` and `rake test` both failing on head). Diagnose that before anything below matters.

Resolved on `47bd94f8d`. The loss was coverage-process state, not missing test assignment: sparse flushes now keep measurement active, and tests that repoint `HiveTestCoverage` restore every process-global path before shard shutdown. The repaired broad checkpoint completed with 13,275 runs and 273,957 assertions; the new exact-head hosted run is in progress.

> Split it. … several components ship before their inputs exist.

Fixed differently on the already-published PR. The branch now has six reversible commits grouped by proof integrity, evidence, nightly admission, advisory latency, documentation, and the adversarial-review repair. The unproven mechanisms were deleted instead of split into dormant PRs:

- runtime/LPT partitioning is gone until a reviewed hosted timing artifact exists;
- docs-only proof skips and installer path filters are gone;
- the empty quarantine and retry dependency are gone;
- changed coverage accepts only mirrored tests or explicit overrides;
- every relaxed functional wait names its separate performance owner;
- the branch is based on current `main`, so the old lockfile collision is absent;
- the redundant `GITHUB_STEP_SUMMARY` interpolation is removed.

The earlier Minitest registration concern does not hold with the locked Minitest 6 API: `init_plugins` accepts a module that responds to `minitest_plugin_init`. A real failing subprocess now proves the reporter emits a portable JSON repro and that the command works from the checkout root. The shallow-checkout change was also removed; every coverage shard retains full history.

PR #1170, opened 2026-08-21
fix(opencode): make prepared startup reliable

ivankuznetsov:
## What changed

- accept OpenCode exact configured prepared-plugin identity when it byte-matches Hive pinned Compound Engineering package
- keep arbitrary configured identities and filesystem shadows conflicting
- give only the verbose model-inventory probe a 30-second deadline; all other local capability checks stay at 10 seconds
- treat exact operator-declared custom models and variants as authoritative and skip the verbose inventory probe entirely for those routes
- preserve inventory validation and fail-closed behavior for undeclared routes
- recognize controller-bound brainstorm answers as intentional input so an automatic recovery can continue without asking the same questions again
- classify OpenCode's nonzero upstream idle-timeout/504 event as a transient timeout rather than a generic CLI defect
- let a current completed state artifact beat an empty terminal assistant message, while incomplete artifacts still fail
- document prepared-overlay, custom-route, retry, timeout, and completion-evidence contracts

## Why

The parallel Ox Alpha Webmail dogfood exposed six integration failures:

1. setup-agents correctly configured the pinned plugin, but Doctor called the prepared configured package foreign until OpenCode materialized a stable install directory.
2. A cold hermetic verbose model inventory emitted about 393 KB and exceeded the generic ten-second probe deadline.
3. The fetch-disabled inventory later omitted the exact custom Ox Alpha route even though that configured route had already succeeded.
4. After the controller answered brainstorm questions, the recovery path still classified the typed artifact as missing answers and would repeat the interview.
5. A real 43-minute plan run preserved 2.7 KB of useful progress but eventually received OpenRouter's typed 504 `Upstream idle timeout exceeded`; OpenCode's strict parser reduced it to `cli_failure`, hiding its transient provider nature from Hive recovery.
6. Its daemon-owned retry exited zero with an empty terminal assistant message after file-tool work. The strict parser correctly called the transcript malformed, but Hive also needs to preserve the stronger case where the current attempt has already written its terminal file artifact.

The final route rule is simpler and cheaper: an exact configured model plus variant does not need a catalog scan. Undeclared routes still require inventory evidence and fail closed when absent.

The observed 504 JSON event is now a versioned fixture. Its normalized outcome is `timed_out`; Hive records agent status and marker reason `timeout`, preserving the partial artifact for scheduler-owned retry without changing the configured model or falling back.

For marker-owned stages, a zero-exit malformed transcript now yields to a current terminal artifact. Previous terminal markers are rotated before launch, so the current marker is attempt-owned. A partial, markerless, or nonzero-exit run still fails; the dogfood retry that exposed this shape correctly remained failed because its plan was incomplete.

## Verification

- OpenCode preparation: 20 runs, 152 assertions, 0 failures
- component gemspec contract: 4 runs, 48 assertions, 0 failures
- managed-skill inspector: 38 runs, 156 assertions, 0 failures
- setup-agents and Doctor focused tests: 13 runs, 66 assertions, 0 failures
- OpenCode result parser + event contract: 21 runs, 404 assertions, 0 failures
- Hive OpenCode lifecycle: 15 runs, 105 assertions, 0 failures
- RuboCop on every changed Ruby file: clean
- three fresh hermetic Ox Alpha preparations: ready, high variant present, auth configured, cleanup complete
- real daemon-owned Webmail recovery: prepared and spawned `opencode run` with `openrouter/stealth/ox-alpha` high; both failed attempts retained their partial plan for autonomous retry
- previous head GitHub CI: 32 of 32 checks green at `247f59398f9bc31d13df76e417e3ba35a73defed`; checks are rerunning for the new dogfood fixes

No model call is made by the regression or preparation probes. This remains the dedicated OpenCode harness-fix PR from the parallel Ox Alpha Webmail trial; it contains no Pi fix.


ivankuznetsov:
Dogfood follow-up at `3c6149389`:

A daemon-owned retry of the completed 118.7 KB Ox Alpha plan failed before model launch because `opencode --version` exceeded the generic 10-second capture bound under sustained host I/O. This was a healthy Bun cold start, not an unavailable binary.

The component now carries a per-profile `version_check_timeout_sec`: OpenCode gets a finite 30-second bound; every other built-in/custom profile retains 10 seconds. Focused verification: component probe 10 runs / 38 assertions; Hive profile 75 / 285; RuboCop clean.

Live combined deployment `f8928d76d` loaded the vendored profile with `30.0`; the next autonomous recovery passed preflight and launched `openrouter/stealth/ox-alpha` through OpenCode (attempt `5b170e0c-1ebe-4f08-9881-f8cefe49476e`).

ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your session limit · resets 12:30am (Europe/London)
```

ivankuznetsov:
Dogfood follow-up at `858fe3f5e`:

The OpenCode copy now completed its 118.7 KB plan and Hive accepted the current `<!-- COMPLETE -->` artifact even though earlier OpenCode exports had ended empty/truncated. Mandatory review then exposed three additional integration gaps, all fixed on this PR:

- `ce-doc-review` omitted OpenCode from the committed capability manifest even though the same pinned Compound Engineering plugin supplies it. Added the mapping plus a manifest regression.
- top-level `models.plan_review*` overrides changed the reviewer actually launched but were absent from review identity. The fingerprint now includes only the three plan-review routing keys, so corrected routes mint a fresh review while unrelated execute-model changes do not.
- the skill probe used the stock ambient OpenCode profile instead of the project-prepared profile, falsely reporting the configured plugin missing. `HiveRunner` now probes with the same project config used for launch.

Focused verification: adapter 29 runs / 158 assertions, policy 10 / 30, manifest 10 / 212, and RuboCop clean. Combined detached deployment `b7acba168` is live. The daemon autonomously minted a fresh review after the routing change and launched the adversarial leg as Codex `gpt-5.6-sol` xhigh; the next route identity is queued to exercise the corrected OpenCode primary probe without a manual advance or waiver.

ivankuznetsov:
Dogfood follow-up for `ca6f20468` (runtime code deployed in combined build `72ca75c4d`):

The daemon retried the same Webmail.sh execute row autonomously after its five-minute recovery cadence; no `workflow.retry`, approval, or waiver was issued. Attempt `f618fd03-6bea-4e9e-b882-0b99d6e9266e` launched the installed OpenCode 1.18.18 binary successfully.

Live `/proc` evidence at launch:

- stdin: owner-private `/tmp/hive-agent-prompt-*`, mode `0600`, 145,764 bytes
- complete argv: 226 bytes, with no prompt payload
- command: `opencode run --auto --model openrouter/stealth/ox-alpha --dir <worktree> --format json`
- process remained CPU-active instead of returning `You must provide a message or a command`

The missing boundary was Hive’s dedicated OpenCode supervisor: the component correctly prepared `stdin_data`, but this specialized spawn path never bound it to `Process.spawn`. The regression fake now reads and records all 150 KB, so merely removing the prompt from argv can no longer pass. Focused verification is 131 runs / 725 assertions plus RuboCop. The component `Unreleased` section remains empty; the exact gemspec contract passes, and no release/version change is included.

ivankuznetsov:
Fresh CI exposed one stale integration fake after OpenCode moved implementation prompts from argv to stdin: the skill-dependent driver still searched `ARGV.last` for `/ce-plan`, so the real transport worked while the test asserted the retired shape. Fixed at `d405c5250`: both OpenCode integration paths now consume stdin, require a non-empty stream, require the prompt to be absent from argv, and preserve the skill invocation assertion against stdin. Verified across seeds 1, 57979, and 65535 (2 runs / 42 assertions each) plus RuboCop clean. This is test/wiki-only; the live `72ca75c4d` runtime remains unchanged and both product workers were left running.

ivankuznetsov:
Dogfood follow-up at `1c902fd7e` (combined local deployment `b077a7089`):

The first real 145 KB Ox Alpha execute turn exposed two more OpenCode integration boundaries:

- File-only workspace write let Ox author the Rails tree but not install gems, run tests, or create the clean commit Hive requires. Scoped OpenCode permissions now accept only qualified `Bash(...)` patterns; bare `Bash` stays rejected, and `Bash(*)` is documented as an explicit full-shell grant with the authority of the Hive OS user.
- OpenCode can exit 0 after tool work while exporting an empty terminal assistant message. Hive still records that malformed protocol result, but execute now accepts stronger repository evidence only when the expected branch is clean and contains a new descendant commit. Dirty, nonzero, wrong-branch, no-commit, and non-descendant cases remain failures.

The run also proved the old `task.md` custody exclusion was stale: agents no longer own `AGENT_WORKING`; the controller owns all execute markers. `task.md` is now a protected anchor, so shell writes are detected/restored. OpenCode file-tool roots also omit the controller task directory.

Focused verification: execute 38 runs / 149 assertions; spawn 58 / 256; OpenCode integration 2 / 44; runtime preparation 22 / 162; lifecycle 16 / 113; Agent 104 / 506; component boundary 11 / 106; release contract 6 / 52; RuboCop clean.

Live autonomous retry `d1d88476-1f26-4213-9dea-4dae490a8b49` is running through installed OpenCode 1.18.18. `/proc` confirms the selected policy grants `bash: {"*":"allow"}`, the task-state directory is absent from external/edit roots, the 145 KB prompt remains off argv, and no manual workflow retry, advance, or waiver was used.

ivankuznetsov:
## Babysitter: CI red was the coverage shard-manifest race, not coverage debt

`coverage (Ruby 3.4)` failed with:

```
coverage shard manifest error: coverage shard manifest does not match its process results
```

`rake test (Ruby 3.4)` is only the aggregation gate (`HIVE_COVERAGE_RESULT: failure`) — derivative, no tests run there.

### Diagnosis

All 6 coverage shards were green. The merge job aborted during manifest
verification, **before** reaching the 100% line gate, so the failure log
contained no coverage number at all.

Downloaded the shard artifacts from run 32547210766 and compared each
`manifest.json` against the marshals actually on disk:

| shard | manifest | on disk |
|---|---|---|
| 0 | 921 | 921 |
| 1 | 491 | 491 |
| 2 | 163 | 163 |
| 3 | 36 | 36 |
| 4 | 49 | 49 |
| **5** | **68** | **69** |

The extra file is `34408-1080.marshal`, with an mtime identical to
`manifest.json` — a subprocess that dumped its coverage *after*
`coverage:collect` had already globbed and frozen the file list into the
manifest. `shard_result_paths` (`test/support/coverage.rb:315`) re-globs in the
merge job, finds one file more than the manifest lists, and raises.

### The real coverage number

Recovered locally from the same artifacts (rewriting the CI path prefix in each
marshal, omitting `HIVE_COVERAGE_EXPECTED_SHARDS` so manifest verification is
skipped):

```
Process results: 1729
Gate:           passed
Lines:          100.00% (97414/97414)
Branches:        87.49% (35758/40871)
```

**The exact-100% line gate passes on this head (`f23e3c2a7`).** There is no
coverage debt on this PR — it is purely the race.

### Action taken

Re-ran the **whole** workflow rather than `--failed`. That distinction matters:
the shard artifact contains both `manifest.json` and the marshals, so re-running
only the merge job would re-download the same stale manifest alongside the same
69 marshals and fail identically. Fresh shard jobs write a fresh manifest.

No code changes and no rebase — the branch is already 0 behind `origin/main`
(base `ee8ae4d8b`), and nothing in `test/support/coverage.rb` should be loosened
to accommodate this (its strictness is asserted by `test/unit/coverage_test.rb`).

### Note on a related, pre-existing bug (not this PR's)

The underlying `exit!`-flush machinery in `test/hive_coverage_boot.rb` also
causes a separate *measurement-loss* failure mode. The fixes for it
(`Coverage.result(stop: false, clear: false)`, dropping the `@dumped` latch, and
the `with_coverage_config` sandbox helper) are **not** on `main` yet —
`test/support/coverage.rb:128` still calls bare `Coverage.result`. That is
out of scope here, but it is why this class of flake keeps recurring.

ivankuznetsov:
Live Webmail dogfood found and fixed one OpenCode-only lifecycle defect at `acbb6bdb3`.

The real Ox Alpha execute attempt launched Puma with `setsid`; it moved to a new session/process group, reparented to the user service manager, and remained listening after the shell command. The existing OpenCode teardown owned only the direct `opencode run` process group, so it could not address that server once ancestry was lost.

OpenCode invocations now receive a random custody ID in their prepared child environment. After the CLI exits, Hive inventories only exact inherited IDs and performs bounded TERM/KILL rounds against PID + start-time identities. This reaches reparented/setsid descendants without pattern matching or touching another invocation. A survivor overrides transcript success as `process_cleanup_failed`. Linux uses exact NUL-delimited procfs environment entries; the portable fallback uses the current user's bounded `ps xeww` inventory.

Proof:
- real-process custody regression forks a child, calls `setsid`, exits its parent, then proves cleanup kills it while a second run's child stays alive;
- OpenCode lifecycle regression proves the custody ID crosses the prepared overlay and the escaped child is gone before a successful result returns;
- cleanup-failure precedence is pinned;
- 104 focused Hive tests / 506 assertions green;
- 4 touched Ruby files RuboCop-clean.

The currently running dogfood worker predates this commit, so I am leaving its live server alone until that attempt finishes. The next daemon-owned attempt will exercise this cleanup without manual workflow intervention.

PR #1175, opened 2026-08-22
perf(patrol): bound admission inventory scan traversal

ivankuznetsov:
## Summary

Patrol admission polling now opens its managed anchor once and keeps directory traversal constant as the record inventory grows. Missing stores remain read-only, record scans remain lock-free, and all existing ordering, bounds, canonical validation, and fail-closed behavior stay intact.

This is the first of the separate daemon-scan optimization PRs. Project configuration caching, Attempts index caching, admission-only imported findings, and changed-task incremental scans remain follow-up changes.

## Design

- One non-creating read session owns the verified root descriptor for the full scan.
- The complete bounded filename inventory is validated before any record is read.
- Caller-validated record names reuse one opened parent directory; per-record reads and full validation remain linear.
- No cache, migration, migration watcher, schema, scheduler-policy, capacity, or task-materialization behavior is added.

Planning selected a one-session design without caching or buffering. Independent review showed that using the inventory lock as that session boundary would create absent stores and convoy readers with writers, so this PR uses a smaller non-creating read-session boundary instead.

## Performance evidence

The dogfood admission store contained 54 records. Each comparison ran three complete `pending` scans.

| Measurement | `origin/main` CPU seconds | Branch CPU seconds | Change |
|---|---:|---:|---:|
| Comparison 1 | 14.312698077 | 13.385584423 | -6.5% |
| Comparison 2 | 11.590105895 | 11.470292061 | -1.0% |
| Aggregate | 25.902803972 | 24.855876484 | -4.0% |

The repeated `AdmissionStore#fetch -> ManagedDirectory#open_root!` stack also disappeared from the daemon profile. Whole-tick CPU was noisy, so this PR makes no total-tick improvement claim. Direct regression tests prove one anchor open per inventory scan and constant parent-directory traversal as record count rises.

## Validation

- Managed directory: 32 runs, 138 assertions, 0 failures
- Admission store: 21 runs, 99 assertions, 0 failures
- Admission scheduler: 17 runs, 97 assertions, 0 failures
- RuboCop: 4 files, no offenses
- Broad checkpoint: 112 runs, 1,151 assertions, 0 failures, 1 error, 1 skip

The broad checkpoint's only error is the unavailable local OpenCode route in `opencode_offline_smoke_test.rb`. The same error reproduces on exact `origin/main`, so it is environment-specific and not introduced by this branch.

Browser testing: skipped because this storage-only change affects no browser route.

### DIFF
--- a/lib/hive/managed_directory.rb
+++ b/lib/hive/managed_directory.rb
@@ -149,6 +149,27 @@ module Hive
       unsafe!
     end
 
+    # Reuses one verified, non-creating root session across nested reads.
+    def with_read_session(missing: false)
+      with_session(create_root: false) do
+        begin
+          yield
+        rescue SystemCallError, IOError, ArgumentError, TypeError => error
+          raise YieldFailure.new(error)
+        end
+      end
+    rescue YieldFailure => failure
+      raise failure.error
+    rescue MissingEntry
+      return nil if missing
+
+      unsafe!
+    rescue Hive::ConfigError
+      raise
+    rescue SystemCallError, IOError, ArgumentError, TypeError
+      unsafe!
+    end
+
     def read(relative, max_bytes:, missing: false)
       snapshot = read_with_metadata(
         relative, max_bytes: max_bytes, missing: missing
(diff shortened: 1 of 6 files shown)


ivankuznetsov:
## Residual Review Findings

No actionable review findings remain.

- Fixed: late `ENOENT` during a missing-aware batch read now fails closed instead of returning a truncated inventory.
- Fixed: admission reads no longer create an absent store.
- Fixed: admission reads no longer take the inventory lock or convoy with `reserve!`; one non-creating read session preserves the one-anchor traversal.
- Proceeded and flagged: plan KTD1 selected the existing shared inventory lock and no new storage API. Independent review proved that mechanism changed read semantics. The implementation therefore uses the narrower `ManagedDirectory#with_read_session` boundary while preserving requirements R1-R6 and the user-selected no-cache, no-buffer, no-migration architecture.

Validation after fixes: 70 focused runs, 334 assertions, 0 failures; RuboCop clean. The broad checkpoint has one local OpenCode inventory error that reproduces on exact `origin/main`.

Issue #1197, opened 2026-08-24
Explore visual-regression evidence without committing PNG baselines


ivankuznetsov:
## Problem

A real coding workflow exposed a gap between review-time visual testing, Hive's clean-worktree invariant, and the `7-artifacts` outcome-evidence stage.

During review of [TopGreenDeals PR #402](https://github.com/ivankuznetsov/topgreendeals.co.uk/pull/402), a review-fix agent ran:

```sh
RABATA_BUCKET=test HEADLESS=true REGENERATE_BASELINES=1 mise exec ruby@3.4.5 node@26.2.0 -- bundle exec bin/rails test test/visual
```

This rewrote all 119 tracked `test/visual_baselines/*.png` files, including pages unrelated to the Solar Calculator task. The regenerated images were mixed with 24 substantive source, test, and wiki paths.

Hive's auto-commit safety scan then rejected several staged PNG blobs larger than 1 MiB and correctly avoided committing them, but the task was stranded at:

```text
6-review
REVIEW_ERROR phase=fix reason=fix_status_check_failed
```

The size guard prevented a bad commit, but it detected the problem late. Hive has no current bridge for treating baseline/current/diff images produced during review as visual evidence rather than source changes.

## Desired product behavior

- PNG baseline, current-render, and visual-diff bytes produced by agents must not be added to the product PR commit.
- Review agents must not silently bless environment-wide rendering drift by regenerating every baseline to make a suite green.
- Relevant visual output should be retained as task evidence and displayed in the Artifacts surface with its implementation identity, instead of remaining as dirty worktree residue.
- Hive should keep the useful code/test/documentation changes recoverable when visual output is rejected or redirected.
- Unrelated regenerated screenshots should be identified and excluded rather than accepted as evidence.
- The flow must remain useful when Screenote is disconnected; optional Screenote support should be fail-soft.

## Existing pieces

Hive already has most of the adjacent capabilities:

- `7-artifacts` stores generation-bound, immutable screenshot/video evidence and independently reviews it.
- Hive Web renders retained evidence.
- `hive connect screenote` provides optional OAuth/MCP screenshot uploads and Screenote links.
- clean-exit and auto-commit guards already prevent unsafe binary residue from being committed.

The missing part is the lifecycle connecting review-time visual-regression output to those artifact capabilities.

## Larger direction — intentionally not designed yet

[...]

PR #1200, opened 2026-08-24
fix(worktree): restore native residue recovery

ivankuznetsov:
## Summary

Hive now resolves task-owned worktrees from the project root and preserves a complete, safe recovery identity whenever CleanExit rejects known residue. Review pre-fix failures use the same native `hive worktree` repair contract as stage-exit and Finalize failures instead of entering an unrecoverable review-specific error state.

## Why this failed

- The worktree command passed an already-resolved root back through `canonical_root`, which appended the configured suffix twice.
- Review collapsed pre-fix CleanExit failures into `REVIEW_ERROR reason=fix_status_check_failed`, but native worktree repair only admits canonical residue markers.
- Failure diagnostics could omit exact paths, redact them into unusable placeholders, or grow beyond the marker reader's 1 MiB tail window.

## Recovery contract

- Small non-sensitive path sets remain inline with a count and SHA-256 identity.
- Large or secret-shaped sets use an owner-private task-local sidecar. The durable marker stays bounded and redacted.
- Signing, configuration, and Git failures retain the exact known-dirty path identity even when no diagnostic path exists.
- Default discard verifies the marker-owned identity. It preserves unrelated meaningful residue and supports literal POSIX backslashes and whitespace filename bytes.
- Git details and full paths are redacted before display truncation.

Generated PNG baselines remain outside commits and still fail the existing oversized-blob safety gate. This PR repairs native recovery; it does not integrate visual evidence into Hive.

## Validation

- Focused unit suite: 184 runs, 863 assertions, 0 failures.
- Targeted Review and Finalize integrations: 4 runs, 33 assertions, 0 failures.
- RuboCop: 16 changed Ruby files, 0 offenses.
- Live read-only TopGreenDeals probe: the task resolves to `/home/asterio/Dev/topgreendeals.co.uk.worktrees/fix-and-comprehensively-verify-the-260814-91b0` with all 143 residue entries visible; no `.worktrees.worktrees` path remains.
- No PNG files are present in the commit.

The broad local checkpoint remains environment-limited: the packaged agent-runtime prerequisite stops on the local OpenCode provider route, and the main suite's isolated remaining failure is a child process that resolves the broken local mise `bundle` shim. CI owns those merge gates.

## Related

Related: #1197

---


ivankuznetsov:

Outcome: failure

Agent tail:

```
You've hit your weekly limit · resets Aug 27, 1pm (Europe/London)
```

PR #1295, opened 2026-08-27
fix(patrol): park secret-blocked publications

ivankuznetsov:
When Patrol Fix publication detects a secret, it now records a durable, sanitized block for that exact generation. Status shows “Publication blocked by secret policy” with the operator as owner. Re-running publication returns the same block without remote calls or cleanup; the daemon cannot retry it automatically.

The block uses the existing publication scan, including the full commit range, so a secret introduced and later removed still prevents publication. Receipts contain safe field names and evidence hashes, never matched secrets.

This PR only adds parking. It has no publication-specific rework action, reopen authorization, or routing to earlier stages. The operator corrects the source/history problem and starts a new task through the normal workflow; the old generation remains parked. The unrelated E2E timing-budget increase has been removed.

Validation:

- Based on current main `9b6ddf64`.
- Focused publication, receipt, projection, action and operational-status checks: 170 tests, 994 assertions passed.
- Broader Patrol workflow and action checks: 327 tests passed.
- Receipt-store regression for older scanner policy versions: 11 tests, 53 assertions passed.
- Root RuboCop: 1,869 files, no offenses.
- Final code review completed on `f96b9d77e9` with no unresolved findings.
- All 34 hosted checks passed on exact head `f96b9d77e938af2c4e8ff3f5010153c90eaed1a5`, including exhaustive coverage, required test gates, and the timing advisory with its original budget.


ivankuznetsov:

Outcome: timeout

Agent tail:

```
The normal five-minute settle window has elapsed with no feedback or CI failures, but the outstanding mergeability race remains. I’ll make one final bounded poll before the run budget expires.
```

PR #1308, opened 2026-08-27
refactor(status): share envelope compatibility handling

ivankuznetsov:
## Summary

`Hive::Bot::StatusWatcher` now delegates strict `hive-status` envelope validation and version-skew messaging to `Hive::StatusEnvelope`. The daemon keeps main's in-process status producer, where subprocess schema skew is not a boundary.

## Validation

- `bundle exec ruby -Itest test/unit/bot/status_watcher_test.rb` — 16 runs, 93 assertions.
- `bundle exec ruby -Itest test/unit/daemon/status_consumer_test.rb` — 29 runs, 185 assertions.
- `bundle exec rubocop lib/hive/bot/status_watcher.rb lib/hive/status_envelope.rb test/unit/bot/status_watcher_test.rb` — no offenses.

### DIFF
--- a/lib/hive/bot/status_watcher.rb
+++ b/lib/hive/bot/status_watcher.rb
@@ -1,6 +1,7 @@
 require "json"
 require "open3"
 require "time"
+require "hive/status_envelope"
 
 module Hive
   module Bot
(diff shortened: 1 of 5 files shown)


ivankuznetsov:
Closing after the open-PR sweep against current main (`34821da3d5`).

The intent is clear: give long-running status consumers one envelope/version-skew policy. The extraction preserves the bot's behavior, but the original sharing opportunity is no longer present: `Daemon::StatusConsumer` now uses an in-process producer and deliberately enforces an exact schema version. `Bot::StatusWatcher` is the only consumer of the proposed `StatusEnvelope`.

Main already performs the strict shape/ok checks, tolerates additive newer versions, and gives the older-binary/restart guidance. Keeping these methods with their sole consumer is simpler than introducing another class with a configurable restart target and error-message callback. The added top-level-array case is already handled by main's Hash guard.

Five review passes found no actionable defect. Focused bot/daemon tests passed on the PR (45 tests, 278 assertions) and on identical current-main status source/test files (45 tests, 274 assertions). This closure avoids an unnecessary extraction; it does not remove any current compatibility behavior. We can revisit a shared owner if another subprocess consumer actually needs the same policy.

PR #1349, opened 2026-08-30
Harden live status connection ownership and disconnect cleanup

ivankuznetsov:
Late subscription setup and reconnect callbacks can outlive the status element that started them, leaving cleanup tied to a shared consumer. This PR gives each mounted source an owner and each setup/retry attempt a dedicated consumer. Retired callbacks cannot affect a successor, unconfirmed releases wait for a bounded disposition, and cleanup continues after an error while retaining the first failure.

The sweep confirmed the problem remains on current main and verified the earlier review comments against the code. It refreshed the branch and combined three identical setup exception paths into one, removing 20 lines without adding an abstraction. Ten review lenses and independent validation found no remaining blocking defect; a test-file-size concern was rejected because no repository contract prohibits the explicit lifecycle matrix.

Validation: the original focused browser matrix passed 27 tests / 292 assertions. On current main, live-status and board browser checks passed 49 tests / 483 assertions; channel/status/broadcaster checks passed 38 tests / 190 assertions. Changed Ruby test files pass lint. No new dependencies, installed-runtime rollout, or many-tab capacity proof is claimed.

Dedicated ownership costs one transport per live source and a fresh handshake on cross-URL navigation. An unconfirmed predecessor may overlap its successor within the bounded cleanup window. Draft #1405 addresses separate server-side StatusChannel ownership.

Hive task #43036: give-each-live-status-connection-an-explic-coding-ed05c6a0.

Head 83e1492175 passed all required checks. The only failing advisory measured the plan scenario 23 ms over its 11-second budget; functional checks passed. Merged as 319d292adecf37216767cc31212af5c76a2a41ab with an identical tree. Post-merge CI 34409168164 and install smoke 34409168160 both passed.


ivankuznetsov:

## High

- [ ] Preserve the same-URL catch-up latch across application retries (`web/app/javascript/status_stream_source.js:270,288`): clearing it on setup, registration, or rejection failure makes the next retry send `refresh_attempted: false`, reopening the refresh loop that the prior behavior kept bounded.
- [ ] Capture sockets even when `connection.open()` throws (`web/app/javascript/status_stream_source.js:182`): Action Cable assigns a CONNECTING socket before later setup can fail, but the wrapper records it only after a successful return, so retirement cannot close the orphan transport.
- [ ] Bound repeated supersession to one retiring predecessor (`web/app/javascript/status_stream_source.js:536`): every unconfirmed old owner independently retains its transport for five seconds, so two rapid attribute changes can leave two predecessors plus the current transport and violate R7/AE15's maximum of two.
- [ ] Retry Turbo stream-source registration along with the Cable attempt (`web/app/javascript/status_stream_source.js:509`): a thrown `Turbo.connectStreamSource()` enters `retry_wait`, but the timer starts only a new Cable attempt, which can become connected without a working Turbo message listener.
- [ ] Make the AE2 lease-count assertion exercise the real broadcaster lifecycle (`web/test/system/status_stream_source_test.rb:25`): queue-only replacements prevent the subscriber count from ever rising, so the final zero check cannot prove the required server lease was acquired and released.
- [ ] Add real server-observable coverage for the AE3 timeout path (`web/test/system/status_stream_source_test.rb:204`): the wholly fake consumer, subscription, and socket prove client cleanup order only and cannot show that a never-confirmed real StatusChannel lease or transport is gone after the bound.

## Medium

## Nit

ivankuznetsov:

# Code review — `give-each-live-status-connection-an-explic-coding-ed05c6a0` (pass 1)

Scope: `git diff origin/main..HEAD` (base `5e2b40ef83`, head `cb16acc9f7`) — 9 files,
`web/app/javascript/status_stream_source.js` (277 -> 639 lines), a new 16-case
`web/test/system/status_stream_source_test.rb`, lifecycle tests moved out of
`kanban_board_test.rb`, and 6 wiki files.

Reviewers: correctness, adversarial, reliability, frontend-races, testing,
maintainability, project-standards, learnings, plus an orchestrator fast pass.
API shapes were verified against the installed `actioncable 8.1.3.1` /
`turbo-rails 2.0.23` assets rather than assumed.

Overall this is a strong implementation. The `CleanupCollector` first-error
transaction, the owner/attempt identity fences, `mayOpen` interception of the real
`ConnectionMonitor` reopen paths, and the exact-once close accounting all hold up
under adversarial interleaving. AE1-AE14 are genuinely asserted (event-order arrays,
sentinel-identity checks, per-phase close counters), and the two plan-mandated
pre-confirmation tests were retained and strengthened. The findings below are
ordering/coverage gaps, not a broken design.

## High

- [ ] `status_stream_source.js:154-158` — the stale-attempt branch calls `retireStaleAttempt(attempt)`, which retires in the default **release-then-close** order, directly contradicting the comment above it ("close it before local release") and KTD3's invariant: the subscription here was just created and is unconfirmed, so its `unsubscribe` can be sent before the dedicated transport closes — the exact overtake race this PR exists to eliminate. Flagged independently by correctness and maintainability, both with the quoted line; fix is `retireAttempt(attempt, cleanup, { transportFirst: true })` plus `warnCleanup`, matching `pendingReleaseTimedOut`. (Reachability is narrow — nothing between `subscriptions.create` and the guard yields in production — but the code and its own stated rationale disagree, and no test drives the path.)
- [ ] `web/test/system/status_stream_source_test.rb` (whole suite) — real-server catch-up verification on the **retry-recovery** path was dropped rather than replaced: the removed `kanban_board_test.rb` cases "a status source retries after asynchronous consumer setup rejects" and "...removes a partial Action Cable registration before retry" proved via `wait_for_status_catch_up` that a genuine `StatusChannel#catch_up` round-trip completes after recovery; that helper is now deleted (0 hits repo-wide) and the new suite contains **zero** `catch_up` observations — its successors at `:1036` and `:1356` use JS fakes only. `StatusChannel#catch_up` has a live `return if data["refresh_attempted"] == true` short-circuit that this coverage guarded. Violates the Verification Contract's "moved tests have explicit replacements rather than deletion" and R2's catch-up/retry preservation.

## Medium

- [ ] `status_stream_source.js:275-291, 312-327, 508-513` — when `Turbo.connectStreamSource(this)` throws, `connectedCallback` routes to `failOwnerSetup`, whose 5s retry goes straight to `startAttempt()` and **never re-registers the stream source**: the retry re-subscribes, sets the `connected` attribute, and dispatches `MessageEvent`s that Turbo is no longer listening for, so the board silently stops updating while reporting healthy — the precise "mounted board silently stops updating" failure the plan's risk table calls out. Flagged by correctness and adversarial. Fix: call an idempotent `Turbo.connectStreamSource(this.source)` at the top of `startAttempt()`. No test stubs `Turbo.connectStreamSource`/`disconnectStreamSource` anywhere in `web/test`, so neither this branch nor `failOwnerSetup`'s no-attempt path has any regression guard.
- [ ] `status_stream_source.js:531-552` — `attributeChangedCallback` swaps `this.statusWarningQueue` to a local `deferredWarnings` array and restores it at `:552` with no `try`/`finally`. `successor.failOwnerSetup(...)` (`:548`) and `new StatusStreamOwner(this)` (`:546`) are the only statements in the method not wrapped in a `CleanupCollector`; if either escapes, the element is left pointing at an abandoned array and **every future** `reportStatusFailure` is silently swallowed for the life of the page. A restoring `finally` satisfies KTD8 (it never throws, so it cannot replace the old error).
- [ ] `web/test/system/status_stream_source_test.rb:1622` — AE15/R7 require instrumented transport counts across "connect, supersession, **navigation**, retry, and detach"; the transport test does one `visit repos_path` up front and then measures connect/supersession/detach/two-sources entirely in-page, and no other test counts consumers across a Turbo navigation. Retry is covered (`:1472` asserts `consumers => 2` + `consumersDiffer`; `:1356` asserts `factoryCalls => 2`, `failedDisconnects => 1`), so navigation is the one unasserted leg — and it is a distinct code path (Turbo permanent-element move / cross-URL element replacement ordering), not a rerun of detach-then-connect.

## Nit

- [ ] `status_stream_source.js:83, 95, 395` — `attempt.id` (`Symbol("status-stream-attempt")`) and `attempt.transportClosed` are write-only: zero readers in the file, the test suite, or anywhere in the repo. Attempt identity is by object reference (`this.currentAttempt === attempt`), so `id` is inert; if `transportClosed` was meant as a double-close guard, that guard is actually supplied by the `socketCanClose` ready-state probes.
- [ ] `status_stream_source.js:110-115` — `setupAttempt`'s `createConsumer()` catch calls `this.failAttempt(attempt, error)` directly, unlike its three sibling error sites (`:120-124`, `:161-165`) and `startAttempt`'s outer catch (`:100-104`), which all branch `isCurrentAttempt ? failAttempt : retireStaleAttempt`. A rejection that settles after the owner is detached or superseded therefore returns silently at `failAttempt`'s own guard with no warning and no retirement. Harmless today (no resources exist yet at that point) but the asymmetry is latent, and no test rejects `createConsumer()` for an already-stale attempt.
- [ ] `web/test/system/status_stream_source_test.rb:152` — the bounded-cleanup test asserts the transport-first ordering and stops; no test delivers a subscription callback (`connected()`/`rejected()`) **after** the pending-release timeout has already fired. The path is inert by construction (disposition nulled, attempt retired), but AE3's "no later socket or server lease survives" has no post-timeout assertion.
- [ ] `wiki/architecture.md`, `wiki/commands/web.md`, `wiki/gaps.md` — the transport delta is documented as concurrent allocation ("one Cable transport per simultaneously live status source"), but not the new **per-navigation handshake cost**: the source is `data-turbo-permanent`, yet a cross-URL Turbo move still fires `disconnectedCallback`/`connectedCallback`, so every navigation now opens a fresh WebSocket instead of reusing turbo-rails' cached consumer, and `connected` + catch-up wait on a full WS + Action Cable handshake. R8/U3 asks the docs to state the delta; this facet is missing.
- [ ] `docs/solutions/` — this change introduces a reusable pattern (ordered best-effort teardown with first-error precedence, identity-fenced retired callbacks, per-attempt dedicated resource instead of a shared cache) that currently lives only in `wiki/`. `docs/solutions/architecture-patterns/` already carries a sibling "warn-and-continue teardown" entry; consider a matching entry so the pattern is discoverable outside the status subsystem.

## Checked and cleared (no finding)

Recorded so a later pass does not re-litigate these:

- **Exact-once close (KTD7/AE8)** holds: `webSocket.close()` moves `readyState` to `CLOSING` synchronously, so both `socketCanClose` probes after `consumer.disconnect()` return false on the OPEN path; the CONNECTING path correctly falls through to the raw close because `Connection#close` only acts when `isActive()`.
- **R5 fencing is complete.** `ConnectionMonitor`'s 200ms visibility timer and `reopen()`'s `setTimeout(this.open, reopenDelay)` both resolve the wrapper as an own instance property at fire time, so `mayOpen` blocks them; all four subscription callbacks early-return on stale/retired attempts.
- **`disconnect()` is not re-enterable** (state guard set before any work), `retireAttempt` is idempotent, two retry timers cannot coexist, and an orphaned timer cannot start an attempt on a superseded owner (`isRetryable` checks `source.statusOwner === this`).
- **The pending-release custodian cannot leak past 5s**, and a confirmation arriving after the timeout is inert.
- **Wiki protocol is compliant**: `wiki/log.d/<ts>-<slug>.md` fragment added, compiled `wiki/log.md` untouched, `wiki/index.md` correctly unchanged (no new page), `wiki/gaps.md` updated with residual uncertainty. `bin/rubocop` config is bare `rubocop-rails-omakase` with no violated rule; `git diff --check` is clean.
- **`failOwnerSetup`'s `if (this.currentAttempt)` branch** is unreachable from both current call sites, but it is a correct defensive guard on a general entry point — deleting it would make a future caller silently skip attempt retirement. Deliberately not flagged.
- **Retry backoff / persistent-failure WebSocket churn** was raised but dropped: KTD2 mandates a fresh consumer per retry and the 5s interval is explicitly out of scope.

## Verdict

**Ready with fixes.** No P0/P1 correctness defect survived scrutiny; the two High
items are an ordering/comment contradiction on a narrow path and a real coverage
regression against the plan's own Verification Contract.

Run artifacts: `/tmp/compound-engineering/ce-code-review/20260830-134404-13be5636/`

ivankuznetsov:

## High

- [ ] Retry-time Turbo registration failures stop recovery (`web/app/javascript/status_stream_source.js:330`): The retry timer clears its handle before calling unguarded `startAttempt()`, so a repeated synchronous registration error escapes and leaves a mounted owner in `retry_wait` with no attempt or future retry, violating R6/AE12.
- [ ] Physical disconnects do not retain pending predecessors (`web/app/javascript/status_stream_source.js:543`): Unlike attribute supersession, `disconnectedCallback()` drops an unconfirmed owner after creating pending-release custody, so rapid remove/reinsert cycles can accumulate more than one retiring transport and violate R7/AE15's two-transport bound.

## Medium

## Nit

ivankuznetsov:
### Reviewer: claude-opus-5-ce-code-review - Pass 02

# Code review — `give-each-live-status-connection-an-explic-coding-ed05c6a0` (pass 2)

Scope: `git diff origin/main..HEAD` (merge-base `5e2b40ef83`, head `d0392ef793`) — 11 files;
`web/app/javascript/status_stream_source.js` (277 -> 679 lines), the 22-case
`web/test/system/status_stream_source_test.rb`, lifecycle tests moved out of
`kanban_board_test.rb`, 6 wiki files, and one `docs/solutions` entry.

Reviewers: correctness, adversarial, testing, frontend-races/project-standards, plus
an orchestrator pass. Behavior was verified against the actual bundled Action Cable
8.1.3 / turbo-rails 2.0.23 assets, not assumed. Pass-1 fixes were re-verified as
landed and are not re-reported.

The two High items below were each independently flagged by three reviewers
(including the Codex pass-2 reviewer) and confirmed against the code. Both are
narrow guard omissions, not design problems; the ownership model, first-error
`CleanupCollector`, `mayOpen` fencing, and exact-once close accounting all hold up.

## High

- [ ] `status_stream_source.js:543` — `disconnectedCallback` reads and clears `statusRetiringOwner` (`:539-541`) but never assigns it, so an owner that enters bounded pending-release custody via DOM detach has no element-level handle and `forcePendingRelease` is unreachable for it; only `attributeChangedCallback:566` registers a predecessor. Because navigation carries identical `channel`/`signed-stream-name` on every page (`application_helper.rb:30-35`), a cross-URL Turbo move of the `data-turbo-permanent` source never reaches `attributeChangedCallback` at all — it is pure disconnect+connect, i.e. exactly the unregistered path. Clicking through the status project nav faster than Cable confirms leaves k+1 live dedicated WebSockets, server connections, and `StatusBroadcaster` leases for one element, against the R7/AE15 bound of two — a bound `wiki/architecture.md:428-430` explicitly claims holds for navigation. Fix: mirror `:566` after `owner.disconnect()`, add a regression test driving >=3 detach/attach cycles with an unconfirmed subscription and asserting peak live consumers <= 2, and keep the architecture claim honest. 

- [ ] `status_stream_source.js:330` — the retry timer calls `startAttempt()` with no try/catch, and `startAttempt` synchronously calls `this.source.connectTurboStreamSource()` (`:82`, added by the pass-1 fix). `connectedCallback` treats that exact call as fallible and routes it to `failOwnerSetup` (`:528-532`); the retry entry point has no equivalent guard. On a throw the callback has already run `this.retryTimer = null` (`:327`), leaving a mounted owner with `state === "retry_wait"`, `currentAttempt === null`, `retryTimer === null`, and no inbound event that can re-arm `scheduleRetry` — a silent terminal state the plan forbids ("no stopped state"; every mounted failure reaches `retry_wait` and recovers), plus one uncaught page error. The existing test at `status_stream_source_test.rb:923` throws only on `connectCalls === 1`, so it exercises the guarded `connectedCallback` path and never the timer. Fix: `try { this.startAttempt() } catch (error) { this.failOwnerSetup(error) }` and extend that test to throw persistently. 

## Medium

- [ ] `web/test/system/status_stream_source_test.rb` (whole suite) — the moved-out case `a pending status subscription is cancelled across disconnect and reconnect` (`kanban_board_test.rb:754` at base) was deleted with no successor, violating the Verification Contract's "moved tests have explicit replacements rather than deletion". It was the only test performing `host.remove()` immediately followed by `document.body.appendChild(host)` on the same node while a consumer-setup promise was still outstanding, and its assertions (`created: 1, unsubscribed: 0`) documented coalescing behavior the owner/attempt model deliberately changes. Nothing in the new suite covers that interleaving: `:1645` removes without re-appending, `:2090` supersedes via `setAttribute`, and `:2241` waits for `#status-stream-source[connected]` before each visit so the predecessor is always confirmed. This is also the coverage hole that hides the High finding above. 

## Nit

- [ ] `status_stream_source.js:138` — `received` is the only Action Cable mixin callback not wrapped in `runAttemptCallback`; `connected`/`disconnected`/`rejected` (`:139-150`) all route a throw into `failAttempt`/`retireStaleAttempt`. A throw from `dispatchMessageEvent` -> `rememberCatchUpRefresh` (`template.innerHTML = data`, `:648`) would escape into actioncable's `Subscriptions#notify` and the raw `webSocket.onmessage` unfenced and unreported. Verified non-triggerable today (`content_security_policy.rb` is fully commented out, so no Trusted Types policy exists, and `dispatchEvent` swallows listener exceptions) and unchanged from the base version, so this is a consistency gap rather than a live defect — worth fencing defensively. 

- [ ] `status_stream_source.js:572` — `if (successorSetup.failed && !this.statusOwner)` is the one `CleanupCollector` consumer with a conditional disposition: if `connectedCallback()` throws *after* assigning `this.statusOwner` (`:527`), the captured error is neither rethrown nor warned. All throw sites past `:527` were traced and none is currently reachable (`owner.connect()` is caught, and `failOwnerSetup` routes every external call through `cleanup.run`), but the branch silently swallows rather than warning. 

- [ ] `status_stream_source.js:277, 303` — `failAttempt` and `retireStaleAttempt` retire release-then-close, while every other pre-confirmation teardown edge passes `transportFirst: true` (`:69`, `:162`, `:361`, `:387`) per the rule the file states at `:63-65`. `failAttempt` is the one edge that can hold a subscription handle with `confirmed === false`, so it can emit `unsubscribe` for a subscribe the server has not processed. Verified latent, not live: every path that reaches `failAttempt` unconfirmed already has a closing socket, so `connection.send` returns false and nothing goes on the wire. Suggested: `{ transportFirst: !attempt.confirmed }`. 

- [ ] `web/test/system/status_stream_source_test.rb:700, 1824` — the client-side failure triggers (`createConsumer()` rejection, `connection.open()` throwing after partial registration) are proven only against hand-rolled JS fakes whose `perform`/`create` always succeed; the deleted `a status source retries after asynchronous consumer setup rejects` and `...removes a partial Action Cable registration before retry` drove a real page and used `with_status_catch_up_observer` for a genuine `StatusChannel#catch_up` round-trip. The pass-1 fix restored that round-trip only for the server-side triggers (`:2331`, `:2387`). The `:700` throwing-open case also stops at `retry_wait` without asserting eventual recovery to `connected`. 

- [ ] `docs/solutions/architecture-patterns/identity-fenced-owned-resource-teardown-2026-08-30.md:5-7` — `module: hive-web` and `component: web` match no sibling in `docs/solutions/`; existing entries use Ruby-module-style `module:` values (`Hive::Tui`, `Hive::Daemon`, `Hive::TaskAction`) or bare `hive`/`hive/bot`, and no prior entry uses `component: web`. No schema gate enforces this, so it is convention drift only. 

## Checked and cleared (no finding)

Recorded so a later pass does not re-litigate these:

- **All 8 pass-1 fixes verified landed**: `transportFirst` on the stale-attempt branch (`:162`), Turbo re-registration in `startAttempt` (`:82`), `finally`-restored `statusWarningQueue` (`:578-579`), socket capture in the open/reopen `finally` blocks (`:189-191`, `:202-204`), dead `attempt.id`/`transportClosed` removed, consistent stale/current branching in the `createConsumer` catch (`:111-118`), real-server AE2/AE3 coverage (`:6`, `:275`), `statusRetiringOwner` bounding attribute supersession (`:563`, `:566`), navigation transport counting (`:2241`), and the restored real `StatusChannel#catch_up` round-trip (`:2361-2368`).
- **Codex pass-1 catch-up latch finding is resolved**: `clearCatchUpRefresh()` was removed from `failAttempt`/`failOwnerSetup`, so `source.catchUpRefresh` survives an application retry — pinned by `recoveredRefreshAttempts => [true]` at `:1450`. The remaining `this.catchUpAttempt = null` (`:276`) is a no-op on every reachable path (`catchUpAttempt` is only set after a successful `perform`, and the sole post-confirmation failure route clears it in `subscriptionDisconnected` anyway, matching base and the intentional "a real Cable disconnect releases the catch-up attempt" contract).
- **`mayOpen` fencing is complete** against bundled actioncable 8.1.3: `Connection#open` is an own bound property the wrapper replaces, `reopen`'s deferred `setTimeout(this.open, ...)` resolves to the wrapper, and both `reconnectIfStale` and `visibilityDidChange` go through `connection.reopen`. `cable.createConsumer` is uncached, so dropping the `setConsumer(undefined)` poison-cache workaround is correct.
- **Exact-once close holds (KTD7/AE8)**: `Connection#close` gates on `isOpen()` (OPEN only), so the raw `socket.close?.()` fallback is genuinely required for CONNECTING, and after `consumer.disconnect()` closes an OPEN socket the synchronous CLOSING transition makes both `socketCanClose` re-checks false.
- **`Turbo.connectStreamSource` is idempotent** (Set-guarded by `streamSourceIsConnected`), so the per-attempt call in `startAttempt` cannot double-register.
- **No shared state across simultaneous elements** — `statusOwner`, `statusRetiringOwner`, `catchUpRefresh`, `statusWarningQueue` are all per-element; steady-state transport count is unchanged and only the churn path regresses.
- **Callback re-entry during `disconnect()`/`retireAttempt` is inert** (`retired`/`pendingReleaseDisposition` cleared before every external call), `disconnect()` is not re-enterable, two retry timers cannot coexist, and pending-release custody cannot outlive its 5s bound.
- **Wiki protocol compliant**: two `wiki/log.d/<ts>-<slug>.md` fragments added, compiled `wiki/log.md` untouched, `wiki/index.md` correctly unchanged (no new page), `gaps.md` honestly retains the many-tab/live-Docker gaps and states the per-navigation handshake delta. The earlier fragment's "16 cases" is a correct append-only historical record superseded by the later "22 cases" — not a defect. `git diff --check` is clean.

## Verdict

**Ready with fixes.** No P0 survived scrutiny. Both High items are one-line guard
omissions with clear fixes and matching regression tests; the Medium is a
Verification Contract compliance gap that the first High fix should close anyway.
Fix order: `disconnectedCallback` registration (High 1) -> retry-timer guard
(High 2) -> deleted-test successor (Medium), then the Nits.

PR #1350, opened 2026-08-30
Centralize TUI new-idea project resolution

ivankuznetsov:
A TUI refresh can change the project at a numeric position while a new idea is being composed. This PR makes Snapshot the shared project-resolution authority for picker, prompt, and plain/rich submission. Numeric scope is consumed once; later checks use the pinned exact name. Invalid or ambiguous selection dispatches nothing, preserves the draft and staged attachments, and requires an explicit valid selection before retry.

The sweep confirmed the problem remains on main and refreshed this branch. It removed the unused project object from resolution results and the test-only name argument to numeric entry. A new regression checks that an unselected picker with eight projects displays the first six without silently highlighting one.

Validation on current main: 601 focused tests / 2,125 assertions; attachment PTY smoke 2 tests / 47 assertions; real CLI `tui_new_idea_editing` passed. The explicit-base changed-coverage run passed 639 tests / 2,363 assertions and exact line coverage for all six changed Ruby sources. All 15 changed Ruby files pass lint. Nine review lenses and independent validation found no remaining blocking issue; repair review confirmed the simplification preserves callers and staging.

The later live registry name lookup remains non-atomic; this PR prevents stale snapshot-position substitution. The focused coverage overrides preserve current main's newer selection machinery and its loaded-source/uncovered-line checks.

Head 6e2153b2b3 passed all 33 checks and merged as cbcaefb760ea0ffb6bd9a16174a9af0be3db42f4 with an identical tree. Post-merge CI 34410658257 and install smoke 34410658250 passed. The post-merge incident timing advisory measured the plan scenario at 11.454 seconds against an 11-second budget; all functional and required checks passed; no installed-runtime rollout is included.

Hive task: introduce-one-internal-new-idea-project-re-coding-a87964bf.


ivankuznetsov:

## High

- [ ] Preserve `disappeared` for a pinned name when the latest snapshot is empty (`lib/hive/tui/snapshot.rb:388`): the unconditional `no_projects` return hides sole-project removal, so submission gives generic setup advice instead of AE2's required disappearance feedback.
- [ ] Move unhealthy recovery classification into `Snapshot` and harden the boundary regression (`lib/hive/tui/update.rb:1083`): `Update` still branches on raw `missing_project_path` values carried through `unhealthy_errors`, violating R2/U1/AE8's sole-authority contract while the current structural test remains green.

## Medium

## Nit

ivankuznetsov:

# ce-code-review — pass 1

Scope: `git diff origin/main..HEAD` (base `5e2b40ef83`, head `5be91ce4bf`, 18 files).
Plan: implementation-ready `ce-unified-plan/v1` "Centralize New Idea Project Resolution" (explicit, supplied by the runner).

Verified locally: `snapshot_test.rb` (37 runs), `update_test.rb` (147), `bubble_model_test.rb` (318),
`new_idea_project_picker_test.rb` (14), `new_idea_prompt_test.rb` (33), `new_idea_prompt`/boundary (3),
`model_test.rb` (24) — all green. `git diff --check` clean; `wiki/log.md` untouched; exactly one
`wiki/log.d/20260830T101613Z-centralize-new-idea-project-resolution.md` fragment.

Requirements completeness (R1-R9, U1-U4): all met in the diff. Snapshot owns admission + typed
resolution (U1), picker/prompt/entry migrated with explicit no-highlight invalidation (U2), one shared
name-only preflight ahead of both plain and rich dispatch with a real task-producing integration retry
test (U3), and wiki authority/residual-race documentation (U4). No plan-required item is missing.

## High

_None._ No correctness defect reachable through the real message flow was found, and no
plan Goal or Requirements-Trace item is unimplemented or undeferred.

## Medium

- [ ] `lib/hive/tui/bubble_model.rb:2734` — the retained-blocked-resolution branch (and the `mode == :new_idea` guard at `:2404`) is unreachable, and it returns a resolution built from an *older* snapshot instead of revalidating: `NEW_IDEA_SUBMITTED` is only emitted from `:new_idea` mode (`key_map.rb:453`), and `:new_idea` is only ever entered with a non-empty pinned name (`update.rb` `apply_open_new_idea_prompt` / `apply_new_idea_project_selected`), so `new_idea_project_name` can never be empty at submit — if a future change makes it reachable, R3 ("revalidate against the latest installed snapshot") is silently violated by returning the stale value. Either drive it through a real transition, or mark the branch defensive-only and assert it revalidates rather than replays.
- [ ] `test/unit/tui/bubble_model_test.rb:1358` and `:1368` — the `invalid_scope` and `selection_required` rows of the AE7 "exhaustive submit failures" matrix are asserted only against hand-built models (`new_idea_project_resolution:` / `name: nil` injected directly), so 2 of the 6 states in both the plain and rich matrices exercise dead code and provide no regression protection for any operator-reachable path; the other four (`disappeared`, `unhealthy`, `ambiguous`, `no_projects`) are genuinely reachable and correctly covered.
- [ ] `test/support/changed_coverage.rb:29,38` — out-of-plan edit to shared test infrastructure that also introduces a staleness trap: the new `SOURCE_TEST_OVERRIDES` entries *replace* convention-based mapping with frozen literal lists for `lib/hive/tui/bubble_model.rb` and `lib/hive/tui/update.rb`, so any future test file added for those sources is silently excluded from `rake coverage:changed` and the local gate will report a coverage hole that the full suite does not have. The plan's U1-U4 file lists do not include this file and the Diff-hygiene gate forbids unrelated edits.
- [ ] `test/integration/tui_new_idea_attachments_smoke_test.rb:14,40,83,87` — out-of-plan harness changes (`RbConfig.ruby` instead of `"ruby"`, `GEM_PATH` injection into the PTY child, `Errno::EIO`/`EOFError` early-return in `read_until`) are unrelated to new-idea project resolution; U3 specifies "verify the existing smoke test", not modify it. They look like correct environment fixes, but they ride in unlabelled and should be gated explicitly rather than absorbed into this refactor's diff.

## Nit

- [ ] `lib/hive/tui/update.rb:1057-1061` — a blocked `:unhealthy`/`:disappeared` flash can advise an impossible recovery: when the pinned project was the registry's only entry, the flash reads "choose another project or re-init it" while the picker it lands on simultaneously renders "No healthy projects available". AE4 explicitly requires truthful recovery copy for the ambiguity case; the same honesty gap exists here, and `blocked_new_idea_submission` already has the admission in hand to pick the emptier-set wording.
- [ ] `lib/hive/tui/snapshot.rb:502` — `NEW_IDEA_AMBIGUOUS_NAME_LIMIT` is reused to bound `unhealthy_errors`; a name-specific constant governing an unrelated error list will read as a copy-paste bug at the next edit. Rename to a neutral `NEW_IDEA_DETAIL_LIMIT` or add a second constant.
- [ ] `lib/hive/tui/snapshot.rb:489` — blank/nil project names produce misleading authority output: two nameless projects classify as `:ambiguous` with an empty `ambiguous_names`, yielding `"duplicate project name  — disambiguate the registry…"` (verified by direct invocation), and one healthy nameless project classifies as `:unhealthy`, yielding "no healthy projects — re-init broken projects…". Low probability (status JSON normally emits a name), but the fail-closed classifier should name the blank-identity case rather than borrow another state's copy.
- [ ] `lib/hive/tui/snapshot.rb:224` — `new_idea_admission` is built eagerly in `initialize`, so every derived snapshot from `scope_to_project_index` / `filter_by_slug` / `visible_projection` (per render frame while scoped or filtered) carries an admission computed over the *projected* project subset. No current consumer reads it from a projection — every call site uses `model.snapshot` — but nothing in the code says so, and a future `visible_snapshot(model).new_idea_admission` would silently return a filtered, wrong authority. Add a comment pinning the invariant, or compute it lazily on the unfiltered snapshot only.

ivankuznetsov:

## High
- [ ] Cover the unhealthy recovery guard at `lib/hive/tui/update.rb:1104`: The required explicit-base `coverage:changed` run exits nonzero on this uncovered branch, leaving R8/U4 and the Definition of Done unmet.

## Medium
- [ ] Preserve malformed-identity feedback for blank scoped rows at `lib/hive/tui/snapshot.rb:401`: With a blank-name row plus a healthy peer, numeric entry resolves as generic `selection_required`, so the user is not told that the selected registry identity is invalid as the plan requires.

## Nit

ivankuznetsov:
### Reviewer: claude-opus-5-ce-code-review - Pass 02

# ce-code-review — pass 2

Scope: `git diff origin/main..HEAD` (base `5e2b40ef83`, head `199aa6d13a`, 18 files, 1830 changed Ruby lines).
Plan: implementation-ready `ce-unified-plan/v1` "Centralize New Idea Project Resolution" (explicit, supplied by the runner).
Reviewers: correctness, adversarial, testing, maintainability, project-standards. Pass-1 findings are all applied
at `199aa6d13a`; the `test/support/changed_coverage.rb` override finding is suppressed (triaged no-fix) and is not
re-raised here.

Verified locally by direct run: `snapshot_test.rb` (40), `update_test.rb` (147), `bubble_model_test.rb` (320),
`model_test.rb` (24), boundary test (3), `new_idea_project_picker_test.rb` (15), `new_idea_prompt_test.rb` (33),
`tui_new_idea_attachments_test.rb` (3) — all green. `git diff --check` clean; `git diff -- wiki/log.md` empty;
both `wiki/log.d` fragments present exactly once. `HIVE_COVERAGE_BASE=5e2b40ef83 bundle exec rake coverage:changed`
**fails** — see the High finding.

Requirements completeness: R1–R7 and R9 are met; U1–U4 are all present in the diff. **R8 is not met** (the plan's
own Changed-coverage gate is red). No plan Goal or Requirements-Trace item is unimplemented-and-undeferred.

## High

- [ ] `lib/hive/tui/update.rb:1104` — the repo's exact-100% line-coverage gate over `lib/**/*.rb` is red on this branch: `HIVE_COVERAGE_BASE=5e2b40ef830af1cd49cac0f3ebd58da6f40e4c1c bundle exec rake coverage:changed` aborts with `lib/hive/tui/update.rb: uncovered lines 1104`, the inner `raise ArgumentError, "unknown new-idea recovery kind: ..."` in `new_idea_admission_flash`. `Snapshot#build_new_idea_admission` always sets `recovery` to `:prune_missing` or `:repair_projects` for `:unhealthy`, so no snapshot-built value reaches it, and `update_test.rb`'s `Struct.new(:state).new(:unknown)` probe only reaches the outer `else` at `:1109`. The plan's Verification Contract makes both the explicit-base changed-coverage gate and the CI-equivalent global exact-100% gate closing conditions, and the Definition of Done requires both — so CI's coverage job fails as-is. Cheapest fix mirroring the existing test: `assert_raises(ArgumentError) { Hive::Tui::Update.new_idea_admission_flash(Hive::Tui::Snapshot::NewIdeaAdmission.new(state: :unhealthy, projects: [])) }` (`recovery:` defaults to `nil` and the Data validator permits it). Alternatively require a recovery kind for `:unhealthy` in `NewIdeaAdmission#initialize` and drop the inner `else`. 

## Medium

- [ ] `lib/hive/tui/snapshot.rb:403` — a blank/nil-named registry row selected by numeric scope produces generic feedback instead of the malformed-identity feedback the wiki claims. Verified by direct invocation: with `projects = [{name: nil}, {name: "healthy"}]`, `resolve_new_idea_entry(scope: 1)` returns `:selection_required` (the blank name falls through `resolve_new_idea_project`'s `candidate.empty?` guard), and because `admission.projects` is non-empty the flash is `"choose a project for the new idea"` — the operator is never told the scoped registry entry itself is invalid. The whole-registry case is handled correctly (`:invalid_identity` copy), so this is the residual half of the pass-1 blank-identity fix, and `wiki/log.d/20260830T133251Z-harden-new-idea-admission-review.md`'s claim that "Blank registry identities fail closed with distinct repair feedback" is only true when no healthy peer exists. Classify a blank scoped row as `:invalid_scope` (or a dedicated invalid-identity state) so R4's state-specific feedback holds. 
- [ ] `lib/hive/tui/snapshot.rb:403` and `:383` — the same Snapshot object can now contradict itself: `filter_by_slug` / `scope_to_project_index` forward the registry-wide `new_idea_admission` verbatim (the pass-1 fix), but `resolve_new_idea_project` and `resolve_new_idea_entry` still scan the *projected* `@projects`. Verified by direct invocation: on `snapshot.scope_to_project_index(1)`, `new_idea_admission.projects` lists `["alpha","beta"]` while `resolve_new_idea_project(name: "beta")` returns `:disappeared` and `resolve_new_idea_entry(scope: 2)` returns `:invalid_scope`. No production consumer passes a derived snapshot today (every call site reads `model.snapshot`), but A1/KD1's "single authority" invariant is exactly what a future `visible_snapshot(model)` consumer would trust — and the resolvers would silently answer from an incomplete subset. Either carry the registry projects alongside the admission, or make the resolvers refuse on a derived snapshot. 
- [ ] `lib/hive/tui/snapshot.rb:513` — a *healthy* project with a duplicate name silently vanishes from the picker whenever any uniquely-named healthy project also exists, with no explanation anywhere. Verified: for `[{dup}, {dup}, {solo}]`, admission is `state: :available`, `projects: ["solo"]`, `ambiguous_names: ["dup"]` — the cause is computed and then never read, because both `Views::NewIdeaProjectPicker.empty_message` and `Update.new_idea_admission_flash` consult `ambiguous_names` only on the empty-admission path. Before this diff both duplicate rows were offered. Excluding duplicates is correct per KTD2, but AE4's requirement that ambiguity "names the duplicate and recovery action" is only honoured when ambiguity is the *sole* cause; surface `ambiguous_names` as an advisory row/flash whenever it is non-empty. 
- [ ] `lib/hive/tui/model.rb:32` — `new_idea_project_resolution` is write-only production state: four `lib/` sites write it (`update.rb:589`, `bubble_model.rb:2752`, plus two clears) and no `lib/` code ever reads it; only tests assert on it. It is also never refreshed on `SnapshotArrived`, so it goes stale the moment a poll lands. This matters because the flash TTL is 5s (`Model::DEFAULT_FLASH_TTL_SECONDS`): five seconds after a blocked submit the operator sits in the picker holding a preserved draft and staged images with nothing on screen explaining why, and `Esc`/`q` then discards it silently — and the retained resolution is precisely the state a picker could render to fix that. Either wire a consumer (persistent reason row + "draft kept" affordance) or delete the field and its write sites; leaving it write-only makes the bubble_model tests that assert on it prove nothing observable. 
- [ ] `test/unit/tui/bubble_model_test.rb:1359` and `test/unit/tui/views/new_idea_project_picker_test.rb:112` — two more rows of the "exhaustive" matrices are asserted only against model/authority shapes no transition can produce, the same defect pass-1 fixed for the `invalid_scope` and `selection_required` rows. `new_idea_submission_cases`'s `:no_projects` entry uses `mode: :new_idea` with `name: nil`; both `mode: :new_idea,` assignment sites in `update.rb` pin a non-empty name (admission excludes blank names), so an operator with an empty registry always reaches `:disappeared` instead — meaning AE7's `no_projects` submit row carries no regression protection on either the plain or rich path. Likewise `test_empty_message_has_a_safe_fallback_for_nonempty_authority_states` hand-builds `NewIdeaAdmission.new(state: :available, projects: [])`, which `build_new_idea_admission` can never emit. Also leftover: `entry[:blocked]` is read at `:1542` and `:1573` but no case supplies that key. Either drive these through real transitions or label them defensive-only, as `test_defensive_missing_pin_revalidates_latest_snapshot_instead_of_replaying_entry_failure` already does. 

## Nit

- [ ] `lib/hive/tui/snapshot.rb:383` — `resolve_new_idea_entry`'s `name:` keyword and its `:invalid_scope` conflict branch have no production caller: the sole call site (`update.rb:577`) passes `scope:` alone, and `apply_open_new_idea_prompt` clears the pin into `fresh` before resolving, so a simultaneous name+scope conflict cannot arise. Only `snapshot_test.rb:775` exercises it. KTD1 describes the classification, but as shipped it is test-only API surface — either narrow the signature or note the branch as reserved. 
- [ ] `lib/hive/tui/update.rb:224` — `apply_new_idea_project_snapshot` dereferences `new_model.snapshot.new_idea_admission` with no nil guard, while the code it replaced (`visible = visible_snapshot(new_model); return new_model if visible.nil?`) degraded gracefully and every sibling branch of `apply_snapshot_arrived` still guards. Unreachable in production — `app.rb:146` emits `SnapshotArrived` only behind `if current && ...` — but a directly-delivered `SnapshotArrived.new(snapshot: nil)` in `:new_idea_project` mode now raises `NoMethodError` where the base returned the model unchanged. Restore the guard or assert the invariant. 
- [ ] `lib/hive/tui/update.rb:615` — pressing `k` from an explicit no-highlight state jumps to the *last* admissible project (and scrolls the window to the bottom for registries larger than `MAX_VISIBLE_PROJECTS`), while `j` goes to the first. Defensible as wrap semantics, but the nil-cursor hint reads a flat `"j/k select"`, so the asymmetry is undiscoverable. Either mirror `j` and return `0`, or say `"j first / k last"` in the no-highlight hint. 
- [ ] `lib/hive/tui/views/new_idea_project_picker.rb:60,71` — `empty_message` is a second `case admission.state` in parallel with `Update.new_idea_admission_flash`, and its `else` silently degrades to `"No selectable projects available"`. The distinct copy is plan-mandated (KTD5 keeps the existing no-healthy wording), so the duplication itself is fine — the trap is that adding a seventh admission state would render generic picker copy with no test failure, because `test_empty_message_has_a_safe_fallback_for_nonempty_authority_states` locks the fallback in as expected behaviour. Make the `else` raise like its `Update` counterparts, so a new state fails loudly in exactly one place. 
- [ ] `wiki/commands/tui.md:93` — the submit-preflight paragraph lists `invalid_scope` among the states that "all fail closed with state-specific feedback and zero dispatch" at submit time, but `Snapshot#resolve_new_idea_project` can never return `:invalid_scope` (it accepts no scope), and `submit_new_idea` returns early unless `mode == :new_idea`, which a blocked entry never is. `invalid_scope` is entry-only; per CLAUDE.md ("Never hallucinate. Ground everything in code") the sentence should say so. 
- [ ] `lib/hive/tui/snapshot.rb:518` — `NEW_IDEA_AMBIGUOUS_NAME_LIMIT = 3` truncation is never exercised; every ambiguity test uses a single duplicate group, so `first(3)` could regress to `first(1)` or be dropped entirely without failing a test. U1 asks for "bounded detail suitable for feedback" — add one case with four distinct duplicate-name groups.

Issue #10, opened 2026-04-27
Fix cursor-vs-snapshot ordering divergence in Render::Grid (P1)


ivankuznetsov:
## Summary

`Hive::Tui::Render::Grid#draw_grouped_rows` indexes rows in **visual order** (sorted by `Hive::Commands::Status::ACTION_LABEL_ORDER`); `Hive::Tui::Snapshot#row_at` indexes them in **JSON-iteration order** (stage-iteration from `Status#collect_rows`). When a project's rows span multiple action_labels, the cursor `>` highlights one row but `at_cursor(snapshot)` returns a different one — Enter or a verb keystroke acts on the wrong task silently.

## Reproduction

A project with at least two rows in different action_labels (e.g., one `agent_running` + one `review_findings`):
1. Press `j`/`k` to position the cursor.
2. Note which row is visually highlighted.
3. Press Enter or a verb key.
4. Observe: the action runs on a different row than the highlighted one.

The unit tests use single-action_label projects so the divergence never surfaces in CI.

## Identified by

ce-code-review run \`20260427-110800-0e135bb6\`, finding F1 (correctness reviewer, validated). See \`.context/compound-engineering/ce-code-review/20260427-110800-0e135bb6/correctness.json\`.

## Why deferred (not auto-fixed)

The fix touches the ordering contract between Render and Snapshot. Two viable approaches:

- **Sort once at render** — `Render::Grid` builds a flat row list in visual order, passes the index back to GridState for cursor translation. Cleanest, but Snapshot stays JSON-ordered.
- **Teach Snapshot/GridState the visual order** — `Snapshot#row_at` (or a new `at_visual_cursor`) walks rows in `ACTION_LABEL_ORDER`. Cleanest from the cursor side, but couples Snapshot to render concerns.

Decision needed before implementing. Plus a new alignment test (\`test/unit/tui/grid_state_test.rb\` or a new integration test that pins "the row dispatched on Enter is the row visually highlighted").

## Files

- \`lib/hive/tui/render/grid.rb\` (around \`draw_grouped_rows\`)
- \`lib/hive/tui/snapshot.rb#row_at\`
- \`lib/hive/tui/grid_state.rb#at_cursor\`

## Severity: P1 (silent action on wrong task)

PR #1352, opened 2026-08-30
feat: add repository-aware brainstorm suggestions

ivankuznetsov:
## Summary

Add repository-aware suggestions for unanswered brainstorm questions without giving them answer or workflow authority. Suggestions use bounded tracked repository/wiki context, isolated workers, validated provenance, and freshness bindings; stale or unsafe results remain hidden.

The TUI, Web, and JSON views now project the same advisory state. Operators can approve, decline, restore, or retry suggestions, while only the existing answer submission paths persist answers. Cleanup removes actionable advisory state when brainstorm ownership ends and before a feature downgrade.

## Validation

- `bundle exec rake coverage` — 14,057 runs, 278,174 assertions, 0 failures/errors; 100.00% lines
- `bundle exec rubocop` — 1,809 files, no offenses
- Focused feature, Web integration, and affected suggestion system suites passed (899/3,478; 326/1,916; 64/646)

The aggregate Web system run was stopped by an unrelated existing Kanban reconnect hang; the affected suggestion system suite passed independently.


ivankuznetsov:

## High
- [ ] Make malformed `hive-suggestion:v1` regions inert without deleting ordinary prose (`lib/hive/brainstorm_suggestions/envelope.rb:52`): An invalid opener is dropped while its body remains parseable as a real answer, and the same substring check silently removes unrelated operator lines containing the reserved token.
- [ ] Reject bare carriage returns during candidate admission (`lib/hive/brainstorm_suggestions/safety.rb:34`): A bare CR evades the per-line structure check, then envelope rendering converts it to LF and materializes hidden Markdown headings or extra Q/A slots.
- [ ] Require and validate `input_epoch` for every fresh record (`lib/hive/brainstorm_suggestions/store.rb:23`, `lib/hive/brainstorm_suggestions/projection.rb:203`): A fresh record without an epoch passes storage validation, skips freshness observation, and exposes its unbound text.
- [ ] Recompute current brainstorm generation and task incarnation during projection (`lib/hive/brainstorm_suggestions/projection.rb:95`): Binding reconstruction trusts persisted identity values, so edits such as adding a later question can leave an old candidate falsely fresh.
- [ ] Route TUI insertion through the shared freshness check (`lib/hive/tui/brainstorm_suggestions.rb:131`): The editor path checks only persisted state and question text, so repository, wiki, or settled-answer changes can still insert stale actionable text.
- [ ] Re-observe selected inputs after the locked result CAS (`lib/hive/daemon/brainstorm_suggestion_scheduler.rb:273`): The current recheck occurs before `publish_result`, allowing an input change in that gap to be persisted as `fresh`.
- [ ] Recheck route and sandbox availability before exposing an existing fresh candidate (`lib/hive/daemon/brainstorm_suggestion_scheduler.rb:365`): Same-epoch fresh records suppress runner construction, so loss of Bubblewrap, profile support, or route configuration leaves old text actionable instead of becoming unavailable.
- [ ] Redact and delimiter-escape question text before provider use (`lib/hive/brainstorm_suggestions/context_bundle.rb:92`, `lib/hive/brainstorm_suggestions/runner.rb:224`): Questions bypass the redaction applied to requests and answers and are interpolated raw into prompt tags, leaking secrets and permitting tag-breakout control text.
- [ ] Replace pathname revalidation with directory-descriptor-relative capture (`lib/hive/brainstorm_suggestions/context_bundle.rb:294`): A parent directory can be swapped to a symlink after component checks, redirecting the open outside the admitted root while the final inode comparison still succeeds.
- [ ] Bound prompt delivery and terminate/reap the worker on every post-spawn exception (`lib/hive/brainstorm_suggestions/runner.rb:241`): The blocking stdin write happens before timeout handling, and write failures close descriptors without killing or reaping the live sandboxed child.
- [ ] Integrate advisory cleanup into question-removal and stage-move transitions (`lib/hive/commands/approve.rb:550`, `lib/hive/daemon/brainstorm_suggestion_scheduler.rb:516`): Direct task renames and partial question pruning rely on a later daemon tick, so sidecars or projected envelopes can enter later stages and Hive history.
- [ ] Enforce an automatic-launch floor across input epochs (`lib/hive/daemon/brainstorm_suggestion_scheduler.rb:365`): Every changed binding resets the attempt budget and is immediately due while coalescing may be zero, allowing repository churn to create an unbounded provider loop.
- [ ] Prefilter due records or use a fair pending queue before reserving workers (`lib/hive/daemon/brainstorm_suggestion_scheduler.rb:232`): The scheduler repeatedly admits the first `max_workers` slots before discovering they are not due, so later questions or tasks can remain `loading` forever.
- [ ] Capture one shared repository/wiki snapshot per task projection (`lib/hive/brainstorm_suggestions/projection.rb:85`): `Observer#call` invokes full context capture once per record under one deadline, turning a multi-question JSON/Web render into N Git/wiki scans and potentially making all cards unavailable.
- [ ] Scan the complete bounded brainstorm file for cleanup fences (`lib/hive/config.rb:3294`, `lib/hive/daemon/brainstorm_suggestion_scheduler.rb:598`): Envelopes after 256 KiB evade both the feature-disable guard and terminal cleanup detection, making downgrade unsafe for the old parser.
- [ ] Add the required real Bubblewrap isolation matrix (`test/unit/brainstorm_suggestions/runner_test.rb:58`): Current tests substitute `/bin/true` and an injected executor, so they never prove live-repository denial, no shell/network tools, controller-only output, or path-escape resistance for the supported route.
- [ ] Restore nested evidence-rework task binding from `origin/main` (`lib/hive/attempts/context.rb:135`): `hive evidence rework TARGET` now resolves literal `rework` as the task, rejecting every authenticated evidence-rework worker and deleting its upstream regression test.
- [ ] Restore `origin/main`'s protected execute-task contract (`templates/execute_prompt.md.erb:37`, `lib/hive/stages/execute.rb:188`): The branch again tells implementers that protected `task.md` is writable and places controller output before the custody check, reverting the fix that prevents `implementer_tampered` failures.

## Medium
- [ ] Hide dismissed fresh payloads in server projections (`lib/hive/brainstorm_suggestions/projection.rb:220`, `web/app/views/tasks/_brainstorm_suggestion.html.erb:3`): JSON and initial/no-JS HTML still expose text, rationale, and provenance until client JavaScript hides the card.
- [ ] Require the current suggestion binding when retrying a now-fresh record (`lib/hive/commands/brainstorm_suggestion.rb:124`): Accepting either the old input binding or the new suggestion binding lets a delayed loading-card Retry clear a newer candidate.
- [ ] Bind the overlaid worktree file mode (`lib/hive/brainstorm_suggestions/context_bundle.rb:180`): Content is read from the worktree but mode always comes from the index, so an unstaged executable-bit change leaves the selected manifest and candidate freshness unchanged.
- [ ] Canonicalize candidate bytes once at admission (`lib/hive/brainstorm_suggestions/envelope.rb:32`): TUI rendering applies `rstrip` while Web copies stored text, so trailing whitespace or newlines produce non-byte-identical editable bodies across surfaces.
- [ ] Avoid rewriting unchanged sidecars on every scheduler poll (`lib/hive/daemon/brainstorm_suggestion_scheduler.rb:170`): `seed_records` always advances document `updated_at`, invalidating the projection cache and forcing unnecessary repository observations and Web updates.
- [ ] Add `hive brainstorm-suggestion` to the canonical CLI command table (`wiki/cli.md:95`): Operators cannot discover the mandatory pre-disable cleanup command from the repository's command index even though the public CLI is installed.

## Nit

ivankuznetsov:

# Code review — repository-aware brainstorm suggestions (pass 1)

Scope: `b4cbc943eb`..`f91691ebe0` (merge-base with `origin/main`), 82 files, ~9.9k diff lines, units U1-U7.
Note: the branch is 3 commits behind `origin/main`; `git diff origin/main..HEAD` also shows main's own work as reversions, so this review used the merge-base diff (the branch's actual changes).

Reviewers: correctness, security, adversarial, reliability, testing, maintainability, project-standards, api-contract, frontend-races, performance.

## High

- [ ] `lib/hive/brainstorm_suggestions/envelope.rb:53-57` — a damaged **open** delimiter promotes the advisory body to a parsed operator answer: verified that one trailing space after `-->` yields `answered=true, answer="Use the repository adapter."` and `unanswered count = 0`, so the round reads as fully answered and the daemon's completeness/auto-resume path acts on text the operator never wrote. `strip` is asymmetric — an *unterminated* region (open matches, close missing) is correctly dropped through the next boundary, but a *damaged open* line is dropped alone while its body is kept verbatim. Hard-wrapping the ~103-char delimiter or truncating the binding reproduce it too. This breaks R10 and AE1, the feature's central invariant.
- [ ] `lib/hive/brainstorm_suggestions/envelope.rb:54-55,63` — `RESERVED_RE = /hive-suggestion:v1/` is unanchored, so `strip` deletes **any** line merely containing the token, including question headings: verified a round with `### Q2. Should we keep hive-suggestion:v1 envelopes?` parses as 1 question instead of 2, silently losing an operator question that is never presented for answering. Anchor the reservation to delimiter-shaped lines.
- [ ] `lib/hive/brainstorm_suggestions/context_bundle.rb:236` + `lib/hive/brainstorm_suggestions/projection.rb:325` — the repository-controlled `.llm-wiki/config.json` `main_wiki_path` is expanded and globbed with no containment check (`canonical_directory` at :445 verifies only "is a real directory"), so a tracked config with `"../../.."` or an absolute path reads arbitrary `.md` files from outside the project into the LLM prompt bundle — and into every Web/JSON projection. Plan R2 specifies an "optional **validated** main wiki"; validate against an operator-owned allowlist.
- [ ] `lib/hive/brainstorm_suggestions/context_bundle.rb:141` + `templates/brainstorm_suggestion_prompt.md.erb` — untrusted repository evidence is interpolated between fixed literal `<untrusted-source>` / `<bound-evidence>` / `<untrusted-question>` tags, so a tracked file containing the matching close tag escapes its own fence; `path` and `class` attributes are unescaped. This skips the repo's established per-generation nonce convention (`Hive::Stages::Base.user_supplied_tag`, used at ~10 prompt sites). Inbound evidence is also screened only by `SecretPatterns` — `Safety`'s prompt-control checks apply to the model's *output*, never to the evidence going in.
- [ ] `lib/hive/brainstorm_suggestions/runner.rb:251` — `input_w.write(launch.stdin)` runs before the stdout reader thread starts and before `wait_for` arms the 120s timeout. `MAX_SELECTED_BYTES = 96 * 1024` and `render_context` inlines every entry, so the routine prompt exceeds Linux's 64 KiB pipe capacity; a worker that emits to stdout before draining stdin deadlocks both pipes with no timeout and no cancellation path, permanently consuming one of `MAX_WORKERS = 4` scheduler threads.
- [ ] `lib/hive/brainstorm_suggestions/projection.rb:86-94` — `Observer#call` runs a full `ContextBundle.capture` (several git subprocesses plus content digesting) **per unanswered question**, against the plan's explicit "at most one deadline-bounded freshness observation per task/input epoch, shared across all slots" (R5, U5, Definition of Done). All captures share one absolute deadline, so a round with several unanswered questions exhausts it and degrades every card to `unavailable`.
- [ ] `test/unit/brainstorm_suggestions/runner_test.rb:90` — the provider-profile isolation matrix required by the plan's Isolation gate does not exist: the test passes `bwrap_path: "/bin/true", executable_resolver: ->(*) { "/bin/true" }` with a stubbed executor, so it asserts the argv string it just built and never runs bubblewrap. U2's Proof requires tests that "attempt live repository/task reads, shell execution, network access, alternate outputs, and path escape for every supported route" — plan-required and undelivered, so the data-only sandbox is unverified rather than known-good.
- [ ] `lib/hive/daemon/brainstorm_suggestion_scheduler.rb:59-76` — `tick`'s only `rescue` is at method level, so one row raising inside `reconcile_row` silently abandons every remaining row for that tick; a single persistently-bad task starves suggestions for all others.
- [ ] `lib/hive/config.rb:109` — `brainstorm.suggestions.enabled` defaults to `true` for every existing project, against this repo's precedent for new subsystems (`patrol.enabled`, `refactor.enabled`, `babysitter.pairing_enabled`). Combined with `validate_brainstorm_suggestion_disable!` turning `enabled: false` into a `ConfigError` that fails *every* hive command until cleanup runs, an operator is opted in by upgrade and cannot simply opt back out.
- [ ] `lib/hive/commands/answer.rb:87-92` — `Projection.call` never consults `brainstorm.suggestions.enabled`, so with the feature disabled every unanswered slot reports a permanent `loading` suggestion in `hive answer --json` and on the Web card instead of `null`.

## Medium

- [ ] `test/integration/run_brainstorm_test.rb:92` — the AE5 "real first-pass producer rerun" sets `HIVE_FAKE_CLAUDE_WRITE_CONTENT = content`, so `assert_equal content, actual` and both `refute_includes` are tautological against a stub told to echo its input. The parser/marker/stage assertions are real; the byte-preservation and no-`## Requirements`/no-`<!-- COMPLETE -->` claims are not.
- [ ] `lib/hive/commands/brainstorm_suggestion.rb:233` — the downgrade fence's `before == after` proof is tautological: both sides are parsed with the envelope-stripping `BrainstormParser`, so a strip that eats operator content is invisible to the guard that exists to detect exactly that. Compute `before` with an envelope-unaware parse (the frozen `test/fixtures/brainstorm_suggestions/pre_feature_parser.rb` grammar).
- [ ] `lib/hive/bot/brainstorm_answer_writer.rb:120` — `parsed` comes from the envelope-stripped view while `question_line_index_for_ordinal(lines, ordinal)` counts raw lines; when `strip` drops a question heading (see the `RESERVED_RE` finding) the two disagree and the answer is written under the wrong question. Derive both from one view.
- [ ] `lib/hive/tui/brainstorm_suggestions.rb:92` — adoption vs dismissal is decided by byte-exact `content.include?(region.source)`, so an editor that normalizes CRLF or trims trailing whitespace silently dismisses an untouched suggestion. The lease captures a `file_digest` that is never checked.
- [ ] `lib/hive/tui/brainstorm_suggestions.rb:92` — answering in the TUI editor leaves the advisory envelope in `brainstorm.md`; `reconcile_editor_exit!` classifies the region `untouched` without first checking `question.answered?`, so cleanup depends on a daemon tick racing the agent it just dispatched (plan FTD #9 requires transition-locked removal).
- [ ] `lib/hive/daemon/brainstorm_suggestion_scheduler.rb:366-367` — `return true unless same_epoch` lets an input-epoch change reset `max_automatic_attempts` and bypass `next_retry_at`/`min_retry_interval_sec`, so only the 5s coalesce window bounds provider spawns under repeated repository activity. R5 requires "repeated repository activity cannot create an unbounded provider loop"; persist a cumulative attempt count.
- [ ] `lib/hive/brainstorm_suggestions/projection.rb:148` — `return result if observable.empty?` returns records that skipped freshness observation while still exposing their text; a `fresh` record with a missing/short `input_epoch` is never re-validated (and `Store#validate_state_payload` does not require one).
- [ ] `web/app/javascript/controllers/answers_controller.js:78-81` — `previousDraft` is keyed by suggestion binding, so after a regeneration the new card's Approve captures the *previous suggestion's text* as the draft; Undo then restores stale suggestion text rather than the operator's writing. Key the original draft by `fieldKey(field)`.
- [ ] `lib/hive/brainstorm_suggestions/runner.rb:67` — `sweep_inactive!` removes every `hive-brainstorm-suggestion-*` tmpdir owned by the uid with no liveness or age check despite its name, so a daemon restart deletes another process's in-flight bundle and copied credentials.
- [ ] `lib/hive/daemon/brainstorm_suggestion_scheduler.rb:199` — `seed_records` rewrites `updated_at` on every tick, and `updated_at` is part of `Projection#observation_identity`'s `sidecar_generation` slice (`projection.rb:292`), so the shared freshness cache is invalidated once per poll and effectively never hits.
- [ ] `lib/hive/brainstorm_suggestions/safety.rb:14` — `CONTROL_RE` screens ASCII controls but not invisible/bidi Unicode (U+200B-200F, U+202A-202E, U+2066-2069, U+FEFF, tag characters), which can hide or reverse text in a candidate the operator is invited to adopt verbatim.
- [ ] `schemas/hive-answer.v1.json:30` — `suggestion` is added to `Slot.required` under `additionalProperties: false` with no version bump, so an old cached v1 validator rejects new payloads and a new validator rejects old ones. The code always emits the key (possibly `null`), so dropping it from `required` restores compatibility at no cost.
- [ ] `lib/hive/commands/brainstorm_suggestion.rb:19-20` — `hive-brainstorm-suggestion-cleanup` / `-action` receipts are emitted with hardcoded `"schema_version" => 1` and no `schemas/*.json`, unregistered in `Hive::Schemas::SCHEMA_VERSIONS`, unlike every sibling command's `--json` contract.
- [ ] `lib/hive/commands/brainstorm_suggestion.rb:39-43` — usage/`Hive::Error` failures raise uncaught, so `--json` callers get an unstructured error where the sibling `Hive::Commands::Answer` emits a structured `emit_error` payload.
- [ ] `lib/hive/commands/brainstorm_suggestion.rb:125` — `binding_matches?` accepts `input_binding` as well as `suggestion_binding`; since `input_binding` is stable across retries within one epoch, Retry/dismiss can act on a different candidate than the one displayed.
- [ ] `lib/hive/brainstorm_suggestions/context_bundle.rb:188` — a single tracked symlink or submodule anywhere in the project aborts the **entire** capture with `unsafe_tracked_entry`, permanently disabling the feature for that project rather than skipping the ineligible entry.
- [ ] `lib/hive/brainstorm_suggestions/context_bundle.rb:397` + `lib/hive/brainstorm_suggestions/projection.rb` — the spawn/deadline/process-group-terminate protocol is copy-pasted between the two new files, bringing the codebase to four independent implementations (alongside `GitOps#run_git_with_timeout` and `AutoCommit.capture_git_with_timeout`, which already differ in kill behavior). Extract one helper.

## Nit

- [ ] `lib/hive/brainstorm_suggestions/context_bundle.rb:261` — `break if total >= max_bytes` inside a `filter_map` block makes the block evaluate to `nil`, so the following `.first(max_files)` raises `NoMethodError` on the one path that reaches it; the preceding `next if` guard already enforces the byte cap, so the line is dead or a crash.
- [ ] `lib/hive/brainstorm_suggestions.rb:25` — `Hive::Error` is referenced without requiring `hive/errors`; verified that `require "hive/brainstorm_parser"` standalone now raises, since the parser requires the envelope. Works only because real entrypoints load `hive` first.
- [ ] `lib/hive/config.rb:3299` — hardcodes `"hive-suggestion:v1"` instead of using `Envelope::RESERVED_RE`, so the disable fence can drift from the grammar it guards.
- [ ] `lib/hive/web/status_feed.rb:448` — re-derives `Store#safe_file?`'s owner/permission predicate instead of sharing it.
- [ ] `lib/hive/brainstorm_suggestions/store.rb` — `validate_document`'s trailing `ensure persisted` is a no-op whose only purpose is consuming an unused keyword argument.
- [ ] `web/app/views/tasks/_brainstorm_suggestion.html.erb:2` — the local `binding` shadows `Kernel#binding` inside an ERB template.
- [ ] `lib/hive/brainstorm_suggestions/projection.rb:148` — returns an unfrozen hash on the early-return path while every other exit freezes.
- [ ] `web/app/views/tasks/_brainstorm_suggestion.html.erb:40` — the Retry link has no pending/disabled state or client-side de-duplication; a double click sends two POSTs (the backend's lock and binding check make the loser a typed 422, so this is UX only).
- [ ] `lib/hive/brainstorm_suggestions/runner.rb:189` — `--disallowedTools default` names no real tool and is a no-op; `--tools ""` is what disables the tool surface. Either spell out the deny list or drop the flag.
- [ ] `lib/hive/tui/help.rb:44` — `u` restores "the first dismissed" suggestion, which is ambiguous on a multi-question round.

## Verified clean

Checked and found sound, recorded so pass 2 need not re-derive them:

- Sandbox argv: `--tools ""`, `--json-schema`, `--setting-sources`, `--safe-mode`, `--disable-slash-commands`, `--no-session-persistence`, `--strict-mcp-config`, and `--permission-mode dontAsk` are all valid on the shipped Claude CLI, and `--tools ""` is documented as "disable all tools". No live repo or task path is bind-mounted.
- Sidecar deserialization is JSON-only with key allowlists — no `YAML.load`, `Marshal`, or key-driven object instantiation (commit `a3361b6f3e` is complete).
- The Web card escapes all model-produced text (no `raw`/`html_safe`), and the retry route is CSRF-protected, auth-gated, archive-blocked, and binding-matched; it cannot write answers.
- `BrainstormParser.parse_text` strips *intact* envelopes correctly, so a well-formed suggestion body never parses as an answer (the High findings above are the malformed cases).
- Wiki protocol: `wiki/log.md` untouched, the `wiki/log.d/20260830T115504Z-*.md` fragment matches sibling naming, and no new wiki page was added so `wiki/index.md` needs no edit.
- TUI keys `u` and `R` collide with no existing binding in `VERB_KEYS` or `help.rb`.
- The downgrade command (`hive brainstorm-suggestion clean`) and its `safe_to_disable` receipt exist and are registered, satisfying plan FTD #10's shape (though see the tautological fence above).

## Verdict

**Not ready.** The envelope grammar's handling of malformed reserved regions defeats the feature's stated central invariant in two verified ways — a damaged open delimiter makes advisory text a parsed answer, and an unanchored token match deletes operator questions outright. Both are narrow one-line fixes in `envelope.rb`, but they need regression tests of the shape "for every single-character mutation of the delimiter, the slot stays unanswered." The plan-required provider-isolation matrix is also undelivered, so the data-only sandbox is currently unproven rather than proven.

Suggested fix order: (1) `envelope.rb` strip symmetry + delimiter anchoring, with a mutation table test; (2) `main_wiki_path` containment and nonce-fenced evidence; (3) `runner.rb` stdin/reader ordering; (4) per-task freshness observation; (5) the real isolation matrix; (6) the config default and disabled-state projection.

ivankuznetsov:

## High
- [ ] Consume nested and orphan `hive-suggestion:v1` delimiters as one corrupt advisory region (`lib/hive/brainstorm_suggestions/envelope.rb:116`): The current scan can strip reserved comments while leaving candidate bytes parseable as an operator answer, which can trigger completion and dispatch.
- [ ] Restrict main-wiki evidence to tracked files (`lib/hive/brainstorm_suggestions/context_bundle.rb:288`): Recursive `Dir.glob` admits ignored and untracked private notes into provider context, violating the tracked-only evidence contract.
- [ ] Detect GitLab personal access tokens in the shared secret policy (`lib/hive/secret_patterns.rb:13`): Common `glpat-` tokens pass context and result screening unchanged and can enter bundles, persisted state, and client projections.
- [ ] Preserve controller-owned provider transport through the data-only sandbox (`lib/hive/brainstorm_suggestions/runner.rb:198`): Running the only supported Claude CLI under `--unshare-net` prevents it from reaching its API, so every production attempt fails despite the route being reported available.
- [ ] Give retryable records without candidates a binding that Retry accepts (`lib/hive/commands/brainstorm_suggestion.rb:135`): Capture failure and attempt exhaustion clear `suggestion_binding` while advertising Retry, but Web submits an input binding that the command rejects and TUI refuses to act.
- [ ] Do not cache failed freshness observations indefinitely (`lib/hive/brainstorm_suggestions/projection.rb:63`): A transient timeout is stored under an identity that excludes retry and candidate-state changes, so later same-input fresh candidates remain hidden until eviction, restart, or unrelated input churn.
- [ ] Move repository and wiki identity scans out of read-side projection and the TUI task lock (`lib/hive/brainstorm_suggestions/projection.rb:166`): The expensive identity is computed before every cache lookup, making each Web render rescan Git/wiki and allowing TUI freshness work to delay answer writes.
- [ ] Escalate timed-out capture children from TERM to KILL (`lib/hive/brainstorm_suggestions/process_capture.rb:61`): A TERM-ignoring child blocks in `Process.wait` forever, defeating the deadline and permanently consuming scheduler or interactive capacity.
- [ ] Run transition cleanup before manual-steering archival (`lib/hive/tui/bubble_model.rb:1666`): The direct rename can move actionable sidecars and projected envelopes into `archived-manual`, beyond scheduler ownership and into archived material.
- [ ] Scope Web Undo snapshots to the suggestion binding (`web/app/javascript/controllers/answers_controller.js:81`): After Retry replaces candidate A with B for the same field, approving and undoing B restores the stale pre-A draft instead of the text present immediately before B.
- [ ] Restore staged reviewer publication and preserve prior successful evidence on failure (`lib/hive/reviewers/agent.rb:111`): The diff deletes canonical Agent and Codex reviewer output before retries and on final error, so a transient rerun destroys the last successful review artifact.
- [ ] Exclude immutable terminal checkpoint history from the mutable attempt budget (`lib/hive/task_projection/store.rb:451`): A task with 101 sealed terminal attempts becomes partial even though routine reads need no attempt-store lookups, reverting the guarded `origin/main` behavior.
- [ ] Keep fresh project initialization aligned with the documented suggestion opt-in (`templates/project_config.yml.erb:120`): The template writes `enabled: true` even though shared defaults, tests, and current wiki documentation require an explicit operator opt-in.
- [ ] Split the newly oversized status-feed test (`test/unit/web/status_feed_test.rb:1044`): This change grows the file from 998 to 1,044 lines, crossing the review's 1,000-line maintainability limit with suggestion-specific cases that belong in a focused test file.

## Medium
- [ ] Bound pre-approval suggestion drafts across Turbo navigation (`web/app/javascript/controllers/answers_controller.js:81`): The module-scoped map adds an entry on approval and removes it only on Undo, retaining unbounded potentially sensitive answer text while the other browser-state maps enforce a 100-entry cap.

## Nit

ivankuznetsov:

## High
- [ ] Keep workflow markers inside reserved envelopes inert (`lib/hive/brainstorm_suggestions/envelope.rb:130`): An exact opening delimiter followed by `<!-- WAITING -->`, candidate text, and the exact closing delimiter makes the real parser return that candidate as an answer; the structural-boundary break exposes the enclosed tail without operator adoption, violating R6/R10. Consume the reserved region across internal markers while preserving genuine subsequent question boundaries.
- [ ] Invalidate changed candidates before throttling regeneration (`lib/hive/daemon/brainstorm_suggestion_scheduler.rb:345`): When selected evidence changes within the default 300-second retry interval, `request_due?` returns false and leaves the old record fresh; Web refresh tokens depend on sidecar bytes, so an already-open card retains actionable stale text. Persist stale/null text and revoke the old binding immediately, applying the interval only to replacement generation (R5/AE2).

## Medium
- [ ] Reject special files without blocking descriptor admission (`lib/hive/brainstorm_suggestions/context_bundle.rb:369-374`): Replacing a tracked text file with a FIFO blocks the read-only open before the regular-file check; cooperative capture deadlines cannot interrupt it, and four such captures exhaust the scheduler worker pool. Open nonblocking and validate descriptor type before reading.
- [ ] Handle hosts without Linux procfs explicitly (`lib/hive/brainstorm_suggestions/context_bundle.rb:371`): Descriptor traversal unconditionally uses `/proc/self/fd`, which is absent on macOS; `stable_tracked_read` rescues the resulting `ENOENT` as deletion, silently dropping all repository/wiki evidence and permitting request-only candidates. Use portable descriptor-relative traversal or return unavailable instead of treating unsupported capture as an empty repository.
- [ ] Enforce an absolute provider-attempt deadline (`lib/hive/brainstorm_suggestions/runner.rb:182-185`): Separate HTTP open/read/write timeouts do not bound total execution, so slow successive response chunks can retain workers and bundles far beyond `timeout_sec`; ownership cancellation has no elapsed-time check and live-job deduplication prevents orphan recovery. Add a monotonic overall deadline that closes transport and finalizes cleanup.
- [ ] Avoid full repository scans on projection cache hits (`lib/hive/brainstorm_suggestions/projection.rb:169-187`): Every unchanged Web/JSON projection invokes the identity factory twice, each running repository-wide `git ls-files` and binary `git diff HEAD` (plus main-wiki scans); caching only avoids bundle capture and still repeats four Git subprocesses per task/render. Share a bounded change observation across consumers so cache hits meet U5's no-scans-per-render requirement.

## Nit

PR #1390, opened 2026-09-02
Guard TERM-to-KILL escalation against reused PIDs

ivankuznetsov:
When a recorded process survives TERM, its PID may be reused before escalation. Single-PID cleanup now rechecks the recorded start identity before KILL: a replacement means the recorded process is gone, while an unreadable identity on a live PID refuses cleanup. Drop preserves task data, logs, worktrees, branches, and PR cleanup whenever any candidate returns `kill_failed`.

The sweep also fixed a bypass where the same PID appeared with both a recorded start identity and an older marker without one. Unknown duplicates are suppressed when a known identity is available; distinct known identities and legacy-only PIDs retain their existing behavior.

Validation: the new mixed-identity regression fails before the repair and passes after in both insertion orders. Focused tests passed (160 tests, 923 assertions); eight changed Ruby files pass lint. Independent repair review found no remaining issue. No real process was signalled by the new regression. The documented identity check-to-signal race remains.

Refreshed onto main through #1380; the same 160 tests and 923 assertions pass, and all eight changed Ruby files pass RuboCop 1.90. All 33 exact-head checks passed. The merged tree matches that tested head; post-merge CI and install smoke also passed.


ivankuznetsov:

## High
- [ ] Preserve distinct recorded identities when a reused PID appears in multiple task folders (`lib/hive/commands/drop.rb:418`): PID-only dedup keeps the first start time, so a stale record can hide the live runner's `kill_failed` result and let Drop delete its recovery state.

## Medium
- [ ] Treat identity-source I/O failures as unavailable (`lib/hive/process_kill.rb:38`): the wrapper only rescues coercion errors, so a post-TERM `Errno::EIO` escapes instead of returning the required `kill_failed` result without KILL.

## Nit
- [ ] Split the unrelated web task-fixture isolation commit from this change (`web/test/integration/tasks_test.rb:15`): the web fixture, its testing-wiki edit, and its log fragment are outside the process-kill plan's explicit final-diff scope.

ivankuznetsov:

## High
- [ ] Split the new identity-escalation tests out of `test/unit/process_kill_test.rb:163`: this diff grows the file from 684 to 1,047 lines, crossing the review rubric's explicit 1,000-line regression threshold; require the companion from the existing file so the plan's focused command remains unchanged.
- [ ] Split the new agent-cleanup tests out of `test/unit/commands/drop_test.rb:108`: this diff grows the file from 910 to 1,153 lines, crossing the review rubric's explicit 1,000-line regression threshold; require the companion from the existing file so the plan's focused command remains unchanged.

## Medium

## Nit
- [ ] Record the residual single-PID identity-check-to-signal race in `wiki/gaps.md` (`wiki/log.d/20260902T003002Z-terminate-process-identity-escalation.md:25`): the project instructions require durable uncertainty tracking, and existing gap 54 covers the distinct descendant snapshot/concurrent-fork race.

PR #1401, opened 2026-09-05
refactor: simplify recovery persistence and runtime test contracts

ivankuznetsov:
Restart-baseline persistence now uses Hive’s existing atomic writer under the daemon’s single-writer custody. The private lock and tempfile implementation are removed; baseline contents, microsecond timestamps, durability, failure diagnostics, and both orphan-temp naming conventions remain.

The change also removes unused source-size accounting from the retirement fixture while preserving the exact legacy-writer and path-override lists. Baselines remain available to protect answered tasks across daemon restarts.

Validation: seven reviews found no defects; focused persistence and dispatcher checks pass (438 tests, 1,557 assertions), and three changed Ruby files pass lint. Required exact-head checks passed before merge. Post-merge verification passed on main including #1404 (CI34425962382 and Install34425962419); the earlier run was cancelled when main advanced.

PR #1404, opened 2026-09-06
Harden the generated configuration defaults reference

ivankuznetsov:
The configuration wiki now derives its complete defaults reference from `Hive::Config::DEFAULTS`. Maintainer refresh and read-only drift verification share one deterministic renderer. It rejects malformed, duplicate, wrapped, or misplaced markers before writing, preserves surrounding bytes, and skips unchanged writes.

The sweep found the change still needed on main. Integration regenerated the documented defaults from current configuration; the stale-page check failed before that refresh. The existing three supervisor test deadline increases remain test-only.

Validation: six reviews found no defects. Generator and supervisor tests pass (40 tests, 317 assertions); four Ruby files pass lint. The actual maintainer command refreshed the page, then a second invocation preserved its bytes and timestamps. Final refresh includes main through #1401; the generated reference remains current and the same 40 tests pass. All 33 checks passed before merge. Post-merge CI and Install smoke passed.


ivankuznetsov:

## High

- [ ] Reject altered markers alongside a valid region (`lib/hive/config_defaults_doc.rb:57-65`): The validator counts only exact marker strings, so appending `<!-- BEGIN GENERATED Config::DEFAULTS -->\nstale duplicate\n<!-- END GENERATED Config::DEFAULTS -->\n` to the current canonical page makes `render(page) == page` pass; this reproduces a malformed-page false pass and violates required R4/R7. Detect malformed Config::DEFAULTS marker candidates throughout the page and add valid-plus-altered cases covering rendering, read-only verification, and rejection before regeneration writes.

## Medium

## Nit

ivankuznetsov:

## High
- [ ] Reject multiline malformed marker candidates (`lib/hive/config_defaults_doc.rb:14-15`): Appending `<!-- BEGIN GENERATED:\nConfig::DEFAULTS -->\nstale\n<!-- END GENERATED:\nConfig::DEFAULTS -->\n` to a current valid page returns byte-identical output because both candidate regexes exclude newlines; the drift guard therefore false-passes an additional malformed region, violating R4/R7. Detect wrapped marker comments and cover them before and after the valid region.
- [ ] Refresh the exact production inventory after the marker fix (`test/fixtures/runtime_control_plane/affected_production.yml:9-16`): The fixture records 76 lines for `config_defaults_doc.rb`, but HEAD contains 81; `RuntimeControlPlaneDeletionContractTest#test_final_affected_inventory_is_exact_and_at_least_twenty_percent_smaller` asserts this exact count, so the default suite and coverage gate fail. Update the file count and derived totals (8545 inventory lines and 397 final lines).

## Medium

## Nit

Review evidence: Reviewed `git diff origin/main..HEAD` against the supplied plan. Read-only Ruby probes confirmed both defects and verified that the actual committed config page matches the shared renderer under Bundler. `git diff --check origin/main..HEAD` passed. Suites were not run because their temporary-file writes exceed the single-file write constraint; the inventory failure was verified by evaluating the exact physical-line counts used by its assertion. An independent read-only reviewer corroborated the malformed-marker finding. The skill's auxiliary artifacts and cross-model job were omitted to honor the same write constraint.

PR #1406, opened 2026-09-06
Make systemd user-service installs replay-safe and network-target-free

ivankuznetsov:
Hive user services no longer depend on system-manager network targets. Installation, removal, and lifecycle commands share one durable transition owner, so interrupted work either verifies the desired or prior state or retains recovery evidence for retry.

The sweep repaired teardown after unverified foreground stops and filesystem failures, custom-configuration recovery, launchd unload verification, and Linux process verification after reload. Removal records whether a foreground stop is required and completed, including when another command resumes recovery. Unused manager compatibility methods were removed.

Validation: 386 focused tests and 2061 assertions pass, with the real-manager case exercised separately. The real-systemd gate passes all 232 tests and 1060 assertions, including offline reconnect, stable process identity, contention, replay, and cleanup. Eight Ruby files pass lint; security and final correctness repair reviews are clear. The broad local checkpoint ran 14,495 tests and found an injected-manager model error plus a wiki wrapping assertion; both were repaired. Follow-up checks pass 327 tests and 1,763 assertions. Final refresh through #1404 passes 338 tests and 1,918 assertions; nine changed Ruby files pass lint. Exact-head hosted CI is pending, including the new real-systemd offline reconnect gate.

The first refreshed hosted run passed its functional gates but found four uncovered lines. Three recovery tests now cover retained journals and replay after unverifiable reloads; an unreachable fallback was removed. Focused UserService line coverage is 874/874. Hosted checks are rerunning on the final repair.

Verification: merged as `789d7e12d9b1fd66d11caf7bd92012e4f7b7f12b`; its tree exactly matches the tested repair. All 34 premerge checks passed. Postmerge CI and install smoke passed; the advisory timing job reported only the plan scenario at 11.089 seconds against an 11-second budget.


ivankuznetsov:

Review scope: `git diff origin/main..HEAD` at `107d3d5d8c`, against the supplied implementation-ready plan. Verdict: not ready.

## High

- [ ] Abort destructive uninstall/purge after busy or ambiguous removal (`lib/hive/commands/uninstall.rb:94`, `:108`, `:289`): `operation_busy` and retained-recovery failures only print "leaving it in place and continuing"; cleanup then returns success and, with `--force-purge-state`, deletes the default `~/.local/state/hive`, including the new live service locks and journals. Stop cleanup on these failures and preserve coordination evidence; otherwise R4/F2's contention and fail-closed guarantees are defeated.
- [ ] Complete the required runtime-cutover ownership routing (`lib/hive/runtime_control_plane/maintenance.rb:68-81`): `transition!` still directly executes systemctl start/stop and launchctl bootout/bootstrap, reached by `Cutover` at `cutover.rb:185`, `:212`, and `:268`. These actions can race a locked installation/removal and invalidate its observations. Route them through UserService lifecycle operations as explicitly required by R4/KTD1/U3; this mandatory caller was left unchanged, not deferred by the plan.
- [ ] Continue the requested install after recovering a different operation (`lib/hive/user_service.rb:107-108`): `apply` immediately returns every recovery result. Interrupt a stop after `lifecycle_acted`, then install with autostart enabled: recovery completes the stop and returns successful `:unchanged`, so the installer reports success while the service remains stopped. After successful unrelated recovery, replan under the same lock and perform the requested apply instead of returning the recovered operation's result.
- [ ] Prove the restoring activation before clearing rollback evidence (`lib/hive/user_service.rb:1153-1155`, caller `:809-823`): if desired process B is running during rollback to A, an interruption after `Manager#restore` reloads A but before its restart leaves A loaded while B still runs. Replay accepts matching enabled/running booleans plus any process identity, skips restoration, and clears the journal as `prior_state_restored`. Persist and verify restoring-process evidence; loaded file metadata alone does not prove the running definition.
- [ ] Retain stop intent when manager observation becomes unavailable (`lib/hive/user_service.rb:1174-1175`, `:1192-1194`): after a failed/timed-out stop, unavailable observations carry `running=false`; both stop predicates interpret that as verified completion. Actual in-memory execution of `complete_lifecycle` returned success and deleted the journal with `manager_available=false`. Require conclusive manager availability at every verification boundary and retain pending evidence when stop cannot be verified.
- [ ] Restart an existing process when promoting a filesystem-only receipt (`lib/hive/user_service.rb:385-394`): start A, write B with `autostart: false`, then let another service install run daemon-reload. Enabling autostart for B now sees a nonnil receipt and current loaded metadata, selects `enable --now`, and leaves A running while recording a managed B receipt. Conservatively reconcile the running process when the receipt does not prove its definition; R3/F5 requires a verified desired file/process pair.

## Medium

- [ ] Allow restart recovery when the prior service was stopped (`lib/hive/user_service.rb:1180-1186`): a stopped service records no prior PID/start identity; after a successful real restart, neither completion branch accepts the new process without that prior identity, so the result is ambiguous and the retained journal blocks subsequent lifecycle operations. Recognize the recorded stopped-to-running transition using valid fresh process evidence, and cover replay from that state.
- [ ] Make the required dynamic gate fail when owned teardown fails (`test/integration/systemd_user_service_offline_test.rb:194-210`): failed UserService removal is silently repaired with direct disable/unlink/reload commands, allowing a removal regression to pass the mandatory gate. Teardown also treats failed manager queries as inactive/PID zero. Keep emergency cleanup, but preserve the failure, require conclusive manager observations, and verify process/unit and coordination-state cleanup as required by U4.

## Nit

None.

Validation: three local review passes, cross-checks of the retained findings, and read-only in-memory Ruby reproductions of transition predicates, lifecycle completion, and uninstall continuation. No repository tests or real service mutations were run. The external peer workflow was not run because it requires artifacts outside the single authorized output file. No code files were changed.

ivankuznetsov:

## High
- [ ] Bind replayed restart verification to the desired loaded definition (`lib/hive/user_service.rb:1174`): After interruption following publication but before reload, systemd can automatically restart the old cached definition; recovery reloads the desired definition and then accepts any PID/start identity different from the original as completed activation, committing a receipt while the old configuration still runs; persist and verify the activation boundary instead of treating PID inequality as proof (R3/KTD3; read-only Ruby reproduction returns `:complete` for this observation).
- [ ] Acquire target ownership before uninstall signals service processes (`lib/hive/commands/uninstall.rb:99`): `deregister_babysitter`, `deregister_daemon`, and `deregister_bot` invoke their foreground-stop helpers before the locked `remove!`, and those PID files also identify supervised instances; a competing uninstall therefore terminates the service or removes PID evidence before reporting `operation_busy`, violating R4/U3's zero-mutation contention contract; move these effects inside the shared ownership scope.
- [ ] Check manager availability before replay mutates the prior file (`lib/hive/user_service.rb:489`): Following interruption at `after_journal_prepared`, recovery creates the backup and publishes desired bytes before its first manager inspection in `complete_apply_transition`, so an absent or indeterminate manager causes the previously consistent file to be replaced before returning pending; inspect availability before forward or rollback file mutations and retain existing evidence unchanged when the recorded manager is unavailable, as required by R3/R4's recovery matrix.
- [ ] Require a conclusively stopped process before completing stop (`lib/hive/user_service.rb:1254`): Manager inspection maps `ActiveState=deactivating` with a live `MainPID` to `running=false`, which both the stop shortcut and recovery endpoint accept as success while shutdown is still draining; preserve transitional state and wait within the declared bound or retain pending intent until the process has stopped, also using that verification before removal unlinks the unit (R4; read-only reproduction with MainPID 123 reports `stop_endpoint_verified: true`).

## Medium
- [ ] Surface lifecycle contention and recovery diagnostics to CLI callers (`lib/hive/commands/service_installer/base.rb:237`): `perform_lifecycle!` stores actionable messages but raises only a generic error, while `Web#run_service_action` never prints those messages and the web adapter replaces the error again; consequently `hive web start/stop` hides whether the operator should retry contention or inspect retained recovery evidence; carry the bounded diagnostics through the raised error or presenter as required by U3.

## Nit

ivankuznetsov:

## High

- [ ] Revalidate manager absence before replaying filesystem-only removal (`lib/hive/user_service.rb:1015-1018`): A removal recorded with `manager_intent: nil` skips availability checks and disable when replay observes an available/running or indeterminate manager, then unlinks the unit at line 1057; reproduced with an in-memory probe that left the service enabled/running and returned `recovery_pending`. Its unchanged nil intent also prevents completion while the manager remains available. Preserve the unit on ambiguity and reconcile newly available manager state before unlinking, as required by R4.

## Medium

- [ ] Finish process-group termination after reaping the manager-command leader (`lib/hive/user_service/manager.rb:642-654`): `terminate_and_reap` returns at line 647 when the leader exits after TERM, bypassing the group KILL; a subprocess reproduction confirmed that a TERM-ignoring child survives the timeout. `run_production` then marks cleanup complete and can release service ownership while that child can still act. Track leader reaping separately and finish bounded group cleanup before returning.

## Nit

None.

PR #1407, opened 2026-09-06
Add Hive-owned daily activity digest

ivankuznetsov:
Hive now keeps a durable daily activity digest across registered projects. One stored record supplies the authenticated Web view, native JSON commands, and opt-in Telegram recap. It distinguishes observed empty days from unavailable data, preserves closed history through amendments, and records explicit pruning without recreating removed digests.

The sweep repaired legacy registration initialization, repeated outage/recovery amendments, duplicate post-prune observations, stale attention after registration removal, and dated gaps appearing on unrelated days. Missing and pruned delivery inputs fail before ledger or transport effects. Web state descriptions use plain wording.

Validation: fourteen reviews and independent finding validation, plus migration and projection repair reviews. After integrating current main, 703 focused tests pass with 3,871 assertions and 126 Ruby files pass lint. Rails checks pass 7 tests/102 assertions; the digest browser flow passes 35 assertions. Focused ProjectSource coverage exercises every line. The broad run completed 14,544 tests with one command-index failure; its documentation fix passes the focused index check and is included in the combined passing run. Exact-head hosted CI is pending. No real Telegram messages were sent.

Merged as `ca5c0b9e6afa8f02cddf2b25fce422422dd2e283`; its tree exactly matches the tested head. Post-merge CI [34429595356](https://github.com/ivankuznetsov/hive/actions/runs/34429595356) and Install smoke [34429595322](https://github.com/ivankuznetsov/hive/actions/runs/34429595322) succeeded.


ivankuznetsov:

## High

- [ ] `lib/hive/daemon/daily_digest_delivery_scheduler.rb:61` — `def cancel(date:)` raises `ArgumentError: unknown keyword: :stage` on every delivery cancel, because `dispatcher.rb:1020` calls `scheduler.cancel(date: ..., stage: ...)` for all global digest schedulers and both `digest_scheduler_base.rb:25` and `daily_digest_close_scheduler.rb:81` declare `cancel(date:, stage: nil)`; the dispatcher tests only exercise a fake scheduler, so the real arity is never called with `stage:`.
- [ ] `lib/hive/daily_digest/coordinator.rb:48` — a date-scoped `hive digest refresh --date D` permanently strands every intermediate day: the `date` branch selects `[ interval ]` (a single element) instead of the chronological prefix that the unscoped branch builds via `due_for_materialization?`, so days between the coverage frontier and `D` are skipped and, once `D` is persisted, never become due again.
- [ ] `lib/hive/daily_digest/materiality.rb:177` — a journal row missing `occurred_at` raises an uncaught `KeyError` that aborts the entire multi-project collection instead of degrading to one scoped gap: `fact` calls `row.fetch("occurred_at")`, `classify`'s rescue at `:132` covers only `JSON::GeneratorError, TypeError, ArgumentError`, and `collector.rb:43` covers only `SourceUnavailable, SystemCallError, IOError` — while sibling `coverage.rb:33` already rescues `ArgumentError, KeyError, TypeError`, so this is a dropped convention.
- [ ] `lib/hive/daily_digest/coordinator.rb:243` — post-prune replay grows the tombstone discard audit without bound, defeating the plan's "bounded, idempotent discard audit": `observed_at` is part of the hashed discard identity, so a replayed fact observed at a new instant is recorded as a *new* discard rather than recognized as the same one.
- [ ] `lib/hive/daily_digest/coordinator.rb:260` — `due_for_materialization?` marks any closed record with a gap as due forever, with no attempt budget or age bound, so a gap that can never resolve (e.g. a replaced project whose old path is gone) keeps re-collecting that day on every cycle and starves catch-up.
- [ ] `lib/hive/daily_digest/collector.rb:32` + `project_source.rb:198` — the per-source frontier that the plan names as the scan-cost mitigation is written but never read back, so materialization is a full fleet scan: `ProjectSource.new` accepts no cursor at any call site and `collect_journal` unconditionally `File.binread`s and re-parses every task journal, so each 5-minute daemon tick re-reads every journal in every registered project regardless of change, and restart catch-up repeats that scan once per missed day.
- [ ] `lib/hive/daily_digest/store.rb:189` — one orphaned record directory (a `<date>/` with no `base.json`, the state a crash between `mkdir` and the atomic write leaves) permanently breaks *every* digest read: `Store#intervals` calls `read(date)` unguarded, so the raised `MissingRecord` propagates out of the reader for all dates, not just the damaged one; `Pruner#call` at `pruner.rb:25` already rescues exactly this.

- [ ] `lib/hive/daily_digest/project_source.rb:220` — the plan's fact-assignment rule "assign a fact by searching the persisted half-open interval sequence using known event time; **otherwise use `observed_at`**" (U3 approach 3) has no implementation: `in_window?` classifies strictly on `occurred_at` (`:444`), so a fact with no event time is never assigned to its observation day — there is no fallback path and no test.

- [ ] `lib/hive/daily_digest/reader.rb:48` — a stale open record still reports `content: "empty"`, violating plan assumption A3 ("a partial record with no known items is `unknown`, never empty") and the named false-empty-record risk: the stale branch sets `view["view_completeness"] = "partial"` but never touches content, so when the materializer is unavailable the CLI, agent JSON, and web view all assert a definitive "nothing happened today" instead of "unknown".

## Medium

- [ ] `lib/hive/daily_digest/reader.rb:13` — `UnknownProject` has no `exit_code` override, so `hive digest --project WRONG` exits 1 instead of the 64/`USAGE` code that `Forget::UsageError`, `Metrics::UsageError`, and `daemon.rb`'s `UNKNOWN_PROJECT` path all use and document as the CLI-wide convention.
- [ ] `lib/hive/commands/digest.rb` / `schemas/hive-digest.v1.json` — the JSON envelope, declared the stable agent contract, omits the remediation command for `missing`/`stale`/`pruned` outcomes even though the CLI text renderer (`digest.rb:166`) and web view (`daily_digest.rb:64`) both compute that exact string, and `Hive::Schemas::ErrorEnvelope` already has a `safe_correction` hook the four digest commands never wire up.
- [ ] `lib/hive/config.rb:1771` — `project_membership_history` is appended on every `register_project`/`unregister_project`/`prune_missing_projects!` and grows forever in `config.yml`, contradicting the plan's "appends a **bounded** before/after membership-history entry"; the sole consumer is `DailyDigest::Coordinator`.
- [ ] `lib/hive/daily_digest/store.rb:37` — the open-record identity guard compares `existing.fetch("local_date") != prepared.fetch("local_date")`, but `local_date` is the directory key both sides were loaded under, so the comparison is tautological and the guard can never fire; it should compare `interval_id` (the test only reaches this branch by stubbing `read_json`).
- [ ] `lib/hive/daily_digest/store.rb:236` — `merge_frontiers` folds the frontier overlay last with no recency check, so an older overlay entry permanently shadows a newer amendment frontier for the same source.
- [ ] `lib/hive/daily_digest/coordinator.rb:134` — `cutover.requested_at` reads `config["time_zone_requested_at"]`, a key nothing ever writes, so it silently falls back to `@clock.call`; because that value feeds `interval_id` in `calendar.rb:146`, the cutover interval's identity is clock-dependent rather than derived from boundary facts.
- [ ] `test/unit/daily_digest/coordinator_test.rb:148` — the only test named for the close/amend race stubs the store out entirely (`close_store.define_singleton_method(:write_base) { raise ImmutableRecord }`), so it exercises the coordinator's rescue clauses rather than the lock, and would pass unchanged if the `flock` in `store.rb` were deleted.
- [ ] `lib/hive/config.rb:2087` — `canonical_membership_value` reimplements the recursive canonicalization already provided by `Hive::WorkflowPackage::CanonicalJSON` (which `Modules::EventLedger` uses for the identical purpose); the same algorithm is now written a third time in `daily_digest/record.rb`.

- [ ] `lib/hive/commands/digest.rb:40` — `--open-web` reports success when no browser launcher exists: the guard is `raise ... if opened == false`, but `system` returns `nil` (not `false`) when the program is not found, so the command prints `Opened <url>` and exits 0 having opened nothing.

## Nit

- [ ] `web/test/system/digest_flow_test.rb:21` — `wiki/commands/web.md:818` claims this test proves "keyboard focus," but it contains no focus or tab-order assertion; add one or narrow the documented claim.
- [ ] `lib/hive/daily_digest/project_source.rb:381` — `attention_item` persists `project`/`task_id`/`task_slug`/`stage` verbatim while the sibling `Materiality.fact` path bounds and control-strips the identical fields; not exploitable (every sink defends itself) but it is a uniformity gap in the immutable record and the `--json` contract.
- [ ] `schemas/hive-digest.v1.json` — `items[].task_url` / `attention[].task_url` are bare relative paths with no origin field, unlike `web_url` which the schema forces to be absolute; an agent consuming the JSON cannot resolve them.
- [ ] `lib/hive/daily_digest/gap.rb:10` — `Gap.build` is a pure pass-through to `Materiality.build_gap` that all six production call sites bypass; its only caller is a unit test.
- [ ] `lib/hive/daily_digest/project_source.rb:65` — `ProjectSource` computes a `health` value its only caller never reads, and `SourceHealth.unavailable` has no production caller in the diff.
- [ ] `lib/hive/daemon/concurrency_controller.rb:155` — vestigial `:digest` default identity left over from the multi-lane generalization; production always passes an explicit identity now.
- [ ] `web/app/views/digests/_attention.html.erb:15` — the digest partials use `Hash#fetch` on persisted record fields, so a record missing `kind`/`name`/`project_id` raises `KeyError`, which `ApplicationController`'s `rescue_from` chain does not cover — a 500 rather than the project's typed error page.
- [ ] `schemas/hive-digest-refresh.v1.json`, `schemas/hive-digest-prune.v1.json` — no JSONSchemer producer-conformance test, unlike the digest and send schemas; drift would not be caught.
- [ ] `lib/hive/daily_digest/recovery.rb` — no production caller: no CLI verb routes to it (`cli.rb:1291-1316` routes only refresh/send/prune) and no daemon path instantiates it; R10's retry/resolution is fully delivered by `Coordinator#refresh`, so the file is dead weight exercised only by its own unit test.
- [ ] `lib/hive/daily_digest/delivery.rb:104` — the `begin/rescue` wraps both the Telegram call and the post-send bookkeeping, so a failure after a successful send (logger or ledger write) drives `settle_send_error`, which attempts an illegal `sent` -> `unknown` transition; narrow the rescue to the transport call only.

ivankuznetsov:

## High
- [ ] Preserve previously observed facts during source outages (`lib/hive/daily_digest/coordinator.rb:193-196`): Replacing an open base uses only the latest collection batch, so an unreadable project removes its already persisted activity and frontiers; an in-memory reproduction changed `[f1]` to `[]` after an outage, and closing then freezes that loss. Retain known facts and committed frontiers while adding the source gap (R3/R10).
- [ ] Route late facts into complete closed days during normal refresh (`lib/hive/daily_digest/coordinator.rb:256-261`): Automatic refresh excludes every complete closed or pruned interval, while the collector for the current interval discards older event times; a merge learned today with yesterday's timestamp therefore never becomes an amendment or tombstone discard unless the operator already knows to refresh that exact historical date. Collect new observations independently of target-day completeness and route them by persisted intervals (R3/U3).
- [ ] Fill coverage holes created by explicit date refresh (`lib/hive/daily_digest/coordinator.rb:43-48`, `:78-84`): With coverage starting August 30 and no stored records, refreshing September 1 writes only that date; subsequent interval enumeration starts from the persisted September 1 record and permanently omits August 30-31, even rejecting them as predating coverage. Generate missing intervals from the configured coverage frontier and preserve chronological catch-up (R2/U3).
- [ ] Clear resolved failures and holds from boundary attention (`lib/hive/daily_digest/project_source.rb:348-354`): A failed session followed by success and stage completion still produces failed attention, including for a task in `9-done`; successful sessions and stage transitions never clear the fold, and native recovery outcomes do not match the accepted clearing values. Observe and fold actual state exits so completed/recovered tasks stop appearing blocked or failed (R4/R5).
- [ ] Isolate structurally malformed journal rows (`lib/hive/daily_digest/project_source.rb:240-241`): An activity row containing valid JSON with `"payload":"broken"` raises `NoMethodError` in `stringify` before classification; neither the per-row rescue nor Collector handles it, so one corrupt project aborts the global refresh. Validate nested row shapes and emit a scoped malformed-source gap while retaining healthy projects' activity (R10).
- [ ] Instrument the required provider, authority, and scheduler-capacity holds (`lib/hive/daily_digest/materiality.rb:65`): `hold_recorded` is declared unsupported legacy and has no producer in `lib/`; terminated-agent resource limits do not cover scheduler provider/capacity or authority holds, and ProjectSource does not consume operational disposition snapshots. Add durable entry/exit observations and boundary reconstruction for these explicitly required, non-deferred holds (R4/R5/U2).
- [ ] Restore quota recovery without deterministic-failure parking (`lib/hive/daemon/recovery_coordinator.rb:1150`, `:1688`): This diff removes the `limits_reached` parking exemption and same-runtime rearming, so repeated provider-quota failures can remain parked after the quota renews; it also replaces the hourly quota cooldown with crash backoff at line 194. Preserve the existing quota-specific recovery behavior rather than introducing this unrelated automation regression.
- [ ] Validate historical task links in CLI JSON and text (`lib/hive/commands/digest.rb:107-114`): CLI serialization passes stored task URLs through without checking current registration or task resolution; removing or replacing a project leaves actionable links that are dead or point at a replacement project's same slug. Web alone performs historical-link validation. Share the resolution check and emit historical, non-actionable identities when the original task no longer resolves, as explicitly required by U5/R12.

## Medium
- [ ] Derive unknown content for stale empty records (`lib/hive/daily_digest/reader.rb:45-48`): Adding a required materializer gap changes completeness to partial but leaves `effective_content: empty`; the reproduced view is simultaneously stale, partial, and empty, and Web still claims every required source was observed successfully. Recompute effective view content as unknown when freshness fails and no items or attention are known (A3/R10).
- [ ] Render actual changed outcomes in human views (`web/app/views/digests/_project.html.erb:14-15`, `lib/hive/commands/digest.rb:226-229`): Web and CLI text show constant summaries without the available outcome details, so passing and failing checks both read “Pull request checks changed,” and stage-transition targets remain invisible. Render bounded transition targets and PR/check/review outcomes so the selected-day surface explains what changed (R6/R12).
- [ ] Retain PR identity fields for check observations (`lib/hive/daily_digest/materiality.rb:86`): The check-event allowlist removes `pr_url`, `pr_state`, and `draft` before constructing the normalized PR; a fully populated check observation consequently produces `pr.url: null`, leaving CI-only activity without its direct PR link. Preserve the required safe PR fields through normalization (R9/AE1).
- [ ] Accept the dispatcher's stage argument when canceling delivery (`lib/hive/daemon/daily_digest_delivery_scheduler.rb:61`): Dispatcher calls `cancel(date:, stage:)`, but this override accepts only `date:`; the call raises `ArgumentError: unknown keyword: :stage`. When delivery capacity is occupied, cancellation fails without clearing the pending marker, which can suppress future recaps until restart. Match the base cancellation signature and test the gated-dispatch path.
- [ ] Bind resumed prepared delivery receipts to the payload actually sent (`lib/hive/daily_digest/delivery_ledger.rb:51`): Missing credentials leave an intent prepared; after a late amendment or Web-origin change, prepare resumes the old hash/frontier unchanged while Delivery sends a newly rendered payload at `delivery.rb:93-94`. The sent receipt then describes different bytes. Atomically refresh prepared payload metadata or resume the original bound payload before transport (U6).

## Nit

ivankuznetsov:

## High

- [ ] Preserve unresolved gaps when skipping unchanged journals (`lib/hive/daily_digest/project_source.rb:219-225`): Collection records a fingerprint before validating the journal; the next unchanged scan returns without reproducing `malformed_journal` or `pr_evidence_incomplete`, and `Coordinator#positive_recovery_evidence?` accepts that fingerprint as recovery, falsely marking broken evidence complete or appending a false resolution amendment; retain cached source health or revalidate unresolved evidence before resolving its gap (R10).

## Medium

- [ ] Isolate activity rows with missing timestamps (`lib/hive/daily_digest/materiality.rb:177`): An activity row with task/payload hashes but no `occurred_at` passes `ProjectSource#valid_activity_shape?`, then raises an uncaught `KeyError` that aborts the global refresh, including healthy projects; validate required fields and apply the planned observation-time fallback or emit a scoped gap (R10; reproduced in memory).
- [ ] Give recurring PR outcomes distinct transition identities (`lib/hive/task_workspace/publication_activity.rb:53-60`): Passing -> failing -> passing checks on the same PR/head reuse the first passing operation ID because it hashes only the destination payload; the journal rejects the returning outcome as an idempotency conflict and the producer suppresses the error, losing a real recovery; identify each durable transition while still deduplicating unchanged polls (R4/R9).
- [ ] Recompute waiting age when retaining unchanged attention (`lib/hive/daily_digest/coordinator.rb:299-310`): Unchanged journals skip boundary reconstruction, so merging retains the previous attention object and its old `waiting_age_seconds`; repeated refreshes and eventual close freeze the age at the last journal change instead of the observation/closing boundary; derive age from durable `waiting_since` at the appropriate boundary (R5).
- [ ] Consume tombstone frontiers and acknowledge recovered gaps once (`lib/hive/daily_digest/coordinator.rb:194`): Pruned records store `source_frontiers`, but collection reads only `effective_source_frontiers`, so every refresh replays their inputs; recovered gaps also remain in the tombstone and receive a new discard ID because each resolution uses the current observation time, causing unbounded repeated recovery audits; read the retained frontier and persist an idempotent resolution acknowledgement (R11/U3; reproduced in memory).
- [ ] Keep store reads free of filesystem initialization and repair (`lib/hive/daily_digest/store.rb:186-231`): `Reader` calls `Store#intervals`, which creates/chmods the store and lock and writes `intervals.json` when absent, so a bare CLI read or Web GET mutates even an uninitialized/disabled projection and requires write access; reserve initialization/index persistence for mutation owners and use a non-writing read path (R7/A2/A9).
- [ ] Recover valid but stale interval indexes after interrupted publication (`lib/hive/daily_digest/store.rb:45-46`): A crash after writing a closed base but before updating an already-valid `intervals.json` leaves that day absent from default lookup/navigation; `interval_index` trusts the stale file and refreshing the existing closed base follows the amendment/frontier path without repairing it, so later index writes retain the hole; make publication recoverable across both writes (R2/R3).
- [ ] Bind prepared delivery ownership to the exact payload (`lib/hive/daily_digest/delivery_ledger.rb:51-56`): Two senders preparing the same date across a late amendment reuse one attempt while the second overwrites its payload hash/frontier; the first can then enter `sending` and send its older payload while the receipt records the newer one as sent; claim preparation exclusively or compare the payload/frontier identity when transitioning (R8/U6; reproduced with an in-memory ledger).
- [ ] Reconcile interrupted deliveries independently of the current recap date (`lib/hive/daemon/daily_digest_delivery_scheduler.rb:67-73`): Failed children only trigger backoff, and leftover `sending` intents become `unknown` only when that exact date is prepared again; restarting after the preceding-day window advances leaves older interrupted receipts permanently `sending`; reconcile dead senders on timeout/startup regardless of the next scheduled date (R8/U6).
- [ ] Preserve the archive route in CLI and JSON task links (`lib/hive/daily_digest/task_links.rb:35`): Successful validation discards the destination's `source: "archive"` and retains the collected URL without `?source=archive`; once a completed task exceeds ordinary retention, the digest's CLI/agent link returns unknown task although the archive route exists; rebuild the URL from the resolved destination while preserving the answer anchor (R5/R7/R12).
- [ ] Restore native-runtime discovery through Codex wrappers (`lib/hive/agent_support/codex/artifact_policy.rb:61-67`): Rejecting every shebang launcher breaks managed capture when the configured Codex executable is an npm launcher or shell shim and no native binary is directly exposed on PATH; version checking succeeds but capture now raises `ConfigError`, whereas `origin/main` used `Runtime.executable` and its doctor/provenance fallback; retain that supported resolution path.

## Nit

None.

ivankuznetsov:

## High
- [ ] Prevent stale concurrent refreshes from replacing newer activity (`lib/hive/daily_digest/store.rb:40`): The coordinator reads, collects, and merges outside the store lock, while `write_base` replaces an open record without comparing revisions; two refreshes can commit X+Y followed by stale X, losing committed activity and regressing its frontier, or closing that stale snapshot. Serialize the read/merge/commit transaction or reject stale revisions and retry against the latest base (R3).
- [ ] Restrict each registration to its actual membership interval (`lib/hive/daily_digest/coordinator.rb:190`, `lib/hive/daily_digest/coverage.rb:78`): Coverage returns every registration overlapping the day, but collection applies the entire day to each; replacing or re-registering the same path at noon attributes journal facts to both registration IDs, and registrations removed before the boundary still contribute attention. Carry active membership bounds into fact collection and boundary selection (R4/R5 and U2).

## Medium
- [ ] Preserve boundary evidence across unchanged-journal cache hits (`lib/hive/daily_digest/project_source.rb:224`): The early return leaves `boundary_records` empty, so the second refresh of an unchanged WAITING task reports `boundary_history_missing` despite its durable question transition; the coordinator retains its attention but adds a false completeness gap. Cache the boundary fold or reconstruct it before checking for missing history.
- [ ] Invalidate cached PR gaps when publication metadata changes (`lib/hive/daily_digest/project_source.rb:224`): PR enrichment reads `pr.md`, but the cache checks only the journal signature; repairing missing metadata without changing the journal replays `pr_evidence_incomplete` indefinitely instead of resolving it through an amendment. Include publication metadata in the cache dependencies or reread it when retrying that gap (R9/R10).
- [ ] Resolve archive links through workflow semantics (`lib/hive/daily_digest/task_links.rb:72`, `web/app/models/daily_digest.rb:158`): Both paths require stage index 9 for `source=archive`, but Content completes at stage 6 and ordinary status hides it after three days; historical digest links then fail even though the archive route exists. Determine archive routing from the task's workflow rather than the coding stage number (R6/R12).
- [ ] Determine precoverage from the persisted local interval (`lib/hive/daily_digest/reader.rb:169`): Comparing the requested label with the coverage instant's UTC date mislabels valid missing history west of UTC; enabling at August 30 20:00 in Los Angeles makes the initial August 30 record appear precoverage because coverage began August 31 UTC. Use the first interval's stable label so Web does not falsely say that refresh cannot reconstruct the day.

## Nit

PR #1410, opened 2026-09-06
fix: keep tasks retryable and workflow restrictions opt-in

ivankuznetsov:
## Summary

- Provider quota failures retry hourly indefinitely instead of becoming operator-owned after repeated identical failures. Already parked quota requests rearm automatically without skipping their cooldown.
- New blank and research workflows inherit project permissions instead of adding an unsolicited read-only policy. Explicit owner-requested restrictions remain supported.

## Test plan

- [x] 149 focused tests, 1,100 assertions passed across recovery coordinator, recovery API/authority, workflow scaffolding, and permission scopes.
- [x] Regression reproduced before the fix: the 26th identical quota failure was blocked; it now remains queued. New tests cover first-failure hourly delay and legacy parked quota recovery.
- [x] RuboCop clean on changed Ruby files; `git diff --check` clean.
- [ ] Broad `bundle exec rake test` checkpoint (running locally).
- [ ] Hosted coverage and dedicated CI merge gates.

## Notes for reviewer

Provider reset timestamps remain informational: a multi-day reset does not prevent hourly readiness retries. Non-provider deterministic failures retain their existing handling. No retry-count migration or new table is needed.

The local Architecture workflow's unwanted scoped allowlist was removed separately with a backup. This PR does not silently strip explicit restrictions from installed third-party workflows. Runtime deployment remains pending.

---

### DIFF
--- a/lib/hive/artifacts/outcome_evidence/identity.rb
+++ b/lib/hive/artifacts/outcome_evidence/identity.rb
@@ -2,6 +2,7 @@ require "digest"
 require "pathname"
 require "hive/agent_git_gate"
 require "hive/draft_pr_receipt"
+require "hive/git_ops"
 require "hive/artifacts/outcome_evidence/document"
 require "hive/worktree"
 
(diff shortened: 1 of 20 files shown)


ivankuznetsov:
CI exposed the old exact production-line-count gate, not a behavior regression. Removed that one-time cutover measurement and its Git-history helpers in 9af1c77ec4; retained all retired-store/schema/constant and bootstrap regression checks. The five remaining deletion-contract tests pass (99 assertions). Historical inventory evidence remains unchanged.

PR #1415, opened 2026-09-07
fix: keep interrupted tasks retrying without runaway dispatch

ivankuznetsov:
## Summary

- Keep failed task transitions retryable without rapid redispatch or an artificial 100-retry ceiling, and show the scheduler's real retry time instead of a provider's reset estimate.
- Recover abandoned empty Git index locks conservatively. Live, recent, nonempty, foreign-owned, or uncertain locks remain untouched; recovered files are quarantined for inspection.

## Test plan

- [x] Focused tests: operation receipts, lock recovery, attempt dispatch, recovery coordinator, operational snapshot/status, and daemon dispatch.
- [x] Regression exercises more than 100 aborted retries, repeated hourly backoff, and a real Git write after lock recovery.
- [x] RuboCop on all changed Ruby files and `git diff --check`.
- [ ] Broad local `bundle exec rake test` (running).
- [ ] Hosted CI coverage and dedicated merge gates.
- [ ] Deploy to dogfood and verify the affected Webmail task; the existing daily dispatch cap may still defer advancement.

## Notes for reviewer

The incident combined an abandoned Git lock with failed transitions that immediately became eligible again. Numbered operation receipts then ran out, and redispatch continued until the project consumed its daily cap. The fix reuses existing operation identity and durable attempt retry charges; it adds no retry store or maintenance service.

Daily caps remain unchanged. Patrol behavior is excluded from the new transition delay. Lock recovery requires successful `ps` and `fuser` probes; missing tools leave the normal Git failure in place.

PR #1416, opened 2026-09-07
fix: recover abandoned Git locks with PSmisc fuser

ivankuznetsov:
## Summary

- Make abandoned Git lock recovery actually work with PSmisc fuser. Live dogfood verification found that its unsupported `--` separator produced usage text, so the safety check correctly declined recovery every time.

## Test plan

- [x] Five lock-recovery tests, 17 assertions, pass.
- [x] New real-process test preserves an open lock, recovers it after close, and completes a real Git write.
- [x] RuboCop and `git diff --check` pass.
- [ ] CI, merge, and dogfood verification of the original Webmail transition.

## Notes for reviewer

The path comes from Git's absolute-path resolver, so it cannot be parsed as an option. Uncertain probe results still leave the lock untouched. This closes the platform behavior missed by the original mocked probe test.

Related: #1415

### DIFF
--- a/lib/hive/git_index_lock.rb
+++ b/lib/hive/git_index_lock.rb
@@ -34,7 +34,9 @@ module Hive
       return false unless status.success? && errors.empty?
       return false if processes.lines.any? { |line| File.basename(line.strip).match?(/\Agit(?:\z|[- ])/) }
 
-      holders, errors, status = Open3.capture3("fuser", "--", path)
+      # rev-parse supplies an absolute path, so no option delimiter is needed.
+      # PSmisc fuser does not accept the conventional `--` delimiter.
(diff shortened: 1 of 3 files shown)


ivankuznetsov:
Live verification passed on dogfood build 794edfb489fc620f526a71bc7e5c62f6f2cca856. The normal hive approve command recovered and quarantined the real September 1 index.lock, returned ok:true / forced:false, and committed webmail.sh:build-the-rails-foundation-for-260708-d6d9 from 7-artifacts to 8-finalize (state commit 8c0106c0). Task readback confirms 8-finalize. Daemon, web, and babysitter services are active. No dispatch cap was changed or reset.

PR #1418, opened 2026-09-08
Make CLI usage contracts command-owned and resolve them once

ivankuznetsov:
Each command now declares the JSON error contract used when argument parsing fails before its handler runs. The launcher resolves that contract once and reuses it for the error class and output, replacing its central registry and selection helpers. A boundary-load or resolver failure keeps the generic usage response and records only the exception class.

The workflow error schemas accept their existing UsageError envelopes. The final main refresh also preserves the digest contract introduced by #1407; its command now owns that declaration, with a regression that failed before the repair.

Validation: nine review passes found no original functional defects. Cold-start seeds 1, 3 and 17 passed. Focused changed coverage passed 2,194 tests / 14,233 assertions with exact line coverage across all 29 then-changed sources. The broad checkpoint passed 14,551 tests / 289,770 assertions (14 skips), plus the separate runtime component suite (131 tests / 1,291 assertions, one skip). After refreshing onto main through #1414 and repairing the newly introduced digest contract, the final focused CLI/digest checks passed 174 tests / 2,500 assertions; changed-file lint is clean. Hosted checks for the final head are pending.

Post-merge verification: the first main run exposed a digest preparer-token GC bug. Corrective #1427 reproduced and repaired it; descendant main `4b6838b5061b34873dc1f5e81a0919b8e7203b21` includes this merge and passed CI 34434802366 and Install smoke 34434802341.


ivankuznetsov:

## High

- [ ] Complete the required U4 coverage and broad checkpoint (`docs/implementation/cli-usage-contracts-baseline.md:197-214`): R6/U4 require passing changed-lib coverage at 100% and `bin/test --all`, but recorded coverage is 93.97% with 364 uncovered lines, the subsequent attempt has 10 registration-order errors, and the broad run was interrupted; the later schema fix passed 70 focused tests but does not satisfy these gates. Resolve the coverage/test failures, complete both required checks on the final revision, and record their results in implementation evidence.

## Medium

None.

## Nit

None.

ivankuznetsov:
### Reviewer: grok-ce-code-review - Pass 01

## High

## Medium
- [ ] wiki/cli.md Error envelopes paragraph still calls hive-setup usage JSON unversioned: the same diff documents hive-setup.v1 on this page and in commands.md, and `SCHEMA_VERSIONS` maps `hive-setup` to 1, so agents reading [[cli]] will mis-classify setup usage envelopes.
- [ ] wiki/log.d/20260908T120000Z-command-usage-boundaries.md still says no public schema changes and that U4 stopped incomplete: that timestamp sorts after the schema-compatibility fragment while the same diff widens lifecycle `error_kind` enums, so the latest log entry is false.

## Nit
- [ ] `setup` / `stage_action` / `finding_toggle` call `declare` with keyword hashes instead of an explicit Hash: this works on Ruby 3.4.10 via implicit keyword-to-last-positional conversion, but every other boundary passes `{ ... }` and that conversion is the brittle form.

PR #1419, opened 2026-09-08
Add durable skill and workflow proposal tracking

ivankuznetsov:
## Summary

Operators can retain and query the full history of skill and workflow proposals, including rejected experiments and later rollbacks. The ledger connects immutable candidate revisions with evaluations, authoritative decisions, lineage, and external revert evidence while leaving active skills and workflows unchanged.

- Task-bound submissions and evaluations use durable attempt bindings, committed source receipts, isolated Git transactions, and bounded crash recovery. Configured lifecycle authorities decide through compare-and-swap over the lifecycle head and observed evaluation digests.
- Live CLI discovery and deterministic wiki views share canonical projections. Corrupt neighbors produce safe quarantine diagnostics; generated wiki pairs publish atomically from a pinned state commit.
- Future proposal-bound attempts receive only bounded typed facts, including fixed-label numeric results. Evidence policy, redaction, quotas, and the shared prompt appendix budget apply before information reaches consumers.

Authority follows Hive's same-user capability model; it is not cryptographic proof of a human caller. Retention is declarative (`enforcement=none`). Dedicated evolution workflows and activation remain follow-up work.

## Test plan

Earlier local validation recorded in the PR reported:

- Focused proposal suite: 734 runs, 4,195 assertions; lifecycle integration: 2 runs, 60 assertions.
- CLI JSON contract: 1 run, 22 assertions; wiki refresh integration: 3 runs, 13 assertions; schema/wiki checks: 214 runs, 1,789 assertions.
- RuboCop: 1,826 files without offenses; exact coverage: 99,086/99,086 lines (100%).
- The broad suite recorded one TUI timing failure under concurrent load; its isolated rerun passed in 1.48 seconds. These earlier results were not rerun during finalization.

Current hosted validation for `97862cc0c02e2dd5063f52b76cb8bdfaeea2dbb5` is **not green**:

- [x] RuboCop, security checks, CLI scenarios, web tests, packaged bootstrap, TUI performance, multi-agent setup, and babysitter security checks passed.
- [ ] Coverage shards 4/6 and 6/6 failed; the aggregate coverage and Ruby test gates consequently failed. Shard 4 includes a failure in “Bundle pinned Betterleaks”; shard 6 failed “Collect coverage shard”. See the [CI run](https://github.com/ivankuznetsov/hive/actions/runs/34290185851).
- [ ] Resolve the failed hosted checks and obtain green exact-head coverage and aggregate test gates before merge.

## Review summary

Two Codex and Grok review passes marked every finding fixed or resolved, with no outstanding user questions in their escalation reports. The fixes cover evidence-policy narrowing, evaluator revocation, committed-state durability, unrelated-index preservation, quota accounting, quarantine isolation, observed-evidence concurrency checks, safe typed context, and pinned wiki publication.

The second pass recorded an automatic commit-scope rejection for five proposal schemas. Those schemas are present in the final committed PR diff. Review completion does not supersede the failed hosted checks above.

## Linked task

Related: Hive task `43059` — `hive:track-skill-and-workflow-proposal-260829-f948`.


ivankuznetsov:

## High
- [ ] Enforce evidence policy on canonical-shaped submission inputs (`lib/hive/proposals.rb:351-382`): A producer can supply `visibility: project`, `retention: {policy: project, enforcement: none}`, and a summary to bypass the configured restricted/task ceiling; the source submission validator reaches this branch and persists the summary, violating R13; separate persisted-record validation from admission or intersect policy on both paths.
- [ ] Exclude free-form labels and metric keys from automatic context (`lib/hive/proposals/context_selector.rb:133-149`): Method labels, revision strings, and metric keys are unrestricted prose, yet are copied into the prompt as supposedly closed typed facts; validation accepts instruction-bearing labels and metric names, so JSON encoding does not satisfy R12's instruction-text exclusion; use controlled identifiers/fixed labels and omit arbitrary keys and prose.
- [ ] Revalidate evaluator authorization before admitting new source events (`lib/hive/proposals/producer.rb:125-135`): Admission uses only the evaluator binding saved at dispatch, so removing an evaluator from configuration still allows that active attempt to commit new evaluations; check current authorization for new receipts while preserving historical authorization for exact retries and replay, as U2 requires.
- [ ] Verify committed terminal state before returning ingestion success (`lib/hive/proposals/ingestor.rb:32-46`): The reusable Producer service does not perform CLI reconciliation, and the ingestor trusts an uncommitted checkout `consumed` marker before taking the commit lock or checking canonical commits; a crash after writing that marker but before committing can therefore make a retry report success without durable canonical history; enforce committed-state cleanup/verification inside the service boundary.
- [ ] Bound or shard the source index before committing it (`lib/hive/proposals/source_event_store.rb:248-249`): `proposal_usage` retains every candidate and the writer has no size limit, while index reads reject files at 256 KiB; 4,000 valid usage rows alone serialize to 280,001 bytes, below default event/storage quotas, so normal growth commits an index that subsequently blocks ingestion and reconciliation; keep the recovery index bounded and reject any oversized write before the source commit.
- [ ] Apply aggregate quotas to the complete proposal namespace (`lib/hive/proposals/source_event_store.rb:46-55`, `lib/hive/proposals/decision_service.rb:50-54`): Accounting includes source receipts/status bytes but excludes canonical records/events and index bytes, while lifecycle mutations bypass aggregate limits and authority rate checks entirely; the configured ceilings therefore do not bound actual storage or lifecycle writes as U2/U3 require.
- [ ] Expose terminal source quarantine through live queries and compiled diagnostics (`lib/hive/proposals/store.rb:190-199`, `lib/hive/proposals/compiler.rb:16`): The normalized snapshot and pinned compiler read only records/events, whereas permanent source failures are stored separately under inbox quarantine; consequently `filter --include-quarantine` and the machine index silently omit these failed admissions, contrary to the required U2/U5 recovery visibility; include their safe terminal diagnostics in the shared query/compile path.
- [ ] Select the newest queued proposal commit by ancestry (`templates/llm-wiki/post-commit-refresh.sh:880-894`, also `.llm-wiki/post-commit-refresh.sh`): Sorting commit timestamps and SHA can select an ancestor when consecutive commits share a timestamp, after which lines 1173-1177 acknowledge and delete every queued source; the newest proposal changes are then omitted without a retry; choose the newest queued descendant rather than lexical SHA order.
- [ ] Protect generated proposal files during ordinary wiki refreshes (`templates/llm-wiki/post-commit-refresh.sh:899`, also `.llm-wiki/post-commit-refresh.sh`): For a batch without proposal changes, compilation returns immediately, but the wiki agent can modify either generated file and broad wiki staging publishes those modifications; restore or regenerate the canonical pair before publication even for ordinary batches so agent edits cannot fabricate facts or publish a torn pair, as U4 requires.
- [ ] Include evaluation evidence and source links in the human wiki summary (`lib/hive/proposals/compiler.rb:129-156`): Evaluation rendering omits evidence digests, method references, provenance/task/commit links, and decision links, although U4 explicitly requires these facts across canonical, JSON, and Markdown views; render the safe references and digests so wiki readers can recover the evidence behind accepted and rejected lessons.

## Medium
- [ ] Make explicit refresh enqueue proposal compilation independently of current HEAD paths (`lib/hive/commands/proposal.rb:185-186`, `lib/hive/commands/proposal.rb:213-217`): The command invokes the ordinary post-commit hook, which exits successfully when the latest state commit changes only inbox/cursor/task-state paths, then reports `queued` despite doing nothing; explicitly enqueue the appropriate pinned proposal source so operators can repair stale views after unrelated state commits.

## Nit

ivankuznetsov:
### Reviewer: grok-ce-code-review - Pass 01

## High
- [ ] Hash-shaped evidence skips classification narrowing: `Proposals.evidence!` treats `retention: {policy, enforcement}` as already-persisted and copies producer `visibility`/`summary` without `effective_classification`, so a submit/evaluate artifact can store `visibility=project` plus a summary under the default restricted/task ceiling and publish that text into `wiki/proposals.md` (R13).
- [ ] Malformed event filenames poison the whole proposal: `Store#next_version_unlocked` reserves any `^\d+-` prefix, then `Projection#validate_history!` requires versions `1..n` with no gaps, so a junk `7-not-an-event.json` plus a valid v8 event yields `inconsistent_history` and drops the candidate from CLI/wiki/context even though the record and valid event remain (R10).
- [ ] Exact-max source receipts cannot be replayed: `admit!` allows a canonical body of `MAX_FILE_BYTES`, but `read_bytes` rejects `stat.size >= max_bytes`, so a 256KiB inbox receipt is durable then `fetch`/`ingest` raise oversize/unavailable.
- [ ] Consume at the byte cap wedges the inbox: `enforce_quotas!` charges only the receipt, then `mark_consumed!`/`quarantine!` charge the status JSON and raise `QuotaExceeded`; Reconciler does not treat that as permanent, so the source stays pending and aborts every later batch item.
- [ ] Crash recovery terminally quarantines TEMPFAIL misses: `SourceUnavailable` (exit TEMPFAIL) is in `Reconciler::PERMANENT_ERRORS`, so a missing HEAD blob or ingest between inbox commit and canonical append writes `inbox/quarantine/<id>.json` and never replays the committed receipt.
- [ ] Context overflow can drop the bound proposal: `ContextSelector` does `truncated = true; next` when a ranked item exceeds the budget, so a too-large exact match is skipped and a smaller related/same-subject lesson is packed instead of prefix-truncating (R12).

## Medium
- [ ] `no_evaluation` can ignore existing evaluations: `Event.normalize_decision` allows an empty considered set whenever the category is `no_evaluation`, and `DecisionService` does not require `projection.evaluations` to be empty, so an authority can reject while leaving evaluations unconsidered.
- [ ] Canonical records/events are outside the advertised byte quota: `Store.create_immutable` only enforces per-file `MAX_FILE_BYTES`, and lifecycle `append_event!` never calls `SourceEventStore.enforce_quotas!`, so `max_project_bytes` does not count the records/events tree the plan said must be included.
- [ ] Inbox `index.json` will quarantine the ledger before the 10k event cap: `proposal_usage` is never pruned and is read with the 256KiB receipt cap, so a few thousand unique proposal IDs make `pending_ids`/`admit!` raise `QuarantinedSource` while `max_project_events` still has headroom.
- [ ] Failed source commits leave `logs/` staged: `hive_commit` always `git add logs`, but Producer rescue unstages only the task folder and inbox paths, so the next unrelated `hive_commit` can last-writer-wins leftover log index entries.
- [ ] Reconciler holds `.commit-lock` for the full batch: `reconcile!` wraps up to 64 nested `ingest!`/`hive_commit` calls in one lock, so a concurrent `decide`/`supersede`/`rollback` waits out `COMMIT_LOCK_TIMEOUT_SEC` instead of interleaving with evaluation ingest.
- [ ] `ConfiguredDispatcher.dispatch` never reconciles: only `dispatch_request` and proposal CLI/list paths call `Reconciler`, so a crash-left inbox receipt is not recovered on ordinary foreground attempt launch.
- [ ] Store/SourceEventStore `flock(LOCK_EX)` has no deadline: unlike `Hive::Lock.with_commit_lock`, a stuck peer can block admit/ingest/lifecycle forever on the shared `/tmp/hive-proposal-<hash>.lock`.
- [ ] Pinned compile materializes git mode `120000` as live `File.symlink` targets from blob bytes, so a committed symlink record/event becomes a real symlink in the compile tmpdir (Store then quarantines it, but the target is attacker-controlled).
- [ ] Producer/ingestor failure tests stub `hive_commit`/`run_git!` before `git add`, so they cannot catch leftover staged logs or `:nothing_to_commit` after a durable write; wiki proposal-only compile failure paths are similarly untested.

## Nit
- [ ] `Attempts::Record` reimplements proposal-binding validation beside `SourceEvent.normalize_proposal_binding`, so the two closed hashes can drift.
- [ ] `Query#relations` and `ContextSelector#relation_ids` duplicate the same lineage tuple.
- [ ] `hive proposal list`/`filter` JSON schema ternary is a no-op (`"hive-proposal-list"` on both branches).
- [ ] Proposal config validation was inlined into already-oversized `lib/hive/config.rb` instead of a proposal-local validator.
- [ ] Human `wiki/cli.md` error-envelope list was not updated for the new proposal JSON schemas.

ivankuznetsov:

## High

- [ ] Redact quarantine filenames before publishing diagnostics (`lib/hive/proposals/store.rb:474`): `logical_path` is copied verbatim into diagnostics and then CLI/compiler JSON and Markdown; a malformed filename containing a token recognized by `Hive::SecretPatterns` exposes that token in generated wiki history. Apply bounded, encoding-safe secret redaction before constructing the diagnostic, as required by R10/R13.
- [ ] Isolate the Git index for proposal transactions (`lib/hive/proposals/producer.rb:124`, `lib/hive/proposals/ingestor.rb:74`): `GitIndexSnapshot` is restored only on failure, while `GitOps#hive_commit` commits the entire existing index; an unrelated pre-staged task journal therefore enters a successful source/canonical commit. Preserve unrelated staging and commit only the authorized normal task paths plus exact proposal paths, as required by U2.
- [ ] Terminally quarantine oversized committed inbox receipts (`lib/hive/proposals/reconciler.rb:60`): the bounded Git blob reader returns `nil` for oversized receipts, which becomes retryable `SourceUnavailable`; these entries remain pending forever and can occupy every slot in the repeatedly selected batch, starving valid receipts. Distinguish permanent oversize corruption from transient unavailability and commit its quarantine status.
- [ ] Implement the required numeric evaluation facts in automatic context (`lib/hive/proposals/context_selector.rb:128`): `typed_fact` drops every `result.metrics` value and emits only enum outcomes, so the numeric threshold outcome explicitly required by R12/U5 and AE2 can never reach a future proposal run. Add a closed numeric representation with fixed labels while continuing to exclude arbitrary metric keys and free-form text.
- [ ] Bind decision observations to the considered evaluations' original digests (`lib/hive/proposals/decision_service.rb:178`): callers supply evaluation IDs only, the lifecycle head excludes evaluations, and considered result digests are calculated from current bytes; a schema-valid rewritten evaluation retaining its ID therefore passes an older observation instead of returning stale. R7 explicitly requires detecting changed considered evidence; compare caller-observed evaluation digests without invalidating decisions for unrelated appends.

## Medium

- [ ] Keep malformed canonical events from breaking quota admission (`lib/hive/proposals/source_event_store.rb:346`): the rate-limit scan calls `dig` on unchecked parsed JSON; an event containing `[]` raises `TypeError` and `null` raises `NoMethodError`, neither handled by this loop, so a quarantined neighbor blocks unrelated submissions and lifecycle writes. Validate the object shape and isolate malformed entries while retaining their quota accounting.
- [ ] Validate retry predecessor references (`lib/hive/proposals/store.rb:321`): retry edges enter cycle detection without checking that the predecessor exists, and submission/ingestion do not supply that check; a candidate retrying an arbitrary nonexistent opaque ID is consumed and compiled as valid without a diagnostic. Reject dangling retry links at admission and quarantine them during projection replay.
- [ ] Record provenance when the context budget excludes every lesson (`lib/hive/context_provenance.rb:79`): activity recording is inside `unless proposal_selection.empty?`, so a bound selection with `truncated=true` and `reason=budget_exhausted` loses its effective-budget and truncation evidence when the first item cannot fit. Persist selection provenance even when no proposal text is appended.
- [ ] Compare refresh checks against the published wiki pair (`lib/hive/commands/proposal.rb:198`): `refresh --check` compares scratch output with primary-checkout files, but managed refresh publishes only the separate wiki branch and leaves that checkout unchanged; a successful refresh can therefore still report stale. Read the pair from the managed publication ref and test refresh followed by check without manually copying generated files into the primary checkout.

## Nit

None.

ivankuznetsov:
### Reviewer: grok-ce-code-review - Pass 02

## High
- [ ] Reserved-version gaps drop valid projections: `Projection#validate_history!` requires every reserved `\d+-` prefix to be contiguous, so a junk later slot (valid v1 plus `00000000000000000007-not-an-event.json`) raises `InconsistentHistory` and Store hides the candidate from CLI/wiki/context (R10).
- [ ] Lifecycle mutations report success without a durable commit: `DecisionService#with_lifecycle_commit` ignores `hive_commit`'s return and never verifies the event blob at HEAD, so `:nothing_to_commit` can yield `applied: true` and Reconciler later deletes the uncommitted event.
- [ ] Empty evaluator allowlists admit every workflow/stage/profile: `EvaluatorAuthority#match!` treats `[]` as a wildcard, so a named evaluator with empty lists is fail-open despite documented fail-closed admission (R2).
- [ ] Inbox writes follow symlink parent directories: `SourceEventStore#create_immutable` / `#with_lock` use `FileUtils.mkdir_p` plus `AtomicFile.create` without `Store#ensure_safe_directory!`, so a planted `inbox/` symlink redirects durable receipts.

## Medium
- [ ] Admission quota and hourly rate-limit still walk consumed history: `namespace_usage_unlocked` `Find.find`s the whole `proposals/v1` tree and `recent_actor_events_unlocked` globs every inbox receipt plus every event file, and Producer runs that under the 30s hive-state commit lock.
- [ ] Ingest idempotency reloads the entire canonical store: `Store#find_by_source_event_unlocked` calls `load_unlocked` per `create_record!`/`append_event!`, so a reconcile batch of 64 holds the commit lock across 64 full-namespace loads.
- [ ] Producer rollback can re-plant a task-journal symlink after append-through: `PathSnapshot#restore!` does `File.symlink` of captured roots, so a planted journal symlink is followed by WorkLedger append and then restored while the target keeps the extra line.
- [ ] Lifecycle GitOps tests cannot catch the missing durable-commit check: `decision_service_test` FakeGitOps returns `:committed` without `git add`, and `fail_reset` raises in `write-tree` before yield, so restore and HEAD verification never run.
- [ ] Exclusive `pathspecs:` `hive_commit` commits the whole index: leftover staged proposal files from a killed ingest can ride into the next decide/submit commit because `git add -- path` does not reset HEAD first.
- [ ] Failed `GitIndexSnapshot.restore!` is swallowed (`rescue nil`) in Producer/Ingestor/DecisionService, so a dirty index can be published by the next `hive_commit` after the lock is released.
- [ ] A failed quarantine `hive_commit` aborts the rest of `reconcile!`: `quarantine_source!` is not isolated, so one permanent receipt's commit error skips remaining pending neighbors in the same batch.
- [ ] CLI list/show/mutation `proposal_id`/`event_id` patterns (`^prp-[0-9a-f-]{36}$`) are looser than canonical UUID-v4 `PROPOSAL_ID`/`EVENT_ID` in the record/event schemas.
- [ ] `wiki/testing.md` proposal verification still documents serial `bundle exec rake test` / `coverage:changed` instead of `bin/test --changed` then `bin/test --all` (AGENTS.md local test loop).
- [ ] `Attempts::Record::SCHEMA_VERSION` stays 4 while `subject.proposal` is added to the closed task-stage key set, so older readers fail closed on new rows without a version bump.
- [ ] Oversize committed blobs retry forever: `read_hive_state_blob_at` returns nil when `size >= max_bytes`, Ingestor maps that to `SourceUnavailable`, and Reconciler leaves the receipt pending.
- [ ] No test covers `StaleObservation` when a decision names a missing considered evaluation ID (`decision_service.rb` `considered_evaluations!`).

## Nit
- [ ] `ContextSelector` reports `reason=budget_exhausted` when the bound subject has no terminal matches (draft-only), not only when a ranked item overflows the byte budget.
- [ ] `hive-proposal-mutation` `error_kind` is an unconstrained string rather than the closed set `Commands::Proposal#error_kind` actually emits.
- [ ] Inbox `hive-proposal-source-index` / `hive-proposal-source-status` schema names are used on disk but not published under `schemas/` or `SCHEMA_VERSIONS`.
- [ ] Authority and evaluator row shape is validated twice (ConfigValidator vs Authority/EvaluatorAuthority) and can drift on capabilities/array bounds.
- [ ] Lineage cycle detection is duplicated in `DecisionService#validate_supersession!` and `Store#cycle_nodes`.
- [ ] Hash-backed `Producer` construction test only asserts class, not that `subject.proposal` is the trusted binding.

PR #1438, opened 2026-09-14
Stop benchmark execution when the planner fails

ivankuznetsov:
A benchmark planner that timed out or exited with an error could leave a non-empty plan file. The pipeline then sent that partial plan to the executor and could report the result as `generated`.

Check the planner's status before executor dispatch. Failed planners now return `plan_failed` with no candidate diff; provider-limit failures retain precedence and return `limit_hit`. Successful planners still need a non-empty plan.

Validation at `080c004912b767d707532a7697c0d3ebcd1b70a7`, including current main:
- Bench pipeline, workflow, and installation tests: 58 tests, 544 assertions, no failures or errors.
- Regression coverage includes timeouts and non-zero exits with partial plans, successful generation, and provider-limit precedence for both `:error` and `:timeout` results using the production `provider_errors` field. Parked planners never dispatch the executor.
- During review, the current-main pipeline reproduced both partial-plan failures; the PR guard fixed them.
- Changed test file passes RuboCop; `git diff --check` passes.
- All 34 hosted checks passed on this exact head, including full coverage, browser suites, CLI scenarios, and advisory timing checks.

Source: ordinary_patrol `architecture-templates-builtins-part-7-20260914T000031Z-9cbfb4f6-1`.


ivankuznetsov:

Outcome: timeout

Agent tail:

```
Approaching the end of the bounded watch, there are still no new events. The remote SHA safety check was not needed because no push was appropriate.
```

